PR #5, opened 2016-07-12
Control Flow Graph

mrrrgn:
This isn't attached to the Function object in any way at the moment. I'm planning to support that in a later patch.


mrrrgn:
Arg, I accidentally 'git add'd some code that wasn't supposed to go in.

PR #6, opened 2016-07-12
Control Flow Graph

mrrrgn:
The CFG here is not associated with the Function yet though I'd planned on adding an `Option<ControlFlowGraph>` to the Function struct at a later time.

One bothersome thing here is that the relationship between `Function` and `ControlFlowGraph` is weak. To fix this I think it may be worthwhile to add a `UUID` to each `Function` created. This id could be associated with a CFG at creation and checked each time an operation is performed to ensure that we aren't operating on the wrong `Function`.

### DIFF
--- a/src/libcretonne/cfg.rs
+++ b/src/libcretonne/cfg.rs
@@ -43,76 +43,42 @@ impl ControlFlowGraph {
     /// During initialization mappings will be generated for any existing
     /// blocks within the CFG's associated function. Basic sanity checks will
     /// also be performed to ensure that the blocks are well formed.
-    pub fn new(func: &Function) -> Result<ControlFlowGraph, String> {
+    pub fn new(func: &Function) -> ControlFlowGraph {
         let mut cfg = ControlFlowGraph { data: BTreeMap::new() };
 
         // Even ebbs without predecessors should show up in the CFG, albeit
         // with no entires.
         for ebb in func.ebbs_numerically() {
-            try!(cfg.init_ebb(ebb));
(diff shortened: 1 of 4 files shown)


src/libcretonne/cfg.rs
```diff
@@ -0,0 +1,270 @@
...
+        };
+
+        if success {
+            Ok(())
+        } else {
+            let mut predecessors = try!(self.init_ebb(ebb));
+            if predecessors.insert(predecessor) {
+                return Ok(());
+            }
+            Err("Predecessor insertion failed.")
```
stoklund: I think this can safely be an assertion, and then this function and `new` amd `init_ebb` don't ned to return results.

src/libcretonne/cfg.rs
```diff
@@ -0,0 +1,270 @@
...
+            try!(cfg.init_ebb(ebb));
+        }
+
+        for ebb in func.ebbs_numerically() {
+            // Flips to true when a terminating instruction is seen. So that if additional
+            // instructions occur an error may be returned.
+            let mut terminated = false;
+            for inst in func.ebb_insts(ebb) {
+                if terminated {
+                    return Err(format!("{} contains unreachable instructions.", ebb));
```
stoklund: This is correct, but I think we should leave this check to the verifier. It doesn't actually affect whether we can compute a control flow graph or not.

stoklund:
This looks good to me!

Please go ahead and rebase and merge after changing the `Result`s to assertions.

PR #7, opened 2016-07-13
Add a print-cfg subcommand.

mrrrgn:
The command returns parses a .cton file, builds a CFG, and prints it to
stdout in graphviz format.

### DIFF
--- a/src/tools/main.rs
+++ b/src/tools/main.rs
@@ -11,12 +11,14 @@ use std::process;
 
 
 mod cat;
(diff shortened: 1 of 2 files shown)


stoklund:
Nice!

How about using this to write some tests for the CFG module à la `tests/parser`?

mrrrgn:
"How about using this to write some tests for the CFG module à la tests/parser?"

On it! :]

PR #14, opened 2016-10-28
Use `items` instead of `iteritems` for Python 3 compatibility

swgillespie:
Ran into this while building with my (somewhat unusual) configuration of having `python` refer to `python3` on my machine. `iteritems` was renamed to `items` in Python 3 (and `items` already exists in Python 2) so this code works on both versions (if being slightly more inefficient on Python 2).


stoklund:
Thanks. I try to keep the Python code compatible with both 2.7 and 3.x, but I don't regularly test with Python 3. I try to catch these issues in the `lib/cretonne/meta/check.sh` script.

If you know of any pylint settings that can help with this, patches welcome!

Issue #13, opened 2016-09-21
Conditional compilation of ISA targets


stoklund:
By default, Cretonne is a cross compiler with all supported target ISAs compiled in. That is overkill in many use cases (JITs) that only need to generate native machine code, and it causes some bloat in the code size of the Cretonne library.

We should add Cargo configuration settings to control which target ISAs are compiled into the Cretonne library. Two settings make sense:
1. Include all target ISAs to get a full cross compiler.
2. Include only the native ISA that Cretonne is being built for.

The `build.rs` build script should interpret the setting and select the set of ISAs to build using the `TARGET` environment variable set by Cargo. Then emit `cargo:rustc-cfg=...` lines for the compiler.

The `isa::lookup()` function needs to return two different errors: "I've never heard of that ISA" and "I know that ISA, but it's not been compiled in". This distinction is important for those file tests that request an unsupported ISA. These tests should be skipped silently. On the other hand, tests that request an unknown ISA should always fail.

PR #32, opened 2016-12-04
conditional compilation of ISA targets

dinfuehr:
Cretonne does not need to compile all available ISAs. This is useful for
e.g. JIT compilers, which only need to support the native ISA.

Targets can be specified with the environment variable `CRETONNE_TARGETS`.
There are three supported values:

1. `all` - compile all ISAs
2. `native` - compile only current ISA
3.  non-empty comma-separated list of ISAs

implements #13

### DIFF
--- a/lib/cretonne/build.rs
+++ b/lib/cretonne/build.rs
@@ -18,6 +18,8 @@ use std::process;
 fn main() {
     let out_dir = env::var("OUT_DIR").expect("The OUT_DIR environment variable must be set");
 
+    output_target_cfgs();
+
     println!("Build script generating files in {}", out_dir);
 
     let cur_dir = env::current_dir().expect("Can't access current working directory");
@@ -44,3 +46,55 @@ fn main() {
         process::exit(status.code().unwrap());
     }
 }
+
+fn output_target_cfgs() {
+    let config = env::var("CRETONNE_TARGETS").unwrap_or("all".into());
+    let possible_isas = ["riscv", "intel", "arm32", "arm64"];
+
+    match config.as_ref() {
+        "all" => {
+            for isa in &possible_isas {
+                println!("cargo:rustc-cfg=build_{}", isa);
+            }
+        }
+
+        "native" => {
+            let target = env::var("TARGET").expect("The TARGET environment variable must be set");
+            let arch = target.split("-").next().unwrap();
+
+            let isa = match arch {
+                "aarch64" => "arm64",
+                "arm" | "armv7" | "armv7s" => "arm32",
+                "x86_64" | "i386" | "i586" | "i686" => "intel",
+                "riscv" => "riscv",
+                _ => {
+                    println!("no supported isa found for target triple `{}`", target);
+                    process::exit(1);
+                }
+            };
+
+            println!("cargo:rustc-cfg=build_{}", isa);
+        }
+
+        _ => {
+            let mut found = false;
+
+            for isa in config.split(',') {
+                if possible_isas.contains(&isa) {
+                    println!("cargo:rustc-cfg=build_{}", isa);
+                    found = true;
+
+                } else {
+                    println!("unknown isa `{}`", isa);
+                    process::exit(1);
+                }
+            }
+
+            if !found {
+                println!("no supported isa found in `{}`", config);
+                process::exit(1);
+            }
+        }
+    };
+}
+
--- a/lib/cretonne/src/isa/mod.rs
+++ b/lib/cretonne/src/isa/mod.rs
@@ -27,10 +27,10 @@
 //! let shared_flags = settings::Flags::new(&shared_builder);
 //!
 //! match isa::lookup("riscv") {
-//!     None => {
+//!     Err(_) => {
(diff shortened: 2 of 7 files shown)


stoklund:
Thanks for working on this! I'll review this later today. 

Would it make sense to control this with cargo features instead of an environment variable? I imagine somebody using the Cretonne crate would just want to request a "native" feature in their Cargo.toml.

dinfuehr:
Sure, sounds definitely the better solution. I will look into it.

lib/cretonne/build.rs
```diff
@@ -44,3 +46,55 @@ fn main() {
...
+            }
+        }
+
+        "native" => {
+            let target = env::var("TARGET").expect("The TARGET environment variable must be set");
+            let arch = target.split("-").next().unwrap();
+
+            let isa = match arch {
+                "aarch64" => "arm64",
+                "arm" | "armv7" | "armv7s" => "arm32",
```
stoklund: Maybe we should follow [Clang's triple parser](http://www.llvm.org/docs/doxygen/html/Triple_8cpp_source.html#l00335) here? There's a number of ARM variants, so it just defers to `starts_with("arm")` (or thumb).

lib/cretonne/build.rs
```diff
@@ -44,3 +46,55 @@ fn main() {
...
+                "x86_64" | "i386" | "i586" | "i686" => "intel",
+                "riscv" => "riscv",
+                _ => {
+                    println!("no supported isa found for target triple `{}`", target);
+                    process::exit(1);
```
stoklund: Did you check that this message actually appears on screen? Cargo normally swallows the build script output. Should it be written to stderr instead?

lib/cretonne/build.rs
```diff
@@ -44,3 +46,55 @@ fn main() {
...
+            let mut found = false;
+
+            for isa in config.split(',') {
+                if possible_isas.contains(&isa) {
+                    println!("cargo:rustc-cfg=build_{}", isa);
+                    found = true;
+
+                } else {
+                    println!("unknown isa `{}`", isa);
+                    process::exit(1);
```
stoklund: Maybe we should have a `fn fail(msg: String) -> !` function?

lib/cretonne/src/isa/mod.rs
```diff
@@ -45,41 +45,83 @@ pub use isa::registers::{RegUnit, RegBank, RegInfo};
 use settings;
 use ir::{InstructionData, DataFlowGraph};
 
+#[cfg(build_riscv)]
 pub mod riscv;
```
stoklund: Nice. This is exactly what I had in mind.

lib/cretonne/src/isa/mod.rs
```diff
@@ -45,41 +45,83 @@ pub use isa::registers::{RegUnit, RegBank, RegInfo};
...
-    Some(riscv::isa_builder())
+#[cfg(build_riscv)]
+fn riscv_builder() -> Result<Builder, LookupError> {
+    Ok(riscv::isa_builder())
+}
+
+#[cfg(not(build_riscv))]
+fn riscv_builder() -> Result<Builder, LookupError>{
+    Err(LookupError::Unsupported)
+}
```
stoklund: Yep. This was the reason I split out these little per-target functions. If you can think of a better way of implementing this, I'd be very interested.

dinfuehr: Hmm, if you want we could create all those functions at compile-time with Python.

dinfuehr:
Thanks for your feedback! I will try to address your points over the next few days.

lib/cretonne/src/isa/mod.rs
stoklund: That's probably not worth it. It's not that often we add a new target ISA, so it's OK if you have to write some boilerplate.

Let's just keep your functions. It's not too bad.

lib/reader/src/error.rs
```diff
@@ -19,6 +19,8 @@ pub struct Error {
     pub location: Location,
     /// Error message.
     pub message: String,
+    /// Error because of unsupported ISA
+    pub unsupported: bool,
```
stoklund: Hmm. I think it shouldn't be an error if a test file contains an unsupported ISA.

A test file [can contain multiple ISA specs](http://cretonne.readthedocs.io/en/latest/testing.html#file-tests) and it is possible that some of them are supported while others are unsupported. In that case, we should run the test on the supported ISAs and ignore the unsupported ISAs.

I think the `IsaSpec` enum can represent everything we need:

- The file has no ISA specs: Use `IsaSpec::None(..)`.
- The file has ISA specs, but none are supported: Use `IsaSpec::Some([])`.
- The file has ISA specs, not all supported: Use `IsaSpec::Some([a, b, ...])` with just the supported ISAs. Any unsupported ISA specs are ignored and omitted.

It should still be a parser error if we see an *unknown* ISA.

stoklund:
I'm going to close this since @zummenix has taken over in #126, based in this work.

Thanks for getting us this far, @dinfuehr !

PR #36, opened 2017-01-17
Parse value alias

### LINKED ISSUE #23: Parse value aliases
The `write_value_aliases()` function will print out any value aliases used by an instruction on lines before the instruction itself:

```
vx3 -> v4
v1 = iadd_imm vx3, 17
```

The parser in `lib/reader` should be taught to handle these aliases.

Since the parser is used mostly for test cases, value aliases in the input should be translated to value aliases in the in-memory IR. We don't want to resolve aliases in the parser, even though that would be easy.

- Create value aliases as they are encountered in the source.
- The same alias may appear multiple times. Verify that the instances are consistent.
- A value alias may be created before the source value has been defined, so all of the aliases need to be rewritten, just like other value and EBB references.

KiChjang:
This is a WIP PR, making this for some feedback about whether the general direction is correct.
Things that still need to be done:
* Tests to ensure parsing value aliases work;
* Verification that repeated value aliasing statements are aliasing to the same value;
* Rewriting aliased values as they are defined in the source code

Fixes #23.

### DIFF
--- a/lib/cretonne/src/ir/dfg.rs
+++ b/lib/cretonne/src/ir/dfg.rs
@@ -234,6 +234,19 @@ impl DataFlowGraph {
             panic!("Cannot change direct value {} into an alias", dest);
         }
     }
+
+    /// Create a new value alias.
+    ///
+    /// Note that this function should only be called by the parser.
+    pub fn make_value_alias(&mut self, src: Value) -> Value {
+        let ty = self.value_type(src);
+
+        let data = ValueData::Alias {
+            ty: ty,
+            original: src,
+        };
+        self.make_value(data)
+    }
 }
 
 /// Where did a value come from?
--- a/lib/reader/src/parser.rs
+++ b/lib/reader/src/parser.rs
@@ -188,6 +188,24 @@ impl<'a> Context<'a> {
     // numbering. These references need to be rewritten after parsing is complete since forward
     // references are allowed.
     fn rewrite_references(&mut self) -> Result<()> {
+        for (&source_from, &(source_to, source_loc)) in &self.aliases {
+            let ir_to = match self.map.get_value(source_to) {
+                Some(v) => v,
+                None => {
+                    return err!(source_loc,
+                                "IR destination value alias not found for {}",
+                                source_to);
+                }
+            };
+            let dest_loc = self.map
+                .location(AnyEntity::from(ir_to))
+                .expect(&*format!("Error in looking up location of IR destination value alias \
+                                   for {}",
+                                  ir_to));
+            let ir_from = self.function.dfg.make_value_alias(ir_to);
+            self.map.def_value(source_from, ir_from, &dest_loc)?;
+        }
+
         for ebb in self.function.layout.ebbs() {
             for inst in self.function.layout.ebb_insts(ebb) {
                 let loc = inst.into();
--- a/lib/reader/src/parser.rs
+++ b/lib/reader/src/parser.rs
@@ -5,6 +5,7 @@
 //
 // ====--------------------------------------------------------------------------------------====//
 
+use std::collections::HashMap;
 use std::str::FromStr;
 use std::{u16, u32};
 use std::mem;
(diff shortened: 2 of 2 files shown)


lib/cretonne/src/ir/dfg.rs
```diff
@@ -218,6 +218,33 @@ impl DataFlowGraph {
             panic!("Cannot change direct value {} into an alias", dest);
         }
     }
+
+    /// Create a new value alias.
+    pub fn make_value_alias(&mut self, dest: Value, src: Value) -> Value {
```
stoklund: This function should not take a `dest` argument. That is the newly created value it will be returning.

This also means that it condenses down to calling `value_type` and `make_value`.

I think it would be a good idea to mention in the function's doc comment that it only is used by the parser. There shouldn't be another reason for creating aliases during compilation. (Other than `change_to_alias`).

KiChjang:
Whew! Sorry it took so long. I've incorporated feedback from your comments into my code. Now what's left is adding a unit test and a file test... will get to that soon.

KiChjang:
Hmm... so I'm running into issues about type inference. Consider the example code you gave in the linked issue:
```
function qux() {
ebb0:
  v4 = iconst.i8 6
  vx3 -> v4
  v1 = iadd_imm vx3, 17
}
```
The `iadd_imm` instruction needs to know what the type of `vx3` is, but since we parse instructions during recursive descent and resolve source-to-IR mappings at a later pass, we do not have enough information about what the type of `vx3` is during instruction parsing.

I think rustc has the same problem and they use the concept of type variables to represent an unknown type, which would be resolved at a later pass. In other words, we'll need to have another pass that specifically resolves type variables, if we are to follow and copy what rustc does.

lib/reader/src/parser.rs
```diff
@@ -158,6 +162,14 @@ impl<'a> Context<'a> {
         self.map.def_ebb(src_ebb, ebb, loc).and(Ok(ebb))
     }
 
+    fn add_alias(&mut self, src: Value, dest: Value, loc: &Location) -> Result<()> {
+        if self.aliases.insert(src, dest).is_some() {
+            err!(loc, "duplicate alias: {} -> {}", src, dest)
```
stoklund: Aliases are printed immediately before an instruction that uses them, so it is likely that there will be duplicates. We should only complain if there are *inconsistent* duplicates. That is, definitions of the same alias pointing to different values.

lib/reader/src/parser.rs
```diff
@@ -158,10 +162,31 @@ impl<'a> Context<'a> {
...
+    }
+
     // The parser creates all instructions with Ebb and Value references using the source file
     // numbering. These references need to be rewritten after parsing is complete since forward
     // references are allowed.
     fn rewrite_references(&mut self) -> Result<()> {
+        for (&source_from, &source_to) in &self.aliases {
+            let ir_to = self.map
+                .get_value(source_to)
+                .expect(&*format!("IR destination value alias not found for {}", source_to));
```
stoklund: Could this be reported as a parser error instead of a panic?

stoklund:
Good to have you back!

About the issues with type inference. I wrote some more about it in #63.

In general, the Cretonne parser doesn't try to be too fancy about type inference. It's a language for writing test cases, so it is usually best to be as explicit as possible. In practice, the parser can only infer types from values that it has already seen defined earlier in the text.

I think this principle can be extended to aliases too:

```
  v4 = iconst.i8 6
  vx3 -> v4
  v1 = iadd_imm vx3, 17
```

When looking for the type of `vx3`, `infer_typevar()` finds that it is not present in the source map and fails. This check could be extended to alternatively check if the source value is an alias, and then check if the aliased value has a known type. If the aliased value has not been defined yet, just fail the same way.

The idea is "best effort, not too fancy".

lib/reader/src/parser.rs
KiChjang: I've considered it, but we need the source location in order to report it as a parser error. I suppose I can add those information into the alias map, but I'm not sure if that would be the right way to go.

stoklund: I see what you mean.

Source locations for aliases are a bit tricky since the same alias can be defined in many places (as long as it's consistent). I think a source location that simply represents the first time you saw the alias would be fine, though.

Storing the first-seen source locations in the hash map makes sense to me.

stoklund:
About the Travis CI failure: Cretonne is currently checking against rustfmt 0.7.1. I see that 0.8.0 is out, so I'll upgrade it now.

KiChjang:
Alright! This should be ready to go. I'm not sure if it needs a file test though...

stoklund:
Awesome, thanks!

For the parser, it is usually easier to write file tests than unit tests when testing more complicated things. You can't test parser errors that way, though.

PR #42, opened 2017-02-14
Coalesce live range intervals in adjacent EBBs

### LINKED ISSUE #37: Live range compression by coalescing intervals
*Note: This is not the classical copy coalescing compiler optimization*.

The `LiveRange` struct represents the live range of an SSA value as a list of intervals, one for each EBB where the value is live-in. In many cases, a value will be live through consecutive EBBs, and we can represent the live range more compactly in that case by coalescing the intervals for adjacent EBBs.

# Example

```
ebb0:
   v1 = foo    ; inst1
   jump ebb1   ; inst9

ebb1:
   jump ebb2  ; inst19

ebb2:
   jump ebb3  ; inst29

ebb3:
   use v1     ; inst31
   return     ; inst39
```

Here, v1 has the def-interval `inst1-inst9` and the live-in intervals `ebb1-inst19, ebb2-inst29, ebb3-inst31`, and that is how the intervals are currently stored in `LiveRange`. Since there is nothing interesting going on between adjacent EBBs, this could be represented with a single live-in interval `ebb1-inst31`. The def-interval stays the same.

# Implementation

Coalescing two live-in intervals is only valid when they represent adjacent EBBs and the earlier interval extends all the way to the terminator of its EBB. We need a mechanism for recognizing this inter-EBB gap. The `ProgramOrder` trait is the right place for that functionality:

```rust
trait ProgramOrder {
    ...
    /// Is the range from `inst` to `ebb` just the gap between consecutive EBBs?
    ///
    /// This returns true if `inst` is the terminator in the EBB immediately before `ebb`.
    fn is_ebb_gap(&self, inst: Inst, ebb: Ebb) -> bool;
}
```

The actual coalescing happens in `LiveRange::extend_in_ebb()` when a new live-in interval is added. There is four possibilities:

1. The new interval can be coalesced with both the previous *and* the next intervals, causing the `liveins` vector to shrink by 1.
2. The new interval can be coalesced with the previous interval only. 
3. The new interval can be coalesced with the next interval only.
4.  The new interval can't be coalesced. The `liveins` vector grows by one.

Note that in cases 2. and 3., the elements in the `liveins` vector don't need to be shifted. The existing interval can be modified in place.

cyndis:
LiveRanges represent the live-in range of a value as a sorted
list of intervals. Each interval starts at an EBB and continues
to an instruction. Before this commit, the LiveRange would store
an interval for each EBB. This commit changes the representation
such that intervals continuing from one EBB to another are coalesced
into one.

Fixes #37.

### DIFF
--- a/lib/cretonne/src/ir/layout.rs
+++ b/lib/cretonne/src/ir/layout.rs
@@ -102,6 +102,13 @@ impl ProgramOrder for Layout {
         let b_seq = self.seq(b);
         a_seq.cmp(&b_seq)
     }
+
+    fn is_ebb_gap(&self, inst: Inst, ebb: Ebb) -> bool {
+        let i = &self.insts[inst];
+        let e = &self.ebbs[ebb];
+
+        i.next.is_none() && i.ebb == e.prev
+    }
 }
 
 // Private methods for dealing with sequence numbers.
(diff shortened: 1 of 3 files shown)


cyndis:
Thanks :)

Issue #24, opened 2016-11-05
Parse encodings


stoklund:
The `write_instruction()` function will print out the encoding of an instruction if it has been set:

```
[R#0c]        v5 = iadd v1, v2
```

The parser should decode these annotations and fill out the `Function::encodings` table.

Since the encoding recipe (`R`) is ISA-dependent, we probably can't support encodings in a test file with multiple ISAs.

See the `DisplayEncoding` struct for details of the format.

angusholder:
I've been reading the codebase and think I'd be able to tackle this

angusholder:
So the recipe system is currently only used with the RISC-V backend, am I right in thinking all the other backends are very much incomplete at this point?

angusholder:
This encoding would surely have to be variable length? One ISA instruction doesn't necessarily correspond to one Cretonne opcode, eg RISC-V doesn't have IaddCarry so you need a branch and second add.

stoklund:
Yes, at the moment only RISC-V has any encodings. The other ISAs are very incomplete. RISC-V is also very incomplete, just a little bit less.

The encoding that gets printed out like `[R#0c]` is a representation of an `Encoding` as defined in `isa/encoding.rs` It consists of a *recipe* and some *bits*:

```rust
pub struct Encoding {
    recipe: u16,
    bits: u16,
}
```

The *recipe* is printed out as a name, `R` in this case. The bits are printed in hexadecimal as the `#0c` part. The recipe names depend on the ISA, so they are represented as a `u16` that indexes into the slice returned by `TargetIsa::recipe_names()`. (Also `TargetIsa::recipe_constraints()`, but that is only for the register allocator. These are the current RISC-V recipe names (from the generated `encoding-riscv.rs`):

```rust
pub static RECIPE_NAMES: [&'static str; 4] = [
    "R",
    "I",
    "Rshamt",
    "Iret",
];
```

The `Encoding` values assigned to instructions cont correspond directly to the final machine code encoding. Instead, each recipe maps to a function that can emit the machine code for an instruction given the following:

- The `InstructionData` object with the `Opcode` and values of immediate operands.
- The bits from the `Encoding`.
- The `ValueLoc` (usually a physical register) that was assigned to each input operand and the results.

It is the job of the legalizer to make sure that every Cretonne instruction in use maps to a single ISA opcode. For example, `IaddCarry`, will be expanded on RISC-V using this transformation:

```python
expand.legalize(
[...]

Issue #10, opened 2016-07-29
Incremental CFG updates


stoklund:
We need a mechanism for updating the control flow graph when the code is changed. An example is an architecture that doesn't have a `cmov` instruction:

```
ebb0:
    v10 = select v1, v2, v3
    v11 = iadd_imm v10, 5
```

Legalizes into:

```
ebb0:
    brnz v1, ebb1(v2)
    jump ebb1(v3)

ebb1(vx0: i32):
    v11 = iadd_imm vx0, 5
```

This would require updating the CFG since `ebb1` is a whole new EBB which inherits some of `ebb0`'s successors.

There are three ways of updating the CFG after a program change:
1. Recompute the CFG from scratch. We can do that now, but it is quite expensive in large functions.
2. Invalidate and recompute the part of the CFG that is affected by a single EBB.
3. Invalidate and recompute the part of the CFG that is affected by a single instruction.

Instruction-level CFG repair is potentially difficult. In the example above, `ebb0` can have arbitrarily many CFG edges after the `select` instruction that is converted. These edges would all need to be moved to `ebb1` after the change. It would not be much faster than recomputing all of the edges connecting to `ebb0` and `ebb1`.

We should provide the second option as a CFG function for recomputing CFG edges affecting a single EBB. After performing the program change above, it should be possible to repair the CFG with code like this:

``` rust
func.cfg.recompute_ebb(ebb0);
func.cfg.recompute_ebb(ebb1);
```

The call `recompute_ebb(ebb0)` should assume that `ebb0` has changed, so repair the CFG in these steps:
1. Go through the cached successor list and remove `ebb0` as a predecessor of each successor block. Don't depend on the instructions currently in `ebb0` to find edges. The instructions may have changed.
2. Clear the successor list for `ebb0`.
3. Scan the instructions now in `ebb0` and add new CFG edges as they are discovered.

This algorithm recomputes all CFG edges _from_ `ebb0` while leaving edges _to_ `ebb0` intact.

stoklund:
Fixed by @angusholder!

PR #45, opened 2017-02-22
Remove Opcode::NotAnOpcode and replace with use of Option where applicable

angusholder:


lib/cretonne/src/ir/instructions.rs
```diff
@@ -43,12 +43,8 @@ impl Display for Opcode {
...
 impl Opcode {
     /// Get the instruction format for this opcode.
-    pub fn format(self) -> Option<InstructionFormat> {
-        if self == Opcode::NotAnOpcode {
-            None
-        } else {
-            Some(OPCODE_FORMAT[self as usize - 1])
-        }
+    pub fn format(self) -> InstructionFormat {
+        OPCODE_FORMAT[self as usize - 1]
```
stoklund: Nice. I didn't like having an `Option` there.

lib/cretonne/src/ir/instructions.rs
```diff
@@ -69,23 +65,21 @@ impl FromStr for Opcode {
...
                 self.len()
             }
 
             fn key(&self, idx: usize) -> Option<&'a str> {
-                if self[idx] == Opcode::NotAnOpcode {
-                    None
-                } else {
-                    Some(opcode_name(self[idx]))
-                }
+                self[idx].map(opcode_name)
```
stoklund: Awesome.

lib/cretonne/meta/gen_instr.py
```diff
@@ -269,7 +273,13 @@ def gen_opcodes(groups, fmt):
                                 'Type inferred from {}.'
                                 .format(i.ins[i.format.typevar_operand]))
                 # Enum variant itself.
-                fmt.line(i.camel_name + ',')
+                if is_first_opcode:
+                    fmt.doc_comment('We explicitly set this to 1 to allow the NonZero optimization,')
+                    fmt.doc_comment('meaning that SIZEOF(Option<Opcode>) == SIZEOF(Opcode)')
```
stoklund: It looks like `flake8` is complaining that these two lines are longer than 79 characters. That's why the Travis build is failing.

I actually think it would be OK to just delete them (or change them into a Python comment) since you already explain what's going on in the block comment above. The non-zero details don't need to go into the opcode documentation we're generating here.

lib/cretonne/src/ir/instructions.rs
```diff
@@ -753,6 +755,12 @@ mod tests {
         assert_eq!("iadd\0".parse::<Opcode>(), Err("Unknown opcode"));
         assert_eq!("".parse::<Opcode>(), Err("Unknown opcode"));
         assert_eq!("\0".parse::<Opcode>(), Err("Unknown opcode"));
+
+        // Opcode is a single byte, and because Option<Opcode> originally came to 2 bytes, early on
+        // Opcode included a variant NotAnOpcode to avoid the unnecessary bloat. Since then the Rust
+        // compiler has brought in NonZero optimization, meaning that an enum not using the 0 value
+        // can be optional for no size cost. We want to ensure Option<Opcode> remains small.
+        assert_eq!(mem::size_of::<Opcode>(), mem::size_of::<Option<Opcode>>());
```
stoklund: Thanks. I hope the non-zero optimization has made it into the stable compiler.

angusholder:
Great, thanks for the feedback, you have a nice codebase!

lib/cretonne/src/ir/instructions.rs
angusholder: Apparently it landed in 1.14, and the tests all pass so looks good

stoklund:
Awesome. Thanks for fixing this!

If you are interested in helping out with something more challenging, #24 and #10 are possibilities. I'll be happy to help clarify things if necessary.

lib/cretonne/meta/gen_instr.py
```diff
@@ -254,9 +254,13 @@ def gen_opcodes(groups, fmt):
     fmt.doc_comment('All instructions from all supported ISAs are present.')
     fmt.line('#[derive(Copy, Clone, PartialEq, Eq, Debug)]')
     instrs = []
+
+    # We explicitly set the discriminant of the first variant to 1, which
+    # allows us to take advantage of the NonZero optimization, meaning that
+    # wrapping enums can use the 0 discriminant instead of increasing the size
+    # if the whole type, and so SIZEOF(Option<Opcode>>) == SIZEOF(Opcode)
```
bjorn3: s/if/of

PR #48, opened 2017-02-23
Lexer can now scan names, hex sequences, brackets and minus signs.

angusholder:
#24 
Not complete yet, I want to add some tests, just putting this here so you can see what it looks like so far.

### DIFF
--- a/lib/reader/src/lexer.rs
+++ b/lib/reader/src/lexer.rs
@@ -230,7 +230,7 @@ impl<'a> Lexer<'a> {
             self.next_ch();
 
             if let Some(c) = self.lookahead {
-                // If the next character won't parse as a number, we conservatively return Token::Minus
+                // If the next character won't parse as a number, we return Token::Minus
                 if !c.is_alphanumeric() && c != '.' {
                     return token(Token::Minus, loc);
                 }
--- a/lib/reader/src/lexer.rs
+++ b/lib/reader/src/lexer.rs
@@ -22,6 +22,9 @@ pub enum Token<'a> {
     RPar, // ')'
     LBrace, // '{'
     RBrace, // '}'
+    LBracket, // '['
+    RBracket, // ']'
+    Minus, // '-'
     Comma, // ','
     Dot, // '.'
     Colon, // ':'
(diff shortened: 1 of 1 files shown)


lib/reader/src/lexer.rs
```diff
@@ -326,6 +338,39 @@ impl<'a> Lexer<'a> {
...
     }
 
+    fn scan_name(&mut self) -> Result<LocatedToken<'a>, LocatedError> {
+        let loc = self.loc();
+        let begin = self.pos;
+
+        assert!(self.lookahead == Some('%'));
+
+        while let Some(c) = self.next_ch() {
+            if !c.is_alphanumeric() && c != '_' {
```
stoklund: I think in this specific case, we should restrict to ASCII characters only (i.e., reject other Unicode chars). The `FuncName` printer will do the same.

It's not that important if Cretonne accepts Unicode names or not. The `.cton` files are just for writing tests.

lib/reader/src/lexer.rs
```diff
@@ -326,6 +338,39 @@ impl<'a> Lexer<'a> {
...
+
+    fn scan_hex_sequence(&mut self) -> Result<LocatedToken<'a>, LocatedError> {
+        let loc = self.loc();
+        let begin = self.pos;
+
+        assert!(self.lookahead == Some('#'));
+
+        while let Some(c) = self.next_ch() {
+            match c {
+                'a'...'f' | 'A'...'F' | '0'...'9' => {},
```
stoklund: See `char::is_digit()`.

stoklund:
Looking good so far.

lib/reader/src/lexer.rs
angusholder: I can't seem to find any stdlib functions to check for ascii-only alphabetic characters, you don't know if there are any do you?

stoklund: There's an example in `ir/funcname.rs`.

angusholder:
I'm fairly happy with the lexer as-is now, do you want me to make separate pull request for the parser work then?

stoklund:
> I'm fairly happy with the lexer as-is now, do you want me to make separate pull request for the parser work then?
> 
Thanks, I'll review it in the morning.

Go ahead and start a new PR for the parser.

PR #50, opened 2017-02-26
Converts all try! macros to ? syntax.

### LINKED ISSUE #46: Replace `try!` with `?`
Recently, the stable Rust compiler supports using `?` in place of the `Try!` macro. We should switch.

This mostly affects the parser and `write.rs`, but there are scattered uses elsewhere.

rep-nop:
Fixes #46

### DIFF
--- a/lib/cretonne/meta/gen_settings.py
+++ b/lib/cretonne/meta/gen_settings.py
@@ -171,13 +171,13 @@ def gen_display(sgrp, fmt):
         with fmt.indented(
                 'fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {',
                 '}'):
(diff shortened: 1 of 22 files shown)


stoklund:
Thanks for working on this. If you `cargo install rustfmt` locally, you can check your formatting before pushing.

rep-nop:
@stoklund Yep, thanks. I realized I had forgotten to run `cargo fmt` on the project. Also had fun trying to get the format-all.sh running, as I'm on Windows lol. Manually `cargo fmt`'d them and fixed the couple of errors that were because currently the `?` operator doesn't propagate `Result<()>` from functions that don't return a value.

PR #53, opened 2017-03-03
Added parsing of instruction encodings and result registers specification

angusholder:
Currently only using result_registers to verify that the same number of result registers are specified as there are results, and then throwing it away. I can see they should be parsed with TargetIsa::register_info().parse_regunit() but where then should they be stored?

I'm not sure about adding find_recipe_index to TargetIsa, it seems like something that implementations would want to implement as a hashtable lookup if there are lots of recipes, so maybe is suitable in the trait?

Can we have encodings without registers and registers without encodings? Which of the following definitions would be correct?
	encoding ::= "[" (encoding_literal | "-") , (result_registers | "-") "]"
or
	encoding ::= "[" encoding_literal, result_registers "]" |
	             "[-,-]"

### DIFF
--- a/lib/reader/src/parser.rs
+++ b/lib/reader/src/parser.rs
@@ -44,16 +44,7 @@ pub fn parse_test<'a>(text: &'a str) -> Result<TestFile<'a>> {
     let commands = parser.parse_test_commands();
     let isa_spec = parser.parse_isa_specs()?;
     let preamble_comments = parser.take_comments();
-
-    let functions = {
-        let mut unique_isa = None;
-        if let isaspec::IsaSpec::Some(ref isa_vec) = isa_spec {
-            if isa_vec.len() == 1 {
-                unique_isa = Some(&*isa_vec[0]);
-            }
-        }
-        parser.parse_function_list(unique_isa)?
-    };
+    let functions = parser.parse_function_list(isa_spec.unique_isa())?;
 
     Ok(TestFile {
         commands: commands,
@@ -102,7 +93,7 @@ impl<'a> Context<'a> {
         Context {
             function: f,
             map: SourceMap::new(),
-            unique_isa: unique_isa
+            unique_isa: unique_isa,
         }
     }
 
@@ -592,8 +583,9 @@ impl<'a> Parser<'a> {
     /// Parse a list of function definitions.
     ///
     /// This is the top-level parse function matching the whole contents of a file.
-    pub fn parse_function_list(&mut self, unique_isa: Option<&TargetIsa>)
-        -> Result<Vec<(Function, Details<'a>)>> {
+    pub fn parse_function_list(&mut self,
+                               unique_isa: Option<&TargetIsa>)
+                               -> Result<Vec<(Function, Details<'a>)>> {
         let mut list = Vec::new();
         while self.token().is_some() {
             list.push(self.parse_function(unique_isa)?);
@@ -605,7 +597,9 @@ impl<'a> Parser<'a> {
     //
     // function ::= * function-spec "{" preamble function-body "}"
     //
-    fn parse_function(&mut self, unique_isa: Option<&TargetIsa>) -> Result<(Function, Details<'a>)> {
+    fn parse_function(&mut self,
+                      unique_isa: Option<&TargetIsa>)
+                      -> Result<(Function, Details<'a>)> {
         // Begin gathering comments.
         // Make sure we don't include any comments before the `function` keyword.
         self.token();
@@ -916,6 +910,7 @@ impl<'a> Parser<'a> {
         while match self.token() {
             Some(Token::Value(_)) => true,
             Some(Token::Identifier(_)) => true,
+            Some(Token::LBracket) => true,
             _ => false,
         } {
             self.parse_instruction(ctx, ebb)?;
@@ -964,8 +959,9 @@ impl<'a> Parser<'a> {
         ctx.map.def_value(vx, value, &vx_location)
     }
 
-    fn parse_instruction_encoding(&mut self, ctx: &Context)
-        -> Result<(Option<Encoding>, Option<Vec<&'a str>>)> {
+    fn parse_instruction_encoding(&mut self,
+                                  ctx: &Context)
+                                  -> Result<(Option<Encoding>, Option<Vec<&'a str>>)> {
         let (mut encoding, mut result_registers) = (None, None);
 
         // encoding ::= "[" encoding_literal result_registers "]"
@@ -977,6 +973,11 @@ impl<'a> Parser<'a> {
 
                 if let Some(recipe_index) = ctx.find_recipe_index(recipe) {
                     encoding = Some(Encoding::new(recipe_index, bits));
+                } else if ctx.unique_isa.is_some() {
+                    return err!(self.loc, "invalid instruction recipe");
+                } else {
+                    return err!(self.loc,
+                                "provided instruction encoding for unspecified ISA");
                 }
             }
 
@@ -1525,7 +1526,7 @@ mod tests {
                                        ss3 = stack_slot 13
                                        ss1 = stack_slot 1
                                      }")
-            .parse_function()
+            .parse_function(None)
             .unwrap();
         assert_eq!(func.name.to_string(), "foo");
         let mut iter = func.stack_slots.keys();
@@ -1542,7 +1543,7 @@ mod tests {
                                     ss1  = stack_slot 13
                                     ss1  = stack_slot 1
                                 }")
-                       .parse_function()
+                       .parse_function(None)
                        .unwrap_err()
                        .to_string(),
                    "3: duplicate stack slot: ss1");
@@ -1554,7 +1555,7 @@ mod tests {
                                      ebb0:
                                      ebb4(vx3: i32):
                                      }")
-            .parse_function()
+            .parse_function(None)
             .unwrap();
         assert_eq!(func.name.to_string(), "ebbs");
 
@@ -1583,7 +1584,7 @@ mod tests {
                          trap ; Instruction
                          } ; Trailing.
                          ; More trailing.")
-                .parse_function()
+                .parse_function(None)
                 .unwrap();
         assert_eq!(func.name.to_string(), "comment");
         assert_eq!(comments.len(), 8); // no 'before' comment.
--- a/lib/reader/src/parser.rs
+++ b/lib/reader/src/parser.rs
@@ -6,7 +6,7 @@
 // ====--------------------------------------------------------------------------------------====//
 
 use std::str::FromStr;
-use std::u32;
+use std::{u16, u32};
 use std::mem;
 use cretonne::ir::{Function, Ebb, Opcode, Value, Type, FunctionName, StackSlotData, JumpTable,
                    JumpTableData, Signature, ArgumentType, ArgumentExtension, ExtFuncData, SigRef,
@@ -17,7 +17,7 @@ use cretonne::ir::entities::AnyEntity;
 use cretonne::ir::instructions::{InstructionFormat, InstructionData, VariableArgs,
                                  TernaryOverflowData, JumpData, BranchData, CallData,
                                  IndirectCallData, ReturnData, ReturnRegData};
-use cretonne::isa;
+use cretonne::isa::{self, TargetIsa, Encoding};
 use cretonne::settings;
 use testfile::{TestFile, Details, Comment};
 use error::{Location, Error, Result};
@@ -40,11 +40,26 @@ pub fn parse_test<'a>(text: &'a str) -> Result<TestFile<'a>> {
     let mut parser = Parser::new(text);
     // Gather the preamble comments as 'Function'.
     parser.gather_comments(AnyEntity::Function);
+
+    let commands = parser.parse_test_commands();
+    let isa_spec = parser.parse_isa_specs()?;
+    let preamble_comments = parser.take_comments();
+
+    let functions = {
+        let mut unique_isa = None;
+        if let isaspec::IsaSpec::Some(ref isa_vec) = isa_spec {
+            if isa_vec.len() == 1 {
+                unique_isa = Some(&*isa_vec[0]);
+            }
+        }
+        parser.parse_function_list(unique_isa)?
+    };
+
     Ok(TestFile {
-        commands: parser.parse_test_commands(),
-        isa_spec: parser.parse_isa_specs()?,
-        preamble_comments: parser.take_comments(),
-        functions: parser.parse_function_list()?,
+        commands: commands,
+        isa_spec: isa_spec,
+        preamble_comments: preamble_comments,
+        functions: functions,
     })
 }
 
@@ -71,16 +86,35 @@ pub struct Parser<'a> {
 //
 // Many entities like values, stack slots, and function signatures are referenced in the `.cton`
 // file by number. We need to map these numbers to real references.
-struct Context {
+struct Context<'a> {
     function: Function,
     map: SourceMap,
+
+    // Reference to the unique_isa for things like parsing ISA-specific instruction encoding
+    // information. This is only `Some` if exactly one set of `isa` directives were found in the
+    // prologue (it is valid to have directives for multiple different ISAs, but in that case we
+    // couldn't know which ISA the provided encodings are intended for)
+    unique_isa: Option<&'a TargetIsa>,
 }
 
-impl Context {
-    fn new(f: Function) -> Context {
+impl<'a> Context<'a> {
+    fn new(f: Function, unique_isa: Option<&'a TargetIsa>) -> Context<'a> {
         Context {
             function: f,
             map: SourceMap::new(),
+            unique_isa: unique_isa
+        }
+    }
+
+    // Get the index of a recipe name if it exists.
+    fn find_recipe_index(&self, recipe_name: &str) -> Option<u16> {
+        if let Some(unique_isa) = self.unique_isa {
+            unique_isa.recipe_names()
+                .iter()
+                .position(|&name| name == recipe_name)
+                .map(|idx| idx as u16)
+        } else {
+            None
         }
     }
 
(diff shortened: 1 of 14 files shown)


angusholder:
What formatting checks should I do before committing to pass Travis-CI? If I run format-all.sh it makes an avalanche of changes on lots of files I haven't touched, do you run something different normally?

stoklund:
> What formatting checks should I do before committing to pass Travis-CI? If I run format-all.sh it makes an avalanche of changes on lots of files I haven't touched, do you run something different normally?
> 
You probably have a different version of `rustfmt` installed, and unfortunately this tool (and the ‘correct’ formatting) is still evolving.

You should run the `test-all.sh` script, that is what Travis CI does. It currently checks if you have `rustfmt 0.7.1` installed, and uses that. Otherwise it skips the format check.

Once you have the right version of rustfmt, you can use `format-all.sh` too.

lib/cretonne/src/isa/encoding.rs
```diff
@@ -76,3 +77,5 @@ impl fmt::Display for DisplayEncoding {
         }
     }
 }
+
+impl PrimaryEntityData for Encoding {}
```
stoklund: This should not be needed. Use `EntityMap::ensure()` instead. See comment below.

lib/cretonne/src/isa/mod.rs
```diff
@@ -148,6 +148,14 @@ pub trait TargetIsa {
     /// This is just used for printing and parsing encodings in the textual IL format.
     fn recipe_names(&self) -> &'static [&'static str];
 
+    /// Get the index of a recipe name if it exists.
+    fn find_recipe_index(&self, recipe_name: &str) -> Option<u16> {
```
stoklund: I don't think this needs to go in `TargetIsa`. ISAs are not going to have very long lists of recipes, and the parser is not performance critical, so a linear search should be good enough for everybody. (We can always do this later, if it turns out I'm wrong.)

I would prefer this as a function in `parser.rs` instead.

lib/reader/src/parser.rs
```diff
@@ -40,11 +40,32 @@ pub fn parse_test<'a>(text: &'a str) -> Result<TestFile<'a>> {
...
     // Gather the preamble comments as 'Function'.
     parser.gather_comments(AnyEntity::Function);
+
+    let commands = parser.parse_test_commands();
+    let mut isa_spec = parser.parse_isa_specs()?;
+
+    if let isaspec::IsaSpec::Some(ref mut isa_vec) = isa_spec {
+        if isa_vec.len() == 1 {
+            // Remove it from the isa_vec just while we parse functions, avoiding a .clone()
+            parser.target_isa = Some(isa_vec.remove(0));
```
stoklund: The parser can just borrow the `TargetIsa` as a reference instead of owning a box. Then you don't need to move it around.

lib/reader/src/parser.rs
```diff
@@ -65,6 +86,12 @@ pub struct Parser<'a> {
 
     // Comments collected so far.
     comments: Vec<Comment<'a>>,
+
+    // Reference to the target_isa for things like parsing ISA-specific instruction encoding
+    // information. This is only `Some` if exactly one set of `isa` directives were found in the
+    // prologue (it is valid to have directives for multiple different ISAs, but in that case we
+    // couldn't know which ISA the provided encodings are intended for)
+    target_isa: Option<Box<TargetIsa>>,
```
stoklund: Here. Use this instead:

```rust
unique_isa: Option<&TargetIsa>,
```

(I prefer the name `unique_isa` so people won't be surprised when it is missing.)

lib/reader/src/parser.rs
```diff
@@ -903,6 +931,68 @@ impl<'a> Parser<'a> {
...
+        // encoding ::= "[" (encoding_literal | "-") , (result_registers | "-") "]"
+        if self.optional(Token::LBracket) {
+            let mut encoding_literal: Option<(&'a str, u16)> = None;
+
+            // encoding_literal ::= Identifier HexSequence
+            if let Some(Token::Identifier(recipe)) = self.token() {
+                self.consume();
+
+                if let Some(Token::HexSequence(bits_str)) = self.token() {
+                    self.consume();
```
stoklund: I would recommend adding `match_foo()` methods to avoid this error pyramid.

- `match_any_identifier()` is like `match_identifier()`, but allows any identifier text.
- `match_hex_sequence()` for `HexSequence`.

lib/reader/src/parser.rs
```diff
@@ -903,6 +931,68 @@ impl<'a> Parser<'a> {
...
+
+                    // The only error we anticipate from this parse is overflow, the lexer should
+                    // already have ensured that the string doesn't contain invalid characters, and
+                    // isn't empty or negative.
+                    if let Ok(bits) = u16::from_str_radix(bits_str, 16) {
+                        encoding_literal = Some((recipe, bits));
+                    } else {
+                        return err!(self.loc, "the hex sequence given overflows the u16 type");
+                    }
```
stoklund: Maybe even a `match_hex16() -> Result<u16>` method? See the existing `match_uimm8()`.

lib/reader/src/parser.rs
```diff
@@ -903,6 +931,68 @@ impl<'a> Parser<'a> {
...
+                } else {
+                    return err!(self.loc, "expected a hex sequence");
+                }
+            } else {
+                self.match_token(Token::Minus, "expected instruction encoding or '-'")?;
+            }
+
+            // result_registers ::= Name { "," Name }
+            if self.token() == Some(Token::Comma) {
+                self.consume();
```
stoklund: Use `self.optional()`.

lib/reader/src/parser.rs
```diff
@@ -903,6 +931,68 @@ impl<'a> Parser<'a> {
...
+
+                let mut result = Vec::new();
+
+                while let Some(Token::Name(s)) = self.token() {
+                    self.consume();
```
stoklund: Here too. Compare to `parse_argument_list`.

lib/reader/src/parser.rs
```diff
@@ -955,6 +1052,12 @@ impl<'a> Parser<'a> {
         ctx.function.layout.append_inst(inst, ebb);
         ctx.map.def_entity(inst.into(), &opcode_loc).expect("duplicate inst references created");
 
+        if let Some(encoding) = encoding {
+            ctx.function.encodings.push(encoding);
+        } else {
+            ctx.function.encodings.push(Encoding::default());
+        }
```
stoklund: Here, the `encodings` table is a *secondary* entity map. Update it like this:

```rust
*ctx.function.encodings.ensure(inst) = encoding;
```

You don't need the default arm. That's already the default value.

lib/reader/src/parser.rs
```diff
@@ -903,6 +931,68 @@ impl<'a> Parser<'a> {
         // Collect comments for the next instruction to be allocated.
         self.gather_comments(ctx.function.dfg.next_inst());
 
+        let mut encoding: Option<Encoding> = None;
+        let mut result_registers: Option<Vec<&'a str>> = None;
+
+        // encoding ::= "[" (encoding_literal | "-") , (result_registers | "-") "]"
```
stoklund: This is the right grammar. Allow either to be `"-"` independently. We can always verify this more carefully in the verifier.

lib/reader/src/parser.rs
stoklund: So the whole thing goes like this:

```rust
while self.optional(Token::Comma) {
    list.push(self.match_value_location()?);
}
```

lib/reader/src/parser.rs
angusholder: Honestly I did that because I wasn't sure how to make it borrow check correctly with the references. I've just been having another look, even explicitly dropping the parser earlier doesn't help:

`    if let isaspec::IsaSpec::Some(ref mut isa_vec) = isa_spec {
        if isa_vec.len() == 1 {
            parser.unique_isa = Some(&*isa_vec[0]);
        }
    }

    let preamble_comments = parser.take_comments();
    let functions = parser.parse_function_list()?;

    drop(parser);

    Ok(TestFile {
        commands: commands,
        isa_spec: isa_spec,
        preamble_comments: preamble_comments,
        functions: functions,
    })`

`error: `(isa_spec:isaspec::IsaSpec::Some).0` does not live long enough
  --> lib\reader\src\parser.rs:64:1
   |
47 |     if let isaspec::IsaSpec::Some(ref mut isa_vec) = isa_spec {
   |                                   --------------- borrow occurs here
...
64 | }
   | ^ `(isa_spec:isaspec::IsaSpec::Some).0` dropped here while still borrowed
   |
   = note: values in a scope are dropped in the opposite order they are created`

stoklund: Ugh, yeah you would need a whole other lifetime on `Parser`.

Another option would be to pass `unique_isa` as an argument to `parse_function_list()` and `parse_function()`. Then you can stick it in the `Context`.

stoklund:
The rest looks good, BTW

angusholder:
Was just gonna add some tests and wrap it up

lib/reader/src/parser.rs
```diff
@@ -42,24 +42,18 @@ pub fn parse_test<'a>(text: &'a str) -> Result<TestFile<'a>> {
...
 
-    // Put unique_isa back in isa_spec if we took it
-    if let isaspec::IsaSpec::Some(ref mut isa_vec) = isa_spec {
-        if let Some(unique_isa) = parser.unique_isa.take() {
-            isa_vec.insert(0, unique_isa);
+    let functions = {
+        let mut unique_isa = None;
+        if let isaspec::IsaSpec::Some(ref isa_vec) = isa_spec {
+            if isa_vec.len() == 1 {
+                unique_isa = Some(&*isa_vec[0]);
```
stoklund: Very cool. You could turn this whole bit into an `IsaSpec::unique_isa()` method.

stoklund:
An update to mypy exposed a bunch of bugs in my Python code, causing your tests to fail here. I've fixed it on the master branch, so your next push should pass CI.

lib/reader/src/parser.rs
```diff
@@ -1653,4 +1649,39 @@ mod tests {
             }
         }
     }
+
+    #[test]
+    fn instruction_encoding() {
+        use cretonne::ir::Inst;
+        use cretonne::entity_map::EntityRef;
+
+        let test_file = parse_test("isa riscv supports_m
```
stoklund: This is fine, but note that it would probably be easier to write this as a file test under `filetests/parser/`. This would let you round-trip these parser changes fairly simply.

angusholder: So would all I need to do be drop that IR into a file filetests/parser/instruction_encoding.rs?

Is squashing commits into one a GitHub feature? Still pretty new to pull requests

stoklund:
> +        let test_file = parse_test("isa riscv supports_m
> So would all I need to do be drop that IR into a file filetests/parser/instruction_encoding.rs?
> 
Yes, and then add `check:` lines to verify that when the function is printed out again, all the information is there.

See the existing tests in that directory and http://cretonne.readthedocs.io/en/latest/testing.html#test-cat
> Is squashing commits into one a GitHub feature? Still pretty new to pull requests
> 
It’s a git feature. Use `git rebase -i` to squash all of your commits into one locally, then push to your branch, and the pull request will be updated.

You may have to `git push -f` because you’re rewriting history instead of just adding new commits.

lib/reader/src/parser.rs
stoklund:
```rust
       let test_file = parse_test("isa riscv supports_m
```

> So would all I need to do be drop that IR into a file `filetests/parser/instruction_encoding.rs`?

Yes, and then add `check:` lines to verify that when the function is printed out again, all the information is there.

See the existing tests in that directory and http://cretonne.readthedocs.io/en/latest/testing.html#test-cat

> Is squashing commits into one a GitHub feature? Still pretty new to pull requests

It’s a git feature. Use `git rebase -i` to squash all of your commits into one locally, then push to your branch, and the pull request will be updated.

You may have to `git push -f` because you’re rewriting history instead of just adding new commits.

angusholder:
I'm having trouble with filecheck, this is what I have so far:

    test cat
    
    isa riscv
    function foo(i32, i32) {
    ebb1(v0: i32, v1: i32):
        [-,-] v2 = iadd v0, v1
        [-] trap
        [R#1234, %x5, %x11] v6, v7 = iadd_cout v2, v0
        [Rshamt#beef, %x25] v8 = ishl_imm v6, 2
        v9 = iadd v8, v7
        [Iret#5] return v0, v8
    }
    ; sameln: function foo(i32, i32) {
    ; nextln: $ebb1($v0: i32, $v1: i32):
    ; nextln:     [-] $v2 = iadd $v0, $v1
    ; nextln:     [-] trap
    ; nextln:     [R#1234, %x5, %x11] $v6, $v7 = iadd_cout $v2, $v0
    ; nextln:     [Rshamt#beef, %x25] $v8 = ishl_imm $v6, 2
    ; nextln:     [-] $v9 = iadd $v8, $v7
    ; nextln:     [Iret#5] return $v0,$v8
    ; nextln: }

I have played around with it and fixed a few errors, but one I still get is that the line `[-] $v2 = iadd...` doesn't match and I've ran out of ideas why that would be.

stoklund:
Maybe it's a problem with matching whitespace? Try this:

```
; regex: WS=[ \t]*

function foo(i32, i32) {
ebb1(v0: i32, v1: i32):
    [-,-] v2 = iadd v0, v1
...
; sameln: function foo(i32, i32) {
; nextln: $ebb1($v0: i32, $v1: i32):
; nextln:     [-]$WS $v2 = iadd $v0, $v1
; nextln:     [-]$WS trap
```

I thought about making filecheck treat all whitespace as equivalent. I think LLVM's FIleCheck tool does that.

angusholder:
The error becomes even less clear when I do that:

    FAIL filetests/parser/instruction_encoding.cton: cat(foo): filecheck failed:
    #0 regex: WS=[ \t]*
    #1 sameln: function foo(i32, i32) {
    #2 nextln: $(ebb1)($(v0): i32, $(v1): i32):
    #3 nextln: [-]$(WS) $(v2) = iadd $(v0), $(v1)
    #4 nextln: [-]$(WS) trap
    #5 nextln: [R#1234, %x5, %x11] $(v6), $(v7) = iadd_cout $(v2), $(v0)
    #6 nextln: [Rshamt#beef, %x25] $(v8) = ishl_imm $(v6), 2
    #7 nextln: [-] $(v9) = iadd $(v8), $(v7)
    #8 nextln: [Iret#5] return $(v0),$(v8)
    #9 nextln: }
    > function foo(i32, i32) {
      ^~~~~~~~~~~~~~~~~~~~~~~~
    Matched #1: \bfunction foo\(i32, i32\) \{
    >                     ebb0(vx0: i32, vx1: i32):
                          ^~~~~~~~~~~~~~~~~~~~~~~~~
    Matched #2: (?:\bebb0\b)\((?:\bvx0\b): i32, (?:\bvx1\b): i32\):
    > [-]                     v0 = iadd vx0, vx1
    ) (?:\bv0\b) = iadd (?:\bvx0\b), (?:\bvx1\b)
    > [-]                     trap
    > [0#1234]                v2, vx2 = iadd_cout v0, vx0
    > [2#beef]                v3 = ishl_imm v2, 2
    > [-]                     v4 = iadd v3, vx2
    > [3#05]                  return vx0, v3
    > }

    slow: 0.008 filetests/parser/branch.cton
    slow: 0.008 filetests/isa/riscv/abi.cton
    21 tests
    1 failure

stoklund:
Huh. Did GitHub trim some of that error message? Perhaps you could push your progress so far, then I can try to reproduce here.

angusholder:
Nah it's right, one line starts with a ')'

stoklund:
Wow. I think you found a bug in filecheck. That doesn't look right at all. Let me take a look.

stoklund:
Thanks. I can reproduce it here, I get the same message. Here's the output piped into `less`:

```
>                     ebb0(vx0: i32, vx1: i32):
                      ^~~~~~~~~~~~~~~~~~~~~~~~~
Matched #2: (?:\bebb0\b)\((?:\bvx0\b): i32, (?:\bvx1\b): i32\):
> [-]                     v0 = iadd vx0, vx1
Missed #3: \[-\](?:[ \t]*^M) (?:\bv0\b) = iadd (?:\bvx0\b), (?:\bvx1\b)
> [-]                     trap
```

Your `.cton` file uses Windows line endings (CRLF), and that it confusing filecheck, it seems.

angusholder:
Ah brilliant! That's the second time Windows has given me trouble, apparently Bash is totally confounded by CRLF endings in shell scrips which surprised me

stoklund:
Yeah, many Unix tools aren't even trying to handle CRLF. I'll push a fix to filecheck as soon as I write a unit test.

angusholder:
So the next error I encounter is that the text output doesn't contain the ISA register assignments of results, but we aren't saving that information yet, I believe because it isn't a part of TargetISA yet. Instruction recipes also output as an index instead of the string, I remember you saying you needed to change infrastructure for that to happen. Shall I leave them both in and push this as a failing test or?

angusholder:
I definitely get the impression we should loosen up the rules on whitespace, things like `return v0,v1` isn't valid and must be `return v0, v1` which is really subtle.

stoklund:
The register assignments, you can leave for now. We can deal with that in a later commit.

The problem with the encodings is probably that the `test cat` filetest command doesn't anticipate a `TargetIsa`, so it doesn't pass one to `write_function`. I'll fix that.

Perhaps you could file an issue about whitespace in filecheck?

angusholder:
I've squashed all my commits now, do I want to pull your latest changes from master or can you do that when accepting a pull request? If I pull from upstream it always adds another commit as far as I can see

stoklund:
Thanks. I can rebase when I merge your commit.

You can use `git pull --rebase` if you want to do it yourself. It does the equivalent of `git pull` followed by `git rebase`.

stoklund:
I fixed `test cat` in 7a45aeebebf1820564ccfdf615633d1e8800f6f2.

Issue #49, opened 2017-02-24
Write and parse ABI annotations on function signatures


stoklund:
In c8be39fa9d06ccf7fb5765c125c16672cebce992 I added support for ABI annotations on function signatures. These annotations indicate the location (stack or register) of individual function arguments and return values. We need:

- Support for writing the ABI annotations using the correct ISA names for registers.
- Support for parsing the ABI annotations in test files with a single unique ISA. Related to #24.

# Writing ABI annotations

The new syntax for function signatures is partially implemented:

```
function average(i32 [%x10], i32 inreg [%x11]) -> f32 [%f10] {
```

If a signature has ABI annotations, they appear in brackets after each argument or return value. This is used everywhere function signatures appear:

- In the function definition.
- In a `signature` declaration in the preamble.
- In a `function` declaration in the preamble.

Each argument is annotated as follows:

- Registers: `[%x10]` using the `display_regunit()` function provided by the `TargetIsa`. Note that the currently committed implementation is incorrect, it just prints the register unit number `[%10]`. This is a stopgap until we can provide a `&TargetIsa` reference. It is not a syntax we want to support going forward.
- Stack arguments: `[24]`, where the number in brackets is the argument's byte offset in the argument array.
- Unassigned arguments don't get an annotation in the text format.

The `Signature` struct also has an `argument_bytes` field now which indicates the needed size of the argument array. This field does not need to appear in the text format since it can be computed from the arguments.

# Parsing ABI annotations

The parser should accept the annotations above and fill in the `location` field of the `ArgumentType` structs when present.

When parsing a file that doesn't specify exactly one ISA, the ABI annotations should be ignored.

After parsing a signature containing with ABI annotations, call the `Signature::compute_argument_bytes()` method.

stoklund:
I've implemented the display part of this one, with a test case in `filetests/isa/riscv/abi.cton`. Maybe the remaining parser work is for @angusholder ?

angusholder:
I'll get onto this now

stoklund:
Fixed by @angusholder !

PR #55, opened 2017-03-09
Parse ValueLoc for each SSA value result of instructions

angusholder:
Issue #24

Some work left to do, and I am currently trying to work out a bug; it seems I am not currently filling in the `function.locations` `EntityMap` properly, as some tests crash with an out-of-bounds access in `write.rs:189:write_instruction()` when trying to stringify a function.

I also want to do some more cleanup, such as possibly moving the code that fills out `function.locations` at the end of `Parser::parse_instruction` into `Parser::add_values`.

I was also going to look at parsing value locations of function arguments (issue #49), but maybe this should be another pull request.

### DIFF
--- a/lib/reader/src/parser.rs
+++ b/lib/reader/src/parser.rs
@@ -10,7 +10,7 @@ use std::{u16, u32};
 use std::mem;
 use cretonne::ir::{Function, Ebb, Opcode, Value, Type, FunctionName, StackSlotData, JumpTable,
                    JumpTableData, Signature, ArgumentType, ArgumentExtension, ExtFuncData, SigRef,
-                   FuncRef};
+                   FuncRef, ValueLoc};
 use cretonne::ir::types::VOID;
 use cretonne::ir::immediates::{Imm64, Ieee32, Ieee64};
 use cretonne::ir::entities::AnyEntity;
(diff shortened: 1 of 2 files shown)


lib/reader/src/parser.rs
```diff
@@ -959,9 +959,32 @@ impl<'a> Parser<'a> {
...
+                    .ok_or(self.error("invalid register value location"))
+            }
+            _ => err!(self.loc, "invalid value location"),
+        }
+    }
+
     fn parse_instruction_encoding(&mut self,
                                   ctx: &Context)
-                                  -> Result<(Option<Encoding>, Option<Vec<&'a str>>)> {
+                                  -> Result<(Option<Encoding>, Option<Vec<ValueLoc>>)> {
```
stoklund: I wonder if the second return value should simply be `Vec<ValueLoc>`? I don't think there is a difference in meaning between `None` and `Some(vec![])`.

lib/reader/src/parser.rs
```diff
@@ -1086,8 +1104,29 @@ impl<'a> Parser<'a> {
...
+                    // We are safe to unwrap here because all values should have been added to the
+                    // map in the above call to self.add_values()
+                    let value = ctx.map.get_value(src_value).unwrap();
+                    *ctx.function.locations.ensure(value) = loc;
+                }
+            }
+        } else {
+            for &src_value in results.iter() {
+                let value = ctx.map.get_value(src_value).unwrap();
+                *ctx.function.locations.ensure(value) = ValueLoc::Unassigned;
```
stoklund: This should not be necessary since `ValueLoc::Unassigned` is the default anyway.

lib/reader/src/parser.rs
```diff
@@ -959,9 +959,32 @@ impl<'a> Parser<'a> {
...
+                         src_num)
+                }
+            }
+            Some(Token::Name(name)) => {
+                self.consume();
+                reginfo.parse_regunit(name)
+                    .map(ValueLoc::Reg)
+                    .ok_or(self.error("invalid register value location"))
+            }
+            _ => err!(self.loc, "invalid value location"),
```
stoklund: I think you should parse a `-` into `ValueLoc::Unassigned` here.

lib/reader/src/parser.rs
```diff
@@ -1086,8 +1104,29 @@ impl<'a> Parser<'a> {
...
-                        ctx.function.dfg.inst_results(inst))
+                        results.iter().map(|v| *v),
+                        ctx.function.dfg.inst_results(inst))?;
+
+        if let Some(ref result_locations) = result_locations {
+            if result_locations.len() != results.len() {
+                return err!(self.loc,
+                            "must have same number of result locations as results");
+            } else {
+                for (&src_value, &loc) in results.iter().zip(result_locations) {
```
stoklund: Since the real result values are on the instruction now, you can use:

```rust
for (&value, &loc) in ctx.function.dfg.inst_results(inst).zip(result_locations) {
```

Then you don't need the `ctx.map.get_value` lookup below.

lib/reader/src/parser.rs
stoklund: I pushed a fix for your crash in bc1901b766782176c9c2edede466de09387eee15. It is related to what you're doing here: The `write_function` code checks `if !func.locations.is_empty()`, and here we're filling default values into the map so it is no longer empty. It's a subtle difference, is the locations map empty or full of default values.

It affects if `write_function` will write a bunch of default value locations, or nothing at all.

lib/reader/src/parser.rs
angusholder: I think there is a difference, eg you can have a `return` instruction which specifies no locations because it has no results, and you can have an `iadd` instruction that just omits the encoding information for its 1 result. Both of those would have `vec.len() == 0` but should be treated differently.

stoklund: That's true, but this function can't tell the difference, right? It would return `None` in both cases, it can't produce the `Some(vec![])` result.

stoklund:
This looks great, thanks!

PR #57, opened 2017-03-09
Added parsing of argument locations for functions and signatures.

angusholder:
Issue #49

This is pretty much complete, the only thing I wasn't sure about was when to call `Signature::compute_argument_bytes()`, so I went for only calling it if all argument locations were assigned.

### DIFF
--- a/filetests/isa/riscv/abi.cton
+++ b/filetests/isa/riscv/abi.cton
@@ -61,3 +61,36 @@ ebb0(v0: i64x4):
     v1 = iadd v0, v0
     return v0
 }
+
+function parse_encoding(i32 [%x5]) -> i32 [%x10] {
+; check: function parse_encoding(i32 [%x5]) -> i32 [%x10] {
+
+    sig0 = signature(i32 [%x10]) -> i32 [%x10]
+; check: sig0 = signature(i32 [%x10]) -> i32 [%x10]
+
+    sig1 = signature(i32 [%x10], i32 [%x11]) -> b1 [%x10]
+; check: sig1 = signature(i32 [%x10], i32 [%x11]) -> b1 [%x10]
+
+    sig2 = signature(f32 [%f10], i32 [%x12], i32 [%x13]) -> f64 [%f10]
+; check: sig2 = signature(f32 [%f10], i32 [%x12], i32 [%x13]) -> f64 [%f10]
+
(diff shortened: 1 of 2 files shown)


lib/reader/src/parser.rs
```diff
@@ -661,17 +676,21 @@ impl<'a> Parser<'a> {
...
-            sig.argument_types = self.parse_argument_list()?;
+            sig.argument_types = self.parse_argument_list(unique_isa)?;
         }
         self.match_token(Token::RPar, "expected ')' after function arguments")?;
         if self.optional(Token::Arrow) {
-            sig.return_types = self.parse_argument_list()?;
+            sig.return_types = self.parse_argument_list(unique_isa)?;
+        }
+
+        if sig.argument_types.all(ArgumentLoc::is_assigned) {
```
stoklund: This approach makes sense to me. (You want `argument_types.iter().all(...`)

filetests/isa/riscv/abi.cton
```diff
@@ -61,3 +61,31 @@ ebb0(v0: i64x4):
...
+    sig2 = signature(f32 [%f10], i32 [%x12], i32 [%x13]) -> f64 [%f10]
+; check: sig2 = signature(f32 [%f10], i32 [%x12], i32 [%x13]) -> f64 [%f10]
+
+; Arguments on stack where not necessary
+    sig3 = signature(f64 [%f10], i32 [0], i32 [4]) -> f64 [%f10]
+; check: sig3 = signature(f64 [%f10], i32 [0], i32 [4]) -> f64 [%f10]
+
+; Stack argument before register argument
+    sig4 = signature(f32 [72], i32 [%x10])
+; check: sig4 = signature(f32 [72], i32 [%x10])
```
stoklund: These tests are great. I think you should add a test for the short-cut signature+function declaration in one:

```
fn8 = function bar(i32) -> b1
```

This will be printed as a separate signature and function declaration. See `filetests/parser/call.cton`.

lib/reader/src/parser.rs
angusholder: Oh dear, can't believe I forgot to compile a last time before commit, sorry about that

angusholder:
Ok all sorted and squashed into 1 commit

PR #62, opened 2017-03-13
Verifier now checks integrity of instruction's results and references

angusholder:
Issue #3 

This is still a work-in-progress currently, just wanting to know if I'm going about it the right way?

### DIFF
--- a/lib/cretonne/src/ir/dfg.rs
+++ b/lib/cretonne/src/ir/dfg.rs
@@ -82,6 +82,11 @@ impl DataFlowGraph {
     pub fn num_ebbs(&self) -> usize {
         self.ebbs.len()
     }
+
+    /// Returns `true` if the given ebb reference is valid.
+    pub fn ebb_is_valid(&self, ebb: Ebb) -> bool {
+        self.ebbs.is_valid(ebb)
+    }
 }
 
 /// Handling values.
@@ -93,6 +98,14 @@ impl DataFlowGraph {
         let vref = Value::new_table(self.extended_values.len());
         self.extended_values.push(data);
         vref
+    }
+
+    /// Check if a value reference is valid.
+    pub fn value_is_valid(&self, v: Value) -> bool {
+        match v.expand() {
+            ExpandedValue::Direct(inst) => self.insts.is_valid(inst),
+            ExpandedValue::Table(index) => index < self.extended_values.len(),
+        }
     }
 
     /// Get the type of a value.
--- a/lib/cretonne/src/verifier.rs
+++ b/lib/cretonne/src/verifier.rs
@@ -12,17 +12,18 @@
 //!   Instruction integrity
 //!
 //!    - The instruction format must match the opcode.
-//! TODO:
 //!    - All result values must be created for multi-valued instructions.
 //!    - Instructions with no results must have a VOID `first_type()`.
 //!    - All referenced entities must exist. (Values, EBBs, stack slots, ...)
 //!
 //!   SSA form
 //!
+//! TODO:
 //!    - Values must be defined by an instruction that exists and that is inserted in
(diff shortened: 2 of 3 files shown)


lib/cretonne/src/ir/dfg.rs
```diff
@@ -95,6 +95,14 @@ impl DataFlowGraph {
         vref
     }
 
+    /// Check if a value reference is valid.
+    pub fn value_is_valid(&self, v: Value) -> bool {
```
stoklund: Nice, this is a good idea.

I think that in the future, the `extended_values` table will have a way of representing dangling values that are no longer valid. We can catch those here too.

lib/cretonne/src/verifier.rs
```diff
@@ -147,12 +147,109 @@ impl<'a> Verifier<'a> {
...
+        macro_rules! verify_func_ref {
+            ($f:expr) => {
+                if !self.func.dfg.ext_funcs.is_valid(f) {
+                    return err!(inst, "invalid function reference");
+                }
+            };
+        }
+        macro_rules! verify_value_list {
+            ($l:expr) => {
+                if self.func.dfg.value_lists.len_of(l) == None {
```
stoklund: Huh. The `len_of()` method wasn't supposed to be accessible outside the `entity_list` module. I think I would prefer if we could add some kind of `is_valid()` method instead.

angusholder: ah I seem to have missed some files off the last commit, I did actually make `is_valid` in the end.

lib/cretonne/src/verifier.rs
```diff
@@ -185,56 +185,28 @@ impl<'a> Verifier<'a> {
 
     fn verify_entity_references(&self, inst: Inst) -> Result<()> {
         use ir::instructions::InstructionData::*;
+
+        for &arg in self.func.dfg[inst].arguments(&self.func.dfg.value_lists) {
+            self.verify_value(inst, arg)?;
+        }
```
stoklund: Yep. You can do the same thing with `second_result`. But you could also just let `DataFlowGraph::inst_results()` handle that and simply check the values returned by that iterator.

This means that you will get some assertion failures instead of the nicer verifier errors. I think that is OK if the fundamental data structures are corrupt: Linked lists of instruction results or EBB arguments.

lib/cretonne/src/verifier.rs
```diff
@@ -261,12 +232,7 @@ impl<'a> Verifier<'a> {
                 self.verify_sig_ref(inst, sig_ref)?;
                 self.verify_value_list(inst, args)?;
             }
-            // These have no entity references to check. Explicitly listed so we get a compile error
-            // if more variants are ever added.
-            &Nullary { .. } |
-            &UnaryImm { .. } |
-            &UnaryIeee32 { .. } |
-            &UnaryIeee64 { .. } => {}
+            _ => {}
```
stoklund: I think it is a good idea to keep the exhaustive list of instruction formats here and no do a catch-all match. It is pretty common to add new formats, and I keep forgetting what needs to be updated for a new format.

stoklund:
I like this direction.

Are you able to run the `./test-all.sh` script? There's a bunch of file tests failing now.

If you want me to merge some of this, just roll it up into one commit and let me know.

angusholder:
Yeah I'm not totally sure why they're failing, the parser should thoroughly and correctly initialise references right?

stoklund:
> Yeah I'm not totally sure why they're failing, the parser should thoroughly and correctly initialise references right?
> 
Yes, the parser should not be able to trigger the kind of integrity checks you have added here.

angusholder:
Ok I've sorted everything out now, just had a couple of silly logic errors. My commits are squashed and ready to be merged.

PR #64, opened 2017-03-14
Verify that values have a valid reference to who defined them

angusholder:


lib/cretonne/src/verifier.rs
```diff
@@ -284,11 +284,34 @@ impl<'a> Verifier<'a> {
...
+        if !dfg.value_is_valid(v) {
+            return err!(inst, "invalid value reference {}", v);
         }
+
+        // SSA form
+        match dfg.value_def(v) {
+            // Value is defined by an instruction that exists and is inserted in an EBB.
+            ValueDef::Res(inst, _) => {
+                if !dfg.insts.is_valid(inst) {
+                    return err!(v, "value is defined by invalid instruction {}", inst);
```
stoklund: The entity the you pass as the first argument to `err!` works like a source location. When possible, we will even use the parser's source map to convert it to a real line number.

So here it would be better to go:

```rust
            ValueDef::Res(def_inst, _) => {
                if !dfg.insts.is_valid(def_inst) {
                    return err!(inst, "{} is defined by invalid instruction {}", v, inst);
```

We won't be able to map an invalidly defined value to a proper source location.

Other times, using a value as a source location is just fine.

angusholder: Ah that makes sense, amended.

stoklund:
Oh, my bad, I didn't edit the instruction names properly. There's two different instructions in play:

- `loc_inst` is the instruction using the value. It's the location you're passing to `verify_value`.
- `def_inst` is the instruction where the value is defined.

With explicit instruction names:

```rust
fn verify_value(&self, loc_inst: Inst, v: Value) -> Result<()> {
...
ValueDef::Res(def_inst, _) => {
    if !dfg.insts.is_valid(def_inst) {
         return err!(loc_inst, "{} is defined by invalid instruction {}", v, def_inst);
```

stoklund:
BTW, I updated the `test-all.sh` script to use rustfmt 0.8.0 which apparently does whitespace differently in many cases.

```sh
cargo install --force rustfmt
```

angusholder:
Ok fixed for real this time!

lib/cretonne/src/dominator_tree.rs
```diff
@@ -68,8 +68,20 @@ impl DominatorTree {
     /// is unreachable.
     ///
     /// An instruction is considered to dominate itself.
-    pub fn dominates(&self, a: Inst, mut b: Inst, layout: &Layout) -> bool {
+    pub fn dominates(&self, a: Inst, b: Inst, layout: &Layout) -> bool {
         let ebb_a = layout.inst_ebb(a).expect("Instruction not in layout.");
+        self.ebb_dominates(ebb_a, b, layout) && layout.cmp(a, b) != Ordering::Greater
```
stoklund: Nice code reuse!

lib/cretonne/src/dominator_tree.rs
```diff
@@ -68,8 +68,20 @@ impl DominatorTree {
...
+    }
+
+    /// Returns `true` if `ebb_a` dominates `b`.
+    ///
+    /// This means that every control-flow path from the function entry to `b` must go through
+    /// `ebb_a`.
```
stoklund: I think we should be more precise with the language here. What does it mean to "go through" an EBB? In this case it means "go through the `ebb_a` header". An EBB header can dominate instructions that are not dominated by later instructions in the EBB.

lib/cretonne/src/verifier.rs
```diff
@@ -101,11 +103,19 @@ pub fn verify_function(func: &Function) -> Result<()> {
...
     func: &'a Function,
+    cfg: ControlFlowGraph,
+    domtree: DominatorTree,
 }
 
 impl<'a> Verifier<'a> {
     pub fn new(func: &'a Function) -> Verifier {
-        Verifier { func: func }
+        let cfg = ControlFlowGraph::with_function(func);
+        let domtree = DominatorTree::with_function(func, &cfg);
```
stoklund: Yep

lib/cretonne/src/verifier.rs
```diff
@@ -291,32 +301,42 @@ impl<'a> Verifier<'a> {
...
                 }
+                // Defining instruction is inserted in an EBB.
                 if self.func.layout.inst_ebb(def_inst) == None {
                     return err!(loc_inst,
                                 "{} is defined by {} which has no EBB",
                                 v,
                                 def_inst);
                 }
+                // Defining instruction dominates the instruction that uses the value.
+                if !self.domtree.dominates(def_inst, loc_inst, &self.func.layout) {
```
stoklund: IIRC, we defined dominance so an instruction dominates itself. It is not valid for an instruction to use its own result, though. Maybe that should be a separate check.

lib/cretonne/src/verifier.rs
```diff
@@ -291,32 +301,42 @@ impl<'a> Verifier<'a> {
...
+                // Defining EBB is inserted in the layout
                 if !self.func.layout.is_ebb_inserted(ebb) {
                     return err!(loc_inst,
                                 "{} is defined by {} which is not in the layout",
                                 v,
                                 ebb);
                 }
+                // The defining EBB dominates the instruction using this value.
+                if !self.domtree.ebb_dominates(ebb, loc_inst, &self.func.layout) {
+                    return err!(loc_inst, "uses value arg from non-dominating {}", ebb);
```
stoklund: Yep.

PR #65, opened 2017-03-22
Limit type inference for controlling type variables in write.rs

### LINKED ISSUE #63: Only assume local type inference when writing text format IL
Cretonne's IL parser supports a limited form of type inference for some polymorphic instructions where the controlling type variable can be inferred from one of the input operands:

```
    v0 = iconst.i32 2
    v1 = iconst.i8 6
    v2 = ishl v0, v1
```

The `ishl` instruction is polymorphic, but its `i32` result type can be inferred from the type of the `v0` operand.

The parser's type inference is best effort only, and in particular, it only knows the types of values that have been defined earlier in the input text. If `v0` were defined in an EBB later in the text format, we would have to use:

```
    v2 = ishl.i32 v0, v1
```

It would be technically possible for the parser to defer type resolution until everything has been parsed so that more missing types could be inferred. I don't think this is desirable, though. Cretonne IL is not a programming language, it is a language for writing unit tests, so obvious typing is better. It is also possible to write test cases in Cretonne IL that are somewhat malformed. For example, defines don't have to dominate uses, and it is useful to be able to write verifier test cases like that.

## Local type inference

The limited type inference is useful for reducing clutter, though. If a value is defined previously in the same EBB, its type is usually obvious, so there is no need repeat its type.

Currently, the instruction writer in `write.rs` is assuming a very smart type inference mechanism, even smarter than what the parser can do. It should be fixed to only depend on local type inference.

*Only omit the type variable when printing an instruction with a type variable operand that is defined in the same EBB.*

There is a TODO in the `type_suffix()` function.

KiChjang:
Fixes #63.

`./test-all.sh` actually reports 1 failure right now during filetests, and I think it is `traps_early.cton` that causes problems.

### DIFF
--- a/filetests/parser/branch.cton
+++ b/filetests/parser/branch.cton
@@ -62,7 +62,7 @@ ebb1:
 ; nextln:     brz vx0, ebb1
 ; nextln: 
 ; nextln: ebb1:
-; nextln:     brnz vx0, ebb1
+; nextln:     brnz.i32 vx0, ebb1
 ; nextln: }
 
 function twoargs(i32, f32) {
@@ -77,7 +77,7 @@ ebb1(vx2: i32, vx3: f32):
 ; nextln:     brz vx0, ebb1(vx0, vx1)
 ; nextln: 
 ; nextln: ebb1(vx2: i32, vx3: f32):
-; nextln:     brnz vx0, ebb0(vx2, vx3)
+; nextln:     brnz.i32 vx0, ebb0(vx2, vx3)
 ; nextln: }
 
 function jumptable(i32) {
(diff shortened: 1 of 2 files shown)


stoklund:
The code below your change needs to handle more kinds of instructions now. You can use `inst_data.ctrl_typevar(dfg)` instead of requiring the instruction to produce a result.

KiChjang:
Interesting, some of the filetests actually failed with my latest changes because they were assuming a more-than-capable parser. I've also updated the test expectation files so that it produces the correct result.

stoklund:
Nice, thanks!

Yeah, the parser is a bit more forgiving. It allows you to omit type variables as long as the values are defined earlier in the file.

PR #66, opened 2017-03-29
Type checking and Dominator Tree integrity checks in Verifier

angusholder:
It felt that often you want to treat fixed arguments and variable arguments differently, so I added methods for that to `InstructionData` and `DataFlowGraph` if that looks alright? 

Filetests are failing at the moment, but it doesn't seem to have anything to do with the verifier as far as I can see, I guess it's binemit?

I just want to do the return instruction arguments and entry block arguments checks now.


lib/cretonne/src/verifier.rs
```diff
@@ -403,11 +403,173 @@ impl<'a> Verifier<'a> {
...
+                .dfg
+                .inst_fixed_args(inst)
+                .iter()
+                .enumerate() {
+            let arg_type = self.func.dfg.value_type(arg);
+            match constraints.value_argument_constraint(i, ctrl_type) {
+                ResolvedConstraint::Bound(expected_type) => {
+                    if arg_type != expected_type {
+                        return err!(inst,
+                                    "arg {} has type {}, expected {}",
```
stoklund: Could you include `arg` itself in this error message? Same for the next one.

lib/cretonne/src/verifier.rs
```diff
@@ -403,11 +403,173 @@ impl<'a> Verifier<'a> {
...
+                                    i,
+                                    arg_type,
+                                    expected_type);
+                    }
+                }
+                ResolvedConstraint::Free(type_set) => {
+                    if !type_set.contains(arg_type) {
+                        return err!(inst,
+                                    "arg {} with type {} failed to satisfy type set",
```
stoklund: Maybe also include the expected type set in the error message here? Type sets don't print very nicely at the moment, but we could fix that.

lib/cretonne/src/verifier.rs
```diff
@@ -403,11 +403,173 @@ impl<'a> Verifier<'a> {
...
+                                    arg_type);
+                    }
+                }
+            }
+        }
+        Ok(())
+    }
+
+    fn typecheck_variable_args(&self, inst: Inst) -> Result<()> {
+        match self.func.dfg[inst] {
```
stoklund: Here you may be able to use the two existing functions `analyze_branch` and `analyze_call` instead of matching on all the instruction formats. See for example the existing `check_call_signature` function in the legalizer.

lib/cretonne/src/verifier.rs
```diff
@@ -403,11 +403,173 @@ impl<'a> Verifier<'a> {
...
+                            "arg {} has type {}, expected {}",
+                            i,
+                            arg_type,
+                            expected_type);
+            }
+            i += 1;
+        }
+        if i != variable_args.len() {
+            return err!(inst,
+                        "mismatched result count, got {}, expected {}",
```
stoklund: Not "result", right?

lib/cretonne/src/ir/dfg.rs
```diff
@@ -344,14 +344,36 @@ impl DataFlowGraph {
         DisplayInst(self, inst)
     }
 
-    /// Get the value arguments on `inst` as a slice.
+    /// Get all value arguments on `inst` as a slice.
     pub fn inst_args(&self, inst: Inst) -> &[Value] {
-        self.insts[inst].arguments(&self.value_lists)
+        let (fixed, variable) = self.insts[inst].arguments(&self.value_lists);
+        if fixed.len() != 0 { fixed } else { variable }
```
stoklund: This is not correct since some instructions have both fixed and variable arguments. Indirect calls, for example.

lib/cretonne/src/ir/dfg.rs
```diff
@@ -344,14 +344,36 @@ impl DataFlowGraph {
...
-    /// Get the value arguments on `inst` as a mutable slice.
+    /// Get all value arguments on `inst` as a mutable slice.
     pub fn inst_args_mut(&mut self, inst: Inst) -> &mut [Value] {
-        self.insts[inst].arguments_mut(&mut self.value_lists)
+        let (fixed, variable) = self.insts[inst].arguments_mut(&mut self.value_lists);
+        if fixed.len() != 0 { fixed } else { variable }
+    }
+
+    /// Get the fixed value arguments on `inst` as a slice.
+    pub fn inst_fixed_args(&self, inst: Inst) -> &[Value] {
```
stoklund: I think it is OK to add this method, but don't change the type returned from `InstructionData::arguments()`. I changed this method recently so it always returns a single slice, and that really simplifies those cases that don't care about the difference between fixed and variable arguments.

This method could simply return `self.inst_args(inst)[0..fixed_values]`.

lib/cretonne/src/verifier.rs
angusholder: Do you want me to add a `Display` implementation for the type sets in this PR? The Python impl is quite simple, and since the types we have isn't going to change much, I'm guessing it's fine not to have codegen to keep them in sync?

lib/cretonne/src/ir/dfg.rs
```diff
@@ -344,16 +344,40 @@ impl DataFlowGraph {
...
+    /// Get all value arguments on `inst` as a mutable slice.
     pub fn inst_args_mut(&mut self, inst: Inst) -> &mut [Value] {
         self.insts[inst].arguments_mut(&mut self.value_lists)
     }
 
+    /// Get the fixed value arguments on `inst` as a slice.
+    pub fn inst_fixed_args(&self, inst: Inst) -> &[Value] {
+        let fixed_args = self[inst].opcode().constraints().fixed_value_arguments();
+        &self.inst_args(inst)[..fixed_args]
+    }
```
stoklund: Yup. Looks great

lib/cretonne/src/legalizer/boundary.rs
```diff
@@ -346,12 +346,8 @@ fn check_call_signature(dfg: &DataFlowGraph, inst: Inst) -> Result<(), SigRef> {
 
 /// Check if the arguments of the return `inst` match the signature.
 fn check_return_signature(dfg: &DataFlowGraph, inst: Inst, sig: &Signature) -> bool {
-    let fixed_values = dfg[inst].opcode().constraints().fixed_value_arguments();
     check_arg_types(dfg,
-                    dfg.inst_args(inst)
-                        .iter()
-                        .skip(fixed_values)
-                        .cloned(),
+                    dfg.inst_variable_args(inst).iter().cloned(),
```
stoklund: Yeah, using `inst_variable_args` here is definitely an improvement.

lib/cretonne/src/verifier.rs
```diff
@@ -377,11 +385,225 @@ impl<'a> Verifier<'a> {
...
+                    .ebb_args(ebb)
+                    .map(|v| self.func.dfg.value_type(v));
+                self.typecheck_variable_args_iterator(inst, iter)?;
+            }
+            BranchInfo::Table(table) => {
+                for (_, ebb) in self.func.jump_tables[table].entries() {
+                    let arg_count = self.func
+                        .dfg
+                        .ebb_args(ebb)
+                        .count();
```
stoklund: There's a `dfg.num_ebb_args(ebb)` method.

lib/cretonne/meta/gen_instr.py
```diff
@@ -67,7 +67,7 @@ def gen_arguments_method(fmt, is_mut):
     with fmt.indented(
             'pub fn {f}<\'a>(&\'a {m}self, pool: &\'a {m}ValueListPool) -> '
             '&{m}[Value] {{'
-            .format(f=method, m=mut), '}'):
+                    .format(f=method, m=mut), '}'):
```
stoklund: Are these formatting changes intentional?

angusholder: Ah no, seems my editor took things into its own hands, I'll revert that.

stoklund:
> Do you want me to add a Display implementation for the type sets in this PR? The Python impl is quite simple, and since the types we have isn't going to change much, I'm guessing it's fine not to have codegen to keep them in sync?
> 
No, you don’t need to do it for this PR.

I think the type sets can be displayed without any help from the Python side by using a format that describes the permitted types instead of using the type variable name.

I guess the Python __repr__ is a start, although it’s no too pretty:

    def __repr__(self):
        # type: () -> str
        s = 'TypeSet(lanes=({}, {})'.format(self.min_lanes, self.max_lanes)
        if self.min_int is not None:
            s += ', ints=({}, {})'.format(self.min_int, self.max_int)
        if self.min_float is not None:
            s += ', floats=({}, {})'.format(self.min_float, self.max_float)
        if self.min_bool is not None:
            s += ', bools=({}, {})'.format(self.min_bool, self.max_bool)
        return s + ')'

stoklund:
I would prefer to merge all of this as a single commit since there's some back and forth with the `arguments()` method.

Do you mind squashing it?

stoklund:
Never mind, GitHub has a "squash and merge feature".

angusholder:
Ah I was just wrestling with rebase right now!

Issue #67, opened 2017-03-29
`verify_context()` should also verify CFG integrity


stoklund:
The `verifier::verify_context()` function checks that the dominator tree in the context matches a newly computed one. We should do the same thing for the control flow graph.

There is an existing `cfg_integrity` method, but it just checks the just-computed CFG against the function. It doesn't look at the CFG in the context.

cc @angusholder

angusholder:
Should I verify for just the reachable EBBs using `postorder_ebbs()`?

stoklund:
No, I think we can require that the CFG is valid for all EBBs that are in the layout.

angusholder:
Should the postorder be the same between them?

stoklund:
No, the postorder does not have to be the same either.

If you consider the CFG as a *graph*, the two graphs should be equivalent. It should be enough to verify that the predecessor sets are identical *as sets*, and the successor sets are identical *as sets*.

angusholder:
Ok got it, thank you

stoklund:
Fixed by @angusholder in #70.

PR #70, opened 2017-03-29
Verify integrity of the existing control flow graph of the context.

angusholder:
Issue #67

### DIFF
--- a/lib/cretonne/src/verifier.rs
+++ b/lib/cretonne/src/verifier.rs
@@ -106,6 +106,7 @@ pub fn verify_function(func: &Function) -> Result<()> {
 pub fn verify_context(ctx: &Context) -> Result<()> {
     let verifier = Verifier::new(&ctx.func);
     verifier.domtree_integrity(&ctx.domtree)?;
+    verifier.cfg_integrity(&ctx.cfg)?;
     verifier.run()
 }
 
@@ -350,7 +351,7 @@ impl<'a> Verifier<'a> {
         Ok(())
     }
 
-    fn cfg_integrity(&self, ebb: Ebb) -> Result<()> {
+    fn control_flow_integrity(&self, ebb: Ebb) -> Result<()> {
         for &(pred_ebb, pred_inst) in self.cfg.get_predecessors(ebb) {
             // All predecessors in the CFG must be branches to the EBB
             match self.func.dfg[pred_inst].analyze_branch(&self.func.dfg.value_lists) {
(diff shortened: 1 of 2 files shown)


lib/cretonne/src/verifier.rs
```diff
@@ -350,7 +351,7 @@ impl<'a> Verifier<'a> {
         Ok(())
     }
 
-    fn cfg_integrity(&self, ebb: Ebb) -> Result<()> {
+    fn control_flow_integrity(&self, ebb: Ebb) -> Result<()> {
```
stoklund: Maybe we should just delete this function?

Assuming that we trust the CFG implementation to be correct, it isn't really adding anything. We have specific CFG tests both as unit tests and `print-cfg` file tests. The value of the verifier is to check that the CFG stored in the context has been kept up to data with any program changes made by optimizations.

lib/cretonne/src/verifier.rs
```diff
@@ -614,6 +615,28 @@ impl<'a> Verifier<'a> {
         Ok(())
     }
 
+    fn cfg_integrity(&self, cfg: &ControlFlowGraph) -> Result<()> {
+        for ebb in self.cfg.ebbs() {
```
stoklund: I think you should use `layout.ebbs()` instead here. The `cfg.ebbs()` method probably shouldn't be public.

lib/cretonne/src/verifier.rs
```diff
@@ -614,6 +615,28 @@ impl<'a> Verifier<'a> {
...
+            let cfg_successors = cfg.get_successors(ebb);
+            let cfg_predecessors = cfg.get_predecessors(ebb);
+            for &succ in self.cfg.get_successors(ebb) {
+                if !cfg_successors.contains(&succ) {
```
stoklund: This is correct, but only half way there. This code is checking that `self.cfg.get_successors(ebb)` is a subset of `cfg_successors`. There could be extra elements in cfg_successors` and we wouldn't notice.

Also be careful about introducing accidentally quadratic code. It is theoretically possible for both successor and predecessor sets to get very large, and this algorithm would be quadratic in the size of the sets.

It might be easier to use something like `HashSet::symmetric_difference`. (Or the `BTreeSet` equivalent for nicely ordered EBBs in the error messages).

I expect the CFG implementation to use our own B-tree data structures in the future, but for now it is fine to use the standard library ones in the verifier.

lib/cretonne/src/verifier.rs
angusholder: That makes sense, I'll do that.

lib/cretonne/src/verifier.rs
angusholder: I went for doing two uses of `difference`, to distinguish the cases when we have excess members and when we have missing members. There's no `Ord` impl for `BasicBlock` though, so I can't currently use BTreeSet. Should I just throw a basic one on it for now? (order by ebb then by inst)

stoklund: That makes sense to me.

Go ahead and add `Ord` to BasicBlock. I bet you can even `#[derive(PartialOrd, Ord)]` which should give you a lexicographical order.

Alternatively, you can just build a set of the predecessor *instructions* only. The EBB part of `BasicBlock` is actually redundant since you can get it from `layout.inst_ebb()`. (Maybe that's one more thing to verify...)

angusholder:
I've just been adding tests for the type checking, am I alright to put it into this PR?

PR #71, opened 2017-03-30
Added tests for verifier type checking.

angusholder:
I'm not sure if I'm missing something or if I've found a bug in filetests, it doesn't like my error directive on the jump instruction, which is identical to the one on the line before on the branch instruction.

### DIFF
--- a/filetests/verifier/type_check.cton
+++ b/filetests/verifier/type_check.cton
@@ -1,0 +1,88 @@
+test verifier
+
+function entry_block_signature_mismatch(i32) {
+    ebb0: ; error: entry block arguments must match function signature
+        return
+}
+
+function entry_block_arg_type(i32) {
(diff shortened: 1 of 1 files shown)


stoklund:
I'll try to reproduce here.

stoklund:
I'm getting this failure:

```
FAIL filetests/verifier/type_check.cton: verifier(jump_args): cannot handle multiple error: directives
```

The `test verifier` command will only accept one `error:` directive per function. That is because the verifier is only capable of returning a single failure. There is no way it can report an error for both instructions, even if they both have errors.

angusholder:
Ah of course, sorry that makes sense

stoklund:
FWIW, it would be pretty cool if the verifier were capable of reporting multiple errors at a time.

stoklund:
Landed in 874f95dc2ac24476d1441093442b3e09c5274abc with some updates.

Thanks!

PR #77, opened 2017-05-09
Make `srem` have the sign of the dividend.

sunfishcode:
This is how remainder is defined in C (as of C99), C++ (as of C++11), Rust,
and WebAssembly, for example.


stoklund:
Thanks!

PR #78, opened 2017-05-10
Use write! in utility code, rather than calling write_function directly.

sunfishcode:
This changes uses of `write_function` in utility code to use the more general `write!` interface, which ends up calling `write_function` internally. It seems to make sense for utility code to use the more general interface, but feel free to close this if not.


stoklund:
The tests are failing because of formatting issues. You can run `rustfmt` yourself as part of the `./test-all.sh` script by installing the required version with `./check-rustfmt.sh --install`.

PR #79, opened 2017-05-10
Start a very simple GVN pass

sunfishcode:
This is a really experimental pass that I put together as a way to start exploring the code. It's pretty incomplete right now, but it does implement enough to replace a redundant instruction in a simple testcase :-).

I'm still quite new to Rust, and Cretonne, so I'd be very interested in better ways to do any of the things the code is doing.

### DIFF
--- a/lib/cretonne/meta/cdsl/instructions.py
+++ b/lib/cretonne/meta/cdsl/instructions.py
@@ -85,6 +85,8 @@ class Instruction(object):
     :param is_call: This is a call instruction.
     :param is_return: This is a return instruction.
     :param can_trap: This instruction can trap.
+    :param can_load: This instruction can load from memory.
+    :param can_store: This instruction can store to memory.
     """
 
     # Boolean instruction attributes that can be passed as keyword arguments to
@@ -95,6 +97,8 @@ class Instruction(object):
             'is_branch': 'True for all branch or jump instructions.',
             'is_call': 'Is this a call instruction?',
             'is_return': 'Is this a return instruction?',
+            'can_load': 'Can this instruction read from memory?',
+            'can_store': 'Can this instruction write to memory?',
             'can_trap': 'Can this instruction cause a trap?',
             }
 
--- a/lib/cretonne/src/context.rs
+++ b/lib/cretonne/src/context.rs
@@ -51,25 +51,26 @@ impl Context {
     ///
     /// Also check that the dominator tree and control flow graph are consistent with the function.
     ///
-    /// The `TargetIsa` argument is currently unused, but the verifier will soon be able to also
+    /// The `isa` argument is currently unused, but the verifier will soon be able to also
     /// check ISA-dependent constraints.
-    pub fn verify<'a, ISA: Into<Option<&'a TargetIsa>>>(&self, isa: ISA) -> verifier::Result {
-        verifier::verify_context(&self.func, &self.cfg, &self.domtree, isa.into())
+    pub fn verify<'a>(&self, isa: Option<&TargetIsa>) -> verifier::Result {
+        verifier::verify_context(&self.func, &self.cfg, &self.domtree, isa)
     }
 
     /// Run the verifier only if the `enable_verifier` setting is true.
-    pub fn verify_if(&self, isa: &TargetIsa) -> CtonResult {
-        if isa.flags().enable_verifier() {
-            self.verify(isa).map_err(Into::into)
-        } else {
-            Ok(())
+    pub fn verify_if(&self, isa: Option<&TargetIsa>) -> CtonResult {
+        if let Some(isa) = isa {
+            if isa.flags().enable_verifier() {
+                return self.verify(Some(isa)).map_err(Into::into);
+            }
         }
+        Ok(())
     }
 
     /// Run the legalizer for `isa` on the function.
     pub fn legalize(&mut self, isa: &TargetIsa) -> CtonResult {
         legalize_function(&mut self.func, &mut self.cfg, isa);
-        self.verify_if(isa)
+        self.verify_if(Some(isa))
     }
 
     /// Recompute the control flow graph and dominator tree.
--- a/lib/cretonne/src/context.rs
+++ b/lib/cretonne/src/context.rs
@@ -17,6 +17,7 @@ use legalize_function;
 use regalloc;
 use result::CtonResult;
 use verifier;
+use simple_gvn::do_simple_gvn;
 
 /// Persistent data structures and compilation pipeline.
 pub struct Context {
@@ -77,6 +78,12 @@ impl Context {
     pub fn flowgraph(&mut self) {
         self.cfg.compute(&self.func);
         self.domtree.compute(&self.func, &self.cfg);
+    }
+
+    /// Perform simple GVN on the function.
+    pub fn simple_gvn(&mut self) -> CtonResult {
+        do_simple_gvn(&mut self.func, &mut self.cfg);
+        self.verify_if(None)
     }
 
     /// Run the register allocator.
--- a/lib/cretonne/src/context.rs
+++ b/lib/cretonne/src/context.rs
@@ -59,19 +59,18 @@ impl Context {
     }
 
     /// Run the verifier only if the `enable_verifier` setting is true.
-    pub fn verify_if(&self, isa: Option<&TargetIsa>) -> CtonResult {
-        if let Some(isa) = isa {
-            if isa.flags().enable_verifier() {
-                return self.verify(Some(isa)).map_err(Into::into);
-            }
+    pub fn verify_if(&self, isa: &TargetIsa) -> CtonResult {
+        if isa.flags().enable_verifier() {
+            self.verify(Some(isa)).map_err(Into::into)
+        } else {
+            Ok(())
         }
-        Ok(())
     }
 
     /// Run the legalizer for `isa` on the function.
     pub fn legalize(&mut self, isa: &TargetIsa) -> CtonResult {
         legalize_function(&mut self.func, &mut self.cfg, isa);
-        self.verify_if(Some(isa))
+        self.verify_if(isa)
     }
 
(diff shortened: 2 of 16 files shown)


lib/cretonne/meta/cdsl/instructions.py
```diff
@@ -95,6 +97,8 @@ class Instruction(object):
             'is_branch': 'True for all branch or jump instructions.',
             'is_call': 'Is this a call instruction?',
             'is_return': 'Is this a return instruction?',
+            'can_load': 'Can this instruction read from memory?',
+            'can_store': 'Can this instruction write to memory?',
```
stoklund: We should have a way of catching when you use `can_load` in one place and `may_load` in another place.

The `Instruction` constructor should complain about unknown flags.

lib/cretonne/src/context.rs
```diff
@@ -78,6 +80,12 @@ impl Context {
         self.domtree.compute(&self.func, &self.cfg);
     }
 
+    /// Perform simple GVN on the function.
+    pub fn simple_gvn(&mut self) -> CtonResult {
+        do_simple_gvn(&mut self.func, &mut self.cfg);
+        self.verify_if(None)
```
stoklund: This call doesn't do anything, with the `verify_if` implementation above.

I haven't thought much about tests without an ISA. It should always be possible to get at a `Flags` instance that we can give the context for those passes that don't need a full `TargetIsa`. See the definition of `IsaSpec` in the parser crate.

lib/cretonne/src/simple_gvn.rs
```diff
@@ -0,0 +1,133 @@
...
+    opcode.is_return() || opcode.can_trap()
+}
+
+#[derive(Debug)]
+struct VisibleInst(InstructionData);
+
+impl Hash for VisibleInst {
+    fn hash<H: Hasher>(&self, state: &mut H) {
+        // TODO: Is a better way to do this? Should we just make InstructionData
+        // implement Hash itself?
```
stoklund: You can `#[derive(Hash)]` too.

The problem is those instructions that store their arguments in a `ValueList`. They will hash the list reference, not the contents. It's possible you don't care about any of those instructions: jumps, calls, return.

lib/cretonne/src/simple_gvn.rs
```diff
@@ -0,0 +1,133 @@
...
+    fn eq(&self, other: &VisibleInst) -> bool {
+        if self.0.opcode() != other.0.opcode() {
+            return false;
+        }
+
+        // TODO: Surely there is a better way to do this? Should we provide a
+        // PartialEq implementation for InstructionData? Note that it isn't as
+        // simple as just using the default implementation, since
+        // InstructionData contains things like floating-point literals which
+        // don't have Eq defined.
```
stoklund: Right now, those floating point literals are represented as `Ieee32` and `Ieee64` structs which contain floats. I have been wanting to change the internal representation to `u32`/`u64` instead. That would also enable you to `#[derive(PartialEq, Eq)]`, with the same caveat about value lists.

We don't want GVN to merge +0.0 and -0.0 either.

lib/cretonne/src/simple_gvn.rs
```diff
@@ -0,0 +1,133 @@
...
+/// Replace the values of new_inst with the values of old_inst.
+///
+fn replace_values(dfg: &mut DataFlowGraph, new_inst: Inst, old_inst: Inst) {
+    let mut result = 0;
+    let new_results = dfg.detach_results(new_inst);
+    while let Some(new_val) = new_results.get(result, &dfg.value_lists) {
+        let old_val = dfg.inst_results(old_inst)[result];
+        dfg.change_to_alias(new_val, old_val);
+        result += 1;
+    }
```
stoklund: Sometimes the borrow checker is really annoying.

If you implement this as a method on `DataFlowGraph`, the borrow checker will be able to understand that `dfg.value_lists` and `dfg.values` don't alias, and you can iterate over the two slices of results instead.

Rust does not fully support functional decomposition...

```rust
for (&new, &old) in new_results.as_slice(...).iter().zip(self.results[old_inst].as_slice(...)) {
   // inline `change_to_alias()` here...
}
```

lib/cretonne/src/simple_gvn.rs
```diff
@@ -0,0 +1,133 @@
...
+            use std::collections::hash_map::Entry::*;
+            match entry {
+                Occupied(mut entry) => {
+                    if domtree.dominates(*entry.get(), inst, &pos.layout) {
+                        // TODO: Delete inst, while we're here.
+                        replace_values(&mut func.dfg, inst, *entry.get());
+                    } else {
+                        // The prior instruction doesn't dominate inst, so it
+                        // won't dominate any subsequent instructions we'll
+                        // visit, so just replace it.
```
stoklund: That's not exactly true, but I see the problem.

stoklund: Sorry, you're right. Nothing to see here.

It would be nice to periodically prune the hash table for entries that will no longer be useful so we don't end up with the whole program in there.

stoklund:
I think your Rust style looks good, at least as far as I am able to recognize good style. You should install `rustfmt` to get the whitespace to match what Travis expects. The `test-all.sh` script picks it up and verifies formatting too.

# Context

The `Context` struct doesn't have good support for passes that don't require an ISA reference. I think it would be reasonable for those passes to take a `&settings::Flags` reference instead of the ISA. You may need to modify some of the test driver code to get at those settings, but the parser always returns them somehow. When you have an ISA, the flags are available as `isa.flags()`.

# Hashing instructions

I don't think the standard `HashMap` will make it into production. It doesn't give you good control over allocation, and it's not flexible enough for a fast GVN. If it works well enough for you, I think it is fine to `#[derive(Hash, Eq)]` on `InstructionData`. Changing the `Ieee*` immediate types to contain an integer instead of a float should be trivial.

I think eventually you will end up with a custom hash-map implementation based on something like a `Vec<PackedOption<Inst>>` where each entry is only 32 bits. If you need both tombstones and empty slots, we can figure something out. When you have your own hash table, you can prune out the stale entries before growing it. That ought to keep it fairly small without using a scoped hash table.

I don't know if Rust gives you a way of controlling the padding bits in an enum, but if you knew they were always zero, you could just transmute the `InstructionData` to a `[u64;2]` and hash that instead.

Note that value numbers are stable in some sense, so it wouldn't be outrageous to sort the arguments on commutative instructions. Value numbers should be the same on identical compilations. (But the parser does not preserve value numbers, so round-tripping to text is problematic).

Alternatively, it would also be possible to define a total order on `ValueDef` by using the `ProgramOrder` provided by `Layout`. That would be quite fast and the order would survive a text round-trip.

# Aliases

I think the `replace_values` function could live in `DataFlowGraph` next to `change_to_alias`. It can be more efficient there for two reasons:

1. You can borrow slices with `self.results[inst].as_slice(&self.value_lists)` without locking the value table for writing. This means you only have to compute the length of the result lists once.
2. You know that you are dealing with instruction results, so you don't have to check for existing aliases and detect loops etc. You can just drop in the `ValueData::Alias` variant immediately.

I think you need to resolve aliases in the instruction arguments before hashing it. Otherwise you won't be able to merge chains of instructions. This would also be useful as a method on `DataFlowGraph`.

lib/cretonne/meta/cdsl/instructions.py
sunfishcode: I've now opened #80 to make it complain, and fixed the discrepancy here.

lib/cretonne/src/ir/dfg.rs
```diff
@@ -243,6 +243,41 @@ impl DataFlowGraph {
         self.values[dest] = ValueData::Alias { ty, original };
     }
 
+    /// Replace the values of one instruction with aliases to the values of another.
```
stoklund: I usually use the term "results" for the values produced by an instruction. Or "result values".

lib/cretonne/src/ir/dfg.rs
```diff
@@ -243,6 +243,41 @@ impl DataFlowGraph {
...
+            assert_eq!(self.value_type(dest),
+                       ty,
+                       "Aliasing {} to {} would change its type {} to {}",
+                       dest,
+                       src,
+                       self.value_type(dest),
+                       ty);
+
+            self.values[dest] = ValueData::Alias { ty, original };
+        }
```
stoklund: I think you should call `self.clear_results(dest)` when you're done. Otherwise, the result list would be left in an invalid state.

lib/cretonne/src/ir/immediates.rs
```diff
@@ -558,6 +564,19 @@ impl FromStr for Ieee32 {
     }
 }
 
+impl Hash for Ieee32 {
```
stoklund: Alternatively, change the representation to `struct Ieee32(u32)` and derive these traits. I don't think we ever need that `f32` representation for anything.

stoklund: The only place we ever use `Ieee32::new` is in the test cases below.

lib/cretonne/src/simple_gvn.rs
```diff
@@ -0,0 +1,65 @@
...
+            }
+
+            let key = func.dfg[inst].clone();
+            let entry = visible_values.entry(key);
+            use std::collections::hash_map::Entry::*;
+            match entry {
+                Occupied(mut entry) => {
+                    if domtree.dominates(*entry.get(), inst, &pos.layout) {
+                        func.dfg.replace_with_aliases(inst, *entry.get());
+                        pos.remove_inst();
```
stoklund: This will leave the cursor pointing to the next instruction, so the `pos.next_inst()` at the top of the loop will skip it.

Maybe we should have two versions of `Cursor::remove_inst()`? One that leaves the cursor at the previous position, and one that leaves it at the next one.

PR #80, opened 2017-05-11
Check for unknown instruction attributes.

sunfishcode:
Per review comment in #79, this adds a check for unknown instruction attributes.

### DIFF
--- a/lib/cretonne/meta/cdsl/instructions.py
+++ b/lib/cretonne/meta/cdsl/instructions.py
@@ -124,7 +124,7 @@ class Instruction(object):
         self._verify_polymorphic()
         for attr in kwargs:
             if attr not in Instruction.ATTRIBS:
-                raise RuntimeError(
+                raise AssertionError(
                         "unknown instruction attribute '" + attr + "'")
         for attr in Instruction.ATTRIBS:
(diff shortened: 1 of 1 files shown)


lib/cretonne/meta/cdsl/instructions.py
```diff
@@ -122,6 +122,10 @@ def __init__(self, name, doc, ins=(), outs=(), **kwargs):
                 i for i, o in enumerate(self.ins) if o.is_immediate())
 
         self._verify_polymorphic()
+        for attr in kwargs:
+            if not Instruction.ATTRIBS.has_key(attr):
```
stoklund: Flake8 says to use

```python
if attr not in Instruction.ATTRIBS:
```

lib/cretonne/meta/cdsl/instructions.py
```diff
@@ -122,6 +122,10 @@ def __init__(self, name, doc, ins=(), outs=(), **kwargs):
...
+                raise RuntimeError(
+                        "unknown instruction attribute '" + attr + "'")
```
stoklund: I tend to use `AssertionError` for these things, but I don't know if it makes a difference.

PR #81, opened 2017-05-15
Update rustfmt to 0.8.4

bnjbvr:
I was just looking around and saw that a new version of rustfmt had been released. Here's the result of updating the check-format script and run the format-all script.

I noticed it sometimes wrapped a few lines into one, which makes things slightly harder to read in some places. Are there any tweaks you'd like me to do in addition to the raw output of format-all?

### DIFF
--- a/check-rustfmt.sh
+++ b/check-rustfmt.sh
@@ -15,7 +15,7 @@
 # With the --install option, also tries to install the right version.
 
 # This version should always be bumped to the newest version available.
-VERS="0.8.3"
+VERS="0.8.4"
 
 if cargo install --list | grep -q "^rustfmt v$VERS"; then
     exit 0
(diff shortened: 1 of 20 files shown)


stoklund:
Thanks!

I would prefer to keep using `rustfmt` without any options, so we arrive at some kind of standard formatting, like what Go has. In my opinion, having a standardized style is more important than personal preferences about white-space—I don't always agree with `rustfmt`'s choices either.

It is a bit of a bumpy ride to follow `rustfmt` these days, though. There has been some quite significant changes to formatting recently. These white-space commits can be a problem when somebody has a significant branch that hasn't landed yet. I think @sunfishcode has one right now, so I'll give them a chance to protest before we land this.

I wonder if we could provide a script that can update a branch to a new formatting style, using `git filter-branch` for example?

bnjbvr:
Thanks for the merge! I'd have been happy waiting for @sunfishcode's MR to be merged first, since reformatting is obviously trivial.

I've tried to figure out a way to reformat source on existing patches and patches not yet upstreamed, but couldn't figure any way (using simple git commands like `git stash`).

PR #82, opened 2017-05-19
Implement simple JSON serialization for Cretonne IL.

sunfishcode:
This is part of an experiment I'm putting together. If you think JSON serialization sounds like a generally desirable thing to have, we can merge it, but otherwise we can wait until we know how the experiments turn out :-).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,6 +14,7 @@ path = "src/cton-util.rs"
 
 [dependencies]
 cretonne = { path = "lib/cretonne" }
+cretonne-json = { path = "lib/json" }
 cretonne-reader = { path = "lib/reader" }
 filecheck = { path = "lib/filecheck" }
 docopt = "0.6.86"
--- a/lib/cretonne/src/ir/function.rs
+++ b/lib/cretonne/src/ir/function.rs
@@ -10,6 +10,7 @@ use ir::{FunctionName, Signature, Value, Inst, Ebb, StackSlot, StackSlotData, Ju
 use isa::{TargetIsa, Encoding};
 use std::fmt::{self, Display, Debug, Formatter};
 use write::write_function;
+use json::json_function;
 
 /// A function.
 ///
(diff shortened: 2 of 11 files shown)


mbebenita:
I think this would be very useful for other tools. Can you emit instructions in a form that doesn't require a consumer to parse them. For example, currently you emit:

```
{"instruction":"v5 = imul_imm v4, 4"}
```
would it make more sense to emit

```
{"instruction":["v5", "imul_imm", "v4", 4]}
```

bjorn3:
You could use serde + serde_json for serialization.

mbebenita:
On second thought, perhaps we should stick to a single textual IR representation and write parsers in each tool that needs to use it.

sunfishcode:
The serde + serde_json scheme sounds like it could be a good idea. It'll take a little more work to implement, because Cretonne's EntityLists aren't serde-friendly, but I think we could make it work by creating a separate set of data structures which would be a kind of a "display-form" IR where everything is expanded, that you could easily translate into from the regular "optimized" IR, and which could easily implement serde's Serialize, and have nicer Debug methods. Possibly we could do deserialization this way too.

stoklund:
At a minimum, I think it should live in a separate crate in the `lib/` directory, like the parser. That way, you don't have to pay for it unless you use it. I also try to keep the main `cretonne` crate free of external dependencies.

The immediate operands on instructions are 64-bit integers and IEEE floats including all NaN bit-patterns. As I understand it, JSON can't represent either of those except as strings, so it may not be the best general format for this?

I try to keep to a minimum the number of large switches over the instruction formats. We still have to add new instruction formats somewhat frequently, and each switch adds to the work required. Whenever possible, I use existing functions such as `dfg.inst_args(inst)` and `analyze_branch` to avoid a format switch. Other code, like the `InstBuilder` is generated.

I'm not very familiar with serde, it sounds like it could be useful. But again, I do want to control the binary size of the core code generator, so I'd rather not take a dependency on serde from that crate or include a lot of dead serialization code. A cretonne-json crate that depends on serde is just fine.

A more old-school approach is to generate your own serialization code. The Python scripts have detailed knowledge about the `InstructionData` type. Take a look at the `gen_format_constructor()` function in `meta/gen_instr.py`. It generates the `InstBuilder` methods in ` target/debug/build/cretonne-*/out/builder.rs`.

sunfishcode:
For now, I've added a patch which moves the code into a separate crate, just to make sure I know how to do that, and which prints out instructions broken down into their various parts rather than a single string, which should make it a little better for simple experimentation.

I think a cretonne-serde crate sounds like a useful thing to explore after this.

sunfishcode:
Closing; if we implement serialization, it's likely to take a different form.

PR #83, opened 2017-05-19
Fix more GVN issues

sunfishcode:
This addresses a few more items from https://github.com/stoklund/cretonne/pull/79; it resolves aliases as it iterates, and it fixes the cursor position after removing an instruction. With these changes, basic cascading optimizations now work.

### DIFF
--- a/lib/cretonne/src/ir/dfg.rs
+++ b/lib/cretonne/src/ir/dfg.rs
@@ -112,7 +112,7 @@ impl DataFlowGraph {
 fn resolve_aliases(values: &EntityMap<Value, ValueData>, value: Value) -> Value {
     let mut v = value;
 
-    // Note that extended_values may be empty here.
+    // Note that values may be empty here.
     for _ in 0..1 + values.len() {
         if let ValueData::Alias { original, .. } = values[v] {
             v = original;
--- a/lib/cretonne/src/ir/dfg.rs
+++ b/lib/cretonne/src/ir/dfg.rs
@@ -106,6 +106,23 @@ impl DataFlowGraph {
     }
 }
 
+/// Resolve value aliases.
+///
+/// Find the original SSA value that `value` aliases.
+fn resolve_aliases(values: &EntityMap<Value, ValueData>, value: Value) -> Value {
+    let mut v = value;
+
+    // Note that extended_values may be empty here.
+    for _ in 0..1 + values.len() {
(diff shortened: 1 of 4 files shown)


lib/cretonne/src/ir/dfg.rs
```diff
@@ -106,6 +106,23 @@ impl DataFlowGraph {
     }
 }
 
+/// Resolve value aliases.
+///
+/// Find the original SSA value that `value` aliases.
+fn resolve_aliases(values: &EntityMap<Value, ValueData>, value: Value) -> Value {
```
stoklund: Nice. I figured something like this would be necessary to make the borrow checker happy.

lib/cretonne/src/ir/layout.rs
```diff
@@ -923,6 +923,18 @@ impl<'f> Cursor<'f> {
         inst
     }
 
+    /// Remove the instruction under the cursor.
+    ///
+    /// The cursor is left pointing at the position preceding the current instruction.
+    ///
+    /// Return the instruction that was removed.
+    pub fn remove_inst_and_step_back(&mut self) -> Inst {
```
stoklund: I approve of this name ;-)
I couldn't think of a good one.

lib/cretonne/src/ir/dfg.rs
```diff
@@ -106,6 +106,23 @@ impl DataFlowGraph {
...
+    let mut v = value;
+
+    // Note that extended_values may be empty here.
```
stoklund: Oops, `extended_values` is the old name for the `values` table back from when there was two kinds of values.

lib/cretonne/src/ir/dfg.rs
stoklund: I wonder if we would get better code if this function takes a `value: &mut Value` argument instead of returning the new value?

When inlined into `resolve_aliases_in_arguments`, there would be one less test.

lib/cretonne/src/ir/dfg.rs
sunfishcode: Fixed.

lib/cretonne/src/ir/dfg.rs
sunfishcode: I compared the generated release-build code before and after making this change, and it doesn't seem significantly different. Since other places in the code find the current return-the-new-value interface more convenient, my sense is to leave it as is for now.

PR #84, opened 2017-05-26
LICM pass implementation

denismerigoux:
I've implemented an LICM pass that uses Tarjan's strongly connected components algorithm to detect loops. However, nested loops are not yet supported (the Tarjan's algorithm is not called recursively as it should be).


stoklund:
Hi Denis,

Thanks for working on this!

Your loop analysis would be really useful for other optimizations too. I would like to see that broken out into an independent analysis, like the dominator tree.

The code needs a bit of massaging to be production ready. Two points in particular:

1. Avoid recursion unless you have tight control over the recursion depth. Recursive code generally uses much more stack space than an equivalent iterative algorithm uses heap space. And since we have limited stack space, it is not safe to recurse over unknown data structures. This means you can't recurse over things like the CFG, the dominator tree, or even a loop tree. They can cause arbitrarily deep recursion. The `ControlFlowGraph::postorder_ebbs()` function is an example of a recursive algorithm rewritten to be iterative.
2. Avoid allocating and freeing memory as much as possible. Even with jemalloc, this is quite expensive. The `cretonne::Context` struct is meant to hold data structures whose memory allocations can be reused between functions. The `DominatorTree` struct is a good example of how this should work. The `nodes` entity map will hold on to its allocated memory and reuse it when `DominatorTree::compute()` is called for a new function. The `ControlFlowGraph` does not do this right. Each `CFGNode` contains `Vec`s that free their memory at the top of `ControlFlowGraph::compute()`. This should be fixed at some point, probably by using `EntityList` instead of `Vec`.

Would it be possible to pull the loop analysis into a separate analysis modeled after the dominator tree, avoiding recursion and deallocation?

denismerigoux:
Hi Jakob,

Yes I can definitely move the loop analysis out of the LICM pass: it will take as input the function, its CFG and its dominator tree, and return at set (`Vec` or `EntityList` ?) of tuples `(Ebb,HashSet<Ebb>)` describing the loops by their header and the EBBs in the loop.

Concerning your points:

1. I'll try to re-write Tarjan's strongly connected components algorithm in an iterative manner. I wrote this recursive version because it wanted to avoid errors by following closely the theoretical description of the algorithm.

2. OK now I understand better the purpose of `EntityList` and `EntityMap`. Would it be better if I allocated a single `ListPool` for the whole pass, and used `EntityList`s with this pool for my intermediary data structures ?

The last improvement point is to detect inner loops, but that should not be too hard to do given that it is just a matter of applying the SCC algorithm multiple times.

stoklund:
Thanks,

If you look at the dominator tree interface, it computes an internal data structure, and lets you query that data structure via the dominates* functions. It doesn't return any data structures to you, because that forces you to later deallocate them. This approach also means that several passes can reuse the same dominator tree computation.

The control flow graph works in the same way, except for the `postorder_ebbs()` function which forces you to deallocate a `Vec`. I don't think it should, and I'm changing it to return a `&[Ebb]` instead now.

For a loop tree, I imagine that each loop would get a unique ID, represented as a `Loop` `EntityRef`. Then you would have something like:

```rust
struct LoopTree {
    loops: EntityMap<Loop, LoopData>,
    ebbs: EntityMap<Ebb, PackedOption<Loop>>,
}

struct LoopData {
    header: Ebb,
    parent: PackedOption<Loop>,
    next: PackedOption<Loop>,
}
```

The is probably the minimum data structure needed to determine which loop contains a given `Ebb`, while making it possible to traverse the loop tree.

It is worth giving it some thought if you really need a list of EBBs per loop. This is actually problematic because it potentially requires quadratic space in the number of EBBs. It is really necessary for an LICM optimization to have those lists?

I am not sure if this data structure is sufficient to compute the loop tree. You may need some helpers.

stoklund:
BTW, is Tarjan's algorithm going to be O(nested loops * EBBs in inner loop)? That could be problematic.

denismerigoux:
I confirm, the LICM pass needs to have for each loop the list of EBBs in the loop. Indeed, for each loop you start from the header EBB and follow the CFG but you need to know which branches exit the loop or not (because you only want to analyse the instructions inside the loop).

From what I understood an Ebb is just a `usize` ref to the EBB's id, are our space constraints so tight that storing these lists is a constraint? 

A single pass of Tarjan's algortihm is O(number of EBBs + number of edges). So if we run it multiple times to detect inner loops the total complexity is indeed O(nested loops * EBBs in inner loop).

The other solution is to use natural loop detection, but this implies one DFS (traversing all the EBBs in the loop) for each loop so the complexity is the same. The advantage with Tarjan's is that it identifies precisely the tree structure of nested loops, so we can limit complexity by limiting the loop nesting level that we are going to optimize.

denismerigoux:
Oh I just understood what you were suggesting with `EntityMap<Ebb, PackedOption<Loop>>`, so that you don't have to store full lists. This could work.

lib/cretonne/src/flowgraph.rs
```diff
@@ -167,10 +168,108 @@ impl ControlFlowGraph {
...
+
+    // Helper recursive function for the Tarjan's SCC algorithm
+    fn strong_connect(&self,
+                      node: &Ebb,
+                      stack: &mut Vec<Ebb>,
+                      components: &mut Vec<HashSet<Ebb>>,
+                      properties: &mut HashMap<Ebb, (u32, u32, bool)>,
+                      index: u32) {
+        properties.insert(node.clone(), (index, index, true));
+        stack.push(node.clone());
```
sunfishcode: FWIW, [Nuutila et al](http://citeseerx.ist.psu.edu/viewdoc/summary?doi=10.1.1.53.5911) give some minor improvements to Tarjan's original algorithm. It's not urgent that we implement these now, but it would be nice to mention them in a TODO comment or so until we get to them.

stoklund:
> From what I understood an Ebb is just a usize ref to the EBB's id, are our space constraints so tight that storing these lists is a constraint?

That's right, an `Ebb` is a 32-bit EBB id. An `EntityMap` is just an array in disguise, indexed by the id. This makes it very efficient, at least if it is fairly densely populated.

We can store the lists if we have to. But it is always better to avoid it if possible.

stoklund:
I was thinking about how we can compute the loop tree in linear time (more or less). Please don't take this too seriously, I probably overlooked something.

Start by discovering the loop headers: Traverse the CFG in a reverse post-order and identify the loop headers as those EBBs that dominate one of their predecessors. The RPO discovery also means that your loop ids form a topological order of the loop tree (right?).

Second, discover the loop blocks for each loop by traversing the loop headers in the reverse order they were discovered. For each header, follow the back-edges and then keep going up the CFG (DFS-like) until you hit the loop header or an EBB that already belongs to a sub-loop. When you hit an EBB from a sub-loop, follow its parent loop pointer until you reach a loop with no parent (or the current loop). Then mark that loop as a child of the current loop, and keep going up the CFG from the loop's header.

The `LoopTree::ebbs` map would be used to keep track of which EBBs have already been visited. I don't think additional data structures are necessary.

denismerigoux:
I've implemented your algorithm @stoklund and it seems to be working quite well, at least on my test cases that contain a single-block loop, a multi-block loop and two nested loops.

lib/cretonne/src/loop_analysis.rs
```diff
@@ -0,0 +1,324 @@
...
+                return true;
+            } else {
+                ebb_loop = self.get_loop_parent(ebb_loop.unwrap());
+            }
+        }
+        false
+    }
+
+    /// Return ebbs from a loop in post-order, starting from an entry point in the block.
+    pub fn postorder_ebbs_loop(&self, cfg: &ControlFlowGraph, lp: Loop) -> Vec<Ebb> {
```
stoklund: I'm a little skeptical about this method, it's throwing around a lot of big data structures.

Perhaps you could move it into the LICM module? I don't think it would be used by anything else.

lib/cretonne/src/ir/entities.rs
```diff
@@ -117,6 +117,11 @@ entity_impl!(FuncRef, "fn");
 pub struct SigRef(u32);
 entity_impl!(SigRef, "sig");
 
+/// A opaque reference to a code loop.
+#[derive(Copy, Clone, PartialEq, Eq, Hash, Debug)]
+pub struct Loop(u32);
+entity_impl!(Loop, "loop");
```
stoklund: Oh, you *do* need `ReservedValue` so you can make a `PackedOption<Loop>`.

lib/cretonne/src/loop_analysis.rs
```diff
@@ -0,0 +1,324 @@
...
+use packed_option::PackedOption;
+use std::collections::HashSet;
+
+/// Loop tree information for a single function.
+///
+/// Loops are referenced by the Loop object, and for each loop you can access its header EBB,
+/// its eventual parent in the loop tree and all the EBB belonging to the loop.
+pub struct LoopAnalysis {
+    loops: EntityMap<Loop, LoopData>,
+    ebbs: EntityMap<Ebb, PackedOption<Loop>>,
```
denismerigoux: What about `ebb_loop_map` ?

lib/cretonne/src/loop_analysis.rs
```diff
@@ -0,0 +1,324 @@
...
+    /// a function.
+    pub fn new() -> LoopAnalysis {
+        LoopAnalysis {
+            loops: EntityMap::new(),
+            ebbs: EntityMap::new(),
+        }
+    }
+
+    /// Returns all the loops contained in a function.
+    pub fn get_loops(&self) -> Vec<Loop> {
```
denismerigoux: In fact I need to iterate the `Loop`s in reverse order for the second pass of the loop detections so I would need to implement `std::iter::DoubleEndedIterator` for `entity_map::Keys`.

stoklund: I think both are simple to implement. The `Keys` iterator is pretty trivial.

lib/cretonne/src/loop_analysis.rs
denismerigoux: This method is basically a copy-paste of `flowgraph::postorder_ebbs` with the difference that we start by a specified entry blocks and we ignore basic blocks that are not in a certain loop. To reduce code duplication, the best solution would be to include in `flowgraph` a postorder traversal taking as argument an entry block and a function `Ebb -> bool` specifying which `Ebb`s to traverse. Then `postorder_ebbs` could just call this function with the entry block of the CFG and the function always returning true, while `postorder_ebbs_loop` could pass in its own parameters. I had tried to do this but had difficulties with the borrow checker to pass a function using the `LoopAnalysis` data structure as an argument to another function.

As a workaround, I will move this in the LICM module.

lib/cretonne/src/lib.rs
```diff
@@ -15,6 +15,7 @@ pub mod dbg;
 
 pub mod binemit;
 pub mod dominator_tree;
+pub mod loop_analysis;
```
denismerigoux: Done.

lib/cretonne/src/lib.rs
```diff
@@ -36,5 +37,6 @@ mod partition_slice;
 mod predicates;
 mod ref_slice;
 mod simple_gvn;
+mod licm;
```
denismerigoux: Fixed.

lib/cretonne/src/ir/entities.rs
denismerigoux: Moved into `loop_analysis`.

lib/cretonne/src/licm.rs
```diff
@@ -0,0 +1,175 @@
...
+            .ins(&mut pos)
+            .jump(header, pre_header_args_value.as_slice(pool));
+    }
+    pre_header
+}
+
+
+// Change the destination of a jump or branch instruction. Panics if called with a non-jump
+// or non-branch instruction.
+fn change_branch_jump_destination(inst: Inst, new_ebb: Ebb, func: &mut Function) {
```
denismerigoux: Used `InstructionData::branch_destination_mut`.

lib/cretonne/src/loop_analysis.rs
```diff
@@ -0,0 +1,324 @@
+//! A loop analysis respresented as mappings of loops to their header Ebb
```
denismerigoux: Fixed.

lib/cretonne/src/loop_analysis.rs
denismerigoux: Implemented `std::iter::DoubleEndedIterator`.

lib/cretonne/src/loop_analysis.rs
```diff
@@ -0,0 +1,324 @@
...
+    /// Returns all the loops contained in a function.
+    pub fn get_loops(&self) -> Vec<Loop> {
+        self.loops.keys().collect()
+    }
+
+    /// Returns the header EBB of a particular loop.
+    ///
+    /// The characteristic property of a loop header block is that it dominates some of its
+    /// predecessors.
+    pub fn get_loop_header(&self, lp: Loop) -> Ebb {
```
denismerigoux: Removed all `get_` prefixes.

lib/cretonne/src/loop_analysis.rs
```diff
@@ -0,0 +1,324 @@
...
+        self.loops[lp].parent.expand()
+    }
+
+    /// Determine if an Ebb belongs to a loop by running a finger along the loop tree.
+    ///
+    /// Returns `true` if `ebb` is in loop `lp`.
+    pub fn is_in_loop(&self, ebb: Ebb, lp: Loop) -> bool {
+        let mut ebb_loop = self.ebbs[ebb].expand();
+        while ebb_loop != None {
+            if ebb_loop.unwrap() == lp {
```
denismerigoux: Used `while let` and moved loop into its separate helper function.

lib/cretonne/src/loop_analysis.rs
```diff
@@ -0,0 +1,324 @@
...
+        // We traverse the CFg in reverse postorder
+        for ebb in cfg.postorder_ebbs().iter().rev() {
+            for &(_, pred_inst) in cfg.get_predecessors(*ebb) {
+                // If the ebb dominates one of its predecessors it is a back edge
+                if domtree.ebb_dominates(ebb.clone(), pred_inst, layout) {
+                    // This ebb is a loop header, so we create its associated loop
+                    let lp = self.loops
+                        .push(LoopData {
+                                  header: *ebb,
+                                  parent: PackedOption::from(None),
```
denismerigoux: Created `LoopData::new()`.

lib/cretonne/src/loop_analysis.rs
```diff
@@ -0,0 +1,324 @@
...
+                              });
+                    self.ebbs[*ebb] = PackedOption::from(lp);
```
denismerigoux: Fixed.

lib/cretonne/src/loop_analysis.rs
```diff
@@ -0,0 +1,324 @@
...
+    // discovers all the ebb belonging to the loop and its inner loops. After a call to this
+    // function, the loop tree is fully constructed.
+    fn discover_loop_blocks(&mut self,
+                            cfg: &ControlFlowGraph,
+                            domtree: &DominatorTree,
+                            layout: &Layout) {
+        let mut stack: Vec<Ebb> = Vec::new();
+        // We handle each loop header in reverse order, corresponding to a pesudo postorder
+        // traversal of the graph.
+        for lp in self.get_loops().iter().rev() {
```
denismerigoux: Implemented `std::iter::DoubleEndedIterator`.

lib/cretonne/src/loop_analysis.rs
```diff
@@ -0,0 +1,324 @@
...
+            for &(pred, pred_inst) in cfg.get_predecessors(self.loops[*lp].header) {
+                // We follow the back edges
+                if domtree.ebb_dominates(self.loops[*lp].header, pred_inst, layout) {
+                    stack.push(pred);
+                }
+            }
+            while !stack.is_empty() {
+                let node = stack.pop().unwrap();
```
denismerigoux: Fixed.

lib/cretonne/src/loop_analysis.rs
```diff
@@ -0,0 +1,324 @@
...
+                let continue_dfs: Option<Ebb>;
+                if self.ebbs[node].is_none() {
```
denismerigoux: Fixed.

lib/cretonne/src/loop_analysis.rs
```diff
@@ -0,0 +1,324 @@
...
+                                // If lp is a one-block loop then we make sure we stop
+                                continue_dfs = None
+                            }
+                        }
+                    }
+                }
+                // Now we have handled the popped node and need to continue the DFS by adding the
+                // predecessors of that node
+                if continue_dfs != None {
+                    for &(pred, _) in cfg.get_predecessors(continue_dfs.unwrap()) {
```
denismerigoux: Fixed.

denismerigoux:
I've fixed all of the review items and I'll try to work my way around git to separate the LICM PR from another loop analysis PR.

stoklund:
Thanks!

denismerigoux:
Pull request obsolete, replaced with https://github.com/stoklund/cretonne/pull/86.

PR #86, opened 2017-06-02
Loop analysis of the IL

denismerigoux:
* Implemented in two passes
* First pass discovers the loops headers (they dominate one of their predecessors)
* Second pass traverses the blocks of each loop
* Discovers the loop tree structure
* Offers a new LoopAnalysis data structure queried from outside the module

### DIFF
--- a/lib/cretonne/src/loop_analysis.rs
+++ b/lib/cretonne/src/loop_analysis.rs
@@ -1,0 +1,336 @@
+//! A loop analysis represented as mappings of loops to their header Ebb
+//! and parent in the loop tree.
+
+use ir::{Function, Ebb, Layout};
+use flowgraph::ControlFlowGraph;
+use dominator_tree::DominatorTree;
+use entity_map::{EntityMap, PrimaryEntityData};
+use packed_option::{PackedOption, ReservedValue};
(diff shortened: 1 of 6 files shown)


denismerigoux:
When this pull request is approved, I will propose another for the LICM pass (https://github.com/denismerigoux/cretonne/tree/licm).

lib/cretonne/src/loop_analysis.rs
```diff
@@ -0,0 +1,336 @@
...
+    }
+
+    // Traverses the CFG in reverse postorder and create a loop object for every EBB having a
+    // back edge.
+    fn find_loop_headers(&mut self,
+                         cfg: &ControlFlowGraph,
+                         domtree: &DominatorTree,
+                         layout: &Layout) {
+        // We traverse the CFg in reverse postorder
+        for ebb in cfg.postorder_ebbs().iter().rev() {
```
stoklund: Use `&ebb` here, so you can avoid `ebb.clone()` and `*ebb` when you use it below.

denismerigoux:
For historic purposes, most of the discussion concerning the implementation is in https://github.com/stoklund/cretonne/pull/84.

PR #87, opened 2017-06-02
LICM pass 

denismerigoux:
* Uses loop analysis to detect loop tree
* For each loop (starting with the inner ones), create a pre-header and move there loop-invariant instructions
* An instruction is loop invariant if it does not use as argument a value defined earlier in the loop
* File tests to check LICM's correctness

I'm sorry, for some reason GitHub detects conflict for the small style change implemented by 7b7bd7d.

### DIFF
--- a/lib/cretonne/src/licm.rs
+++ b/lib/cretonne/src/licm.rs
@@ -1,6 +1,6 @@
 //! A Loop Invariant Code Motion optimization pass
 
-use ir::{Function, Ebb, Inst, Value, Cursor, Type, InstBuilder};
+use ir::{Function, Ebb, Inst, Value, Cursor, Type, InstBuilder, Layout};
 use flowgraph::ControlFlowGraph;
 use std::collections::HashSet;
 use dominator_tree::DominatorTree;
@@ -22,11 +22,21 @@ pub fn do_licm(func: &mut Function,
         // Then we create the loop's pre-header and fill it with the invariant instructions
         // Then we remove the invariant instructions from the loop body
         if invariant_inst.len() > 0 {
-            let pre_header =
-                create_pre_header(loop_analysis.loop_header(lp).clone(), func, cfg, domtree);
+            // If the loop has a natural pre-header we use it, otherwise we create it.
+            let pre_header = match has_pre_header(&func.layout,
+                                                  cfg,
+                                                  domtree,
+                                                  loop_analysis.loop_header(lp).clone()) {
+                None => {
+                    create_pre_header(loop_analysis.loop_header(lp).clone(), func, cfg, domtree)
+                }
+                Some(pre_hd) => pre_hd,
+            };
             let mut pos = Cursor::new(&mut func.layout);
-            pos.goto_top(pre_header);
-            pos.next_inst();
+            pos.goto_bottom(pre_header);
+            // The last instruction of the pre-header is the termination instruction (usually
+            // a jump) so we need to insert just before this.
+            pos.prev_inst();
             for inst in invariant_inst.iter() {
                 pos.insert_inst(inst.clone());
             }
@@ -77,6 +87,32 @@ fn create_pre_header(header: Ebb,
             .jump(header, pre_header_args_value.as_slice(pool));
     }
     pre_header
+}
+
+// Detects if a loop header has a natural pre-header.
+//
+// A loop header has a pre-header if there is only one predecessor that the header doesn't
+// dominate.
+fn has_pre_header(layout: &Layout,
+                  cfg: &ControlFlowGraph,
+                  domtree: &DominatorTree,
+                  header: Ebb)
+                  -> Option<Ebb> {
+    let mut result = None;
+    let mut found = false;
+    for &(pred_ebb, last_inst) in cfg.get_predecessors(header) {
+        // We only count normal edges (not the back edges)
+        if !domtree.ebb_dominates(header.clone(), last_inst, layout) {
+            if found {
+                // We have already found one, there are more than one
+                return None;
+            } else {
+                result = Some(pred_ebb);
+                found = true;
+            }
+        }
+    }
+    result
 }
 
 
--- a/lib/cretonne/src/licm.rs
+++ b/lib/cretonne/src/licm.rs
@@ -1,0 +1,162 @@
+//! A Loop Invariant Code Motion optimization pass
+
+use ir::{Function, Ebb, Inst, Value, Cursor, Type, InstBuilder};
+use flowgraph::ControlFlowGraph;
+use std::collections::HashSet;
+use dominator_tree::DominatorTree;
+use entity_list::{EntityList, ListPool};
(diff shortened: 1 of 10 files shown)


denismerigoux:
Most of the discussion concerning the implementation is in #84.

stoklund:
Just some high-level comments:

- It looks like you create a pre-header even if a loop already has a pre-header. Could that be avoided by identifying an existing pre-header?
- I think this algorithm has the same quadratic behavior as the first version of the loop analysis. It's O(EBBs in inner loop * nesting depth).

Is there some way of modifying the iteration so you don't have to visit those inner loop blocks over and over?

denismerigoux:
- It is true that if the loop header has only one predecessor that it doesn't dominate, then this predecessor is the pre-header and we can just insert invariant instructions at the end of it without recreating a new pre-header. I can implement that.
- Actually the algorithm only visit each inner-loop block once: indeed, when the traversal discover a block belonging to a child loop, it jumps directly to the predecessors of the header block of the inner loop (or the most outer inner loop by following the already constructed loop tree).

The second point is the reason why your algorithm seems to be more efficient than O(EBBs in inner loop * nesting depth).

lib/cretonne/src/licm.rs
```diff
@@ -79,6 +89,32 @@ fn create_pre_header(header: Ebb,
...
+// A loop header has a pre-header if there is only one predecessor that the header doesn't
+// dominate.
+fn has_pre_header(layout: &Layout,
+                  cfg: &ControlFlowGraph,
+                  domtree: &DominatorTree,
+                  header: Ebb)
+                  -> Option<Ebb> {
+    let mut result = None;
+    let mut found = false;
+    for &(pred_ebb, last_inst) in cfg.get_predecessors(header) {
```
stoklund: You should verify that `last_inst` really is the terminator of `pred_ebb`. Otherwise you could be inserting code in a place that isn't post-dominated by the loop header.

It's probably also a good idea to add a test case for this special case.

denismerigoux: Indeed, I had thought about this this weekend and was about to fix it.

stoklund:
I think we could land this in its current state if you rebase it.

denismerigoux:
I tried `git rebase licm master` and then `git push` but it was rejected and I had to do a merge commit.

denismerigoux:
And it's rebased !

lib/cretonne/src/licm.rs
```diff
@@ -0,0 +1,208 @@
...
+                // If there is a natural pre-header we insert new instructions just before the
+                // related jumping instruction (which is not necessarily at the end).
+                Some((_, last_inst)) => {
+                    pos = Cursor::new(&mut func.layout);
+                    pos.goto_inst(last_inst);
+                }
+            };
+            // The last instruction of the pre-header is the termination instruction (usually
+            // a jump) so we need to insert just before this.
+            for inst in invariant_inst.iter() {
```
sunfishcode: The `.iter()` here isn't necessary.

lib/cretonne/src/licm.rs
```diff
@@ -0,0 +1,208 @@
...
+// Change the destination of a jump or branch instruction. Does nothing if called with a non-jump
+// or non-branch instruction.
+fn change_branch_jump_destination(inst: Inst, new_ebb: Ebb, func: &mut Function) {
+    match func.dfg[inst].branch_destination_mut() {
+        None => (),
+        Some(instruction_dest) => *instruction_dest = new_ebb,
+    }
+}
+
+// Traverses a loop in reverse post-order from a header EBB and identify lopp-invariant
```
sunfishcode: Typo: lopp

lib/cretonne/src/licm.rs
```diff
@@ -0,0 +1,208 @@
...
+// instructions. Theseloop-invariant instructions are then removed from the code and returned
```
sunfishcode: Typo: Theseloop

lib/cretonne/src/licm.rs
```diff
@@ -0,0 +1,208 @@
...
+               func.dfg
+                   .inst_args(inst)
+                   .into_iter()
+                   .all(|arg| !loop_values.contains(arg)) {
+                // If all the instruction's argument are defined outside the loop
+                // then this instruction is loop-invariant
+                invariant_inst.push(inst);
+                // We remove it from the loop
+                pos.remove_inst();
+                pos.prev_inst();
```
sunfishcode: You can do `remove_inst()` and `prev_inst()` at the same time, with `remove_inst_and_step_back()` (and if you can think of a better name for that function, I'm all ears :-)).

lib/cretonne/src/licm.rs
```diff
@@ -0,0 +1,208 @@
...
+fn create_pre_header(header: Ebb,
+                     func: &mut Function,
+                     cfg: &mut ControlFlowGraph,
+                     domtree: &DominatorTree)
+                     -> Ebb {
+    let pool = &mut ListPool::<Value>::new();
+    let header_args_values: Vec<Value> = func.dfg
+        .ebb_args(header)
+        .into_iter()
+        .map(|val| *val)
```
sunfishcode: You can use `.cloned()` instead of `.map(|val| *val)`.

lib/cretonne/src/licm.rs
```diff
@@ -0,0 +1,208 @@
...
+        }
+    }
+    invariant_inst
+}
+
+/// Return ebbs from a loop in post-order, starting from an entry point in the block.
+pub fn postorder_ebbs_loop(loop_analysis: &LoopAnalysis,
+                           cfg: &ControlFlowGraph,
+                           lp: Loop)
+                           -> Vec<Ebb> {
```
sunfishcode: `postorder_ebbs_loop` currently is only used by licm, so as things are right now, the `pub` here isn't needed. If you're anticipating reusing this function elsewhere, I suggest moving it to a different file so that licm.rs doesn't have things unrelated to LICM depending on it.

lib/cretonne/src/licm.rs
denismerigoux: Fixed.

lib/cretonne/src/licm.rs
denismerigoux: Fixed.

lib/cretonne/src/licm.rs
denismerigoux: Fixed.

lib/cretonne/src/licm.rs
denismerigoux: Fixed, and I like explicit names ;)

lib/cretonne/src/licm.rs
denismerigoux: Removed the pub, I don't see much use for a postorder traversal of a specific group of nodes starting from an arbitrary entry point.

PR #90, opened 2017-06-03
Updated the regex crate to 0.2.2 as per issue #88

### LINKED ISSUE #88: Upgrade regex crate in filecheck
Currently in `lib/filecheck/Cargo.toml` we have `regex = "0.1.71"`, the plan is:
- replace it by `regex = "0.2.2"`
- `cargo build` and follow-fix compiler errors
- `cargo test` to make sure tests pass

This is a tedious task but suitable for starters as it doesn't require deep knowledge about the project (I think). Also https://github.com/rust-lang/regex/blob/master/CHANGELOG.md can help you.

codri:
Hi,

This should hopefully resolve #88, the tests are passing.
Please let me know if I missed something.

Thanks

### DIFF
--- a/lib/filecheck/src/checker.rs
+++ b/lib/filecheck/src/checker.rs
@@ -30,8 +30,8 @@ const DIRECTIVE_RX: &str = r"\b(check|sameln|nextln|unordered|not|regex):\s+(.*)
 impl Directive {
     /// Create a new directive from a `DIRECTIVE_RX` match.
     fn new(caps: Captures) -> Result<Directive> {
-        let cmd = caps.get(1).map_or("", |m| m.as_str());
-        let rest = caps.get(2).map_or("", |m| m.as_str());
+        let cmd = caps.get(1).map(|m| m.as_str()).expect("group 1 must match");
+        let rest = caps.get(2).map(|m| m.as_str()).expect("group 2 must match");
 
         if cmd == "regex" {
             return Directive::regex(rest);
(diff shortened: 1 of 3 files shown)


lib/filecheck/src/checker.rs
```diff
@@ -30,8 +30,8 @@ const DIRECTIVE_RX: &str = r"\b(check|sameln|nextln|unordered|not|regex):\s+(.*)
 impl Directive {
     /// Create a new directive from a `DIRECTIVE_RX` match.
     fn new(caps: Captures) -> Result<Directive> {
-        let cmd = caps.at(1).expect("group 1 must match");
-        let rest = caps.at(2).expect("group 2 must match");
+        let cmd = caps.get(1).map_or("", |m| m.as_str());
```
codri: I wonder if a map and and expect would more closely match(not pun intended) the original.

codri:
Not sure what the fail means. I've tried an ran cargo fmt and it didn't change any files, assuming it likes it this way.

Is it a formatting lint or something?

codri:
It seems to have got past the formatting error, now it seems to fail with valid errors, most likely introduced by me, will try to investigate and fix.

lib/filecheck/src/checker.rs
```diff
@@ -337,23 +338,23 @@ impl<'a> State<'a> {
         } else {
             // We need the captures to define variables.
             rx.captures(txt).map(|caps| {
-                let matched_range = caps.pos(0).expect("whole expression must match");
+                let matched_range = caps.get(0).expect("whole expression must match");
                 for var in defs {
-                    let txtval = caps.name(var).unwrap_or("");
+                    let txtval = caps.name(var).map(|_| "var").unwrap_or("");
```
stoklund: Huh? I'm pretty sure this is the cause of the test failures.

stoklund:
Thanks for working on this. You can run tests locally with `./test-all.sh`.

lib/filecheck/src/checker.rs
codri: oh. yeah, lol. I meant var. since name() now returns an option of Match.
let me patch this quickly.

codri:
Fingers crossed. This time hopefully it will work, as test-all seems to have passed locally.

codri:
Will try to pick up some of the other things.
Definitely not all of them are as easy as this one.

Issue #20, opened 2016-11-02
Document NaN propagation semantics


stoklund:
The language reference is not so clear on what happens when floating point arithmetic instructions produce a NaN. We should specify that.

- Can instructions return a signaling NaN or only quiet NaNs?
- When propagating a NaN input, can the payload bits change?

The specification should be tight enough that WebAssembly semantics are simple to represent, and loose enough that we don't need lots of NaN adjustment code emitted for `fadd` on common targets.

Copying the WebAssembly semantics is probably a good choice.

serprex:
For reference: https://github.com/WebAssembly/design/pull/660

sunfishcode:
Some background questions:

 - Can Cretonne depend on the IEEE 754-2008-recommended encoding for signaling NaN? This would simplify some things in the NaN rules, with the only practical downside being incompatibility with older (though still in use) MIPS CPUs. For reference, [WebAssembly has this dependency](https://github.com/WebAssembly/design/issues/976), though Rust doesn't.

 - Should Cretonne keep the non-associative `minnum`/`maxnum` operators?
     - On one hand, these map to instructions on ARM and Power.
     - On the other, they need special-casing in the NaN rules, IEEE 754-2018 is expected to remove them, WebAssembly omitted them, and [Rust doesn't depend on their particular sNaN behavior](https://github.com/rust-lang/rust/pull/42430).

stoklund:
- I think we should follow WebAssembly on NaN representations for now. We can revisit if old MIPS chips become relevant for us.
- I would be OK with removing `fmaxnum` and `fminnum`. We should add a note to the remaining `fmin` and `fmax` to make it clear that they are not the functions from IEEE 754-2008.

PR #91, opened 2017-06-04
Binary function names

### LINKED ISSUE #47: Binary function names
Cretonne compiles functions independently, so function names are used differently than in LLVM. They serve two purposes:

1. In `.cton` test cases, the function names are all ASCII, and are used to identify individual test cases in the same file.
2. When Cretonne is embedded as a JIT compiler, function names are used to identify other functions that may be called. This identifier can be any sequence of bytes, it doesn't have to be ASCII or UTF-8. Cretonne doesn't interpret these function names, they are opaque identifiers.

The binary function names are not well supported. They get printed out as quoted ASCII with a bunch of backslash escapes.

- The parser doesn't supported the quoted format of function names.
- If the name is a binary encoding of some identifier, ASCII with escapes is not a good representation.
- Right now, function names like `v7` or `ebb4` get printed without quotes, and the lexer recognizes them as value and EB tokens.

# Alternative function name syntax.

Over in #24, I proposed two new identifier/data tokens: `%nnnn` and `#xxxx`. We should use these to represent function names everywhere:

1. If the function name consists entirely of ASCII alphanumerical characters and `_`, use the `%nnnn` notation. Note that this also allows for names like `%0`. There is no need to give special treatment to the first character.
2. If the function name contains any other characters, use the `#xxxx` hexadecimal representation of the function name bytes.
3. If the name is empty, use  a special syntax, maybe `noname` (no `%`).

With these changes, the parser should stop accepting unquoted identifiers as function names.

# Binary name representation.

Currently, the `FuncName` struct contains a `String`:

```rust
pub struct FunctionName(String);
```

This restricts us to names in UTF-8 form. We should accept any sequence of bytes, so change this to:
```rust
pub struct FunctionName(Vec<u8>);
```

# Allocation-free representation

For extra credit: Cretonne tries to minimize heap allocations everywhere, so an internal short-string optimization would make a lot of sense:

```rust
enum NameRepr {
    Short {
        length: u8,
        bytes: [u8;12],
    }
    Long(Vec<u8>),
}
```

zummenix:
Eventually will close #47

I hope I move in right direction.

### DIFF
--- a/lib/cretonne/src/ir/funcname.rs
+++ b/lib/cretonne/src/ir/funcname.rs
@@ -6,8 +6,6 @@
 use std::fmt::{self, Write};
 use std::ascii::AsciiExt;
 
-const HEX_CHARS: &'static [u8] = b"0123456789abcdef";
-
 /// The name of a function can be any sequence of bytes.
 ///
 /// Function names are mostly a testing and debugging tool.
@@ -57,9 +55,8 @@ impl fmt::Display for FunctionName {
             write!(f, "%{}", name)
         } else {
             f.write_char('#')?;
-            for i in self.0.iter().map(|&b| b as usize) {
-                f.write_char(HEX_CHARS[i >> 4] as char)?;
-                f.write_char(HEX_CHARS[i & 0xf] as char)?;
+            for byte in &self.0 {
+                write!(f, "{:02x}", byte)?;
             }
             Ok(())
         }
--- a/lib/cretonne/src/ir/funcname.rs
+++ b/lib/cretonne/src/ir/funcname.rs
@@ -65,8 +65,11 @@ mod tests {
         assert_eq!(FunctionName::with_str("x").to_string(), "%x");
         assert_eq!(FunctionName::with_str("x_1").to_string(), "%x_1");
         assert_eq!(FunctionName::with_str(" ").to_string(), "#20");
-        assert_eq!(FunctionName::with_str("кретон").to_string(), "#d0bad180d0b5d182d0bed0bd");
-        assert_eq!(FunctionName::with_str("印花棉布").to_string(), "#e58db0e88ab1e6a389e5b883");
-        assert_eq!(FunctionName::new(vec![0, 1, 2, 3, 4, 5]).to_string(), "#000102030405");
+        assert_eq!(FunctionName::with_str("кретон").to_string(),
+                   "#d0bad180d0b5d182d0bed0bd");
+        assert_eq!(FunctionName::with_str("印花棉布").to_string(),
+                   "#e58db0e88ab1e6a389e5b883");
+        assert_eq!(FunctionName::new(vec![0, 1, 2, 3, 4, 5]).to_string(),
+                   "#000102030405");
     }
 }
--- a/lib/cretonne/src/ir/funcname.rs
+++ b/lib/cretonne/src/ir/funcname.rs
@@ -16,14 +16,25 @@ const HEX_CHARS: &'static [u8] = b"0123456789abcdef";
 pub struct FunctionName(Vec<u8>);
 
 impl FunctionName {
-    /// Creates a new function name from a string.
-    pub fn with_str(s: &str) -> FunctionName {
-        FunctionName::new(s.into())
-    }
-
     /// Creates a new function name from a sequence of bytes.
-    pub fn new(bytes: Vec<u8>) -> FunctionName {
-        FunctionName(bytes.into())
+    ///
+    /// # Examples
+    ///
+    /// ```rust
+    /// # use cretonne::ir::FunctionName;
+    /// // Create `FunctionName` from a string.
+    /// let name = FunctionName::new("hello");
+    /// assert_eq!(name.to_string(), "%hello");
+    ///
+    /// // Create `FunctionName` from a sequence of bytes.
+    /// let bytes: &[u8] = &[10, 9, 8];
+    /// let name = FunctionName::new(bytes);
+    /// assert_eq!(name.to_string(), "#0a0908");
+    /// ```
+    pub fn new<T>(v: T) -> FunctionName
+        where T: Into<Vec<u8>>
+    {
+        FunctionName(v.into())
     }
 }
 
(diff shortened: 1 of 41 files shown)


lib/cretonne/src/write.rs
```diff
@@ -365,26 +365,26 @@ mod tests {
     #[test]
     fn basic() {
         let mut f = Function::new();
-        assert_eq!(f.to_string(), "function \"\"() {\n}\n");
+        assert_eq!(f.to_string(), "function %() {\n}\n");
```
stoklund: Yep. I think this is an OK syntax for the empty function name.

lib/cretonne/src/ir/funcname.rs
```diff
@@ -14,65 +13,30 @@ use std::ascii::AsciiExt;
 pub struct FunctionName(String);
 
 impl FunctionName {
-    /// Create new function name equal to `s`.
-    pub fn new<S: Into<String>>(s: S) -> FunctionName {
+    /// Creates a new function name from a string.
+    ///
+    /// Caller should validate that the string contains only
+    /// ASCII alphanumerical characters and `_`.
+    pub fn from_string(s: &str) -> FunctionName {
```
stoklund: I think `with_str` would be a better name for this constructor. The`from_string` name implies a *conversion constructor* which would be expected to consume a `String`. See the [Rust naming conventions](https://doc.rust-lang.org/1.0.0/style/style/naming/README.html).

I don't think the comment is correct. The caller should be allowed to pass any string, and it will be printed in `#xxx` form if it is not pure ASCII.

lib/cretonne/src/ir/funcname.rs
```diff
@@ -14,65 +13,30 @@ use std::ascii::AsciiExt;
...
-            f.write_char('"')?;
-            for c in self.0.chars().flat_map(char::escape_default) {
-                f.write_char(c)?;
-            }
-            f.write_char('"')
-        } else {
-            f.write_str(&self.0)
-        }
+        f.write_char('%')?;
+        f.write_str(&self.0)
```
stoklund: Perhaps easier: `write!(f, "%{}", self.0)`.

I would like for the `Display` implementation to still detect if the function name is pure ASCII so it can be printed with the `%` syntax. If it can't, it should be printed in hexadecimal instead with `#xxxx`. The legal characters in the `%xx` notating is ASCII alphanumerical characters and `_`.

lib/cretonne/src/ir/funcname.rs
```diff
@@ -14,65 +13,30 @@ use std::ascii::AsciiExt;
...
-        assert_eq!(FunctionName::new("").to_string(), "\"\"");
-        assert_eq!(FunctionName::new("x").to_string(), "x");
-        assert_eq!(FunctionName::new(" ").to_string(), "\" \"");
-        assert_eq!(FunctionName::new(" \n").to_string(), "\" \\n\"");
-        assert_eq!(FunctionName::new("a\u{1000}v").to_string(),
-                   "\"a\\u{1000}v\"");
+    fn formatting_string() {
+        assert_eq!(FunctionName::from_string("").to_string(), "%");
+        assert_eq!(FunctionName::from_string("x").to_string(), "%x");
+        assert_eq!(FunctionName::from_string(" ").to_string(), "% ");
```
stoklund: The first two are correct, the second one should print as `#20`.

stoklund:
BTW, some of this may be easier if you switch the internal representation from a `String` to a `Vec<u8>`. Then you won't have to worry about invalid UTF-8 etc.

It's still useful for the parser to have a `with_str` constructor that takes a UTF-8 `&str` like the one you added. See `str::as_bytes()`.

stoklund:
Sorry, didn't mean to close this!

zummenix:
Thank you for the review!

Initially I thought that `FunctionName` will be an enum with cases like `Name` and `Hex` since lexer recognises `%nnnn` or `#xxxx` and these sequences are valid to create `FunctionName`.

So my next actions are:
- `FunctionName` will accept any bytes (two constructors `new` and `with_str`). 
- for `Token::Name` we'll convert `str::as_bytes()`
- for `Token::HexSequence` we'll convert from hex to bytes
- If `FunctionName` can `Display` itself as `%nnnn` it will, if not it'll fallback to `#xxxx`

Is that correct?

stoklund:
Yes, you got it right

zummenix:
@stoklund, It would be helpful to have some dependency or module that can convert to/from hex. I'm hesitant to copy code from other crates.

lib/cretonne/src/ir/funcname.rs
```diff
@@ -6,73 +6,70 @@
...
-}
-
-fn is_id_start(c: char) -> bool {
-    c.is_ascii() && (c == '_' || c.is_alphabetic())
-}
 
-fn is_id_continue(c: char) -> bool {
-    c.is_ascii() && (c == '_' || c.is_alphanumeric())
+    /// Creates a new function name from a sequence of bytes.
+    pub fn new(bytes: Vec<u8>) -> FunctionName {
```
stoklund: I think you should just take a `bytes: &[u8]` slice here. We don't want to force the caller to allocate a `Vec`.

lib/cretonne/src/ir/funcname.rs
```diff
@@ -6,73 +6,70 @@
...
-            }
-            f.write_char('"')
+        if let Some(name) = try_as_name(&self.0) {
+            write!(f, "%{}", name)
         } else {
-            f.write_str(&self.0)
+            f.write_char('#')?;
+            for i in self.0.iter().map(|&b| b as usize) {
+                f.write_char(HEX_CHARS[i >> 4] as char)?;
+                f.write_char(HEX_CHARS[i & 0xf] as char)?;
```
stoklund: You can print a hexadecimal byte like this:

```rust
write!(f, "{:02x}", b)?
```

stoklund:
For parsing the hexadecimal string, you can use `u8::from_str_radix()` after you split the string into two-character parts.

lib/cretonne/src/ir/funcname.rs
zummenix: Thanks, didn't think of it.

lib/cretonne/src/ir/funcname.rs
zummenix: Let's leave one generic constructor:`pub fn new<T>(v: T) -> FunctionName where T: Into<Vec<u8>>` ?

stoklund: Do you mean in addition to a `&[u8]` constructor? That would be OK.

zummenix: No, I mean only one constructor. That way we can create `FunctionName` using `&str` and `&[u8]`

stoklund: I see what you mean. As long as we can construct from both an `&str` and an `&[u8]`, that is fine.

PR #93, opened 2017-06-05
SSA construction on the fly

denismerigoux:
In order to feed WebAssembly as an input to Cretonne and produce native code from it, we need to translate WebAssembly into Cretonne IL. One of the challenges it poses is SSA construction (WebAssembly is not in SSA form).

Here is a proposal for an SSA construction API that can be invoked on the fly while constructing the Cretonne IL control flow graph, block by block.

### DIFF
--- a/lib/cretonne/src/entity_map.rs
+++ b/lib/cretonne/src/entity_map.rs
@@ -57,6 +57,11 @@ impl<K, V> EntityMap<K, V>
         self.elems.get(k.index())
     }
 
+    /// Get a mutable reference to the element at `k` if it exists.
+    pub fn get_mut(&mut self, k: K) -> Option<&mut V> {
+        self.elems.get_mut(k.index())
+    }
+
     /// Is this map completely empty?
     pub fn is_empty(&self) -> bool {
         self.elems.is_empty()
--- a/lib/cretonne/src/entity_map.rs
+++ b/lib/cretonne/src/entity_map.rs
@@ -57,11 +57,6 @@ impl<K, V> EntityMap<K, V>
         self.elems.get(k.index())
     }
 
-    /// Get a mutable reference to the element at `k` if it exists.
-    pub fn get_mut(&mut self, k: K) -> Option<&mut V> {
-        self.elems.get_mut(k.index())
-    }
-
     /// Is this map completely empty?
     pub fn is_empty(&self) -> bool {
         self.elems.is_empty()
--- a/lib/cretonne/src/ir/dfg.rs
+++ b/lib/cretonne/src/ir/dfg.rs
@@ -3,7 +3,7 @@
 use entity_map::{EntityMap, PrimaryEntityData};
 use ir::builder::{InsertBuilder, ReplaceBuilder};
 use ir::extfunc::ExtFuncData;
-use ir::instructions::{Opcode, InstructionData, CallInfo};
+use ir::instructions::{Opcode, InstructionData, CallInfo, BranchInfo};
 use ir::layout::Cursor;
 use ir::types;
 use ir::{Ebb, Inst, Value, Type, SigRef, Signature, FuncRef, ValueList, ValueListPool};
--- a/lib/cretonne/src/ir/dfg.rs
+++ b/lib/cretonne/src/ir/dfg.rs
@@ -3,7 +3,7 @@
 use entity_map::{EntityMap, PrimaryEntityData};
 use ir::builder::{InsertBuilder, ReplaceBuilder};
 use ir::extfunc::ExtFuncData;
-use ir::instructions::{Opcode, InstructionData, CallInfo, BranchInfo};
+use ir::instructions::{Opcode, InstructionData, CallInfo};
 use ir::layout::Cursor;
 use ir::types;
 use ir::{Ebb, Inst, Value, Type, SigRef, Signature, FuncRef, ValueList, ValueListPool};
@@ -718,7 +718,6 @@ impl DataFlowGraph {
             } else {
                 panic!("{} should be an Ebb argument but is not", last_arg_val);
             }
-            true
         }
         num as usize
     }
--- a/lib/cretonne/src/ir/dfg.rs
+++ b/lib/cretonne/src/ir/dfg.rs
@@ -312,6 +312,14 @@ impl DataFlowGraph {
         let data = ValueData::Alias { ty, original: src };
         self.make_value(data)
     }
+
+    /// Makes a value the alias of another.
+    ///
+    /// Erase all information about the value you're transforming to an alias.
+    pub fn alias_value(&mut self, alias: Value, original: Value) {
+        let ty = self.value_type(original);
+        self.values[alias] = ValueData::Alias { ty, original };
+    }
 }
 
 /// Where did a value come from?
--- a/lib/cretonne/src/ir/dfg.rs
+++ b/lib/cretonne/src/ir/dfg.rs
@@ -312,14 +312,6 @@ impl DataFlowGraph {
         let data = ValueData::Alias { ty, original: src };
         self.make_value(data)
     }
-
-    /// Makes a value the alias of another.
-    ///
-    /// Erase all information about the value you're transforming to an alias.
-    pub fn alias_value(&mut self, alias: Value, original: Value) {
-        let ty = self.value_type(original);
-        self.values[alias] = ValueData::Alias { ty, original };
-    }
 }
 
 /// Where did a value come from?
@@ -726,6 +718,7 @@ impl DataFlowGraph {
             } else {
                 panic!("{} should be an Ebb argument but is not", last_arg_val);
             }
+            true
         }
         num as usize
     }
--- a/lib/cretonne/src/ir/instructions.rs
+++ b/lib/cretonne/src/ir/instructions.rs
@@ -333,56 +333,6 @@ impl InstructionData {
         }
     }
 
-    /// Get a mutable reference to the value list of jump arguments of this branch.
-    ///
-    /// Multi-destination branches like `br_table` return `None`.
-    /// TODO: deal with multi-branch instructions
-    pub fn branch_arguments_mut(&mut self, dest: Ebb) -> Option<&mut ValueList> {
-        match self {
-            &mut InstructionData::Jump {
-                     ref mut args,
-                     destination: dest_alt,
-                     ..
-                 } => if dest_alt == dest { Some(args) } else { None },
-            &mut InstructionData::Branch {
-                     ref mut args,
-                     destination: dest_alt,
-                     ..
-                 } => if dest_alt == dest { Some(args) } else { None },
-            &mut InstructionData::BranchIcmp {
-                     ref mut args,
-                     destination: dest_alt,
-                     ..
-                 } => if dest_alt == dest { Some(args) } else { None },
-            _ => None,
-        }
-    }
-
-    /// Get a reference to the value list of jump arguments of this branch.
-    ///
-    /// Multi-destination branches like `br_table` return `None`.
-    /// TODO: deal with multi-branch instructions
-    pub fn branch_arguments(&self, dest: Ebb) -> Option<&ValueList> {
-        match self {
-            &InstructionData::Jump {
-                ref args,
-                destination: dest_alt,
-                ..
-            } => if dest_alt == dest { Some(args) } else { None },
-            &InstructionData::Branch {
-                ref args,
-                destination: dest_alt,
-                ..
-            } => if dest_alt == dest { Some(args) } else { None },
-            &InstructionData::BranchIcmp {
-                ref args,
-                destination: dest_alt,
-                ..
-            } => if dest_alt == dest { Some(args) } else { None },
-            _ => None,
-        }
-    }
-
     /// Return information about a call instruction.
     ///
     /// Any instruction that can call another function reveals its call signature here.
--- a/lib/cretonne/src/ir/instructions.rs
+++ b/lib/cretonne/src/ir/instructions.rs
@@ -358,6 +358,31 @@ impl InstructionData {
         }
     }
 
+    /// Get a reference to the value list of jump arguments of this branch.
+    ///
+    /// Multi-destination branches like `br_table` return `None`.
+    /// TODO: deal with multi-branch instructions
+    pub fn branch_arguments(&self, dest: Ebb) -> Option<&ValueList> {
+        match self {
+            &InstructionData::Jump {
+                ref args,
+                destination: dest_alt,
+                ..
+            } => if dest_alt == dest { Some(args) } else { None },
+            &InstructionData::Branch {
+                ref args,
+                destination: dest_alt,
+                ..
+            } => if dest_alt == dest { Some(args) } else { None },
+            &InstructionData::BranchIcmp {
+                ref args,
+                destination: dest_alt,
+                ..
+            } => if dest_alt == dest { Some(args) } else { None },
+            _ => None,
+        }
+    }
+
     /// Return information about a call instruction.
     ///
     /// Any instruction that can call another function reveals its call signature here.
--- a/lib/cretonne/src/ir/instructions.rs
+++ b/lib/cretonne/src/ir/instructions.rs
@@ -333,6 +333,31 @@ impl InstructionData {
         }
     }
 
+    /// Get a mutable reference to the value list of jump arguments of this branch.
+    ///
+    /// Multi-destination branches like `br_table` return `None`.
+    /// TODO: deal with multi-branch instructions
+    pub fn branch_arguments_mut(&mut self, dest: Ebb) -> Option<&mut ValueList> {
+        match self {
+            &mut InstructionData::Jump {
+                     ref mut args,
+                     destination: dest_alt,
+                     ..
+                 } => if dest_alt == dest { Some(args) } else { None },
+            &mut InstructionData::Branch {
+                     ref mut args,
+                     destination: dest_alt,
+                     ..
+                 } => if dest_alt == dest { Some(args) } else { None },
+            &mut InstructionData::BranchIcmp {
+                     ref mut args,
+                     destination: dest_alt,
+                     ..
+                 } => if dest_alt == dest { Some(args) } else { None },
+            _ => None,
+        }
+    }
+
     /// Return information about a call instruction.
     ///
     /// Any instruction that can call another function reveals its call signature here.
--- a/lib/cretonne/src/ssa.rs
+++ b/lib/cretonne/src/ssa.rs
@@ -409,14 +409,13 @@ impl<Variable> SSABuilder<Variable>
     // Panics if called with a non-header block.
     fn resolve_undef_vars(&mut self, block: Block, dfg: &mut DataFlowGraph) {
         // TODO: find a way to not allocate vectors
-        let (predecessors, undef_vars, ebb): (Vec<(Block, Inst)>,
-                                              Vec<(Variable, Value)>,
-                                              Ebb) = match self.blocks[block] {
-            BlockData::EbbBody(_, _) => panic!("this should not happen"),
-            BlockData::EbbHeader(ref predecessors, _, ebb, ref undef_vars) => {
-                (predecessors.iter().map(|(&x, &y)| (x, y)).collect(), undef_vars.clone(), ebb)
-            }
-        };
+        let (predecessors, undef_vars): (Vec<(Block, Inst)>, Vec<(Variable, Value)>) =
+            match self.blocks[block] {
+                BlockData::EbbBody(_, _) => panic!("this should not happen"),
+                BlockData::EbbHeader(ref predecessors, _, _, ref undef_vars) => {
+                    (predecessors.iter().map(|(&x, &y)| (x, y)).collect(), undef_vars.clone())
+                }
+            };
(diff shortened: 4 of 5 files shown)


lib/cretonne/src/ssa.rs
```diff
@@ -0,0 +1,145 @@
...
+    /// Returns the SSA value corresponding to the current SSA definition of this variable.
+    pub fn use_var(var: Variable, block: Block) -> Value {
+        unimplemented!()
+    }
+
+    /// Declares a new basic block belonging to a certain `Ebb`. If it is the first basic block
+    /// in the `Ebb`, pass `None` for the second argument, otherwise give the predecessor.
+    pub fn declare_block(ebb: Ebb, pred: Option<Block>) -> Block {
+        unimplemented!()
+    }
```
sunfishcode: Are blocks which are not the first in their `Ebb` implicitly sealed by the `declare_block` function? If so, it'd be nice to explicitly document that.

As a related idea, it might clarify this to split this function into two functions, one for the case of the first block in an ebb, and one for blocks that aren't the first, rather than having an Option.

lib/cretonne/src/ssa.rs
```diff
@@ -0,0 +1,145 @@
...
+/// Structure containing the data relevant the construction of SSA for a given function.
+pub struct SSABuilder {
+    // Records for every variable its type and, for every revelant block, the last definition of
+    // the variable in the block.
+    variables: EntityMap<Variable, VariableData>,
+    // Records the position of the basic blocks and the list of values used but not defined in the
+    // block.
+    blocks: EntityMap<Block, BlockData>,
+    // Records the basic blocks at the beginning of the `Ebb`s.
+    ebb_headers: SparseMap<Ebb, (Ebb, Block)>,
```
sunfishcode: What is the `Ebb` in the `(Ebb, Block)` here?

sunfishcode: Ah, I see now. It's because SparseMap needs to be able to compute the key from the value.

lib/cretonne/src/ssa.rs
```diff
@@ -0,0 +1,145 @@
...
+}
+
+struct BlockData {
+    block_predecessor: PackedOption<Block>,
+    ebb: Ebb,
+    undef_values: ValueList,
+}
+
+struct VariableData {
+    current_defs: SparseMap<Block, (Block, Value)>,
```
sunfishcode: And that's the explanation here too.

lib/cretonne/src/ssa.rs
denismerigoux: Indeed, that's the reason. I think this is the tradeoff between `EntityMap` and `SparseMap`, `SparseMap` takes less space if the keys are sparse but the values have to be larger to be able to compute the keys from them.

lib/cretonne/src/ssa.rs
denismerigoux: Yes, I should leave a comment to explain that because it's not intuitive.

stoklund:
I wonder if it is necessary to make a distinction between `Block` and `Ebb`? It seems to me that an algorithm concerned with building SSA form will only ever care about basic blocks that are at the top of an EBB. An internal basic block will never need a phi.

The `EntityMap` and `SparseMap` data structures are somewhat specialized. They are based on arrays indexed by the `EntityRef`, so they use a fair bit of memory. They are best used at the top level, like you are doing in `struct SSABuilder`. Using them in `struct VariableData` is problematic, I think, since you will have many instances of that. Since you are just hashing out the interface now, you could just use standard library data structures in `struct VariableData` for now, and I can nag you about an efficient representation later.

About the `Variable` entity reference: Somebody using this library to build SSA form will probably already have their own representation of local variables. For example, a WebAssembly translator will have local variable numbers. Since you are allocating the `Variable` references in `make_var`, the library user is forced to maintain a mapping from their own variable representation to yours.

An alternative would be to allow the user to provide their own `Variable` references by making it a generic parameter:

```rust
pub struct SSABuilder<Variable>
    where Variable: EntityRef {
```

You can then use the user-provided `Variable` references to index into an `EntityMap` or `SparseMap`. This is very similar to how you are using the user-provided `Ebb` references already.

lib/cretonne/src/ssa.rs
denismerigoux: Yes the blocks which are not the first `Ebb` are implicitely sealed at creation. I'll document this explicitely.

I'll also split this into two functions and have another `declare_ebb_header_predecessor`. Indeed, our SSA algorithm needs to know which basic blocks are the predecessors of an `Ebb`, whereas the control flow graph only records the `Ebb` predecessors.

denismerigoux:
Concerning the distinction between `Block` and `Ebb`: yes, we need it. Indeed, imagine you `use_var` in a basic block `b` and there is no definition of the variable inside `b`. Then you will proceed to search for a definition in the predecessors of `b`. Let `pred` be a predecessor of this basic block. If `pred` is an `Ebb`, it might contain the definition we're looking for but this definition could be after the instruction jumping to `b`, and in that case we don't want to take it into account. 

That is the reason we need `Block` which contains more information than `Ebb`. Nevertheless you're right, an internal basic block of an `Ebb` will never have a phi because it has an unique predecessor.

Thanks for your advice on the data structures, I'll proceed as you said. Very good idea for switching `Variable` to a module parameter, it does implement more efficiently what I was seeking.

lib/cretonne/src/entity_map.rs
```diff
@@ -57,6 +57,11 @@ impl<K, V> EntityMap<K, V>
         self.elems.get(k.index())
     }
 
+    /// Get a mutable reference to the element at `k` if it exists.
+    pub fn get_mut(&mut self, k: K) -> Option<&mut V> {
```
stoklund: You're probably not using this method any more.

lib/cretonne/src/ir/instructions.rs
```diff
@@ -333,6 +333,56 @@ impl InstructionData {
...
+                 } => if dest_alt == dest { Some(args) } else { None },
+            _ => None,
+        }
+    }
+
+    /// Get a reference to the value list of jump arguments of this branch.
+    ///
+    /// Multi-destination branches like `br_table` return `None`.
+    /// TODO: deal with multi-branch instructions
+    pub fn branch_arguments(&self, dest: Ebb) -> Option<&ValueList> {
```
stoklund: See `analyze_branch`.

lib/cretonne/src/ir/instructions.rs
```diff
@@ -333,6 +333,56 @@ impl InstructionData {
         }
     }
 
+    /// Get a mutable reference to the value list of jump arguments of this branch.
+    ///
+    /// Multi-destination branches like `br_table` return `None`.
+    /// TODO: deal with multi-branch instructions
+    pub fn branch_arguments_mut(&mut self, dest: Ebb) -> Option<&mut ValueList> {
```
stoklund: This is a little dangerous because the returned value list contains *all* of the instruction's arguments, including one or two values controlling the branch. So it is not just the EBB arguments.

There is already a `take_value_list()` method which can be used for manipulating these lists. See `legalizer/split.rs` for an example. We could generate a `value_list_mut()` method, if that matches your needs better.

lib/cretonne/src/entity_map.rs
denismerigoux: Indeed, I removed it.

lib/cretonne/src/ir/instructions.rs
denismerigoux: Ok, managed a solution using `take_value_list()` and `analyze_branch()` because I only need to append an value to the argument value list, No need to generate `value_list_mut()`.

lib/cretonne/src/ir/instructions.rs
denismerigoux: Removed, as well as `branch_arguments_mut()`.

lib/cretonne/src/ssa.rs
```diff
@@ -0,0 +1,520 @@
...
+    // The fields are `(predecessor, ebb)`.
+    EbbBody(Block, Ebb),
+}
+impl<Variable> PrimaryEntityData for BlockData<Variable> {}
+
+impl<Variable> BlockData<Variable> {
+    pub fn add_predecessor(&mut self, pred: Block, inst: Inst) {
+        match self {
+            &mut BlockData::EbbBody(_, _) => assert!(false),
+            &mut BlockData::EbbHeader(ref mut predecessors, _, _, _) => {
```
sunfishcode: Can we do `assert!(!sealed);` here, to catch front-ends mistakenly adding a predecessor to a sealed block?

denismerigoux:
I had an idea: using the `SSABuilder` API is quite cumbersome for now, but actually a lot of the calls related to instructions arguments and results could be hidden by an enhanced `InstBuilder` API parametrized by `Variable`, the representation of variables in the original language.

lib/cretonne/src/ssa.rs
```diff
@@ -0,0 +1,520 @@
...
+    pub fn seal_ebb_header_block(&mut self, ebb: Ebb, dfg: &mut DataFlowGraph) {
+        let block = match self.ebb_headers.get(ebb) {
+            None => panic!("this ebb has no block header defined"),
+            Some(&(_, block)) => block,
+        };
+        // TODO: find a way to not allocate vectors
+        let (predecessors, undef_vars): (Vec<(Block, Inst)>, Vec<(Variable, Value)>) =
+            match self.blocks[block] {
+                BlockData::EbbBody(_, _) => panic!("this should not happen"),
+                BlockData::EbbHeader(ref predecessors, _, _, ref undef_vars) => {
```
sunfishcode: Similarly, can we `assert!(!sealed)` here too? It seems like frontends shouldn't be attempting to seal any block twice.

lib/cretonne/src/ssa.rs
denismerigoux: Yes we can.

lib/cretonne/src/ssa.rs
denismerigoux: Hum, so here there is a catch. Indeed, so far I have dissociated _filling_ from _sealing_ when using the API, meaning that you can `use_var` and `def_var` on a sealed block. However, using `use_var` on a sealed block could actually call the function `seal_block` because `seal_block` recursively looks in the predecessors of the `Ebb`, which is exactly what we want.

So a frontend can seal a block twice and it will have no effect since the second time the `undef_vars` list will always be empty.

sunfishcode: Would it make sense to have a separate function be the public API, which can `debug_assert!(!sealed)` or so, with the private one doing the main work? I think you could also move the `self.ebb_headers.get(ebb)` lookup into the public one, because the callsite in `use_var` already knows the `Block`.

lib/cretonne/src/ssa.rs
```diff
@@ -0,0 +1,653 @@
...
+            // The block has multiple predecessors, we register the ebb argument as the current
+            // definition for the variable.
+            Some((val, sealed, ebb)) => {
+                self.def_var(var, val, block);
+                if sealed {
+                    self.seal_ebb_header_block(ebb, dfg)
+                }
+                val
+            }
+        }
```
sunfishcode: Hrm. On the other hand, this may not be easy to do, because `def_var` needs a `&mut self`.

sunfishcode:
I like your idea of making an `InstBuilder` with SSA construction built in. That sounds fun :-).

lib/cretonne/src/ssa.rs
denismerigoux: Yes I can make that, as a sanity check helper for the user calling the API.

lib/cretonne/src/ssa.rs
```diff
@@ -0,0 +1,653 @@
...
+            None => (),
+            Some(var_data) => {
+                match var_data.current_defs.get(&block) {
+                    None => (),
+                    Some(val) => {
+                        return *val;
+                    }
+                }
+            }
+        };
```
denismerigoux: Fixed.

lib/cretonne/src/ssa.rs
```diff
@@ -0,0 +1,653 @@
...
+                    let val = dfg.append_ebb_arg(ebb, ty);
+                    undef_variables.push((var, val));
+                    (block, Some((val, sealed, ebb)))
+                }
+            }
+            BlockData::EbbBody(pred, _) => (pred, None),
+        };
+        match possible_val {
+            // The block has a single predecessor, we look into it.
+            None => self.use_var(dfg, var, ty, new_block),
```
denismerigoux: Oh yes, I don't know why I don't followed the paper for this, memoization seems a good idea here.

lib/cretonne/src/ssa.rs
```diff
@@ -0,0 +1,653 @@
...
+            // The block has multiple predecessors, we register the ebb argument as the current
+            // definition for the variable.
+            Some((val, sealed, ebb)) => {
+                self.def_var(var, val, block);
+                if sealed {
+                    self.seal_ebb_header_block(ebb, dfg)
```
denismerigoux: I added a TODO.

lib/cretonne/src/ssa.rs
denismerigoux: Yeah I know this piece of code is not straightforward, however the first `match` borrows `self.blocks[block]` immutably and it's not possible to borrow it mutably for `def_var` afterwards. That's the cleanliest way I found to achieve this, if you have something better I can change.

lib/cretonne/src/ir/dfg.rs
```diff
@@ -311,6 +311,14 @@ impl DataFlowGraph {
         let data = ValueData::Alias { ty, original: src };
         self.make_value(data)
     }
+
+    /// Makes a value the alias of another.
+    ///
+    /// Erase all information about the value you're transforming to an alias.
+    pub fn alias_value(&mut self, alias: Value, original: Value) {
```
stoklund: See the existing `change_to_alias()` method.

lib/cretonne/src/ir/dfg.rs
```diff
@@ -657,6 +665,28 @@ impl DataFlowGraph {
                         })
     }
 
+    /// Removes the argument `val` from `ebb`. Returns true if the removal is sucessful.
+    pub fn remove_ebb_arg(&mut self, ebb: Ebb, val: Value) -> bool {
```
stoklund: I think this is worth its own PR. A couple points:

- Please include some unit tests in the `dfg::tests` module.
- You may want to take a look at an existing method like `replace_ebb_arg()` for some ideas. 
- You'll need to update the value table. Each EBB argument value knows its argument number, so you can't just slide arguments around. The `value_is_attached()` method should be useful for writing tests of this.
- It could be worthwhile to use semantics similar to the `Vec::swap_remove` method. It is easy to introduce accidental quadratic behavior without it. Of course, it needs to be clear how EBB arguments are moved around by this function so you can manipulate the predecessors accordingly.

lib/cretonne/src/ir/dfg.rs
denismerigoux: Thanks, I didn't see it when I looked for a aliasing method.

lib/cretonne/src/ir/dfg.rs
denismerigoux: Ok, I'll transform this into a proper method. 

* OK for the tests.
* Ok fro the value table.
* Ok for the `swap_remove` semantics, it works for the use in `ssa::SSABuilder` because we process ebb arguments in order and when we remove an arg from an ebb we haven't appended jump arguments corresponding to it or the last ebb argument in the branch instruction.

denismerigoux: I have a semantics problem here. If I want to update the value table after the removal of the Ebb arg, by what should I replace the `ValueData::Arg::...`? Should I transform the method into `remove_ebb_arg_and_change_to_alias`, which is my use case here?

stoklund: You don't need to update the value table entry for the removed argument, but you need to make sure it is consistent for the remaining arguments.

The value table entry points to the EBB argument list, and the EBB argument list points back to the value table. For attached values, this two-way pointing must be consistent. That is what `is_value_attached` tests. When you remove a value from the EBB argument list, it becomes *detached*, which means that `is_value_attached` returns false.

All of the values that appear in the EBB argument list must be attached, meaning that their value table entries points back to their position in the argument list.

denismerigoux: OK, got it!

denismerigoux: I created a pull request concerning only the DFG changes here: https://github.com/stoklund/cretonne/pull/95.

denismerigoux:
So now I need to optimize my data structures and create a optimized representation for pairs of `(Block, Value)`, `(Block, Inst)` and `(Variable,Value)` where `Block`, `Inst`, `Variable` and `Value` have the `EntityRef` trait.

The naïve way of doing that would be to have three new opaque refs `BlockValuePair`, `VariableValuePair`, `BlockInstPair` and 3 associated `EntityMaps`, so that we can store the new opaque refs into `EntityList` inside each `BlockData` or `VariableData`. Is it fine do to that or is it too much memory and reference allocation?

stoklund:
I think it would be good to collect some real-world data before we try to optimize anything.

That means landing this PR as is and building a WebAssembly frontend.

lib/cretonne/src/ssa.rs
```diff
@@ -0,0 +1,716 @@
...
+    // Records for every variable its type and, for every revelant block, the last definition of
+    // the variable in the block.
+    variables: SparseMap<Variable, VariableData<Variable>>,
+    // Records the position of the basic blocks and the list of values used but not defined in the
+    // block.
+    blocks: EntityMap<Block, BlockData<Variable>>,
+    // Records the basic blocks at the beginning of the `Ebb`s.
+    // The value is `(Ebb,Block)` because in a `SparseMap`, the key has to be computable from the
+    // value.
+    ebb_headers: SparseMap<Ebb, (Ebb, Block)>,
```
stoklund: You're probably better off with an `EntityMap<Ebb, PackedOption<Block>>`. At least in terms of memory use.

The use of a tuple here is a bit cryptic.

lib/cretonne/src/ssa.rs
```diff
@@ -0,0 +1,716 @@
...
+}
+
+// Describes the current position of a basic block in the control flow graph.
+enum BlockData<Variable> {
+    // A block at the top of an `Ebb`. Contains the list of known predecessors of this block
+    // and a boolean that indicates if the block is sealed or not.
+    // A block is sealed if all of its predecessors have been declared.
+    // The fields are `(predecessors, sealed, ebb, undef_values)`.
+    EbbHeader(HashMap<Block, Inst>, bool, Ebb, Vec<(Variable, Value)>),
```
stoklund: This is a little too complex for a tuple to my taste. I would prefer a struct with named fields.

lib/cretonne/src/ssa.rs
```diff
@@ -0,0 +1,716 @@
...
+///
+/// - when you have constructed all the predecessor to a basic block at the beginning of an `Ebb`,
+///   call `seal_ebb_header_block` on it with the `Function` that you are building.
+///
+/// This API will give you the correct SSA values to use as arguments of your instructions,
+/// as well as modify the jump instruction and `Ebb` headers arguments to account for the SSA
+/// Phi functions.
+///
+/// # Examples
+/// Here a small programm written in a non-SSA Cretonne IL.
```
stoklund: typo

lib/cretonne/src/ssa.rs
```diff
@@ -0,0 +1,716 @@
...
+/// SSA values to feed to `ebb1`.
+
+impl<Variable> SSABuilder<Variable>
+    where Variable: EntityRef + Hash
+{
+    /// Declares a new definition of a variable in a given basic block.
+    /// The SSA value is passed as an argument because it should be created with
+    /// `ir::DataFlowGraph::append_result`.
+    pub fn def_var(&mut self, var: Variable, val: Value, block: Block) {
+        if match self.variables.get_mut(var) {
```
stoklund: This whole thing would be simpler if `self.variables` were and `EntityMap` and `VariableData` implemented `Default`. Then you could just use `self.variables.ensure(var).current_defs.insert(...)`.

lib/cretonne/src/ssa.rs
denismerigoux: Yes you're right, since eventually all `Ebb`s are going to be declared here. I switched to an `EntityMap`.

lib/cretonne/src/ssa.rs
denismerigoux: Done, created `EbbHeaderBlockData`.

lib/cretonne/src/ssa.rs
denismerigoux: Fixed.

lib/cretonne/src/ssa.rs
denismerigoux: Ok, I abandonned the `SparseMap` and switched to an `EntityMap` which makes more sense, as long as `Variable` implementes `Default`.

denismerigoux:
Outdated, closed in favor of https://github.com/stoklund/cretonne/pull/97

PR #95, opened 2017-06-09
Improved DFG API

denismerigoux:
Added `remove_ebb_args` and `append_branch_argument`. `remove_ebb_args` has `Vec::swap_remove` semantics.

### DIFF
--- a/lib/cretonne/src/ir/dfg.rs
+++ b/lib/cretonne/src/ir/dfg.rs
@@ -3,7 +3,7 @@
 use entity_map::{EntityMap, PrimaryEntityData};
 use ir::builder::{InsertBuilder, ReplaceBuilder};
 use ir::extfunc::ExtFuncData;
-use ir::instructions::{Opcode, InstructionData, CallInfo, BranchInfo};
+use ir::instructions::{Opcode, InstructionData, CallInfo};
 use ir::layout::Cursor;
 use ir::types;
 use ir::{Ebb, Inst, Value, Type, SigRef, Signature, FuncRef, ValueList, ValueListPool};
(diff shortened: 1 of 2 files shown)


lib/cretonne/src/ir/dfg.rs
```diff
@@ -562,6 +562,22 @@ impl DataFlowGraph {
                         })
     }
 
+    /// Append a new value argument to the branch jump to a given destination.
+    ///
+    /// Panics if the instruction does not branch.
+    pub fn append_branch_argument(&mut self, inst: Inst, new_arg: Value) {
```
stoklund: Should we generalize this to an `append_inst_arg()` function? I could see it being useful for calls too, perhaps.

lib/cretonne/src/ir/dfg.rs
```diff
@@ -682,6 +698,51 @@ impl DataFlowGraph {
...
+    /// Removes `val` as an `Ebb` argument and transform it into an alias of `alias_val`. This
+    /// means that all the uses of `val` will refer to `alias_val`.
+    ///
+    /// *Important*: to ensure O(1) deletion, this method swaps the removed argument with the
+    /// last `Ebb` argument. This can disrupt all the branch instructions jumping to this
+    /// `Ebb` for which you have to change the jump argument order if necessary.
+    ///
+    /// Panics if `val` is not an `Ebb` argument. Returns `true` if `Ebb` arguments have been
+    /// swapped.
+    pub fn remove_ebb_arg(&mut self, val: Value) -> bool {
```
stoklund: For safety, I think we should call this `swap_remove_ebb_arg`. Just so people know how it behaves.

I think it would be useful to return `num as usize` from this function instead of a bool. That will be useful for a caller who needs to update all of the predecessor branches. They can call `swap_remove` with that index (adjusted for non-EBB args).

This would require implementing `swap_remove` for `EntityList`, of course. That method could also simplify your implementation below:

```rust
self.ebbs[ebb].args.swap_remove(num, &mut self.value_lists);
if let Some(last_arg_val) = self.ebbs[ebb].args.get(num, &self.value_lists) {
       // We update the position of the old last arg.
       if let ValueData::Arg { num: ref mut old_num, .. } = self.values[last_arg_val] {
           *old_num = num;
       } else {
         panic!("{} should be an Ebb argument but is not", last_arg_val);
    }
}
```

In the case where `swap_remove` removed the last element, `get(num, ...)` will return `None`.

lib/cretonne/src/ir/dfg.rs
denismerigoux: Yes, can do.

lib/cretonne/src/ir/dfg.rs
denismerigoux: Fixed.

Issue #76, opened 2017-04-24
Enumerate all legal encodings of an instruction


stoklund:
The `TargetIsa::encode()` function returns the first encoding of an instruction where both ISA and instruction predicates are satisfied. We also need a way of enumerating all of the possible encodings of an instruction, probably by creating an iterator that returns encodings.

Instruction encodings can have multiple constraints that are checked at different points in the compiler pipeline:

1. ISA predicates depend only on static settings like available CPU and instruction set features. For example, the `imul.i32` instruction can only be encoded if the RISC-V "M" extension is supported by the CPU.
2. Instruction predicates check if the immediate operands can be encoded. For example, `iadd_imm.i32` can only be encoded as a RISC-V `addi` instruction if the immediate is a 12-bit signed integer.
3. Register constraints. These are described by the `RecipeConstraints` available from `EncInfo` and satisfied by the register allocator.
4. Branch range constraints. These are described by the `BranchRange` constraints also available from `EncInfo`.

Only the first two types of constraints are checked by the iterator, just like `TargetISa::encode()` does it. This iterator would be used by:

- The encoding verifier will permit any of the valid encodings instead of requiring a unique one as it does now.
- The branch relaxation pass (#73) will use it to search for a different branch encoding with more relaxed range constraints. For example, x86 branches can be encoded with an 8-bit or a 32-bit displacement.
- An instruction compression pass (which doesn't exist yet) can look for a smaller encoding of the same instruction that still satisfies register constraints. For example, 32-bit ARM instructions can be represented as 16-bit Thumb2 instructions if the registers are right.

The tables that are generated in files like `encoding-riscv.rs` should already be usable for this. The implementation of `TargetIsa::encode()` stops at the first match. The iterator would keep going until reaching a stop code.

cc @anholt

zummenix:
This can be closed. Implemented in #96

PR #96, opened 2017-06-14
Implement an iterator over encodings

zummenix:
This is my attempt to address #76 

What I don't like here it's that `instp` and `isap` are passed to iterator but it seems to me they should be implementation detail.


stoklund:
Thanks!

The problem with `instp` and `isap` is that each ISA has its own implementation of these functions, and we want to be able to share code between ISAs and still benefit from optimizations like inlining.

For a `TargetIsa` to return an iterator, I think we have to wait for the [impl Trait](https://github.com/rust-lang/rust/issues/34511) feature to become stable. In the meantime, we could use an interface like this:

```rust
trait TargetIsa {
    ...
    fn get_encodings(&self,
              dfg: &ir::DataFlowGraph,
              inst: &ir::InstructionData,
              ctrl_typevar: ir::Type,
              result: &mut Vec<Encoding>);
}
```

This should push all legal encodings onto the `result` vector

stoklund:
Thinking about this some more, it would be really nice if the `legal_encodings()` function could return an iterator instead of adding stuff to a `Vec`. Since it is a method on a trait object, it can't have a generic type that depends on the ISA.

The ISA predicates would be pretty simple to handle. If you look at the generated code in `settings-riscv.rs`, for example, you have this implementation:

```rust
/// User-defined settings.
impl Flags {
    /// Dynamic numbered predicate getter.
    pub fn numbered_predicate(&self, p: usize) -> bool {
        self.bytes[0 + p/8] & (1 << (p%8)) != 0
    }
```

Your `Encodings` iterator could simply hold on to an `&[u8]` slice that is initialized with the ISA's `isa_flags.bytes` array. That gets rid of the `isap` closure.

The `instp` closure is a bit more complicated. To begin with, you can just give the `Encodings` iterator a function pointer instead of a closure. Then it can call the ISA's `enc_tables::check_instp` function through that pointer.

A slightly better version that is more work to implement is to replace the generated `check_instp` functions with a generated array of function pointers, one per instruction predicate. The iterator can reference the array of predicate functions as a slice. This avoids the overhead of the big switch in the generated function.

lib/cretonne/meta/gen_settings.py
```diff
@@ -87,6 +87,9 @@ def gen_getters(sgrp, fmt):
     """
     fmt.doc_comment("User-defined settings.")
     with fmt.indented('impl Flags {', '}'):
+        fmt.doc_comment('Returns inner slice of bytes.')
+        with fmt.indented('pub fn bytes(&self) -> &[u8] {', '}'):
+            fmt.line('&self.bytes')
```
stoklund: The `bytes` array contains a number of non-boolean settings at the front that we don't want included here. Notice how the `numbered_predicate` function below adds `self.boolean_offset` to the byte array offset.

Two changes here:

- Rename `bytes()` to `predicate_bytes()` to make it clear that the byte-sized settings are not included.
- Return the `&self.bytes[boolean_offset]` sub-slice instead of the full array.

stoklund:
This looks great. You just need to fix a white space problem that flake8 is complaining about.

Use "pip install flake8" to check locally

zummenix:
Should be good now. Didn't install flake8, will do it later.

I wish I'd written better documentation, but I still don't understand a lot of parts.

stoklund:
Thanks!

PR #97, opened 2017-06-14
Cretonne IL frontend: ILBuilder

denismerigoux:
* API used to build Cretonne IL `Functions` without having to worry about `Layout` and `DataFlowGraph`.
* Supports non-SSA variables with `def_var` and `use_var` that can be used on the fly.
* Lots of sanity checks on the fly that help the user build valid Cretonne IL `Function`s.
* Hides `Ebb` arguments to the user (they are filled via SSA construction).
* Includes one method per instruction like `InstBuilder` to easily insert instructions to the function.
This pull request is based on https://github.com/stoklund/cretonne/pull/93.


stoklund:
These patches are giving me a bit of a "you ain't gonna need it" vibe. I think it is difficult to try and make life easier for code that hasn't even been written yet. It's hard to anticipate which problems are going to show up ahead of time. We've iterated over this by fixing problems I could think of, but it would be much better to iterate the design with real problems that show up in real code.

I think it is premature to design this interface and land it in Cretonne, and I would much prefer if the `ILBuilder` were developed alongside the code that translates from wasm or MIR into Cretonne IL. Once we have an actual working translator, we can evaluate if there are common patterns that would be good to extract and land in Cretonne.

A few specific comments:

- The `cretonne::Context` data structure is meant to be reused for compiling multiple functions. The overhead of allocating and deallocating these data structures can get quite large when compiling small functions, so it is important to be able to reuse them. The builder interface here doesn't allow for reusing data structures across function compilations.
- The sanity checks in the builder interface are duplicating functionality that is already in the verifier. Just use the verifier for catching those bugs. I think it would be fine to generate some additional `debug_assert!()` calls in the `InstBuilder` methods so you can get early assertion failures with better stack traces.
- I think it would be much simpler to have an `ins()` function that returns some `InstBuilder` instance. Duplicating that interface generates a ton of code and the benefit is not great. You could even provide your own implementation of `InstBuilderBase` if needed.

denismerigoux:
I agree with your general feelings: I will keep this code around somewhere and use it if I actually need it for the wasm to Cretonne IL translation. 

Concerning the specific points:

* I can make `ILBuilder` reusable by having it `clear()` its data structures when exporting its `Function`.
* I can change all the `panic!` into `debug_assert!`.
* I wasn't very familiar with the design of `InstBuilder` but I think I understand now, indeed providing my own implementation of `InstBuilderBase` is a much better solution.

stoklund:
Thanks, Denis

About reusing data structures: A `Function` object contains a bunch of `Vec`s once you peel off the abstraction layers, and each `Vec` will hold on to its allocated memory even when you call `clear()`. That is working as intended: when we're done compiling one function and move on to the next one, we'll call `clear()` and start over on the same `Function` object. There's a good chance the `Vec`s won't need to allocate more memory than what they already have after the `clear()`.

If we call `Function::with_name_signature()` to create a new object for each function we're translating, that strategy fails. The new `Function` object doesn't have any pre-allocated memory in the vectors, so it will have to allocate all over again.

There's two tactics you can use to reuse these memory allocations:

1. Have the `ILBuilder` operate on a borrowed `&mut Function` instead of owning the function.
2. Use `mem::replace` or `mem::swap` to move an existing `Function` object into the builder, pre-allocated buffers and all. Then move the function object back into the `Context` when it's built.

I tend to use the first tactic inside Cretonne. It can be tricky to control the lifetimes of the borrowed function reference when you also have data structures of your own that you want to reuse too. Take a look at `src/regalloc/coloring.rs` for an example of how I handle that. I split the data into two structs:

One struct contains data structures that I want to reuse to avoid repeated memory allocations:

```rust
/// Data structures for the coloring pass.
///
/// These are scratch space data structures that can be reused between invocations.
pub struct Coloring {
    divert: RegDiversions,
    solver: Solver,
}
```

The second struct is transient and fast to create and destroy. I make one for each pass:

```rust
/// Bundle of references that the coloring algorithm needs.
///
/// Some of the needed mutable references are passed around as explicit function arguments so we
/// can avoid many fights with the borrow checker over mutable borrows of `self`. This includes the
/// `Function` and `LiveValueTracker` references.
///
/// Immutable context information and mutable references that don't need to be borrowed across
/// method calls should go in this struct.
struct Context<'a> {
    // Cached ISA information.
    // We save it here to avoid frequent virtual function calls on the `TargetIsa` trait object.
    reginfo: RegInfo,
    encinfo: EncInfo,

    // References to contextual data structures we need.
    domtree: &'a DominatorTree,
    liveness: &'a mut Liveness,

    // References to working set data structures.
    // If we need to borrow out of a data structure across a method call, it must be passed as a
    // function argument instead, see the `LiveValueTracker` arguments.
    topo: &'a mut TopoOrder,
    divert: &'a mut RegDiversions,
    solver: &'a mut Solver,

    // Pristine set of registers that the allocator can use.
    // This set remains immutable, we make clones.
    usable_regs: AllocatableSet,
}
```

Unfortunately, limitations of the Rust borrow checker makes it hard to keep a `&mut Function` reference in the context struct. The problem is that you can't call *any* `&self` methods if you have mutably borrowed any member of the context. The borrow checker can only track member borrows inside a function body, not across calls.

Hopefully, [this will be fixed in the future](https://internals.rust-lang.org/t/partially-borrowed-moved-struct-types/5392/2).

It's possible you can get away with a `&mut Function` borrow in the IL builder. It's a good place to start, at least.

denismerigoux:
Thanks for the explanation, I think I understood now. I'll create `ILBuilder`  and `FunctionBuilder` to mimic respectively `Coloring` and `Context`, so that memory is not reallocated each time.

lib/cretonne/src/ir/frontend.rs
```diff
@@ -0,0 +1,764 @@
...
+    }
+
+    /// Returns the result values of an instruction.
+    pub fn inst_results(&self, inst: Inst) -> &[Value] {
+        self.func.dfg.inst_results(inst)
+    }
+
+    /// Changes the destination of a jump instruction after creation.
+    ///
+    /// **Note:** You are responsible for maintaining the coherence with the jump arguments.
```
sunfishcode: ... with the arguments of *other* jump instructions.

lib/cretonne/src/ir/frontend.rs
```diff
@@ -0,0 +1,764 @@
+//! A frontend for building Cretonne IL from other languages.
+//!
+//! # Example
+//!
+//! Here is a pseudo-program we want to transform in Cretonne IL:
```
denismerigoux: Fixed.

lib/cretonne/src/ir/frontend.rs
```diff
@@ -0,0 +1,764 @@
...
+//!
+//! ```cton
+//! function(x) {
+//! x, y, z : I32
```
denismerigoux: Fixed.

lib/cretonne/src/ir/frontend.rs
```diff
@@ -0,0 +1,764 @@
...
+//! Here is how you build the corresponding Cretonne IL function using `ILBuilder`:
+//!
+//! ```rust
+//! use cretonne::entity_ref::EntityRef;
+//! use cretonne::ir::{FunctionName, Function, Signature, ArgumentType, InstBuilder};
+//! use cretonne::ir::types::*;
+//! use cretonne::ir::frontend::{ILBuilder, FunctionBuilder};
+//! use cretonne::verifier::verify_function;
+//!
+//! use std::u32;
```
denismerigoux: Yes because we use `u32::MAX`.

lib/cretonne/src/ir/frontend.rs
```diff
@@ -0,0 +1,764 @@
...
+//!
+//! // An opaque reference to variable.
+//! #[derive(Copy, Clone, PartialEq, Eq, Hash, Debug)]
+//! pub struct Variable(u32);
+//! impl EntityRef for Variable {
+//!     fn new(index: usize) -> Self {
+//!         assert!(index < (u32::MAX as usize));
```
denismerigoux: No because you can't compare `usize` and `Variable` which is completely opaque. Besides this code is copied from `entities.rs` and this is how all Cretonne's references are defined.

lib/cretonne/src/ir/frontend.rs
```diff
@@ -0,0 +1,764 @@
...
+//!         Variable(index as u32)
```
denismerigoux: I can do that but you should go check out `entity_ref.rs`, I copied the code from there and those remarks should be addressed more generally to the rest of Cretonne.

lib/cretonne/src/ir/frontend.rs
```diff
@@ -0,0 +1,764 @@
...
+        &self.builder.func.dfg
+    }
+
+    fn data_flow_graph_mut(&mut self) -> &mut DataFlowGraph {
+        &mut self.builder.func.dfg
+    }
+
+    // This implementation is richer than `InsertBuilder` because we use the data of the
+    // instruction being inserted to add related info to the DFG and the SSA building system,
+    /// and perform debug sanity checks.
```
denismerigoux: Fixed.

lib/cretonne/src/ir/frontend.rs
```diff
@@ -0,0 +1,764 @@
...
+                Some(dest_ebb) => {
+                    self.builder.declare_successor(dest_ebb, inst);
+                    // If the user has supplied jump arguments we must adapt the arguments of
+                    // the destination ebb
+                    // TODO: find a way not to allocate a vector
+                    let ty_to_append: Option<Vec<Type>> =
+                        match data.analyze_branch(&self.builder.func.dfg.value_lists) {
+                            BranchInfo::SingleDest(_, args) => {
+                                let dest_ebb_args = self.builder.func.dfg.ebb_args(dest_ebb);
+                                if args.len() != 0 {
```
denismerigoux: OK so here the situation is complicated. Indeed, there are two types of Ebb arguments: those supplied by the user and those produced by the SSA construction. User supplied args are added to the Ebb's arg list when the first jump instruction to this Ebb is inserted. SSA arguments are added to the Ebb's arg list during the insertion of instructions to this Ebb.

That said, here's how the algorithm should work when you insert a jump instruction targeting `ebb`:
- If `ebb` already has a predecessor, then the user supplied args of the instruction being inserted should match the user supplied args of the other jump instructions having `ebb` as dest;
- if `ebb` has no predecessors, then the jump args of thie instruction being inserted are added as user supplied arguments to `ebb`. 

Your comment made me realize that my implementation has a bug: indeed, SSA produced `Ebb` arguments can be mixed with user supplied arguments in the Ebb's arg list because SSA undef values (when the block is not sealed) are directly inserted in the Ebb's arg list instead of being left in a detached state.

I will then use your suggestion to simplify this code in `frontend.rs` and also fix the bug in `ssa.rs`.

lib/cretonne/src/ir/frontend.rs
```diff
@@ -0,0 +1,764 @@
...
+        let ebb = self.func.dfg.make_ebb();
+        self.builder.ssa.declare_ebb_header_block(ebb);
+        *self.builder.ebbs.ensure(ebb) = EbbData {
+            filled: false,
+            sealed: false,
+        };
+        ebb
+    }
+
+    /// After the call to this function, new instructions will be inserted into the designed
```
denismerigoux: Fixed.

lib/cretonne/src/ir/frontend.rs
```diff
@@ -0,0 +1,764 @@
...
+    /// block, in the order they are declared.
+    ///
+    /// When inserting the terminator instruction (which doesn't have a falltrough to its immediate
+    /// successor), the block will be declared filled and it will not be possible to append
+    /// instructions to it.
+    pub fn switch_to_block(&mut self, ebb: Ebb) {
```
denismerigoux: I think we have to take the point of view of the user calling the API here: for him, switching to the first block should not be different from swtiching to any other block. The `self.pristine` and `fill_function_args_values` are just implementation quirks, and in my opinion they should be hidden from the user. This is also subjective though.

lib/cretonne/src/ir/frontend.rs
```diff
@@ -0,0 +1,764 @@
...
+        // Then we change the cursor position.
+        self.position = Position {
+            ebb: ebb,
+            basic_block: basic_block,
+            pristine: true,
+        };
+    }
+
+    /// Function to call with `block` as soon as the last branch instruction to `block` has been
+    /// created. Declares that all the predecessors of this block are known.
```
denismerigoux: Fixed.

lib/cretonne/src/ir/frontend.rs
```diff
@@ -0,0 +1,764 @@
...
+    ///
+    /// Forgetting to call this method on any block will prevent you from retrieving
+    /// the Cretonne IL.
```
denismerigoux: Fixed, I implemented the `Drop` trait that checks sealing and filling.

lib/cretonne/src/ir/frontend.rs
```diff
@@ -0,0 +1,764 @@
...
+    }
+
+
+
+    /// Returns an object with the [`InstBuilder`](../trait.InstBuilder.html) trait that allows to
+    /// conveniently append an instruction to the current `Ebb` being built.
+    pub fn ins<'short>(&'short mut self) -> FuncInstBuilder<'short, 'a, Variable> {
+        let ebb = self.position.ebb;
+        FuncInstBuilder::new(self, ebb)
+    }
```
denismerigoux: I wish I could do that but the signature of the build function is `build(self, ...)` and not `build(&mut self)`. Because `self` is consumed I can't make `self` the whole `FunctionBuilder`, and as a consequence we're stuck with the quirky `builder.ins()....` syntax.

lib/cretonne/src/ir/frontend.rs
```diff
@@ -0,0 +1,764 @@
...
+    /// inserted that target it and the SSA construction.
+    pub fn ebb_args(&self, ebb: Ebb) -> &[Value] {
+        self.func.dfg.ebb_args(ebb)
+    }
+
+    /// Creates a argument for a specific `Ebb` by appending it to the list of already existing
+    /// arguments.
+    ///
+    /// **Note:** this function has to be called at the creation of the `Ebb` before adding
+    /// instructions to it, otherwise this could interfere with SSA construction.
```
denismerigoux: I can't do that because the function has a arbitrary `ebb` argument whereas `self.position.pristine` concerns only the current `Ebb`. To enforce this assert I would need to store the `pristine` parameter inside `EbbData`.

lib/cretonne/src/ir/frontend.rs
denismerigoux: Fixed.

lib/cretonne/src/ir/frontend.rs
```diff
@@ -0,0 +1,764 @@
...
+    }
+
+    /// Changes the destination of a jump instruction after creation.
+    ///
+    /// **Note:** You are responsible for maintaining the coherence with the jump arguments.
+    pub fn change_jump_destination(&mut self, inst: Inst, new_dest: Ebb) {
+        let old_dest = match self.func.dfg[inst].branch_destination_mut() {
+            None => panic!("you want to change the jump destination of a non-jump instruction"),
+            Some(ebb) => ebb,
+        };
```
denismerigoux: Fixed.

lib/cretonne/src/ir/frontend.rs
```diff
@@ -0,0 +1,764 @@
...
+            .declare_ebb_predecessor(new_dest, pred, inst);
+    }
+
+    /// Returns `true` if and only if the current `Ebb` is sealed and has no predecessors declared.
+    pub fn is_unreachable(&self) -> bool {
+        let is_entry = match self.func.layout.entry_block() {
+            None => true,
+            Some(entry) => self.position.ebb != entry,
+        };
+        is_entry && self.builder.ebbs[self.position.ebb].sealed &&
```
denismerigoux: Fixed.

lib/cretonne/src/ir/frontend.rs
```diff
@@ -0,0 +1,764 @@
...
+    fn declare_successor(&mut self, dest_ebb: Ebb, jump_inst: Inst) {
+        self.builder
+            .ssa
+            .declare_ebb_predecessor(dest_ebb, self.position.basic_block, jump_inst);
+    }
+
+    fn check_return_args(&self, args: &[Value]) {
+        debug_assert!(args.len() == self.func.signature.return_types.len(),
+                      format!("the number of returned values doesn't match the function\
+                       signature ({} vs {})",
```
denismerigoux: Fixed.

lib/cretonne/src/ir/frontend.rs
```diff
@@ -0,0 +1,764 @@
...
+                              args.len(),
+                              self.func.signature.return_types.len()));
+        for (i, arg) in args.iter().enumerate() {
+            let valty = self.func.dfg.value_type(*arg);
+            debug_assert!(self.func.signature.return_types[i].value_type == valty,
+                          format!("the types of the values returned don't match the function\
+                           signature ({} vs {})",
```
denismerigoux: Fixed.

lib/cretonne/src/ir/frontend.rs
```diff
@@ -0,0 +1,764 @@
...
+
+        fn index(self) -> usize {
+            self.0 as usize
+        }
+    }
+    impl Default for ExtendedBlock {
+        fn default() -> ExtendedBlock {
+            ExtendedBlock(u32::MAX)
+        }
+    }
```
denismerigoux: Indeed, it was a leftover from a previous implementation.

lib/cretonne/src/ssa.rs
```diff
@@ -0,0 +1,755 @@
...
+}
+
+// Describes the current position of a basic block in the control flow graph.
+enum BlockData<Variable> {
+    // A block at the top of an `Ebb`.
+    EbbHeader(EbbHeaderBlockData<Variable>),
+    // A block inside an `Ebb` with an unique other block as its predecessor.
+    // The block is implicitely sealed at creation.
+    // The fields are `(predecessor, ebb)`.
+    EbbBody(Block, Ebb),
```
denismerigoux: Indeed, fixed.

lib/cretonne/src/ssa.rs
```diff
@@ -0,0 +1,755 @@
...
+            &mut BlockData::EbbBody(_, _) => panic!("should not happen"),
+            &mut BlockData::EbbHeader(ref mut data) => {
+                // This a linear complexity operation but the number of predecessors is low
+                // in all non-pathological cases
+                let pred: Block = match data.predecessors
+                          .iter()
+                          .find(|&(_, &jump_inst)| jump_inst == inst) {
+                    None => panic!("the predecessor you are trying to remove is not declared"),
+                    Some((&b, _)) => b.clone(),
+                };
```
denismerigoux: Actually no because it is the basic block we need here and not the extended basic block.

lib/cretonne/src/ssa.rs
```diff
@@ -0,0 +1,755 @@
...
+        }
+    }
+}
+
+struct EbbHeaderBlockData<Variable> {
+    // The predecessors of the Ebb header block, with the block and branch instruction.
+    predecessors: HashMap<Block, Inst>,
+    // A ebb header block is sealed if all of its predecessors have been declared.
+    sealed: bool,
+    // The ebb of which this block is part of.
```
denismerigoux: Fixed.

lib/cretonne/src/ssa.rs
```diff
@@ -0,0 +1,755 @@
...
+/// Rust sample that shows how to do this while building the Cretonne IL function.
+///
+/// ```
+/// use cretonne::entity_ref::EntityRef;
+/// use cretonne::ir::{Function, InstBuilder, Cursor, Inst};
+/// use cretonne::ir::types::*;
+/// use cretonne::ssa::SSABuilder;
+/// use std::u32;
+///
+/// // First we have to defined what are variables in our original language.
```
denismerigoux: Fixed.

lib/cretonne/src/ir/frontend.rs
denismerigoux: Actually I did just that and now there's the check.

lib/cretonne/src/ir/frontend.rs
sunfishcode: Would it work with `build(mut self, ...)`? I think that still satisfies the `InstBuilderBase` trait, and gives you the mut you need.

sunfishcode: Oh, never mind. I see. The builder is consumed, so it can't be used for subsequent API calls.

lib/cretonne/src/ir/entities.rs
```diff
@@ -23,7 +23,7 @@ use std::fmt;
 use std::u32;
 
 /// An opaque reference to an extended basic block in a function.
-#[derive(Copy, Clone, PartialEq, Eq, Hash, Debug, PartialOrd, Ord)]
+#[derive(Copy, Clone, PartialEq, Eq, Hash, Debug, PartialOrd, Ord, Default)]
```
stoklund: This doesn't seem right to me. What is a default EBB needed for?

lib/cretonne/src/ir/funcname.rs
```diff
@@ -52,7 +52,7 @@ impl FunctionName {
 fn try_as_name(bytes: &[u8]) -> Option<String> {
     let mut name = String::with_capacity(bytes.len());
     for c in bytes.iter().map(|&b| b as char) {
-        if c.is_ascii() && c.is_alphanumeric() || c == '_' {
+        if c.is_ascii() && c.is_alphanumeric() || c == '_' || c == '.' || c == '-' {
```
stoklund: I don't think this would be good for the Cretonne IL syntax.

Cretonne's function names are used for linkage, but WebAssembly doesn't link functions by name — it uses numbers instead. So it is unlikely that we end up generating Cretonne functions with names from the WebAssembly name section anyway.

So there is no need for us to propagate the lispisms from WebAssembly's test suite.

lib/cretonne/src/ir/mod.rs
```diff
@@ -10,7 +10,7 @@ pub mod jumptable;
 pub mod dfg;
 pub mod layout;
 pub mod function;
-mod builder;
+pub mod builder;
```
stoklund: Rather than making the `builder` module public, could you just re-export the type you need? (I'm guessing it's `InstBuilderBase`).

lib/cretonne/src/ir/builder.rs
```diff
@@ -23,6 +23,8 @@ pub trait InstBuilderBase<'f>: Sized {
     /// Get an immutable reference to the data flow graph that will hold the constructed
     /// instructions.
     fn data_flow_graph(&self) -> &DataFlowGraph;
+    /// Get an mutable reference to the data flow graph that will hold the constructed
```
stoklund: s/an/a/

lib/frontend/src/frontend.rs
```diff
@@ -0,0 +1,693 @@
...
+        } else {
+            debug_assert!(!self.builder.builder.ebbs[self.builder.position.ebb].filled,
+                          "you cannot add an instruction to a block already filled");
+        }
+        let inst = self.builder.func.dfg.make_inst(data.clone());
+        self.builder.func.dfg.make_inst_results(inst, ctrl_typevar);
+        {
+            let mut cur = Cursor::new(&mut self.builder.func.layout);
+            cur.goto_bottom(self.ebb);
+            cur.insert_inst(inst);
```
stoklund: Simpler to just call `layout.append_inst()` instead of creating a cursor.

lib/frontend/src/frontend.rs
```diff
@@ -0,0 +1,693 @@
...
+    pub fn declare_var(&mut self, var: Variable, ty: Type) {
+        *self.builder.types.ensure(var) = ty;
+    }
+
+    /// Returns the Cretonne IL value corresponding to the utilization at the current program
+    /// position of a previously defined user variable.
+    pub fn use_var(&mut self, var: Variable) -> Value {
+        let ty = match self.builder.types.get(var) {
+            Some(&ty) => ty,
+            None => panic!("this variable is used but its type has not been declared"),
```
stoklund: Use `expect()` instead of `match`?

lib/frontend/src/ssa.rs
```diff
@@ -0,0 +1,1278 @@
...
+{
+    // Records for every variable and for every revelant block, the last definition of
+    // the variable in the block.
+    variables: EntityMap<Variable, HashMap<Block, Value>>,
+    // Records the position of the basic blocks and the list of values used but not defined in the
+    // block.
+    blocks: EntityMap<Block, BlockData<Variable>>,
+    // Records the basic blocks at the beginning of the `Ebb`s.
+    // The value is `(Ebb,Block)` because in a `SparseMap`, the key has to be computable from the
+    // value.
```
stoklund: This comment is obsolete.

lib/cretonne/src/ir/entities.rs
denismerigoux: That was a corner case between `IlBuilder` initialization and the beginning of the translation for the `Position` struct, It is not used or read so I don't actually need a proper default value.

lib/cretonne/src/ir/funcname.rs
denismerigoux: If you think so I can remove it. This was really important for me to debug my translations, otherwise every function name was #065065AB5468DC5664 and I couldn't figure out where I was in the module.

lib/cretonne/src/ir/mod.rs
denismerigoux: Fixed.

lib/cretonne/src/ir/builder.rs
denismerigoux: Fixed.

lib/frontend/src/frontend.rs
denismerigoux: Fixed.

lib/frontend/src/frontend.rs
denismerigoux: Fixed.

lib/frontend/src/ssa.rs
denismerigoux: Fixed.

lib/cretonne/src/ir/entities.rs
stoklund: This is probably similar to the problem with the `Layout::entry_block()` method. The solution is to use an `Option<Ebb>`.

Implementing `Default` for `Ebb` indicates to users that `ebb0` is somehow a sensible default, but it really isn't.

lib/frontend/src/frontend.rs
```diff
@@ -231,8 +220,8 @@ impl<'a, Variable> FunctionBuilder<'a, Variable>
             func: func,
             builder: builder,
             position: Position {
-                ebb: Ebb::default(),
-                basic_block: Block::default(),
+                ebb: Ebb::new(0),
```
stoklund: Sure, that works too ;-)

denismerigoux: Yeah I never read the value and I didn't want the hassle of an option type (lots of `unwrap()`).

lib/cretonne/src/ir/funcname.rs
stoklund: When you are creating function names for debugging purposes instead of for linking purposes, you can simply map all non-alphanumeric characters to `_`.

PR #98, opened 2017-06-19
Tc

d1m0:
This is a first hash of type-checking for Rtl/XForms in the meta-language - just to get the ball rolling. Most of the code is in:

lib/cretonne/meta/cdsl/tc.py

with example usage in:

lib/cretonne/meta/cdsl/test_tc.py

Specifically tc.py defines these entry points:

tc_rtl(r, m) -> m'  where r : Rtl, m, m': Dict[Var, TypeVar]
tc_xform(x, m) -> m' where x: XForm, m, m': Dict[Var, TypeVar]

The design is that we do as minimum inference as possible - we require the consumer to specify in the initial map m all control variables necessary to monomorphise instructions. The consumer can also optionally specify the types of any free input variables that are not derived from control variables; if not specified, their types will be inferred based on the uses of those vars. The types of all variables defined in the Rtl/XForm will be inferred based on the monomorphic type of their defining instruction.

We don't do unification, but when we check if the actual type of a variable agrees with the monomorphic type of its use location we do check for 'structural' equality of the two types. That is, if they are both derived type vars, we check that the deriving function is the same, and recursively check the equality of the base TVs. This is necessary for cases such as the following instruction sequence:

v0 << iconst(imm0)
v2 << icmp(ne, v1, v0)
v5 << vselect(v2, v3, v4)

Here v0, v1 and v3 are control variables and must be provided by the consumer. The type checker needs to confirm that v0, v1 and v3 have the same number of lanes. For v0 and v1 this is achieved because:

1) Since v1 is the control var for icmp, we monomorphise the polymorphic icmp: (Cond, IxN, IxN) -> as_bool(IxN) to the concrete type (Cond, Tv0, Tv0) -> as_bool(Tv0)
2) Since v0 is passed at the third argument, we check that Tv1 agrees with Tv0. The only way to satisfy this structurally is by Tv1 == Tv0 (since they are both non-derived)

For v3 this holds because:

1) Since v3 is the control var for vselect we monomorphise the type of vselect to (as_bool(Tv3), Tv3, Tv3) -> Tv3

2) Since v2 is passed as the first argument, we check that Tv2 agrees with as_bool(Tv3). However Tv2 in our environment is defined to be as_bool(Tv0). So we check that as_bool(Tv0) agrees with as_bool(Tv3), which works iff Tv0 == Tv3.

This way, the TC will type check this pattern iff v0,v1, v3 are mapped to the same vector integer type.

### DIFF
--- a/lib/cretonne/meta/cdsl/instructions.py
+++ b/lib/cretonne/meta/cdsl/instructions.py
@@ -189,6 +189,7 @@ class Instruction(object):
                 if tv is tv.free_typevar():
                     self.other_typevars = self._verify_ctrl_typevar(tv)
                     self.ctrl_typevar = tv
+                    self.ctrl_typevar_ind = opnum
                     self.use_typevar_operand = True
             except RuntimeError as e:
                 typevar_error = e
@@ -203,10 +204,29 @@ class Instruction(object):
                     raise RuntimeError(
                             "typevar_operand must be a free type variable")
             tv = self.outs[0].typevar
+            self.ctrl_typevar_ind = -1
             if tv is not tv.free_typevar():
                 raise RuntimeError("first result must be a free type variable")
             self.other_typevars = self._verify_ctrl_typevar(tv)
             self.ctrl_typevar = tv
+
+    def get_ctrl_typevar_ind(self):
+        # type: (Instruction) -> Tuple[bool, int]
+        """
+        If this Instruction doesn't have a control TV return None.
+        Otherwsire return a (is_arg, index) tuple where:
+            - is_arg is True if the control TV is one of the arguments and
+              False if its a return
+            - index is the index of the argument/return which has the control
+              TV
+        """
+        if (not hasattr(self, "ctrl_typevar_ind")):
+            return None
+
+        if (self.ctrl_typevar_ind == -1):
+            return (False, 0)
+        else:
+            return (True, self.ctrl_typevar_ind)
 
     def _verify_ctrl_typevar(self, ctrl_typevar):
         # type: (TypeVar) -> List[TypeVar]
--- a/lib/cretonne/meta/cdsl/instructions.py
+++ b/lib/cretonne/meta/cdsl/instructions.py
@@ -189,7 +189,6 @@ class Instruction(object):
                 if tv is tv.free_typevar():
                     self.other_typevars = self._verify_ctrl_typevar(tv)
                     self.ctrl_typevar = tv
-                    self.ctrl_typevar_ind = opnum
                     self.use_typevar_operand = True
             except RuntimeError as e:
                 typevar_error = e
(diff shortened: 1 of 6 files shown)


lib/cretonne/meta/cdsl/instructions.py
```diff
@@ -189,6 +189,7 @@ def _verify_polymorphic(self):
                 if tv is tv.free_typevar():
                     self.other_typevars = self._verify_ctrl_typevar(tv)
                     self.ctrl_typevar = tv
+                    self.ctrl_typevar_ind = opnum
```
d1m0: Fixed. Removed changes from cdsl/instructions.py

lib/cretonne/meta/cdsl/instructions.py
```diff
@@ -203,11 +204,30 @@ def _verify_polymorphic(self):
                     raise RuntimeError(
                             "typevar_operand must be a free type variable")
             tv = self.outs[0].typevar
+            self.ctrl_typevar_ind = -1
             if tv is not tv.free_typevar():
                 raise RuntimeError("first result must be a free type variable")
             self.other_typevars = self._verify_ctrl_typevar(tv)
             self.ctrl_typevar = tv
 
+    def get_ctrl_typevar_ind(self):
```
d1m0: Renamed to get_ctrl_typevar_index and moved into tc.py; Added clarifying doc comment about indexing and now using value_opnums/value_results.

lib/cretonne/meta/cdsl/typevar.py
```diff
@@ -46,6 +52,80 @@ def intersect(a, b):
...
+    (lo, hi) = interval
+
+    if (lo is None and hi is None):
+        return True
+
+    if (lo is None or hi is None):
+        return False
+
+    if (not is_power_of_two(lo) or
+       not is_power_of_two(hi) or
```
d1m0: They shouldn't. Changed those changes to asserts.

lib/cretonne/meta/cdsl/typevar.py
```diff
@@ -46,6 +52,80 @@ def intersect(a, b):
...
+    else:
+        (lo, hi) = interval
+        new_interval = (f(lo), f(hi))
+        assert valid_interval(new_interval, full_range)
+        return new_interval
+
+
+def encode_interval(lo, hi, full_range):
+    # type: (int, int, Interval) -> Interval
+    return map_interval((lo, hi), lambda x:  x, full_range)
```
d1m0: Fixed.

lib/cretonne/meta/cdsl/typevar.py
```diff
@@ -152,7 +228,7 @@ def __eq__(self, other):
             return False
 
     def __repr__(self):
-        # type: () -> str
+        # type: (TypeSet) -> str
```
d1m0: Strangely mypy checks on my machine succeed both with and without the self argument in the signature. I removed the self argument from the signatures of all added/modified code.

d1m0:
Wanted to point out a couple of potential weak points:

1) In tc_def here:

https://github.com/stoklund/cretonne/pull/98/files/b0f3f0c232fb0a083055b70e8f52bb3788b8456e#diff-cc311dd7578ed4df76c1c15d397073a7R383

I am implicitly assuming that all value operands to an instruction are always Vars. Can this be ever violated? (I haven't seen this in the code so far). If this is always true I should assert it.

2) I am adding a new field (symtab) to XForm.

https://github.com/stoklund/cretonne/pull/98/files/b0f3f0c232fb0a083055b70e8f52bb3788b8456e#diff-bffb8ed352cb424b062f6d83bf10a155R86

However it is only used for tests. Perhaps there is a way to better encapsulate this data only in test code?

3) In test_tc.py, when comparing exceptions of types TCNotSubtype and TCDisagree, I am not comparing the concrete inferred types with the expected exception types. I am only comparing the location of the disagreement (which instruction, operand/result). For TCDisagree I seem to be even more coarse grained. Those should prob be addressed to get more precise testing.

d1m0:
This work duplicates existing type inference in the repo. Closing in favor of a more-incremental approach.

PR #102, opened 2017-06-22
Convert interval sets inside TypeSet/ValueTypeSet in general sets

d1m0:
As part of a better type inference we need to compute the image (and pre-image) of TypeSets across derivation functions. This computation across the as_bool function can result in a typeset containing disjoint intervals. E.g.:

TypeSet(..., ints=(8,16), bools=(64,64)).as_bool() = TypeSet(bools={ (8,16) , (64,64) })

This examples suggests that we need to support general sets in each component of the TypeSet datastructure. TypeSet gets lowered down to ValueTypeSet in Rust so a similar change is needed there.

In python we use the built-in set() type. In rust we make a BitSet tuple struct wrapping around a u8/u16 (can be expanded to larger types if needed).

### DIFF
--- a/lib/cretonne/meta/cdsl/typevar.py
+++ b/lib/cretonne/meta/cdsl/typevar.py
@@ -9,7 +9,7 @@ import math
 from . import types, is_power_of_two
 
 try:
-    from typing import Tuple, Union, TYPE_CHECKING # noqa
+    from typing import Tuple, Union, Iterable, Any, Set, TYPE_CHECKING # noqa
     if TYPE_CHECKING:
         from srcgen import Formatter  # noqa
         Interval = Tuple[int, int]
@@ -52,6 +52,7 @@ def is_empty(intv):
 
 
 def encode_bitset(vals, size):
+    # type: (Iterable[int], int) -> int
     res = 0
     assert is_power_of_two(size) and size <= 64
     for v in vals:
@@ -61,6 +62,7 @@ def encode_bitset(vals, size):
 
 
 def pp_set(s):
+    # type: (Iterable[Any]) -> str
     return '{' + ', '.join([repr(x) for x in sorted(s)]) + '}'
 
 
@@ -93,6 +95,7 @@ def decode_interval(intv, full_range, default=None):
 
 
 def interval_to_set(intv):
+    # type: (Interval) -> Set
     if is_empty(intv):
         return set()
 
--- a/lib/cretonne/meta/cdsl/typevar.py
+++ b/lib/cretonne/meta/cdsl/typevar.py
@@ -201,9 +201,9 @@ class TypeSet(object):
                   ('bools', 8))
 
         for (field, bits) in fields:
-            val = [int_log2(x) for x in getattr(self, field)]
+            vals = [int_log2(x) for x in getattr(self, field)]
             fmt.line('{}: BitSet::<u{}>({}),'
-                     .format(field, bits, encode_bitset(val, bits)))
+                     .format(field, bits, encode_bitset(vals, bits)))
 
     def __iand__(self, other):
         # type: (TypeSet) -> TypeSet
--- a/lib/cretonne/meta/cdsl/typevar.py
+++ b/lib/cretonne/meta/cdsl/typevar.py
@@ -53,6 +53,9 @@ def is_empty(intv):
 
 def encode_bitset(vals, size):
     # type: (Iterable[int], int) -> int
+    """
+    Encode a set of values (each between 0 and size) as a bitset of width size.
+    """
     res = 0
     assert is_power_of_two(size) and size <= 64
     for v in vals:
@@ -63,6 +66,9 @@ def encode_bitset(vals, size):
 
 def pp_set(s):
     # type: (Iterable[Any]) -> str
+    """
+    Return a consistent string representation of a set (ordering is fixed)
+    """
     return '{' + ', '.join([repr(x) for x in sorted(s)]) + '}'
 
 
@@ -136,7 +142,7 @@ class TypeSet(object):
     >>> TypeSet(floats=True)
     TypeSet(lanes={1}, floats={32, 64})
     >>> TypeSet(bools=True)
-    TypeSet(lanes={1}, bools={1, 2, 4, 8, 16, 32, 64})
+    TypeSet(lanes={1}, bools={1, 8, 16, 32, 64})
 
     Similarly, passing `True` for the lanes selects all possible scalar and
     vector types:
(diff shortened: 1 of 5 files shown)


lib/cretonne/src/ir/instructions.rs
```diff
@@ -499,17 +500,88 @@ impl OpcodeConstraints {
...
+    T: BitSetInnerType<T> +
+       BitAnd<T, Output=T> +
+       BitOr<T, Output=T> +
+       Shl<u8, Output=T> +
+       Copy +
+       PartialEq
+{
+    /// Check if this BitSet contains the number num
+    pub fn contains(&self, num: u8) -> bool {
+        assert!((num as usize) < T::bits());
```
stoklund: I think `debug_assert!` will suffice here.

It seems that `bits()` doesn't need to be a trait method. Just use `size_of::<T>()*8`.

I actually think that `<<` will trap in debug builds if the shift amount is too large. That may be god enough.

A different approach that may be simpler to handle: Instead of the `BitSetInnerType` trait, just require `T: Into<u32>` and do all of your arithmetic on `u32`. The `Into<u32>` trait is implemented by `u8`, `u16`, and `u32`.

lib/cretonne/src/ir/instructions.rs
```diff
@@ -499,17 +500,88 @@ impl OpcodeConstraints {
...
+        return self.0 & (T::coerce(1) << num) != T::coerce(0)
+    }
+
+    /// Return the smallest number contained in the bitset or T::bits() if
+    /// empty
+    pub fn min(&self) -> u8 {
+        let mut idx: u8 = 0;
+        while (idx as usize) < T::bits() && !self.contains(idx) {
+            idx += 1
+        }
```
stoklund: See `u32::trailing_zeros()`.

lib/cretonne/src/ir/instructions.rs
```diff
@@ -519,11 +591,11 @@ impl ValueTypeSet {
...
         let l2b = scalar.log2_lane_bits();
         if scalar.is_int() {
-            self.min_int <= l2b && l2b < self.max_int
+            self.ints.contains(l2b)
         } else if scalar.is_float() {
-            self.min_float <= l2b && l2b < self.max_float
+            self.floats.contains(l2b)
         } else if scalar.is_bool() {
-            self.min_bool <= l2b && l2b < self.max_bool
+            self.bools.contains(l2b)
```
stoklund: Nice.

lib/cretonne/src/ir/instructions.rs
```diff
@@ -499,17 +500,88 @@ impl OpcodeConstraints {
...
+    /// Construct a BitSet with the range (lo,hi) filled in
+    pub fn from_range(lo: u8, hi: u8) -> BitSet<T> {
+        let mut t: T = T::coerce(0);
+        let mut i: u8 = lo;
+        assert!(lo <= hi);
+
+        while i <= hi {
+            t = t | (T::coerce(1) << i);
+            i += 1;
+        }
```
stoklund: Or `(1 << hi) - (1 << lo)` for a half-open interval.

lib/cretonne/src/ir/instructions.rs
```diff
@@ -499,17 +500,88 @@ impl OpcodeConstraints {
...
+        idx
+    }
+
+    /// Return the largest number contained in the bitset or -1 if
+    /// empty
+    pub fn max(&self) -> i8 {
+        let mut idx: i8 = (T::bits() as i8) - 1;
+        while idx >= 0 && !self.contains(idx as u8) {
+            idx -= 1
+        }
```
stoklund: See `u32::leading_zeros()`.

lib/cretonne/src/ir/instructions.rs
```diff
@@ -709,15 +781,12 @@ mod tests {
...
-            min_int: 3,
-            max_int: 7,
-            min_float: 0,
-            max_float: 0,
-            min_bool: 3,
-            max_bool: 7,
+            lanes: BitSet16::from_range(0,8),
+            ints: BitSet8::from_range(4,7),
+            floats: BitSet8::from_range(0,0),
+            bools: BitSet8::from_range(3,7),
```
stoklund: Note that the old `ValueTypeSet` intervals were half-open: `min <= x < max`. We should document what `BitSet::from_range()` does. Half-open or closed.

lib/cretonne/src/ir/instructions.rs
```diff
@@ -530,14 +538,14 @@ impl<T> BitSet<T> where
...
+/// Check if this BitSet contains the number num
     pub fn contains(&self, num: u8) -> bool {
         assert!((num as usize) < T::bits());
         return self.0 & (T::coerce(1) << num) != T::coerce(0)
     }
 
-    /// Return the smallest number contained in the bitset or T::bits() if
-    /// empty
+/// Return the smallest number contained in the bitset or T::bits() if
+/// empty
```
sunfishcode: FYI, it looks like the awkward formatting for these comments has been fixed in newer versions of `rustfmt`, so I think we can just ignore this for now, and things will get fixed whenever cretonne updates its `rustfmt` version.

lib/cretonne/meta/cdsl/typevar.py
```diff
@@ -46,6 +46,26 @@ def intersect(a, b):
         return (None, None)
 
 
+def is_empty(intv):
+    # type: (Interval) -> bool
+    return intv is None or intv is False or intv == (None, None)
+
+
+def encode_bitset(vals, size):
```
d1m0: Added

d1m0: Added.

lib/cretonne/src/bitset.rs
```diff
@@ -0,0 +1,147 @@
...
+    pub fn max(&self) -> Option<u8> {
+        if self.0.into() == 0 {
+            None
+        } else {
+            let leading_zeroes = self.0.into().leading_zeros() as usize;
+            Some((Self::max_bits() - leading_zeroes - 1) as u8)
+        }
+    }
+
+/// Construct a BitSet with the range (lo,hi) filled in
```
stoklund: You should explain that this is a half-open range.

lib/cretonne/src/bitset.rs
```diff
@@ -0,0 +1,147 @@
...
+    pub fn from_range(lo: u8, hi: u8) -> BitSet<T> {
+        assert!(lo <= hi);
+        assert!((hi as usize) <= Self::bits());
+        let one : T = T::from(1);
+// I can't just do (one << hi) - one here as the shift may overflow
```
stoklund: This indentation looks wrong

lib/cretonne/src/ir/instructions.rs
d1m0: Got it :)

PR #103, opened 2017-06-22
Documentation fixes

sunfishcode:
While reading the docs, I noticed that the documentation for `_imm` instructions isn't entirely consistent. This PR fixes a few examples of this. Are you interested in fixing all the `_imm` instructions this way, or would you prefer to factor this text out somehow?


stoklund:
Eh. I guess it's better to go for consistency in the instruction descriptions.

Another option is to omit these things entirely and let the generated type variable documentation handle the details. That might be too subtle.

PR #104, opened 2017-06-23
Add image computation of typesets; Remove TypeVar.singleton_type

d1m0:
This PR only adds forward image computations of typesets. This allows us to compute the concrete typeset for any derived type var. As a result, TypeVar.singleton_type can be computed from the typeset for any type var. Therefore we can replace it with a identically named function. This (i think) takes care of the following todo in TypeVar.constrain_type():

-            # TODO: What if a.type_set becomes empty?
-            if not a.singleton_type:
-                a.singleton_type = b.singleton_type

### DIFF
--- a/lib/cretonne/meta/cdsl/typevar.py
+++ b/lib/cretonne/meta/cdsl/typevar.py
@@ -7,15 +7,19 @@ polymorphic by using type variables.
 from __future__ import absolute_import
 import math
 from . import types, is_power_of_two
+from copy import deepcopy
+from .types import ValueType, IntType, FloatType, BoolType
 
 try:
     from typing import Tuple, Union, Iterable, Any, Set, TYPE_CHECKING # noqa
+    from typing import cast
     if TYPE_CHECKING:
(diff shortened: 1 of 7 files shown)


lib/cretonne/meta/cdsl/typevar.py
```diff
@@ -241,6 +263,109 @@ def __iand__(self, other):
...
+    def get_singleton(self):
+        # type: () -> types.ValueType
+        """
+        Return the singleton type represented by self. Can only call on
+        typesets containing 1 type.
+        """
+        assert self.size() == 1
+        if len(self.ints) > 0:
+            bits = tuple(self.ints)[0]
+            scalar_type = ValueType.by_name(IntType.get_name(bits))
```
stoklund: If `IntType` had a registry of created integer types and a static `with_bits()` method, you wouldn't need this stringly typed stuff. I think mypy would be happier too.

d1m0: I can clean this up in the next PR

PR #107, opened 2017-06-28
Implement optimization fuel

sunfishcode:
Here's an initial draft of an optimization fuel mechanism. I experimented with a few different ways to route the needed information around, and putting it in the Cretonne Context seems simplest, though I'm open to other ideas here.

### DIFF
--- a/lib/cretonne/src/context.rs
+++ b/lib/cretonne/src/context.rs
@@ -55,7 +55,7 @@ impl Context {
             domtree: DominatorTree::new(),
             regalloc: regalloc::Context::new(),
             loop_analysis: LoopAnalysis::new(),
-            dbg: DebugState { fuel: None },
+            dbg: DebugState::new(),
         }
     }
 
--- a/lib/cretonne/src/context.rs
+++ b/lib/cretonne/src/context.rs
@@ -20,6 +20,7 @@ use result::CtonResult;
 use verifier;
 use simple_gvn::do_simple_gvn;
 use licm::do_licm;
+use dbg::DebugState;
 
 /// Persistent data structures and compilation pipeline.
 pub struct Context {
@@ -37,6 +38,9 @@ pub struct Context {
 
     /// Loop analysis of `func`.
     pub loop_analysis: LoopAnalysis,
+
+    /// Debugging state.
+    pub dbg: DebugState,
 }
 
 impl Context {
@@ -51,6 +55,7 @@ impl Context {
             domtree: DominatorTree::new(),
             regalloc: regalloc::Context::new(),
             loop_analysis: LoopAnalysis::new(),
+            dbg: DebugState { fuel: None },
         }
     }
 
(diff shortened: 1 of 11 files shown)


sunfishcode:
[This link](http://blog.ezyang.com/2011/06/debugging-compilers-with-optimization-fuel/) has background on the optimization fuel concept.

sunfishcode:
Thinking about the API here, I had an idea to generalize it to combine fuel tracking with logging. I just added a patch which implements this, and converts simple_gvn.rs to use it as an example.

Here's the latest code in action. Limitting fuel to 2 makes GVN fail the basic.cton test fail as follows:

```
$ target/debug/cton-util test --fuel=2 --trace=1 filetests/simple_gvn/basic.cton
FAIL filetests/simple_gvn/basic.cton: simple-gvn(%redundancies_on_some_paths): filecheck failed:
#0 check: jump ebb2(v3)
#1 check: jump ebb2(v3)
#2 check: v8 = iadd v6, v3
> function %redundancies_on_some_paths(i32, i32, i32) -> i32 {
> ebb0(v0: i32, v1: i32, v2: i32):
>     v3 = iadd v0, v1
>     brz v3, ebb1
>     jump ebb2(v3)
      ^~~~~~~~~~~~~
Matched #0: \bjump ebb2\(v3\)
> 
> ebb1:
>     jump ebb2(v3)
      ^~~~~~~~~~~~~
Matched #1: \bjump ebb2\(v3\)
Missed #2: \bv8 = iadd v6, v3\b
> 
> ebb2(v6: i32):
>     v7 = iadd.i32 v0, v1
>     v8 = iadd v6, v7
>     return v8
> }

1 tests
1 failure
```

And the trace log shows GVN performing 2 operations and then running out of fuel:
```
$ cat cretonne.dbg.worker0
performing simple_gvn
  simple_gvn opt: replace redundant inst2 with inst0
  simple_gvn opt: replace redundant inst4 with inst0
  simple_gvn opt disabled: replace redundant inst6 with inst0
```

stoklund:
Thanks!

I think I would prefer for the debugging API to be less intrusive, using thread-local variables instead of a `DebugState` object that you would need to pass around everywhere.

Compare the existing `dbg!` macro to the new `trace!` macro.

I also think optimization fuel would be useful at a scope larger than a single function. Threads are a problem, of course. I think we should simply have an option to force a single thread when using it.

sunfishcode:
Another thing to look into here is either using the [log](https://docs.rs/log/) crate for logging. I know Cretonne wants to avoid dependencies, bug `log` is pretty minimal, being factored such that the code generating the log messages (eg. Cretonne) doesn't have a dependency on the code for handling them (eg. a tool using Cretonne).

sunfishcode:
I'm still planning to get back to this, but it'll need a rewrite, so I'll open a new PR when I'm ready.

PR #109, opened 2017-06-30
Fixed bug in verifier

denismerigoux:
Does not check variable def for unreachable code.

Fixes https://github.com/stoklund/cretonne/issues/108.

### DIFF
--- a/lib/cretonne/src/verifier/mod.rs
+++ b/lib/cretonne/src/verifier/mod.rs
@@ -364,7 +364,9 @@ impl<'a> Verifier<'a> {
                 }
                 // Defining instruction dominates the instruction that uses the value.
                 if !self.domtree
-                        .dominates(def_inst, loc_inst, &self.func.layout) {
+                        .dominates(def_inst, loc_inst, &self.func.layout) &&
+                   self.domtree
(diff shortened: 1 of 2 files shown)


lib/cretonne/src/verifier/mod.rs
```diff
@@ -364,7 +364,9 @@ impl<'a> Verifier<'a> {
                 }
                 // Defining instruction dominates the instruction that uses the value.
                 if !self.domtree
-                        .dominates(def_inst, loc_inst, &self.func.layout) {
+                        .dominates(def_inst, loc_inst, &self.func.layout) &&
+                   self.domtree
+                       .is_reachable(self.func.layout.inst_ebb(loc_inst).unwrap()) {
```
stoklund: Note there's a `Layout::pp_ebb()` method which lets you avoid the unwrap. (PP is for program point).

I think it would be better to check for reachability first, so we never call `dominates` with unreachable code. I think it works now, but I could imagine `dominates` asserting on unreachable code in the future.

denismerigoux: Fixed. I also included a file test.

PR #105, opened 2017-06-24
Add inverse image computation for TypeSets and generalize TypeVar.constrain_types

d1m0:
Add computation of inverse image of typeset across a derived function - TypeSet.map_inverse; 
Change TypeVar.constrain_type to perform a more-general computation using inverse images of TypeSets;
Tests for map_inverse;
Cleanup ValueType.get_names to with_bits form previous PR;

PR #110, opened 2017-07-03
Add better type inference and encapsulate it in its own file

d1m0:
This PR builds on top of PRs #104 and PR #105 and fills in the blanks in type inference. Specifically:

1) TI is separated from TypeVar/XForm code in its own file (ti.py)

2) More general unification between derived type variables(tvs) that:
        - normalizes derived TVs that involve bijective functions
        - unifies derived tvs with bijective derivation functions by unifying one's base with the other's inverse image

3) Support for constraints between non-bijective derived tvs. This is useful for both handling equivalence between as_bool() derived tvs, as well as future use cases (e.g. width constraints in uextend, sextend, itrunc)

4) The ti code doesn't depend on having a single controlling tv

5) We add an enumerator for all concrete possible typings for an XForm. The enumerator respects constraints.

Also this adds the renaming of map/map_inverse to image/preimage that somehow slipped from PR #105

### DIFF
--- a/lib/cretonne/meta/cdsl/ti.py
+++ b/lib/cretonne/meta/cdsl/ti.py
@@ -1,0 +1,556 @@
+"""
+Type Inference
+"""
+from .typevar import TypeVar
+from .ast import Def, Var
+from copy import copy
+from itertools import product
+
+try:
+    from typing import Dict, TYPE_CHECKING, Union, Tuple, Optional, Set # noqa
+    from typing import Iterable # noqa
+    from typing import cast, List
+    from .xform import Rtl, XForm # noqa
+    from .ast import Expr # noqa
+    if TYPE_CHECKING:
+        Constraint = Tuple[TypeVar, TypeVar]
+        ConstraintList = List[Constraint]
+        TypeMap = Dict[TypeVar, TypeVar]
+        VarMap = Dict[Var, TypeVar]
+except ImportError:
+    TYPE_CHECKING = False
+    pass
+
+
+class TypeEnv(object):
+    """
+    Class encapsulating the neccessary book keeping for type inference.
+        :attribute type_map: dict holding the equivalence relations between tvs
+        :attribute constraints: a list of accumulated constraints - tuples
+                            (tv1, tv2)) where tv1 and tv2 are equal
+        :attribute ranks: dictionary recording the (optional) ranks for tvs.
+                      tvs corresponding to real variables have explicitly
+                      specified ranks.
+        :attribute vars: a set containing all known Vars
+        :attribute idx: counter used to get fresh ids
+    """
+    def __init__(self, arg=None):
+        # type: (Optional[Tuple[TypeMap, ConstraintList]]) -> None
+        self.ranks = {}  # type: Dict[TypeVar, int]
+        self.vars = set()  # type: Set[Var]
+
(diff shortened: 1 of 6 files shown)


lib/cretonne/meta/cdsl/ti.py
```diff
@@ -0,0 +1,556 @@
...
+
+        if v.is_input():
+            r = 4
+        elif v.is_intermediate():
+            r = 3
+        elif v.is_output():
+            r = 2
+        else:
+            assert(v.is_temp())
+            r = 1
```
stoklund: Is it worthwhile to use global constants for these magic numbers, or do they only appear in this function?

lib/cretonne/meta/cdsl/ti.py
```diff
@@ -0,0 +1,556 @@
...
+
+        self.ranks[v.get_typevar()] = r
+
+    def free_typevars(self):
+        # type: () -> Set[TypeVar]
```
stoklund: Be careful here that we don't introduce a dependency on Python's set ordering. The result of type checking should not depend on it.

lib/cretonne/meta/cdsl/ti.py
```diff
@@ -0,0 +1,556 @@
...
+            s = children.get(t, set())
+            s.add(v)
+            children[t] = s
+
+        for (a, b) in self.type_map.items():
+            s = children.get(b, set())
+            s.add(a)
+            children[b] = s
+
+        for r in list(self.free_typevars()):
```
stoklund: Here, for example. It is not obvious that this algorithm would yield the same result for all set orderings. (I'm not saying it doesn't).

It would be good to either add a comment describing why the algorithm is order-independent, or iterate over a canonical order of the type variables.

lib/cretonne/meta/cdsl/ti.py
```diff
@@ -0,0 +1,556 @@
...
+
+
+def get_error(typing_or_err):
+    # type: (TypingOrError) -> Optional[TypingError]
+    """
+    Helper function to appease mypy when checking the result of typing.
+    """
+    if isinstance(typing_or_err, str):
+        if (TYPE_CHECKING):
+            return cast(TypingError, typing_or_err)
```
stoklund: I'm surprised mypy can't figure this out without a cast.

lib/cretonne/meta/cdsl/ti.py
```diff
@@ -0,0 +1,556 @@
...
+        # Reordering: {HALFWIDTH, DOUBLEWIDTH} commute with {HALFVECTOR,
+        # DOUBLEVECTOR}. Arbitrarily pick WIDTH < VECTOR
+        if df in vector_derives and base_df in width_derives:
+            return normalize_tv(
+                    TypeVar.derived(
+                        TypeVar.derived(tv.base.base, df), base_df))
+
+        # Cancelling: HALFWIDTH, DOUBLEWIDTH and HALFVECTOR, DOUBLEVECTOR
+        # cancel each other. TODO: Does this cancellation hide type
+        # overflow/underflow?
```
stoklund: If the domain of the base type variable was restricted at the time the derived type variable is formed, there shouldn't be a problem. When the type functions cancel out, we *don't* increase the base domain again.

PR #112, opened 2017-07-07
Emit runtime type checks in legalizer.rs

d1m0:
This PR emits runtime type checks for patters in legalizer.rs. Specifically it emits checks under 2 conditions:

1) A most-general type of a var v in a XForm(r1,r2) is more constraint then its most general type in just r1. For these cases, the fact that r1 is valid RTL is not sufficient to determine if XForm applies. We need to emit a check of the type typeof_v \in TS where TS is the typeset of v in the XForm

2) During type checking we may generate constraints between derived type vars that have non-bijective derived functions (e.g. as_bool). For those, we can't simply propagate the constraint up by constraining the root free variables, and instead must add an explicit runtime check.

In the future, we will add more  constraints. The for example we will need constraints for the following instructions:
      uextend/sextend - a constraint that the output type is wider than the input
      itrunc - a constraint that the output type is narrower than the input
      bitcast - TBD

### DIFF
--- a/lib/cretonne/meta/gen_legalizer.py
+++ b/lib/cretonne/meta/gen_legalizer.py
@@ -15,13 +15,14 @@ from cdsl.ti import ti_rtl, TypeEnv, get_type_env, ConstrainTVsEqual,\
     ConstrainTVInTypeset
 from unique_table import UniqueTable
 from gen_instr import gen_typesets_table
+from cdsl.typevar import TypeVar
 
 try:
     from typing import Sequence, List, Dict # noqa
     from cdsl.isa import TargetISA  # noqa
     from cdsl.ast import Def  # noqa
     from cdsl.xform import XForm, XFormGroup  # noqa
-    from cdsl.typevar import TypeVar, TypeSet # noqa
+    from cdsl.typevar import TypeSet # noqa
     from cdsl.ti import TypeConstraint # noqa
 except ImportError:
     pass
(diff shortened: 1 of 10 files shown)


lib/cretonne/meta/gen_legalizer.py
```diff
@@ -203,6 +298,10 @@ def gen_xform(xform, fmt):
     instp = xform.src.rtl[0].expr.inst_predicate()
     assert instp is None, "Instruction predicates not supported in legalizer"
 
+    # Emit any runtime checks. TODO: Does this make the above code obsolete?
```
stoklund: No it doesn't. Instruction predicates are generally limits on immediate fields, like "imm must be a 7-bit signed integer".

lib/cretonne/src/ir/types.rs
```diff
@@ -35,6 +35,22 @@ pub const VOID: Type = Type(0);
 // 512-bit SIMD vectors.
 include!(concat!(env!("OUT_DIR"), "/types.rs"));
 
+/// Possible derivation functions for types. Used in runtime typechecking
+pub enum DerivedFunctions {
```
stoklund: Are you sure this (and the `transform` function below) is necessary?

It seems unlikely that we would need to apply type functions that are unknown at compile time.

d1m0: It isn't neccessary. I added it just to simplify the work of gen_legalizer.py in the face of possible over/underflow in the "derived" types we are comparing. I removed it in 
29b987ac8713a8a8764bf2f26a8fac8a62b6f4dc

d1m0: On second thought, it might make sense to add an assert before the check that the "derived" type is not None (i.e. there is no over/underflow)?

lib/cretonne/meta/gen_legalizer.py
d1m0: Comment removed.

Issue #114, opened 2017-07-10
Add WiderThan constraint


d1m0:
Instructions such as sextend, uextend and ireduce require a type constraint between the widths of the argument and the result value. One way to implement this is to:

- Add a ConstrainWiderThan TypeConstraint in ti.py
- Extend Instruction constructors to allow for constraints amongst formal vars
- Add the instruction specific constraints to the constraints accumulated by typeinference.

~Dimo

stoklund:
This sounds good to me. Also consider the constraints for the `bitcast` instruction. It needs the input and output types to be of the same size, but they are not required to have the same number of vector lanes.

PR #115, opened 2017-07-11
Add fix for #114

d1m0:
This PR does several things:
    1. Add a fix for #114 - Add `ConstrainWiderOrEqual`
    2. Adds the ability for `Instructions` to have constraints, and add a `ConstrainWiderOrEqual` to `ireduce`,`{s,u}extend` and `f{promote,demote}`
    3. Reduce code duplication in `TypeConstraint` subclasses
    4. Modify runtime constraint checks to reject match when they encounter overflow
    5. Fix a small bug in emitting constraint edges in `TypeEnv.dot()`;

### DIFF
--- a/lib/cretonne/meta/gen_legalizer.py
+++ b/lib/cretonne/meta/gen_legalizer.py
@@ -12,7 +12,7 @@ from srcgen import Formatter
 from base import legalize, instructions
 from cdsl.ast import Var
 from cdsl.ti import ti_rtl, TypeEnv, get_type_env, ConstrainTVsEqual,\
-    ConstrainTVInTypeset
+    ConstrainTVInTypeset, ConstrainWiderOrEqual
 from unique_table import UniqueTable
 from gen_instr import gen_typesets_table
 from cdsl.typevar import TypeVar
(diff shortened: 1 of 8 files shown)


lib/cretonne/meta/gen_legalizer.py
```diff
@@ -113,10 +122,28 @@ def build_derived_expr(tv):
...
+            fmt.line('let a = {};'.format(build_derived_expr(check.tv1)))
+            fmt.line('let b = {};'.format(build_derived_expr(check.tv2)))
+
+            fmt.comment('On overflow constraint doesn\'t appply')
+            with fmt.indented('if a.is_none() || b.is_none() {', '};'):
+                fmt.line('return false;')
+
+            with fmt.indented('if (a.lane_count() != b.lane_count()) ||' +
+                              '   (a.lane_bits() < b.lane_bits()) {', '};'):
+                fmt.line('return false;')
```
stoklund: This looks like it should just call a function in types.rs.

d1m0:
One more question: In the current implementation i32 is considered "WiderOrEqual" compared to b8. Should I limit it to require that the two lane types are of the same kind as well?

stoklund:
No need, I think comparing lane bits is good enough

PR #117, opened 2017-07-21
Semantics

d1m0:
This PR includes the basic machinery to specify instruction semantics in terms of a sequence of XForms. In order of commits it includes:

1) Helper routines {Ast,Def,Rtl}.vars(), Def.{definitions(), uses{}), {Ast, Def,Rtl}.substitutions()

2) The new BVType and support for BVType in TypeVar/TypeSet. BVType represents a flat bitvector of the same size as a real cretonne type. Its non-emittable, and only used by primitive instructions to describe the semantics of larger instructions. It corresponds to a smtlib bitvector.

3) Instruction semantics. This includes:
        - semantics/primitives.py - a new InstructionGroup containing the 'primitive' set of instructions in terms of which we describe all other instructions. All of these should correspond almost 1-1 to bitvector functions in smtlib
        - semantics/elaborate.py - routines to repeatedly in-line non-primitive instructions in an Rtl until we've reached a semantically equivalent Rtl with all-primitive instructions
        - Instruction.semantics field, and sanity checking of semantics using verify_semantics()
        - base/semantics.py - sample semantics for enough instructions to describe iadd

### DIFF
--- a/lib/cretonne/meta/base/semantics.py
+++ b/lib/cretonne/meta/base/semantics.py
@@ -2,11 +2,10 @@ from __future__ import absolute_import
 from semantics.primitives import prim_to_bv, prim_from_bv, bvsplit, bvconcat,\
     bvadd
 from .instructions import vsplit, vconcat, iadd
-from cdsl.xform import XForm, Rtl
+from cdsl.xform import Rtl
 from cdsl.ast import Var
 from cdsl.typevar import TypeSet
 from cdsl.ti import InTypeset
-import semantics.types # noqa
 
 x = Var('x')
 y = Var('y')
@@ -28,30 +27,32 @@ bvhi = Var('bvhi')
 ScalarTS = TypeSet(lanes=(1, 1), ints=True, floats=True, bools=True)
 
 vsplit.set_semantics(
-    XForm(Rtl((lo, hi) << vsplit(x)),
-          Rtl(bvx << prim_to_bv(x),
-              (bvlo, bvhi) << bvsplit(bvx),
-              lo << prim_from_bv(bvlo),
-              hi << prim_from_bv(bvhi))))
+    (lo, hi) << vsplit(x),
+    Rtl(
+        bvx << prim_to_bv(x),
+        (bvlo, bvhi) << bvsplit(bvx),
+        lo << prim_from_bv(bvlo),
+        hi << prim_from_bv(bvhi)
+    ))
 
 vconcat.set_semantics(
-    XForm(Rtl(x << vconcat(lo, hi)),
-          Rtl(bvlo << prim_to_bv(lo),
-              bvhi << prim_to_bv(hi),
-              bvx << bvconcat(bvlo, bvhi),
-              x << prim_from_bv(bvx))))
+    x << vconcat(lo, hi),
+    Rtl(
+        bvlo << prim_to_bv(lo),
+        bvhi << prim_to_bv(hi),
+        bvx << bvconcat(bvlo, bvhi),
+        x << prim_from_bv(bvx)
+    ))
 
-iadd.set_semantics([
-     XForm(Rtl(a << iadd(x, y)),
-           Rtl(bvx << prim_to_bv(x),
-               bvy << prim_to_bv(y),
-               bva << bvadd(bvx, bvy),
-               a << prim_from_bv(bva)),
-           constraints=[InTypeset(x.get_typevar(), ScalarTS)]),
-     XForm(Rtl(a << iadd(x, y)),
-           Rtl((xlo, xhi) << vsplit(x),
-               (ylo, yhi) << vsplit(y),
-               alo << iadd(xlo, ylo),
-               ahi << iadd(xhi, yhi),
-               a << vconcat(alo, ahi)))
-])
+iadd.set_semantics(
+    a << iadd(x, y),
+    (Rtl(bvx << prim_to_bv(x),
+         bvy << prim_to_bv(y),
+         bva << bvadd(bvx, bvy),
+         a << prim_from_bv(bva)),
+     [InTypeset(x.get_typevar(), ScalarTS)]),
+    Rtl((xlo, xhi) << vsplit(x),
+        (ylo, yhi) << vsplit(y),
+        alo << iadd(xlo, ylo),
+        ahi << iadd(xhi, yhi),
+        a << vconcat(alo, ahi)))
--- a/lib/cretonne/meta/base/semantics.py
+++ b/lib/cretonne/meta/base/semantics.py
@@ -1,0 +1,57 @@
+from __future__ import absolute_import
+from semantics.primitives import prim_to_bv, prim_from_bv, bvsplit, bvconcat,\
+    bvadd
+from .instructions import vsplit, vconcat, iadd
+from cdsl.xform import XForm, Rtl
+from cdsl.ast import Var
+from cdsl.typevar import TypeSet
+from cdsl.ti import InTypeset
+import semantics.types # noqa
+
+x = Var('x')
+y = Var('y')
+a = Var('a')
+xhi = Var('xhi')
+yhi = Var('yhi')
+ahi = Var('ahi')
+xlo = Var('xlo')
+ylo = Var('ylo')
+alo = Var('alo')
+lo = Var('lo')
+hi = Var('hi')
+bvx = Var('bvx')
+bvy = Var('bvy')
+bva = Var('bva')
+bvlo = Var('bvlo')
+bvhi = Var('bvhi')
+
+ScalarTS = TypeSet(lanes=(1, 1), ints=True, floats=True, bools=True)
+
+vsplit.set_semantics(
+    XForm(Rtl((lo, hi) << vsplit(x)),
+          Rtl(bvx << prim_to_bv(x),
+              (bvlo, bvhi) << bvsplit(bvx),
+              lo << prim_from_bv(bvlo),
+              hi << prim_from_bv(bvhi))))
+
+vconcat.set_semantics(
+    XForm(Rtl(x << vconcat(lo, hi)),
+          Rtl(bvlo << prim_to_bv(lo),
+              bvhi << prim_to_bv(hi),
+              bvx << bvconcat(bvlo, bvhi),
+              x << prim_from_bv(bvx))))
+
+iadd.set_semantics([
+     XForm(Rtl(a << iadd(x, y)),
+           Rtl(bvx << prim_to_bv(x),
+               bvy << prim_to_bv(y),
+               bva << bvadd(bvx, bvy),
+               a << prim_from_bv(bva)),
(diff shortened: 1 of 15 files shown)


lib/cretonne/meta/cdsl/ast.py
```diff
@@ -88,6 +90,40 @@ def copy(self, m):
...
+    def vars(self):
+        # type: () -> Set[Var]
+        """ Return the set of all Vars that appear in self"""
+        return self.definitions().union(self.uses())
+
+    def substitution(self, other, s):
+        # type: (Def, VarMap) -> Optional[VarMap]
+        """
+        If the Defs self and other agree structurally, return a variable
+        substitution to transform self ot other. Two Defs agree structurally
```
stoklund: typo: ot

Also, could you clarify in the documentation what it means to "agree structurally"?

lib/cretonne/meta/cdsl/ast.py
```diff
@@ -332,14 +368,51 @@ def inst_predicate(self):
...
         """
         Return a copy of this Expr with vars replaced with fresh variables,
         in accordance with the map m. Update m as neccessary.
         """
         return Apply(self.inst, tuple(map(lambda e: replace_var(e, m),
                                           self.args)))
 
+    def vars(self):
+        # type: () -> Set[Var]
+        """ Return the set of all Vars that appear in self"""
```
stoklund: Should it be clarified that this is only the Vars corresponding to SSA values?

Vars can also refer to immediate operands:

```python
expand.legalize(
        a << icmp_imm(cc, x, y),
        Rtl(
            a1 << iconst(y),
            a << icmp(cc, x, a1)
        ))
```

lib/cretonne/meta/cdsl/ast.py
```diff
@@ -332,14 +368,51 @@ def inst_predicate(self):
...
+            arg = self.args[i]
+            assert isinstance(arg, Var)
+            res.add(arg)
+        return res
+
+    def substitution(self, other, s):
+        # type: (Apply, VarMap) -> Optional[VarMap]
+        """
+        If the application self and other agree structurally, return a variable
+        substitution to transform self ot other. Two applications agree
```
stoklund: typo: ot
add: "Otherwise, return `None`".

That other mention of "agree structurally" can simply refer to the definition here.

lib/cretonne/meta/cdsl/ast.py
```diff
@@ -332,14 +368,51 @@ def inst_predicate(self):
...
+        structurally if:
+            1) They are over the same instruction
+            2) Every Var v in self, maps to a single Var w in other. I.e for
+               each use of v in self, w is used in the corresponding place in
+               other.
+        """
+        if self.inst != other.inst:
+            return None
+
+        # TODO: Should we check imm/cond codes here as well?
```
stoklund: Probably, yes. Immediate operands (including condition codes) can be either Vars or literals.

Operands can also be entity references, most commonly a destination EBB for a branch.

lib/cretonne/meta/cdsl/types.py
```diff
@@ -84,7 +84,10 @@ def __init__(self, name, membytes, doc):
         self._vectors = dict()  # type: Dict[int, VectorType]
         # Assign numbers starting from 1. (0 is VOID).
         ValueType.all_scalars.append(self)
-        self.number = len(ValueType.all_scalars)
+        # Numbers are only valid for Cretone types that get emitted to Rust.
+        # This excludes BVTypes
+        self.number = len([x for x in ValueType.all_scalars
+                           if not isinstance(x, BVType)])
```
stoklund: Perhaps it wold be better to make a whole new `ValueType` subclass for the bit vectors? They will need special treatment everywhere.

And something called "bit vector" doesn't strike me as belonging to "scalar types" ;-)

lib/cretonne/meta/base/semantics.py
```diff
@@ -0,0 +1,57 @@
...
+hi = Var('hi')
+bvx = Var('bvx')
+bvy = Var('bvy')
+bva = Var('bva')
+bvlo = Var('bvlo')
+bvhi = Var('bvhi')
+
+ScalarTS = TypeSet(lanes=(1, 1), ints=True, floats=True, bools=True)
+
+vsplit.set_semantics(
```
stoklund: I like this syntax!

lib/cretonne/meta/base/semantics.py
```diff
@@ -0,0 +1,57 @@
...
+    XForm(Rtl((lo, hi) << vsplit(x)),
+          Rtl(bvx << prim_to_bv(x),
+              (bvlo, bvhi) << bvsplit(bvx),
+              lo << prim_from_bv(bvlo),
+              hi << prim_from_bv(bvhi))))
+
+vconcat.set_semantics(
+    XForm(Rtl(x << vconcat(lo, hi)),
```
stoklund: Can we reduce some of the `XForm(Rtl(` noise here? Compare to the `legalize` method:

```python
expand.legalize(
        a << iadd_imm(x, y),
        Rtl(
            a1 << iconst(y),
            a << iadd(x, a1)
        ))
```

lib/cretonne/meta/cdsl/instructions.py
```diff
@@ -12,11 +12,12 @@
         from .ast import Expr, Apply, Var  # noqa
         from .typevar import TypeVar  # noqa
         from .ti import TypeConstraint  # noqa
+        from .xform import XForm
         # List of operands for ins/outs:
         OpList = Union[Sequence[Operand], Operand]
         ConstrList = Union[Sequence[TypeConstraint], TypeConstraint]
         MaybeBoundInst = Union['Instruction', 'BoundInstruction']
-        VarTyping = Dict[Var, TypeVar]
+        InstructionSemantics = List[XForm]
```
stoklund: Prefer `Sequence`.

lib/cretonne/meta/cdsl/instructions.py
```diff
@@ -334,6 +336,17 @@ def __call__(self, *args):
         from .ast import Apply  # noqa
         return Apply(self, args)
 
+    def set_semantics(self, sem):
+        # type: (Union[XForm, InstructionSemantics]) -> None
```
stoklund: How about:

```python
 def set_semantics(self, src, *dsts):
+        # type: (Union[Def, Apply], *Rtl) -> None
```

Compare `XFormGroup.legalize()`.

lib/cretonne/meta/semantics/__init__.py
```diff
@@ -0,0 +1,55 @@
...
+            "XForm {} doesn't describe instruction {}."\
+            .format(xform, inst)
+
+    # 2) Any possible typing for the instruction should be covered by
+    #    exactly ONE semantic XForm
+    inst_rtl = sem[0].src
+    typenv = get_type_env(ti_rtl(inst_rtl, TypeEnv()))
+
+    # This bit is awkward. Concrete typing is defined in terms of the vars
+    # of one Rtl. We arbitrarily picked that Rtl to be sem[0].src. For any
```
stoklund: Instead of defining semantics as a sequence of `XForms`, how about defining it as a single source `Rtl` and a sequence of destination `Rtl`s?

lib/cretonne/meta/semantics/types.py
```diff
@@ -0,0 +1,9 @@
+"""
+The semantics.types module predefines all the Cretone primitive bitvector
+types.
+"""
+from cdsl.types import BVType
+from cdsl.typevar import MAX_BITVEC, int_log2
+
+for width in range(0, int_log2(MAX_BITVEC)+1):
+    BVType(2**width)
```
stoklund: The `BVType.with_bits()` is a bit gross, looking up existing types by name.

How about adding a `BVType._registry` dictionary that maps bits to the corresponding bitvector type. Then `with_bits` can generate types on demand and add then to the registry. This is OK for the bitvectors since they don't appear in rust code.

lib/cretonne/meta/semantics/__init__.py
d1m0: Externally - from the view of callers of verify_semantics() and set_semantics() I made the semantics a source Rtl + a sequence of destination Rtls. Internally, it seemed simpler to me to keep Instruction.semantics as a Sequence[XForm].

lib/cretonne/meta/semantics/types.py
d1m0: An advantage of using ValueType._registry and lookup by name is that we can use existing code to bind primitive instructions. E.g. in test_elaborate.py I use prim_bvsplit.bv64 etc. I modified BVType.with_bits() to generate types on the fly, and removed semantics/types.py

lib/cretonne/meta/cdsl/instructions.py
d1m0: Fixed.

lib/cretonne/meta/cdsl/instructions.py
d1m0: Fixed.

lib/cretonne/meta/base/semantics.py
d1m0: Fixed.

lib/cretonne/meta/base/semantics.py
d1m0: Fixed. One small nit: Instead of adding a new constraints field to Rtls, set_semantics accepts either Rtls, or Tuple[Rtl, List[TypeConstraint]]. I am not sure what is the impact of adding TypeConstraints to Rtls to typechecking XForms. If it is a strong preference to include constraints as a part of an Rtl, I can change that.

lib/cretonne/meta/cdsl/types.py
d1m0: Fixed.

lib/cretonne/meta/cdsl/ast.py
d1m0: Fixed.

lib/cretonne/meta/cdsl/ast.py
d1m0: Fixed.

lib/cretonne/meta/cdsl/ast.py
d1m0: Fixed.

lib/cretonne/meta/cdsl/ast.py
d1m0: Do we want to leave the TODO here? Or stop and implement this first?

stoklund: No, you can fix it when you hit problems.

stoklund:
@d1m0 are you not able to run `./test-all.sh` locally?

stoklund:
Tests are passing now, but it looks like you need to rebase before we can merge.

d1m0:
Rebased.

PR #123, opened 2017-07-28
Add more semantics and utilities to emit smtlib queries

d1m0:
This PR adds bug fixes and machinery neccessary to express semantics of RTLs as smtlib queries for equivalence checking. In detail:
      1) bug fixes in TI to make sure tvs in a dest pattern are always expressed as a function of the free tvs in the source pattern
      2) cleanup - moving some routines such as apply(), is_concrete_rtl() and cleanup_concrete_rtl() from semantics/elaborate() to better suited locations in cdsl/xforms.py
      3) smtlib.py - routines for emitting queries for concretely typed Rtls
      4) Stricter constraints for binary conversion instructions
      5) additional semantic primitives and semantics for some arithmetic instructions, sufficient to emit queries checking iadd legalizations.

### DIFF
--- a/lib/cretonne/meta/base/semantics.py
+++ b/lib/cretonne/meta/base/semantics.py
@@ -1,7 +1,8 @@
 from __future__ import absolute_import
 from semantics.primitives import prim_to_bv, prim_from_bv, bvsplit, bvconcat,\
     bvadd, bvult, bvzeroext
-from .instructions import vsplit, vconcat, iadd, iadd_cout, icmp, bextend
+from .instructions import vsplit, vconcat, iadd, iadd_cout, icmp, bextend, \
+    isplit, iconcat, iadd_cin, iadd_carry
 from .immediates import intcc
 from cdsl.xform import Rtl
 from cdsl.ast import Var
@@ -13,18 +14,24 @@ y = Var('y')
 a = Var('a')
 b = Var('b')
 c_out = Var('c_out')
+c_in = Var('c_in')
 bvc_out = Var('bvc_out')
+bvc_in = Var('bvc_in')
 xhi = Var('xhi')
 yhi = Var('yhi')
 ahi = Var('ahi')
+bhi = Var('bhi')
 xlo = Var('xlo')
 ylo = Var('ylo')
 alo = Var('alo')
+blo = Var('blo')
 lo = Var('lo')
 hi = Var('hi')
 bvx = Var('bvx')
 bvy = Var('bvy')
 bva = Var('bva')
+bvt = Var('bvt')
+bvs = Var('bvs')
 bva_wide = Var('bva_wide')
 bvlo = Var('bvlo')
 bvhi = Var('bvhi')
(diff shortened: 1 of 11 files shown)


lib/cretonne/meta/cdsl/ti.py
```diff
@@ -8,12 +8,13 @@
...
 try:
     from typing import Dict, TYPE_CHECKING, Union, Tuple, Optional, Set # noqa
-    from typing import Iterable, List, Any # noqa
+    from typing import Iterable, List, Any, TypeVar as MTypeVar # noqa
     from typing import cast
     from .xform import Rtl, XForm # noqa
     from .ast import Expr # noqa
     from .typevar import TypeSet # noqa
     if TYPE_CHECKING:
+        T = MTypeVar('T')
```
stoklund: This doesn't appear to be used anywhere.

lib/cretonne/meta/base/semantics.py
```diff
@@ -56,3 +61,42 @@
...
+        bvc_out << bvult(bva, bvx),
+        a << prim_from_bv(bva),
+        c_out << prim_from_bv(bvc_out)
+    ))
+
+bextend.set_semantics(
+    a << bextend(x),
+    (Rtl(
+        bvx << prim_to_bv(x),
+        bvy << bvzeroext(bvx),
```
stoklund: I don't know if the bitwise representation of the Boolean types gets exposed anywhere, but this should probably be a `bvsignext`. The Boolean types are represented as the values 0/-1 in registers.

Only a b1 is represented as 0/1 (which I guess is the same as 0/-1 in a 1-bit register)

lib/cretonne/meta/semantics/smtlib.py
```diff
@@ -0,0 +1,150 @@
+"""
+Tools to emit SMTLIB bitvector queries encoding concrete RTLs containing only
+primitive instructions.
```
stoklund: This is ok, but depending on the z3 Python bindings would also be fine. We just don't want z3 to be a build dependency, but it can be a proof dependency.

lib/cretonne/meta/base/semantics.py
```diff
@@ -1,7 +1,8 @@
 from __future__ import absolute_import
 from semantics.primitives import prim_to_bv, prim_from_bv, bvsplit, bvconcat,\
     bvadd, bvult, bvzeroext
-from .instructions import vsplit, vconcat, iadd, iadd_cout, icmp, bextend
+from .instructions import vsplit, vconcat, iadd, iadd_cout, icmp, bextend, \
+    isplit, iconcat, iadd_cin, iadd_carry
```
stoklund: In the encodings files, I tend to go `import .instructions as insts`.

lib/cretonne/meta/semantics/smtlib.py
d1m0: I can file a followup issue to change this. It would definitely be cleaner to depend on z3's python bindings.

PR #126, opened 2017-07-29
[WIP] Conditional compilation of isa targets

### LINKED ISSUE #13: Conditional compilation of ISA targets

zummenix:
Eventually will close #13 

This is partially based on work of @dinfuehr  in #32. I've addressed some code review comments, but I need help to figure out how to skip unsupported isa targets in filecheck runner.

### DIFF
--- a/lib/reader/src/parser.rs
+++ b/lib/reader/src/parser.rs
@@ -621,8 +621,6 @@ impl<'a> Parser<'a> {
                                            &self.loc)?;
                 }
                 "isa" => {
-                    last_set_loc = None;
-                    seen_isa = true;
                     let loc = self.loc;
                     // Grab the whole line so the lexer won't go looking for tokens on the
                     // following lines.
@@ -633,9 +631,16 @@ impl<'a> Parser<'a> {
                         Some(w) => w,
                     };
                     let mut isa_builder = match isa::lookup(isa_name) {
-                        None => return err!(loc, "unknown ISA '{}'", isa_name),
-                        Some(b) => b,
+                        Err(isa::LookupError::Unknown) => {
+                            return err!(loc, "unknown ISA '{}'", isa_name)
+                        }
+                        Err(isa::LookupError::Unsupported) => {
+                            continue;
+                        }
+                        Ok(b) => b,
                     };
+                    last_set_loc = None;
+                    seen_isa = true;
                     // Apply the ISA-specific settings to `isa_builder`.
                     isaspec::parse_options(words, &mut isa_builder, &self.loc)?;
 
@@ -1939,6 +1944,7 @@ mod tests {
     }
 
     #[test]
+    #[cfg(build_riscv)]
     fn isa_spec() {
         assert!(parse_test("isa
                             function %foo() {}")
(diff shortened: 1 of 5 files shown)


stoklund:
Sorry for the late review, thanks for working on this!

If you like, we can land this as is, but there are two things missing:

1. I would prefer to control the configuration with cargo features instead of environment variables.
2. As you mention, the test runner should be aware of configured-out targets.

For the test runner, look at the code in `src/filetest/runone.rs`:

```rust
    // Flags to use for those tests that don't need an ISA.
    // This is the cumulative effect of all the `set` commands in the file.
    let flags = match testfile.isa_spec {
        IsaSpec::None(ref f) => f,
        IsaSpec::Some(ref v) => v.last().expect("Empty ISA list").flags(),
    };
```

As I mentioned in the review of @dinfuehr's patch, there should be three possibilities for the `IsaSpec`:

1. The file has no ISA specs: Use `IsaSpec::None(..)`.
2. The file has ISA specs, but none are supported: Use `IsaSpec::Some([])`.
3. The file has ISA specs, not all supported: Use `IsaSpec::Some([a, b, ...])` with just the supported ISAs. Any unsupported ISA specs are ignored and omitted.

The test runner simply needs to handle the case of `IsaSpec::Some([])`. This corresponmds to a test file that is written specifically for ISAs that are not compiled in. This means that there is nothing to do for the test runner, and it should simply make the test always pass by returning `Ok(..)` early.

lib/reader/src/parser.rs
```diff
@@ -633,9 +631,16 @@ impl<'a> Parser<'a> {
...
                     };
                     let mut isa_builder = match isa::lookup(isa_name) {
-                        None => return err!(loc, "unknown ISA '{}'", isa_name),
-                        Some(b) => b,
+                        Err(isa::LookupError::Unknown) => {
+                            return err!(loc, "unknown ISA '{}'", isa_name)
+                        }
+                        Err(isa::LookupError::Unsupported) => {
+                            continue;
+                        }
```
stoklund: This looks right to me, but it would be good to add a test that verifies the `IsaSpec` returned in the three cases:

1. Test file has no ISA specs.
2. Test file has valid ISA specs, but they are not supported.
3. Test file contains valid and supported ISA specs.

zummenix:
@stoklund, I'm planning to continue this on the weekend, but no guarantee, so you can land this as is if you want and I just continue in new PR.

PR #127, opened 2017-07-31
Added Intel x86-64 encodings for 64bit loads and store instructions

denismerigoux:
Encoding recipes factored with 32 bits instructions, corresponding tests added.

### DIFF
--- a/filetests/isa/intel/binary64.cton
+++ b/filetests/isa/intel/binary64.cton
@@ -167,33 +167,33 @@ ebb0:
     istore8 v2, v1                              ; bin: 40 88 31
 
     ; asm: movq (%rcx), %rdi
-    [-,%rdi]            v460 = load.i64 v1      ; bin: 48 8b 39
+    [-,%rdi]            v120 = load.i64 v1      ; bin: 48 8b 39
     ; asm: movq (%rsi), %rdx
-    [-,%rdx]            v461 = load.i64 v2      ; bin: 48 8b 16
+    [-,%rdx]            v121 = load.i64 v2      ; bin: 48 8b 16
     ; asm: movl (%rcx), %edi
-    [-,%rdi]            v400 = uload32.i64 v1   ; bin: 40 8b 39
+    [-,%rdi]            v122 = uload32.i64 v1   ; bin: 40 8b 39
     ; asm: movl (%rsi), %edx
-    [-,%rdx]            v491 = uload32.i64 v2   ; bin: 40 8b 16
+    [-,%rdx]            v123 = uload32.i64 v2   ; bin: 40 8b 16
     ; asm: movslq (%rcx), %rdi
-    [-,%rdi]            v490 = sload32.i64 v1   ; bin: 48 63 39
+    [-,%rdi]            v124 = sload32.i64 v1   ; bin: 48 63 39
     ; asm: movslq (%rsi), %rdx
-    [-,%rdx]            v401 = sload32.i64 v2   ; bin: 48 63 16
+    [-,%rdx]            v125 = sload32.i64 v2   ; bin: 48 63 16
     ; asm: movzwq (%rcx), %rdi
-    [-,%rdi]            v402 = uload16.i64 v1   ; bin: 48 0f b7 39
+    [-,%rdi]            v126 = uload16.i64 v1   ; bin: 48 0f b7 39
     ; asm: movzwq (%rsi), %rdx
-    [-,%rdx]            v403 = uload16.i64 v2   ; bin: 48 0f b7 16
+    [-,%rdx]            v127 = uload16.i64 v2   ; bin: 48 0f b7 16
     ; asm: movswq (%rcx), %rdi
-    [-,%rdi]            v404 = sload16.i64 v1   ; bin: 48 0f bf 39
+    [-,%rdi]            v128 = sload16.i64 v1   ; bin: 48 0f bf 39
     ; asm: movswq (%rsi), %rdx
-    [-,%rdx]            v405 = sload16.i64 v2   ; bin: 48 0f bf 16
+    [-,%rdx]            v129 = sload16.i64 v2   ; bin: 48 0f bf 16
     ; asm: movzbq (%rcx), %rdi
-    [-,%rdi]            v406 = uload8.i64 v1    ; bin: 48 0f b6 39
+    [-,%rdi]            v130 = uload8.i64 v1    ; bin: 48 0f b6 39
     ; asm: movzbq (%rsi), %rdx
-    [-,%rdx]            v407 = uload8.i64 v2    ; bin: 48 0f b6 16
+    [-,%rdx]            v131 = uload8.i64 v2    ; bin: 48 0f b6 16
     ; asm: movsbq (%rcx), %rdi
-    [-,%rdi]            v408 = sload8.i64 v1    ; bin: 48 0f be 39
+    [-,%rdi]            v132 = sload8.i64 v1    ; bin: 48 0f be 39
     ; asm: movsbq (%rsi), %rdx
-    [-,%rdx]            v409 = sload8.i64 v2    ; bin: 48 0f be 16
+    [-,%rdx]            v133 = sload8.i64 v2    ; bin: 48 0f be 16
 
     ; Register-indirect with 8-bit signed displacement.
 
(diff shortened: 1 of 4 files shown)


lib/cretonne/meta/base/instructions.py
```diff
@@ -271,7 +271,7 @@
         'istore16', r"""
         Store the low 16 bits of ``x`` to memory at ``p + Offset``.
 
-        This is equivalent to ``ireduce.i16`` followed by ``store.i8``.
+        This is equivalent to ``ireduce.i16`` followed by ``store.i16``.
```
stoklund: Thanks!

filetests/isa/intel/binary64.cton
```diff
@@ -145,6 +145,149 @@ ebb0:
...
+    ; asm: movq (%rsi), %rdx
+    [-,%rdx]            v461 = load.i64 v2      ; bin: 48 8b 16
+    ; asm: movl (%rcx), %edi
+    [-,%rdi]            v400 = load.i32 v1      ; bin: 40 8b 39
+    ; asm: movl (%rsi), %edx
+    [-,%rdx]            v401 = load.i32 v2      ; bin: 40 8b 16
+    ; asm: movzwl (%rcx), %edi
+    [-,%rdi]            v402 = uload16.i32 v1   ; bin: 40 0f b7 39
+    ; asm: movzwl (%rsi), %edx
+    [-,%rdx]            v403 = uload16.i32 v2   ; bin: 40 0f b7 16
```
stoklund: I think the loads producing a 32-bit value should be in the `%I32` function below.

Here, you want to test the loads producing 64-bit values: `uload32.i64` and `uload16.i64`.

filetests/isa/intel/binary64.cton
```diff
@@ -145,6 +145,149 @@ ebb0:
...
+    istore16 v2, v1                             ; bin: 66 40 89 31
+    ; asm: movb %cl, (%rsi)
+    istore8 v1, v2                              ; bin: 40 88 0e
+    ; asm: movb %sil, (%rcx)
+    istore8 v2, v1                              ; bin: 40 88 31
+
+    ; asm: movq (%rcx), %rdi
+    [-,%rdi]            v460 = load.i64 v1      ; bin: 48 8b 39
+    ; asm: movq (%rsi), %rdx
+    [-,%rdx]            v461 = load.i64 v2      ; bin: 48 8b 16
```
stoklund: I think it's best to keep the value numbers monotonically increasing, if possible.

filetests/isa/intel/binary64.cton
```diff
@@ -145,6 +145,149 @@ ebb0:
...
+    ; asm: movswq -50(%rsi), %rdx
+    [-,%rdx]            v415 = sload16.i64 v2-50        ; bin: 48 0f bf 56 ce
+    ; asm: movzbq 50(%rcx), %rdi
+    [-,%rdi]            v416 = uload8.i64 v1+50         ; bin: 48 0f b6 79 32
+    ; asm: movzbq -50(%rsi), %rdx
+    [-,%rdx]            v417 = uload8.i64 v2-50         ; bin: 48 0f b6 56 ce
+    ; asm: movsbq 50(%rcx), %rdi
+    [-,%rdi]            v418 = sload8.i64 v1+50         ; bin: 48 0f be 79 32
+    ; asm: movsbq -50(%rsi), %rdx
+    [-,%rdx]            v419 = sload8.i64 v2-50         ; bin: 48 0f be 56 ce
```
stoklund: Exactly, this is the right set of instructions to test here. You should do the same for the no-displacement ones above.

filetests/isa/intel/binary64.cton
```diff
@@ -145,6 +145,149 @@ ebb0:
...
+    ; asm: movswq -50000(%rsi), %rdx
+    [-,%rdx]            v425 = sload16.i64 v2-50000        ; bin: 48 0f bf 96 ffff3cb0
+    ; asm: movzbq 50000(%rcx), %rdi
+    [-,%rdi]            v426 = uload8.i64 v1+50000         ; bin: 48 0f b6 b9 0000c350
+    ; asm: movzbq -50000(%rsi), %rdx
+    [-,%rdx]            v427 = uload8.i64 v2-50000         ; bin: 48 0f b6 96 ffff3cb0
+    ; asm: movsbq 50000(%rcx), %rdi
+    [-,%rdi]            v428 = sload8.i64 v1+50000         ; bin: 48 0f be b9 0000c350
+    ; asm: movsbq -50000(%rsi), %rdx
+    [-,%rdx]            v429 = sload8.i64 v2-50000         ; bin: 48 0f be 96 ffff3cb0
```
stoklund: This is great, but please add similar tests for `*.i32` instructions to the `%I32` function below.

lib/cretonne/meta/isa/intel/encodings.py
```diff
@@ -55,6 +55,23 @@ def enc_i32_i64(inst, recipe, *args, **kwargs):
     I64.enc(inst.i64, *recipe.rex(*args, w=1, **kwargs))
 
 
+def enc_i32_i64_ld_st(inst, w_bit, recipe, *args, **kwargs):
+    # type: (MaybeBoundInst, int, r.TailRecipe, *int, **int) -> None
+    """
+    Add encodings for `inst.i32` to I32.
+    Add encodings for `inst.i32` to I64 with and without REX.
+    Add encodings for `inst.i64` to I64 with a REX.W prefix.
```
stoklund: This is a good idea. Please add an explanation for the `w_bit` argument.

lib/cretonne/meta/isa/intel/encodings.py
```diff
@@ -142,38 +159,52 @@ def enc_flt(inst, recipe, *args, **kwargs):
...
-I32.enc(base.store.i32.any, *r.stDisp8(0x89))
-I32.enc(base.store.i32.any, *r.stDisp32(0x89))
+#
+enc_i32_i64_ld_st(base.store, 1, r.st, 0x89)
+enc_i32_i64_ld_st(base.store, 1, r.stDisp8, 0x89)
+enc_i32_i64_ld_st(base.store, 1, r.stDisp32, 0x89)
+
+I64.enc(base.istore32.i64.any, *r.st.rex(0x89))
+I64.enc(base.istore32.i64.any, *r.stDisp8.rex(0x89))
+I64.enc(base.istore32.i64.any, *r.stDisp32.rex(0x89))
```
stoklund: This is a nice simplification. Very cool!

Do you think it makes sense to handle the three types of displacement that way too?

lib/cretonne/meta/isa/intel/encodings.py
```diff
@@ -142,38 +159,52 @@ def enc_flt(inst, recipe, *args, **kwargs):
...
+
+enc_i32_i64_ld_st(base.istore16, 0, r.st, 0x66, 0x89)
+enc_i32_i64_ld_st(base.istore16, 0, r.stDisp8, 0x66, 0x89)
+enc_i32_i64_ld_st(base.istore16, 0, r.stDisp32, 0x66, 0x89)
+
+enc_i32_i64_ld_st(base.istore8, 0, r.st_abcd, 0x88)
+enc_i32_i64_ld_st(base.istore8, 0, r.stDisp8_abcd, 0x88)
+enc_i32_i64_ld_st(base.istore8, 0, r.stDisp32_abcd, 0x88)
```
stoklund: Aha, so this part is extra tricky because you're using 8-bit registers.

In 32-bit mode, and in 64-bit instructions *without* a REX prefix, the following 8-bit registers can be addressed: AL, CL, DL, BL, AH, CH, DH, BH. The last 4 are the "high" byte registers, corresponding to bits 8-15 of the ABCD GPRs.

In 64-bit mode *with* a REX prefix, the accessible 8-bit registers are: AL, CL, DL, BL, SPL, BPL, SIL, DIL, R8B-R15B. Note the difference in registers 4-7 — they are the low bytes of GPRs instead of the high bytes of ABCD.

In Cretonne, we don't model the high-byte registers. They are never used. This is the reason the `I32` byte store instructions have an `ABCD` register constraint instead of `GPR` like the other instructions. When storing a byte, the value to store must be in one of the ABCD registers because without a REX prefix, we can't store the low byte of `%esi` for example.

However, *with* a REX prefix, a byte store can access the low byte of all 16 registers, so we just want a normal `GPR` register constraint which is much easier for the register allocator to satisfy.

The end result is:
- When accessing 8-bit registers *without* a REX prefix, use the `ABCD` register constraint.
- When accessing 8-bit registers *with* a REX prefix, use the `GPR` register constraint.

filetests/isa/intel/binary64.cton
```diff
@@ -145,6 +145,149 @@ ebb0:
...
+    ; Load/Store instructions.
+
+    ; Register indirect addressing with no displacement.
+
+    ; asm: movq %rcx, (%rsi)
+    store v1, v2                                ; bin: 48 89 0e
+    ; asm: movq %rsi, (%rcx)
+    store v2, v1                                ; bin: 48 89 31
+    ; asm: movl %ecx, (%rsi)
+    istore32 v1, v2                             ; bin: 40 89 0e
```
sunfishcode: `movl %ecx, (%rsi)` doesn't require a REX prefix.

sunfishcode:
In the binary64.cton, several instructions appear to have REX prefixes that don't require them, like this:

```
; asm: movl %ecx, (%rsi)
istore32 v1, v2                             ; bin: 40 89 0e
```

Is this a known limitation?

stoklund:
Removing redundant REX prefixes is treated as an optimization in Cretonne. Such an instruction will have two valid encodings: With and without a REX prefix. The encoding process goes like this:

1. Before register allocation: Pick the first valid encoding for each instruction. The encodings should be arranged such that this constraints the register allocator as little as possible.
2. After register allocation: Pick the smallest valid encoding for each instruction that has register constraints compatible with the register allocator's choices.

This approach means that redundant REX prefixes are removed after register allocation. On ARM, Thumb compression works the same way.

The test binemit mode could test either choice. Currently it picks the first valid encoding, which is the encoding that would be used before register allocation. You could argue that it shold pick the compressed encoding instead since it knows the register assignments.

stoklund:
Ideally, I would like to be able to test both:

```
; asm: movl %ecx, (%rsi)
istore32 v1, v2                             ; bin[early]: 40 89 0e
                                            ; bin[late]:  89 0e
```

lib/cretonne/meta/base/instructions.py
denismerigoux: You're welcome.

filetests/isa/intel/binary64.cton
denismerigoux: Indeed, I'll do a final value renumbering pass when all modifications are done.

stoklund: Thanks. It's fine to leave gaps, BTW.

filetests/isa/intel/binary64.cton
denismerigoux: Oups, my mistake.

filetests/isa/intel/binary64.cton
denismerigoux: Fixed.

filetests/isa/intel/binary64.cton
denismerigoux: Done.

lib/cretonne/meta/isa/intel/encodings.py
denismerigoux: Done.

lib/cretonne/meta/isa/intel/encodings.py
denismerigoux: Ok, got it. I won't use `enc_i32_i64_ld_st` for `istore8` but rather declare the `I32.enc` and `I64.enc` directly with the correct recipes.

lib/cretonne/meta/isa/intel/encodings.py
denismerigoux: I didn't understand you question. Is there another way of handling the three types of displacement?

stoklund: All I meant is that the code is very repetitive, so maybe a function call could reduce the redundancy. It's not necessary.

denismerigoux:
Tell me if it's good before merging and I will renumber the values in `binary64.cton`.

stoklund:
Looks good!

denismerigoux:
For 64bits instructions for which the REX prefix is optional (i.e. the REX.W bit is set to 0), should I generate only the encoding with REX prefix, only the encoding without or both ?

stoklund:
You should generate both, the first one without REX, and the second one with REX.

You will only be able to test the first encoding, per the comments above and #128.

denismerigoux:
Ok, got it.

PR #130, opened 2017-08-02
Avoid floating-point types in Ieee32::new and Ieee64::new.

sunfishcode:
This eliminates the need for unsafe code in code that uses Cretonne, a few
instances of unsafe code in Cretonne itself, and eliminates the only instance
of floating point in Cretonne.

As proposed here, this is an API change, requiring Cretonne users to update. I've prepared a patch [here](https://github.com/sunfishcode/wasm2cretonne/commit/65585485d8c9073dabda65e69d76f709394f67e1) for wasm2cretonne.


stoklund:
See https://doc.rust-lang.org/1.0.0/style/style/naming/README.html for constructor names: `new`, `with_details`, or `from_some_type`.

PR #132, opened 2017-08-02
Add a calling convention to all function signatures.

stoklund:
A CallConv enum on every function signature makes it possible to
generate calls to functions with different calling conventions within
the same ISA / within a single function.

The calling conventions also serve as a way of customizing Cretonne's
behavior when embedded inside a VM. As an example, the SpiderWASM
calling convention is used to compile WebAssembly functions that run
inside the SpiderMonkey virtual machine.

All function signatures must have a calling convention, so this changes
the textual IL syntax. The syntax for declaring a function signature
replaces the "signature" keyword with the calling convention.

Before:

    sig1 = signature(i32, f64) -> f64

After

    sig1 = native(i32, f64) -> f64
    sig2 = spiderwasm(i32)

When printing functions, the signature goes between the function name
and the arguments:

    function %r1 spiderwasm() -> i32, f32 {
    ebb1:
        ...
    }

In the parser, this calling convention is optional and defaults to
"native". This is mostly to avoid updating all the existing test cases
under filetests/. When printing a function, the calling convention is
always included, including for "native" functions.

### DIFF
--- a/docs/example.cton
+++ b/docs/example.cton
@@ -1,6 +1,6 @@
 test verifier
 
-function %average(i32, i32) -> f32 {
+function %average(i32, i32) -> f32 native {
     ss1 = local 8            ; Stack slot for ``sum``.
 
 ebb1(v1: i32, v2: i32):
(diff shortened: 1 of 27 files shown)


sunfishcode:
My preference is to leave the calling convention after the return value, as I instinctively parse
"foo()" as a function named foo. However, things can obviously work either way.

PR #133, opened 2017-08-03
Added last instruction information to loop analysis data

denismerigoux:
Loop analysis now splits `Ebb`s they belong to two or more loops and that the second loop does not extend to the end of the `Ebb`.

`ControlFlowGraph` and `DominatorTree` are recomputed incrementally on the fly. A new numbering system is implemented to allow that in `DominatorTree`.

Adequate testing and increased verifying for dominator tree integrity.

### DIFF
--- a/lib/cretonne/src/dominator_tree.rs
+++ b/lib/cretonne/src/dominator_tree.rs
@@ -317,8 +317,7 @@ impl DominatorTree {
             .filter(|&(pred, _)| self.nodes[pred].rpo_number > 1);
 
         // The RPO must visit at least one predecessor before this node.
-        let mut idom = reachable_preds
-            .next()
+        let mut idom = reachable_preds.next()
             .expect("EBB node must have one reachable predecessor");
 
         for pred in reachable_preds {
@@ -344,11 +343,10 @@ impl DominatorTree {
         }
         // We use the RPO comparison on the postorder list so we invert the operands of the
         // comparison
-        let old_ebb_postorder_index =
-            self.postorder
-                .as_slice()
-                .binary_search_by(|probe| self.rpo_cmp_ebb(old_ebb, *probe))
-                .expect("the old ebb is not declared to the dominator tree");
+        let old_ebb_postorder_index = self.postorder
+            .as_slice()
+            .binary_search_by(|probe| self.rpo_cmp_ebb(old_ebb, *probe))
+            .expect("the old ebb is not declared to the dominator tree");
         let new_ebb_rpo = self.insert_after_rpo(old_ebb, old_ebb_postorder_index, new_ebb);
         self.nodes[new_ebb] = DomNode {
             rpo_number: new_ebb_rpo,
--- a/lib/cretonne/src/dominator_tree.rs
+++ b/lib/cretonne/src/dominator_tree.rs
@@ -317,7 +317,8 @@ impl DominatorTree {
             .filter(|&(pred, _)| self.nodes[pred].rpo_number > 1);
 
         // The RPO must visit at least one predecessor before this node.
-        let mut idom = reachable_preds.next()
+        let mut idom = reachable_preds
+            .next()
             .expect("EBB node must have one reachable predecessor");
 
         for pred in reachable_preds {
@@ -343,10 +344,11 @@ impl DominatorTree {
         }
         // We use the RPO comparison on the postorder list so we invert the operands of the
         // comparison
-        let old_ebb_postorder_index = self.postorder
-            .as_slice()
-            .binary_search_by(|probe| self.rpo_cmp_ebb(old_ebb, *probe))
-            .expect("the old ebb is not declared to the dominator tree");
+        let old_ebb_postorder_index =
+            self.postorder
+                .as_slice()
+                .binary_search_by(|probe| self.rpo_cmp_ebb(old_ebb, *probe))
+                .expect("the old ebb is not declared to the dominator tree");
         let new_ebb_rpo = self.insert_after_rpo(old_ebb, old_ebb_postorder_index, new_ebb);
         self.nodes[new_ebb] = DomNode {
             rpo_number: new_ebb_rpo,
--- a/lib/cretonne/src/dominator_tree.rs
+++ b/lib/cretonne/src/dominator_tree.rs
@@ -12,7 +12,7 @@ use std::cmp::Ordering;
 const STRIDE: u32 = 4;
 
 // Dominator tree node. We keep one of these per EBB.
-#[derive(Clone, Default)]
+#[derive(Clone, Default, Debug)]
 struct DomNode {
     // Number of this node in a reverse post-order traversal of the CFG, starting from 1.
     // This number is monotonic in the reverse postorder but not contiguous, since we leave
@@ -29,6 +29,7 @@ struct DomNode {
 }
 
 /// The dominator tree for a single function.
+#[derive(Debug)]
 pub struct DominatorTree {
     nodes: EntityMap<Ebb, DomNode>,
 
--- a/lib/cretonne/src/loop_analysis.rs
+++ b/lib/cretonne/src/loop_analysis.rs
@@ -5,7 +5,8 @@ use dominator_tree::DominatorTree;
 use entity::{PrimaryMap, Keys};
 use entity::EntityMap;
 use flowgraph::ControlFlowGraph;
-use ir::{Function, Ebb, Layout};
+use ir::{Function, Ebb, Layout, Cursor, InstBuilder};
+use ir::entities::Inst;
 use packed_option::PackedOption;
 
 /// A opaque reference to a code loop.
@@ -18,8 +19,8 @@ entity_impl!(Loop, "loop");
 /// Loops are referenced by the Loop object, and for each loop you can access its header EBB,
 /// its eventual parent in the loop tree and all the EBB belonging to the loop.
 pub struct LoopAnalysis {
-    loops: PrimaryMap<Loop, LoopData>,
-    ebb_loop_map: EntityMap<Ebb, PackedOption<Loop>>,
+    loops: EntityMap<Loop, LoopData>,
+    ebb_loop_map: EntityMap<Ebb, EbbLoopData>,
 }
 
 struct LoopData {
@@ -29,11 +30,30 @@ struct LoopData {
 
 impl LoopData {
     /// Creates a `LoopData` object with the loop header and its eventual parent in the loop tree.
-    pub fn new(header: Ebb, parent: Option<Loop>) -> LoopData {
+    fn new(header: Ebb, parent: Option<Loop>) -> LoopData {
         LoopData {
             header: header,
             parent: parent.into(),
         }
+    }
+}
+
+/// If an `Ebb` is part of a loop, then we record two things: the id of the loop it's part of
+/// and the last instruction in the `Ebb` pertaining to the loop. If the `Ebb` is part of multiple
+/// loops, then we make sure by splitting the `Ebb` that it is part of at most two loops, one being
+/// the direct child of the other, and the parent loop should contain all the `Ebb` up to its
+/// terminator instruction.
+#[derive(Clone,Debug)]
+enum EbbLoopData {
+    NoLoop(),
+    Loop { loop_id: Loop, last_inst: Inst },
+}
+
+impl PrimaryEntityData for EbbLoopData {}
+
+impl Default for EbbLoopData {
+    fn default() -> EbbLoopData {
+        EbbLoopData::NoLoop()
     }
 }
 
@@ -66,14 +86,32 @@ impl LoopAnalysis {
         self.loops[lp].parent.expand()
     }
 
-    /// Determine if an Ebb belongs to a loop by running a finger along the loop tree.
+    /// Determine if an `Ebb` belongs to a loop by running a finger along the loop tree.
     ///
     /// Returns `true` if `ebb` is in loop `lp`.
     pub fn is_in_loop(&self, ebb: Ebb, lp: Loop) -> bool {
-        let ebb_loop = self.ebb_loop_map[ebb];
-        match ebb_loop.expand() {
-            None => false,
-            Some(ebb_loop) => self.is_child_loop(ebb_loop, lp),
+        match self.ebb_loop_map[ebb] {
+            EbbLoopData::NoLoop() => false,
+            EbbLoopData::Loop { loop_id, .. } => self.is_child_loop(loop_id, lp),
+        }
+    }
+
+    /// Determine which region of an `Ebb` belongs to a particular loop.
+    ///
+    /// If `ebb` belongs to `lp`, it returns `None` if the whole `ebb` belongs to `lp` or
+    /// `Some(inst)` where `inst` is the last instruction of `ebb` to belong to `lp`.
+    pub fn last_loop_instruction(&self, ebb: Ebb, lp: Loop) -> Result<Option<Inst>, ()> {
+        match self.ebb_loop_map[ebb] {
+            EbbLoopData::NoLoop() => Err(()),
+            EbbLoopData::Loop { loop_id, last_inst } => {
+                if lp == loop_id {
+                    Ok(Some(last_inst))
+                } else if self.is_child_loop(loop_id, lp) {
+                    Ok(None)
+                } else {
+                    Err(())
+                }
+            }
         }
     }
 
@@ -95,12 +133,15 @@ impl LoopAnalysis {
 
 impl LoopAnalysis {
     /// Detects the loops in a function. Needs the control flow graph and the dominator tree.
-    pub fn compute(&mut self, func: &Function, cfg: &ControlFlowGraph, domtree: &DominatorTree) {
+    pub fn compute(&mut self,
+                   func: &mut Function,
+                   cfg: &mut ControlFlowGraph,
+                   domtree: &mut DominatorTree) {
         self.loops.clear();
         self.ebb_loop_map.clear();
         self.ebb_loop_map.resize(func.dfg.num_ebbs());
         self.find_loop_headers(cfg, domtree, &func.layout);
-        self.discover_loop_blocks(cfg, domtree, &func.layout)
+        self.discover_loop_blocks(cfg, domtree, func)
     }
 
     // Traverses the CFG in reverse postorder and create a loop object for every EBB having a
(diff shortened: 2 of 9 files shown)


lib/cretonne/src/dominator_tree.rs
```diff
@@ -7,10 +7,16 @@ use packed_option::PackedOption;
...
+// RPO number are not first assigned in a contiguous way but as multiples of STRIDE, to leave
+// room for modifications of the dominator tree.
+const STRIDE: u32 = 4;
+
 // Dominator tree node. We keep one of these per EBB.
 #[derive(Clone, Default)]
 struct DomNode {
     // Number of this node in a reverse post-order traversal of the CFG, starting from 1.
+    // This number is monotonic in the reverse postorder but not contiguous, since we leave
+    // holes for later loczalized modifications of the dominator tree.
```
bjorn3: *localized

lib/cretonne/src/dominator_tree.rs
```diff
@@ -323,6 +329,72 @@ impl DominatorTree {
...
+                // We have to renumber
+                let return_value = ebb_rpo_number + 1;
+                let mut current_ebb = prev_postorder_ebb;
+                let mut current_ebb_rpo_number = prev_postorder_ebb_rpo_number;
+                let mut current_ebb_postorder_index = ebb_postorder_index - 1;
+                loop {
+                    // We renumber the current Ebb
+                    self.nodes[current_ebb].rpo_number = current_ebb_rpo_number + 1;
+                    if current_ebb_postorder_index == 0 {
+                        //We have finished here
```
sunfishcode: nit: space after `//`.

lib/cretonne/src/loop_analysis.rs
```diff
@@ -129,77 +169,162 @@ impl LoopAnalysis {
...
                                     continue_dfs = None
+                                } else {
+                                    // Else we proceed to mark the inner loop as child of lp
+                                    // And continue the DFS with the inner loop's header
+                                    self.loops[inner_loop].parent = lp.into();
+                                    continue_dfs = Some(inner_loop_header);
+                                    // Here we perform a modification that's not related to
+                                    // the loop discovery algorithm but rather to the way we
+                                    // store the loop information. Indeed, for each Ebb we record
+                                    // the loop its part of and the last inst in this loop
```
sunfishcode: nit: "it's" rather than "its". And "." at the end of the sentence.

lib/cretonne/src/loop_analysis.rs
```diff
@@ -129,77 +169,162 @@ impl LoopAnalysis {
...
+                                    // But here the Ebb is part of at least two loops, the child
+                                    // loop and the parent loop . So two cases arise:
+                                    // - either the entire ebb is part of the parent loop and this
+                                    // is fine;
+                                    // - either only part of the ebb is part of the parent loop and
+                                    // in that case we can't store the information of where does
+                                    // the parent loop stops.
```
sunfishcode: The word "either" should only appear once. One way to fix this would be to replace the second "either" with "or".

Also, remove the word "does" here.

lib/cretonne/src/loop_analysis.rs
```diff
@@ -129,77 +169,162 @@ impl LoopAnalysis {
...
-                        let mut node_loop = node_loop;
-                        // The node is part of a loop, which can be lp or an inner loop
-                        let mut node_loop_parent_option = self.loops[node_loop].parent;
-                        while let Some(node_loop_parent) = node_loop_parent_option.expand() {
-                            if node_loop_parent == lp {
+                    EbbLoopData::Loop {
+                        loop_id,
+                        last_inst: _,
+                    } => {
+                        // So here the ebb we're visiting is already tagged as being part of a loop
```
sunfishcode: More comment nits: period after this sentence. And if you reflow this whole paragraph it would be tidier.

lib/cretonne/src/dominator_tree.rs
```diff
@@ -7,10 +7,16 @@ use packed_option::PackedOption;
 
 use std::cmp::Ordering;
 
+// RPO number are not first assigned in a contiguous way but as multiples of STRIDE, to leave
```
sunfishcode: nit: "numbers" rather than "number"

lib/cretonne/src/dominator_tree.rs
```diff
@@ -323,6 +329,72 @@ impl DominatorTree {
...
+            ebb_rpo_number + STRIDE
+        } else {
+            let prev_postorder_ebb = self.postorder[ebb_postorder_index - 1];
+            let prev_postorder_ebb_rpo_number = self.nodes[prev_postorder_ebb].rpo_number;
+            if prev_postorder_ebb_rpo_number > ebb_rpo_number + 1 {
+                // There is a gap, we can use it
+                ebb_rpo_number + 1
+            } else {
+                // We have to renumber
+                let return_value = ebb_rpo_number + 1;
```
sunfishcode: Both arms of this 'if' appear to be returning `ebb_rpo_number + 1` which could be simplified.

lib/cretonne/src/loop_analysis.rs
```diff
@@ -129,77 +169,162 @@ impl LoopAnalysis {
...
+                                      func: &mut Function,
+                                      domtree: &mut DominatorTree,
+                                      cfg: &mut ControlFlowGraph) {
+        let new_ebb = func.dfg.make_ebb();
+        let real_split_inst = {
+            let mut cur = Cursor::new(&mut func.layout);
+            cur.goto_inst(split_inst);
+            cur.next_inst()
+                .expect("you cannot split at the last instruction")
+        };
```
stoklund: There already is a `Layout::next_ebb()` method. We should have a `next_inst` too.

lib/cretonne/src/loop_analysis.rs
```diff
@@ -30,14 +31,33 @@ impl PrimaryEntityData for LoopData {}
...
+/// loops, then we make sure by splitting the `Ebb` that it is part of at most two loops, one being
+/// the direct child of the other, and the parent loop should contain all the `Ebb` up to its
+/// terminator instruction.
+#[derive(Clone,Debug)]
+enum EbbLoopData {
+    NoLoop(),
+    Loop { loop_id: Loop, last_inst: Inst },
+}
+
+impl PrimaryEntityData for EbbLoopData {}
```
stoklund: This should not be necessary. You're making a secondary map.

lib/cretonne/src/loop_analysis.rs
```diff
@@ -67,14 +87,32 @@ impl LoopAnalysis {
...
+            EbbLoopData::NoLoop() => false,
+            EbbLoopData::Loop { loop_id, .. } => self.is_child_loop(loop_id, lp),
+        }
+    }
+
+    /// Determine which region of an `Ebb` belongs to a particular loop.
+    ///
+    /// If `ebb` belongs to `lp`, it returns `None` if the whole `ebb` belongs to `lp` or
+    /// `Some(inst)` where `inst` is the last instruction of `ebb` to belong to `lp`.
+    pub fn last_loop_instruction(&self, ebb: Ebb, lp: Loop) -> Result<Option<Inst>, ()> {
```
stoklund: If all of an EBB's instructions belong to the loop, it seems that this function will return None or the EBB terminator somewhat randomly.

Wouldn't it be more consistent if this never returned the terminator? Only return some instruction when the EBB is actually split between two loops?

lib/cretonne/src/loop_analysis.rs
```diff
@@ -30,14 +31,33 @@ impl PrimaryEntityData for LoopData {}
...
 
+/// If an `Ebb` is part of a loop, then we record two things: the id of the loop it's part of
+/// and the last instruction in the `Ebb` pertaining to the loop. If the `Ebb` is part of multiple
+/// loops, then we make sure by splitting the `Ebb` that it is part of at most two loops, one being
+/// the direct child of the other, and the parent loop should contain all the `Ebb` up to its
+/// terminator instruction.
+#[derive(Clone,Debug)]
+enum EbbLoopData {
+    NoLoop(),
+    Loop { loop_id: Loop, last_inst: Inst },
```
stoklund: I think this is better represented as two packed options. Both because the enum makes the data structure 50% larger, and because you naturally have three possibilities:

- None, none. No loop
- Loop, None. Whole EBB belongs to a single loop.
- Loop, last. EBB belongs to two loops.

lib/cretonne/src/dominator_tree.rs

lib/cretonne/src/loop_analysis.rs
```diff
@@ -96,12 +134,15 @@ impl LoopAnalysis {
 
 impl LoopAnalysis {
     /// Detects the loops in a function. Needs the control flow graph and the dominator tree.
```
denismerigoux: Done.

lib/cretonne/src/loop_analysis.rs
denismerigoux: Fixed.

lib/cretonne/src/loop_analysis.rs
denismerigoux: Fixed.

lib/cretonne/src/verifier/mod.rs
```diff
@@ -409,6 +411,33 @@ impl<'a> Verifier<'a> {
...
             }
         }
+        // We also verify if the postorder defined by `DominatorTree` is sane
+        for (index, (&true_ebb, &test_ebb)) in
+            self.domtree
+                .cfg_postorder()
+                .iter()
+                .zip(domtree.cfg_postorder().iter())
+                .enumerate() {
+            if true_ebb != test_ebb {
```
stoklund: Unfortunately, this won't detect if one PO is a prefix if the other. The `zip` iterator stops when either input ends.

Note that slices implement `Eq`, so you can just compare them with `==`.

lib/cretonne/src/loop_analysis.rs
denismerigoux: Fixed.

lib/cretonne/src/loop_analysis.rs
denismerigoux: Do I add the `Layout::next_inst()` in this PR ?

lib/cretonne/src/verifier/mod.rs
```diff
@@ -409,6 +411,33 @@ impl<'a> Verifier<'a> {
...
+                            "invalid domtree, postorder ebb number {} should be {}, got {}",
+                            index,
+                            true_ebb,
+                            test_ebb);
+            }
+        }
+        // We verify rpo_cmp on pairs of adjacent ebbs in the postorder
+        for (&prev_ebb, &next_ebb) in self.domtree.cfg_postorder().iter().adjacent_pairs() {
+            match domtree.rpo_cmp(prev_ebb, next_ebb, &self.func.layout) {
+                Ordering::Greater => (),
```
stoklund: I don't think a match is necessary here: `if rpo_cmp(...) != Ordering::Greater`.

lib/cretonne/src/dominator_tree.rs

lib/cretonne/src/dominator_tree.rs

lib/cretonne/src/dominator_tree.rs

lib/cretonne/src/dominator_tree.rs
```diff
@@ -323,6 +329,72 @@ impl DominatorTree {
     }
 }
 
+impl DominatorTree {
+    /// When splitting an `Ebb` using `Layout::split_ebb`, you can use this method to update
+    /// the dominator tree locally rather than recomputing it.
+    ///
+    /// `old_ebb` is the `Ebb` before splitting, and `new_ebb` is the `Ebb` which now contains
+    /// the second half of `old_ebb`.
+    pub fn recompute_split_ebb(&mut self, old_ebb: Ebb, new_ebb: Ebb, split_jump_inst: Inst) {
```
stoklund: Since this API is a bit tricky, I think it would be a good idea to add a couple `debug_assert!`s here: `split_jump_inst` is the terminator of `old_ebb`, and it is a jump to `new_ebb`.

lib/cretonne/src/dominator_tree.rs
```diff
@@ -323,6 +329,72 @@ impl DominatorTree {
...
+        let old_ebb_postorder_index =
+            self.postorder
+            .as_slice()
+            // We use the RPO comparison on the postorder list so we invert the operands of the
+            // comparison
+            .binary_search_by(|probe| self.rpo_cmp_ebb(old_ebb,*probe,))
+            .expect("the old ebb is not declared to the dominator tree");
```
stoklund: What if it is unreachable?

lib/cretonne/src/loop_analysis.rs
denismerigoux: OK for two `PackedOption`s. Also you're right about the lack of `max` computation, there's currently something wrong.

lib/cretonne/src/dominator_tree.rs
```diff
@@ -323,6 +329,72 @@ impl DominatorTree {
...
+        let new_ebb_rpo = self.insert_after_rpo(old_ebb, old_ebb_postorder_index);
+        *self.nodes.ensure(new_ebb) = DomNode {
+            rpo_number: new_ebb_rpo,
+            idom: Some(split_jump_inst).into(),
+        };
+        // TODO: insert in constant time?
+        self.postorder.insert(0, new_ebb);
```
stoklund: Technically we should, but I don't think we need to worry about this for now.

lib/cretonne/src/loop_analysis.rs
denismerigoux: Oh that is what this trait is for. Understood.

lib/cretonne/src/dominator_tree.rs
```diff
@@ -323,6 +329,72 @@ impl DominatorTree {
...
+    }
+
+    // We want to insert a new ebb just after ref_ebb in the rpo order. This function checks
+    // if there is a gap in rpo numbers; if yes it returns the number in the gap and if
+    // not it renumbers.
+    fn insert_after_rpo(&mut self, ebb: Ebb, ebb_postorder_index: usize) -> u32 {
+        let ebb_rpo_number = self.nodes[ebb].rpo_number;
+        if ebb_postorder_index == 0 {
+            ebb_rpo_number + STRIDE
```
stoklund: Use an early return here to reduce nesting of the remaining function.

lib/cretonne/src/dominator_tree.rs
```diff
@@ -323,6 +329,72 @@ impl DominatorTree {
...
+        } else {
+            let prev_postorder_ebb = self.postorder[ebb_postorder_index - 1];
+            let prev_postorder_ebb_rpo_number = self.nodes[prev_postorder_ebb].rpo_number;
+            if prev_postorder_ebb_rpo_number > ebb_rpo_number + 1 {
+                // There is a gap, we can use it
+                ebb_rpo_number + 1
```
stoklund: Again, just return instead of nesting.

lib/cretonne/src/loop_analysis.rs
denismerigoux: You're right, I'll do this.

lib/cretonne/src/loop_analysis.rs
stoklund: Yeah, it's not a great solution. I'm thinking about splitting `EntityMap` into two separate types instead.

lib/cretonne/src/loop_analysis.rs
stoklund: Sure, that's fine.

stoklund:
Would it make sense to break the dominator tree change out into it's own PR? It seems pretty independent from the loop analysis change, and it is closer to be able to land.

lib/cretonne/src/verifier/mod.rs
denismerigoux: I can use `==` but the drawback is that is doesn't return any interesting location when the two slices are different.

lib/cretonne/src/verifier/mod.rs
denismerigoux: Fixed.

denismerigoux:
I can do that.

lib/cretonne/src/verifier/mod.rs
stoklund: That's true. It's up to you, you're probably the one who gets to see these verifier errors.

denismerigoux: I can have the best of both worlds by doing the `zip` and testing for equal length beforehand.

lib/cretonne/src/dominator_tree.rs
denismerigoux: Good idea.

lib/cretonne/src/dominator_tree.rs
denismerigoux: If `ebb` is unreachable then it stays unreachable after splitting, and `new_ebb` unreachable too. I'll deal with this case.

lib/cretonne/src/dominator_tree.rs
denismerigoux: Ok

lib/cretonne/src/dominator_tree.rs

lib/cretonne/src/dominator_tree.rs

lib/cretonne/src/dominator_tree.rs
denismerigoux: But actually in order to do the `debug_assert()` I would need to add the `dfg` and `layout` as arguments to the function. Is it worth it since they're not going to be used in release mode?

stoklund: No, not worth it. The verifier should catch any errors too.

lib/cretonne/src/loop_analysis.rs
```diff
@@ -96,12 +136,21 @@ impl LoopAnalysis {
...
-    pub fn compute(&mut self, func: &Function, cfg: &ControlFlowGraph, domtree: &DominatorTree) {
+    ///
+    /// Because of loop information storage efficiency, the loop analysis only allows one `Ebb` to
+    /// be part of at most two loops: an inner loop which can end in the middle of the `Ebb` and
+    /// an outer loop that has to contain the full `Ebb`. All `Ebb`s who don't match this criteria
+    /// are split until they do so. That is why `compute` mutates the function, control flow
+    /// graph and moniator tree.
```
sunfishcode: typo: moniator

denismerigoux:
Dominator tree-related changes are now in https://github.com/stoklund/cretonne/pull/135.

denismerigoux:
By the way how do you compare the position of two instructions inside an `Ebb`? I haven't seen any method in `Layout` for that.

stoklund:
Look for `ProgramOrder` which is implemented by `Layout`.

lib/cretonne/src/loop_analysis.rs
denismerigoux: Fixed.

sunfishcode:
Closing; we'll now track this at #393.

PR #135, opened 2017-08-03
Added partial recompute of dominator tree in case of Ebb splitting 

denismerigoux:
Implemented via striding RPO numbers and renumbering in worst case.

### DIFF
--- a/lib/cretonne/src/dominator_tree.rs
+++ b/lib/cretonne/src/dominator_tree.rs
@@ -363,37 +363,24 @@ impl DominatorTree {
     // not it renumbers.
     fn insert_after_rpo(&mut self, ebb: Ebb, ebb_postorder_index: usize) -> u32 {
         let ebb_rpo_number = self.nodes[ebb].rpo_number;
-        if ebb_postorder_index == 0 {
-            return ebb_rpo_number + STRIDE;
-        }
-        let prev_postorder_ebb = self.postorder[ebb_postorder_index - 1];
-        let prev_postorder_ebb_rpo_number = self.nodes[prev_postorder_ebb].rpo_number;
-        if prev_postorder_ebb_rpo_number <= ebb_rpo_number + 1 {
-            // We have to renumber
-            let mut current_ebb = prev_postorder_ebb;
-            let mut current_ebb_rpo_number = prev_postorder_ebb_rpo_number;
-            let mut current_ebb_postorder_index = ebb_postorder_index - 1;
-            loop {
-                // We renumber the current Ebb
-                self.nodes[current_ebb].rpo_number = current_ebb_rpo_number + 1;
-                if current_ebb_postorder_index == 0 {
-                    // We have finished here
-                    break;
-                }
-                let next_ebb = self.postorder[current_ebb_postorder_index - 1];
-                let next_ebb_rpo_number = self.nodes[next_ebb].rpo_number;
-                if next_ebb_rpo_number > current_ebb_rpo_number + 1 {
-                    // There is a gap that we take
-                    break;
-                }
-                // Else we continue looping
-                current_ebb = next_ebb;
-                current_ebb_rpo_number = next_ebb_rpo_number;
-                current_ebb_postorder_index -= 1;
+        let inserted_rpo_number = ebb_rpo_number + 1;
+        // If there is no gaps in RPo numbers to insert this new number, we iterate
+        // forward in RPO numbers and backwards in the postorder list of EBBs, renumbering the Ebbs
+        // until we find a gap
+        for (&current_ebb, current_rpo) in
+            self.postorder[0..ebb_postorder_index]
+                .iter()
+                .rev()
+                .zip(inserted_rpo_number..) {
+            if self.nodes[current_ebb].rpo_number <= current_rpo {
+                // There is no gap, we renumber
+                self.nodes[current_ebb].rpo_number = current_rpo;
+            } else {
+                // There is a gap, we stop the renumbering and exit
+                break;
             }
         }
-        // In all cases we return the same value
-        ebb_rpo_number + 1
+        inserted_rpo_number
     }
 }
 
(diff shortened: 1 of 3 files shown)


lib/cretonne/src/dominator_tree.rs
```diff
@@ -323,6 +329,77 @@ impl DominatorTree {
...
+    /// `old_ebb` is the `Ebb` before splitting, and `new_ebb` is the `Ebb` which now contains
+    /// the second half of `old_ebb`. `split_jump_inst` is the terminator jump instruction of
+    /// `old_ebb` that points to `new_ebb`.
+    pub fn recompute_split_ebb(&mut self, old_ebb: Ebb, new_ebb: Ebb, split_jump_inst: Inst) {
+        if !self.is_reachable(old_ebb) {
+            // old_ebb is unreachable, it stays so and new_ebb is unreachable too
+            *self.nodes.ensure(new_ebb) = DomNode {
+                rpo_number: 0,
+                idom: Some(split_jump_inst).into(),
+            };
```
stoklund: Early return + unindent.

I don't think we should store an idom for unreachable nodes. Just set it to `None`:

```rust
*self.nodes.ensure(new_ebb) = Default::default();
```

lib/cretonne/src/dominator_tree.rs
```diff
@@ -323,6 +329,77 @@ impl DominatorTree {
...
+        } else {
+            let old_ebb_postorder_index =
+                self.postorder
+            .as_slice()
+            // We use the RPO comparison on the postorder list so we invert the operands of the
+            // comparison
+            .binary_search_by(|probe| self.rpo_cmp_ebb(old_ebb,*probe,))
```
stoklund: What's happening with the commas and spaces here? Strange that rustfmt isn't complaining.

The indentation is also weird, but that could be a GitHub thing.

lib/cretonne/src/dominator_tree.rs

lib/cretonne/src/dominator_tree.rs
denismerigoux: Apparently `rustfmt` doesn't support comments in the middle of a sequence of method calls like that.

lib/cretonne/src/dominator_tree.rs
```diff
@@ -323,6 +329,74 @@ impl DominatorTree {
...
+    }
+
+    // We want to insert a new ebb just after ref_ebb in the rpo order. This function checks
+    // if there is a gap in rpo numbers; if yes it returns the number in the gap and if
+    // not it renumbers.
+    fn insert_after_rpo(&mut self, ebb: Ebb, ebb_postorder_index: usize) -> u32 {
+        let ebb_rpo_number = self.nodes[ebb].rpo_number;
+        if ebb_postorder_index == 0 {
+            return ebb_rpo_number + STRIDE;
+        }
```
stoklund: The rest of this function feels like it could be a lot simpler. Does this make sense?

```rust
for (&ebb, rpo) in self.postorder[0..ebb_postorder_index-1].iter().rev().zip(ebb_rpo_number+2..) {
    if self.nodes[ebb].rpo_number < rpo {
        self.nodes[ebb].rpo_number = rpo;
    } else {
        break;
    }
}
```

denismerigoux: It indeed works in a nice Rusty way. I will renamed some variables to make it more understandable tough.

lib/cretonne/src/dominator_tree.rs
```diff
@@ -323,6 +329,61 @@ impl DominatorTree {
...
+                .as_slice()
+                .binary_search_by(|probe| self.rpo_cmp_ebb(old_ebb, *probe))
+                .expect("the old ebb is not declared to the dominator tree");
+        let new_ebb_rpo = self.insert_after_rpo(old_ebb, old_ebb_postorder_index);
+        *self.nodes.ensure(new_ebb) = DomNode {
+            rpo_number: new_ebb_rpo,
+            idom: Some(split_jump_inst).into(),
+        };
+        // TODO: insert in constant time?
+        self.postorder.insert(0, new_ebb);
```
stoklund: This insert index does not look right.

lib/cretonne/src/dominator_tree.rs
```diff
@@ -323,6 +329,61 @@ impl DominatorTree {
...
+        let inserted_rpo_number = ebb_rpo_number + 1;
+        // If there is no gaps in RPo numbers to insert this new number, we iterate
+        // forward in RPO numbers and backwards in the postorder list of EBBs, renumbering the Ebbs
+        // until we find a gap
+        for (&current_ebb, current_rpo) in
+            self.postorder[0..ebb_postorder_index]
+                .iter()
+                .rev()
+                .zip(inserted_rpo_number..) {
+            if self.nodes[current_ebb].rpo_number <= current_rpo {
```
stoklund: Two off-by-one errors here:

1. `zip(inserted_rpo_number..)` should start from `inserted_rpo_number+1` to avoid a duplicate number.
2. `rpo_number <= current_rpo` should be `<`. It is safe to stop on equality.

lib/cretonne/src/dominator_tree.rs
```diff
@@ -323,6 +329,61 @@ impl DominatorTree {
...
+            idom: Some(split_jump_inst).into(),
+        };
+        // TODO: insert in constant time?
+        self.postorder.insert(0, new_ebb);
+    }
+
+    // We want to insert a new ebb just after ref_ebb in the rpo order. This function checks
+    // if there is a gap in rpo numbers; if yes it returns the number in the gap and if
+    // not it renumbers.
+    fn insert_after_rpo(&mut self, ebb: Ebb, ebb_postorder_index: usize) -> u32 {
```
stoklund: This function doesn't actually do any inserting, it only renumbers. Maybe move the `self.postorder.insert` call in here?

lib/cretonne/src/dominator_tree.rs

lib/cretonne/src/dominator_tree.rs
denismerigoux: Yep, good idea.

lib/cretonne/src/dominator_tree.rs
denismerigoux: You're right, I was confused trying to make some sense about the `ebb_rpo_number+2` earlier.

PR #137, opened 2017-08-05
[WIP] Reimplement conditional compilation of isa targets using cargo features

zummenix:
This is a continuation of work started in #126 

I've realised that this design doesn't allow conditional compilation of other project in the tree. For example, if user wants ability to compile `cton_reader` only for concrete target. To support it for all projects we need to define the same features (isa targets) that are defined in `cretonne` for `cton-util`, `cton_reader` and `cton_frontend`. Besides if we want to support native targets each project should have own build.rs file that checks enabled cargo features and sets native isa target if needed (as implemented in this PR in `lib/cretonne/build.rs`)

Maybe there is a better way and I'm missing something?

### DIFF
--- a/lib/cretonne/Cargo.toml
+++ b/lib/cretonne/Cargo.toml
@@ -13,6 +13,15 @@ build = "build.rs"
 [lib]
 name = "cretonne"
 
+[features]
+default = ["riscv", "intel", "arm32", "arm64"]
+
+# Names of the features should be in sync with isa targets defined in build.rs.
+riscv = []
+intel = []
+arm32 = []
+arm64 = []
+
 [dependencies]
 # It is a goal of the cretonne crate to have minimal external dependencies.
 # Please don't add any unless they are essential to the task of creating binary
--- a/lib/cretonne/build.rs
+++ b/lib/cretonne/build.rs
@@ -11,10 +11,6 @@
 // TARGET
 //     Target triple provided by Cargo.
 //
-// CRETONNE_TARGETS (Optional)
-//     A setting for conditional compilation of isa targets. Possible values can be "native" or
-//     known isa targets separated by ','.
-//
 // The build script expects to be run from the directory where this build.rs file lives. The
 // current directory is used to find the sources.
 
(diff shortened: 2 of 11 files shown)


stoklund:
Your work here looks good to me, but I don't have a lot of experience working with cargo features.

I don't quite understand the problem you're describing, could you elaborate a bit with an example?

zummenix:
Sorry for the late response.

Here is an example: suppose we want to conditionally compile `cton_reader` crate for example for testing. To be able to do that we need to add the same cargo features as in `cretonne` crate:
```toml
[features]
default = ["riscv", "intel", "arm32", "arm64"]
riscv = ["cretonne/riscv"]
intel = ["cretonne/intel"]
arm32 = ["cretonne/arm32"]
arm64 = ["cretonne/arm64"]
```
I haven't tried it yet, but according to http://doc.crates.io/manifest.html#the-features-section it should work.

And as I mentioned in the previous comment each crate should have its own `build.rs` with the same logic of selecting a cargo feature to compile for a native target. For this part I think we can create a small crate that will only be used in `build.rs` scripts.

In overall I guess it's not a problem.  My main concern is that there will be a lot of duplication in `Cargo.toml` for each crate.

I hope I explained it better this time.

stoklund:
Thanks for spelling this out for me, I think I understand what you mean now.

I really would prefer if we can avoid the duplication of cargo features, so any features for controlling which targets get built are only on the main Cretonne crate.

I am thinking that a project that uses Cretonne through the front-end crates is going to have some direct interaction with the main Cretonne crate too. This means it will declare Cretonne itself as a dependency and it can specify the features then. I don't know how Cargo behaves when multiple crates depend on a crate with features. I am guessing it computes the union of features requested. That should be ok, right?

One thing to note is that these features don't change the public interface of the Cretonne crate at all. They are enabling code in `isa` sub-modules that is not exposed publicly. In particular, instruction opcodes like `x86_udivmodx` will still exist even when the `intel` target is disabled.

zummenix:
Ok, if I understand you correctly only `cretonne` crate will have these cargo features.

A couple of questions related to testing:
1. How should we proceed with testing regarding cargo features? Now I'm changing the list of default features in `cretonne` and `./test-all.sh`. It's not good.
2. How should we handle isa dependent tests? For example this one in `reader`: https://github.com/stoklund/cretonne/blob/86215ca4125ce0801490a908f22d6011f39c5bb2/lib/reader/src/parser.rs#L2447
I'm thinking to wrap it in isa::lookup("riscv").

sunfishcode:
Hello, sorry for the delay in getting back to you here.

> How should we proceed with testing regarding cargo features? Now I'm changing the list of default features in cretonne and ./test-all.sh. It's not good.
> How should we handle isa dependent tests? For example this one in reader: https://github.com/stoklund/cretonne/blob/86215ca4125ce0801490a908f22d6011f39c5bb2/lib/reader/src/parser.rs#L2447
> I'm thinking to wrap it in isa::lookup("riscv").

For these kinds of tests, does it work to just add `#[cfg(feature = "riscv")]` etc. to the test functions that need it?

zummenix:
> Hello, sorry for the delay in getting back to you here.

No problem!
I need to refresh my memory on this. I will try on the weekend if I have time.

zummenix:
I'm working on filetest runner to skip tests that have unsupported isa target and I don't know a good way to handle this situation in `parse_test` since some methods of the parser require `TargetIsa`.
https://github.com/Cretonne/cretonne/blob/20de74f73bb935437cddbcda67d8bb89435265f8/lib/reader/src/parser.rs#L40-L61

sunfishcode:
I believe all the parts of the parser that use `TargetIsa` look like `unique_isa: Option<&TargetIsa>`. If a test has multiple `isa` directives, then `None` is used.

If a test has exactly one `isa` directive, and that isa is unsupported, the parser should stop and not parse the rest of the file. If the isa is supported, then we process the test as normal. Otherwise, `unique_isa` will be None, and we can just ignore the ISAs that are unsupported. Does that make sense?

zummenix:
I'm sorry, I stuck with other projects and don't have time to continue this. If there is anything left to do and someone willing to take away, I don't mind.

sunfishcode:
This is now implemented in #673!

PR #138, opened 2017-08-07
Optimized LICM pass

denismerigoux:
This new LICM pass now respects the limit of loops inside `Ebb`s, contrarily to the previous implementation which could move upwards instructions that were not in a loop.

Relies heavily on https://github.com/stoklund/cretonne/pull/133. 

Control flow graph, dominator trees and loop analysis information are updated incrementally during the pass, leaving them correct after it is over.

The LICM pass now works by doing a single pass over instructions in reverse postorder: for each instruction belonging to a loop, it determines where is the closest definition of one of its arguments, and depending on that hoists the instruction as far up the loop tree as it can. Subsequent instructions that use its result as argument will see the instruction at its new place, ensuring correctness of the LICM.


lib/cretonne/src/licm.rs
```diff
@@ -15,40 +20,137 @@ pub fn do_licm(func: &mut Function,
...
+// - inside a parent loop which is not `inst_loop`.
+fn closest_def_position(func: &Function,
+                        inst: Inst,
+                        inst_loop: Loop,
+                        loop_analysis: &LoopAnalysis)
+                        -> DefPosition {
+    func.dfg
+        .inst_args(inst)
+        .into_iter()
+        .fold(DefPosition::OutsideLoop(), |acc, arg| {
```
sunfishcode: I do like expressing this as a fold, in the abstract. However, fold isn't short-circuiting. It seems like we'd want to stop iterating once we find an InsideCurrentLoop operand. I'm assuming it's common to use values defined in the same block, so that terminating this loop early could save a significant amount of work.

lib/cretonne/src/licm.rs
```diff
@@ -15,40 +20,137 @@ pub fn do_licm(func: &mut Function,
...
+    let mut inst_pool = ListPool::new();
+    let mut insts_rpo = EntityList::new();
+    // Because Rust doesn't allow us to modify data structures that we are iterating upon, we
+    // first have to do a pass to list all the instruction susceptible to be hoisted.
+    for ebb in domtree.cfg_postorder().iter().rev() {
+        for inst in func.layout.ebb_insts(*ebb) {
+            match loop_analysis.base_loop_inst(inst, &func.layout) {
+                None => (),
+                Some(_) => {
+                    if func.dfg.has_results(inst) {
```
sunfishcode: Should there a check around here for "does inst have side effects", making it unsafe to hoist and therefore not interesting to visit?

If so, does that subsume what the `has_results()` call is testing for?

lib/cretonne/src/licm.rs
```diff
@@ -15,40 +20,137 @@ pub fn do_licm(func: &mut Function,
...
+        };
+        match has_pre_header(&func.layout,
+                             cfg,
+                             domtree,
+                             loop_analysis.loop_header(destination_loop)) {
+            None => {
+                let loop_header = loop_analysis.loop_header(destination_loop);
+                let (pre_header, jump_inst) = create_pre_header(loop_header, func, cfg, domtree);
+                loop_analysis.recompute_loop_preheader(pre_header, loop_header);
+                domtree.recompute_loop_preheader(loop_header, pre_header, jump_inst);
```
sunfishcode: Misc style suggestion: would it make sense to move these recompute_loop_preheader calls into `create_pre_header`?

lib/cretonne/src/licm.rs
```diff
@@ -15,40 +20,137 @@ pub fn do_licm(func: &mut Function,
...
+            }
+            // If there is a natural pre-header we insert new instructions just before the
+            // related jumping instruction (which is not necessarily at the end).
+            Some((_, last_jump_inst)) => {
+                let mut cur = Cursor::new(&mut func.layout);
+                cur.goto_inst(inst);
+                cur.remove_inst();
+                cur.goto_inst(last_jump_inst);
+                cur.insert_inst(inst);
+            }
```
sunfishcode: This code is pretty verbose for doing a conceptually simple task, moving an inst, and it appears twice. I suggest factoring this out into a helper function, either on layout or just a local function.

lib/cretonne/src/dominator_tree.rs
```diff
@@ -354,7 +355,41 @@ impl DominatorTree {
             rpo_number: new_ebb_rpo,
             idom: Some(split_jump_inst).into(),
         };
+    }
 
+    /// When inserting a loop preh-header,  you can use this method to update the dominator tree
```

lib/cretonne/src/licm.rs
denismerigoux: I hesitated to do that because then I would have to add `loop_analysis` and `domtree` as arguments only for that. But I guess it makes more sense.

lib/cretonne/src/licm.rs
denismerigoux: Moved to a local helper function.

lib/cretonne/src/licm.rs
```diff
@@ -15,40 +20,137 @@ pub fn do_licm(func: &mut Function,
...
+        };
+    }
+}
+
+// Given a instruction and the inner-most loop it is part of, determine from the positions of its
+// arguments the closest definition of one of the instruction's argument. This closest position
+// can be:
+// - outside any loop;
+// - inside the `inst_loop`, meaning that the instruction is not loop-invariant;
+// - inside a parent loop which is not `inst_loop`.
```
denismerigoux: Fixed.

lib/cretonne/src/licm.rs
denismerigoux: You're right. That's my OCaml habits resurfacing.

lib/cretonne/src/licm.rs
denismerigoux: I just found the `cretonne::ir::instructions::Opcode::other_side_effects` that should do the job.

sunfishcode: `other_side_effects` doesn't include loads or stores. See `trivially_unsafe_for_gvn` in simple_gvn.rs, for example. Although this prompts me to notice that that function has the inverse lack -- it lacks a call to `other_side_effects()`.

lib/cretonne/src/licm.rs
```diff
@@ -15,78 +22,189 @@ pub fn do_licm(func: &mut Function,
...
+    for arg in func.dfg.inst_args(inst).into_iter() {
+        let new_def_position = match func.dfg.value_def(*arg) {
+            ValueDef::Arg(ebb_def, _) => {
+                match loop_analysis.base_loop_ebb(ebb_def) {
+                    None => DefPosition::OutsideLoop(),
+                    Some(loop_def) => {
+                        if inst_loop != loop_def {
+                            DefPosition::ParentLoop(loop_def)
+                        } else {
+                            DefPosition::InsideCurrentLoop()
```
sunfishcode: I'd find this code easier to follow if there were an explicit `return` here.

lib/cretonne/src/licm.rs
```diff
@@ -15,78 +22,189 @@ pub fn do_licm(func: &mut Function,
...
+            }
+        };
+        // We now find some sort of minimum between `closest_def_position` and `new_def_position`.
+        match (closest_def_position.clone(), new_def_position) {
+            (DefPosition::InsideCurrentLoop(), _) |
+            (_, DefPosition::InsideCurrentLoop()) => {
+                closest_def_position = DefPosition::InsideCurrentLoop();
+                break;
+            }
+            (DefPosition::ParentLoop(_), DefPosition::OutsideLoop()) => break,
```
sunfishcode: Is a `break` out of the loop intended here? It seems like even if one of the args is outside the loop, later args could still be closer.

lib/cretonne/src/licm.rs
```diff
@@ -15,78 +22,189 @@ pub fn do_licm(func: &mut Function,
...
                 }
-            };
-            // The last instruction of the pre-header is the termination instruction (usually
-            // a jump) so we need to insert just before this.
-            for inst in invariant_inst {
-                pos.insert_inst(inst);
             }
+            (DefPosition::OutsideLoop(), DefPosition::OutsideLoop()) => (),
+
         }
```
sunfishcode: In C++ I might expect the code above to look like this:

```
if (new_def_position.kind() < closest_def_position.kind()) {
    closest_def_position = new_def_position;
}
```

instead of a big match/switch on the pair of kind() values. Is something like this possible in Rust?

lib/cretonne/src/licm.rs
```diff
@@ -15,78 +22,189 @@ pub fn do_licm(func: &mut Function,
...
+            Some((_, last_jump_inst)) => move_inst(&mut func.layout, inst, last_jump_inst),
+        };
+    }
+}
+
+/// Test whether the given opcode is unsafe to even consider for LICM.
+fn trivially_unsafe_for_licm(opcode: Opcode) -> bool {
+    opcode.is_call() || opcode.is_branch() || opcode.is_terminator() || opcode.is_return() ||
+    opcode.can_trap() || opcode.other_side_effects()
+}
```
sunfishcode: I neglected to mention this earlier, but in the gvn version, `can_load()` and `can_store()` are handled separately in anticipation of gvn possibly be extended to operate on loads and stores some day. So licm needs to check these somewhere also.

We could probably think about factoring out these concerns into utility functions that could be shared by both, and other things in the future, but I'm ok deferring that for now.

lib/cretonne/src/licm.rs
sunfishcode: I spoke a little too soon; it's also necessary to compare the loops in the case of ParentLoop. But I'm still curious if there's any way to do this in Rust :-).

lib/cretonne/src/licm.rs
denismerigoux: Added checks for loads and stores.

lib/cretonne/src/licm.rs
denismerigoux: Fixed.

lib/cretonne/src/licm.rs
denismerigoux: Fixed.

lib/cretonne/src/licm.rs
denismerigoux: You can use `#[derive(Ord)]` and it will do what you want for enums, using the order in which you have declared the variants.

stoklund:
@denismerigoux made a comment on IRC that the domtree updating might not be working properly:

> jolesen, when you come back : I'm afraid it is not possible to incrementally update the dominator tree like I did and maintain perfect coherence between incrementally-updated and recomputed from scratch postorder https://gist.github.com/denismerigoux/0301009c5ee2bb91bda7274d8997cd97
> the reason is because there can be multiple postorder lists possible for a same dominance relationship (depending on which order you consider successors)

It seems to me that it should be possible to define a *canonical* post-order for Cretonne EBBs. In particular, the `DomTree::cfg_postorder()` function returns a *CFG* post-order, not a dominator tree post-order. (And we make use of that fact that any CFG post-order is also a topological ordering of the dominator tree).

The `DomTree::rpo_cmp()` function is actually defining a post-order for *program points* (i.e., instructions and EBB headers). This post-order should be invariant under EBB splitting, if it isn't we need to fix that.

sunfishcode:
Closing; we'll now track this at #393.

Issue #141, opened 2017-08-10
Better handling of literals in Ast substitution/copy and semantic checking


d1m0:
Rationale: icmp,load,store all have an enumerable immediate field, base on whose value we may want to do something different in the semantics. As a result, we'd like to be able to match a different transform to a concrete piece of rtl, depending on the value of some of the immediate fields. To enable this we need several things:
   0) A base class for all Literals - current (Enumerator, ConstantInt) and future (FlagSet)
   1) support substitutions Var->Literal, Literal->Var (requires changes to {Ast, Def,Rtl}.{copy, substitution}
   2) change verify_semantics() to enumerate all concrete values of enumerable immediates on top of enumerating all concrete typings.

PR #142, opened 2017-08-10
Fix for  #141

d1m0:
This PR proposes a fix for #141. Namely it:
     - Adds a new `Expr` subclass `Literal`, (base class for `ConstantInt`, `Enumerator`...)
     - Adds support for substitutions `Var`->`Literal`, `Literal`->`Var` in {`Ast`, `Def`,`Rtl`}.{`copy()`, `substitution()`}
     - Changes `verify_semantics()` to enumerate all concrete values of enumerable immediates on top of enumerating all concrete typings.
     - As a proof of concept, adds correct semantics for `icmp`, along with the necessary primitives
     - Additional type checking for `Literal` kinds in `Apply.__init__()`

### DIFF
--- a/lib/cretonne/meta/cdsl/operands.py
+++ b/lib/cretonne/meta/cdsl/operands.py
@@ -5,10 +5,10 @@ from .types import ValueType
 from .typevar import TypeVar
 
 try:
-    from typing import Union, Dict, TYPE_CHECKING  # noqa
+    from typing import Union, Dict, TYPE_CHECKING, Iterable  # noqa
     OperandSpec = Union['OperandKind', ValueType, TypeVar]
     if TYPE_CHECKING:
-        from .ast import Enumerator, ConstantInt  # noqa
+        from .ast import Enumerator, ConstantInt, Literal  # noqa
 except ImportError:
     pass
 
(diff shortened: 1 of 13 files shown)


stoklund:
In `meta/isa/intel/legalize.py`, we have code like this:

```python
for ty in [i32, i64]:
    intel_expand.legalize(
            a << insts.sdiv.bind(ty)(x, y),
            Rtl(
                xhi << insts.sshr_imm(x, imm64(ty.lane_bits() - 1)),
                (a, dead) << x86.sdivmodx(x, xhi, y)
            ))
    intel_expand.legalize(
            a << insts.srem.bind(ty)(x, y),
            Rtl(
                xhi << insts.sshr_imm(x, imm64(ty.lane_bits() - 1)),
                (dead, a) << x86.sdivmodx(x, xhi, y)
            ))
```

This is no big deal for only two types, but one could imagine other cases where it may be useful to be able to represent "constant functions". That is, functions whose value depend on the concrete types that a `Var` assumes. So we could write instead:

```python
intel_expand.legalize(
        a << insts.sdiv(x, y),
        Rtl(
            xhi << insts.sshr_imm(x, x.lane_bits() - 1),
            (a, dead) << x86.sdivmodx(x, xhi, y)
        ))
```

We don't need to support this now, but it is something to keep in mind when extending `Expr`.

lib/cretonne/meta/cdsl/operands.py
```diff
@@ -128,6 +128,17 @@ def rust_enumerator(self, value):
         """
         return '{}::{}'.format(self.rust_type, self.values[value])
 
+    def is_enumeratable(self):
```
stoklund: \* enumerable

PR #143, opened 2017-08-10
Transfer from the wasm2cretonne repo

denismerigoux:
Now that it is in a somewhat mature state, the wasm2cretonne code is ready to be transfered in the main Cretonne repo. The original repo is here : https://github.com/denismerigoux/wasm2cretonne.


lib/wasm2cretonne-util/Cargo.toml
```diff
@@ -0,0 +1,22 @@
+[package]
+name = "wasm2cretonne-util"
+version = "0.0.0"
+authors = ["The Cretonne Project Developers"]
+publish = false
```
sunfishcode: Could you add description and license fields to this and the other Cargo.toml files?

denismerigoux: Done.

stoklund:
Thanks @denismerigoux, this is great work!

I would like to rearrange things a bit before we land this in the Cretonne repo:

- The `wasm2cretonne` library crate should follow the existing naming conventions, so the crate name should be `cretonne-wasm` and its sources should be in `lib/wasm`.
- I would rather not have binary files checked in as test cases. We don't need to land the tests right away, but when we do, I would like to have `.wast` files that are translated to `.wasm` by a script that runs `wast2wasm`.
- I think these `.wast` tests should live in a top-level directory `wasmtests` next to `filetests`.
- I would prefer to keep `wasmstandalone` and the corresponding parts of `wasm2cretonne-util` in a separate repo for now.
- The parts of `wasm2cretonne-util` that are used for running the tests can be merged into the `cton-util` executable.

denismerigoux:
All the suggestions have been done.

PR #150, opened 2017-09-01
Replace SSABuilder's variables HashMaps with EntityMaps.

sunfishcode:
Current measurements show that memory usage is approximately the same,
and it's about 20% faster.

### DIFF
--- a/lib/cretonne/src/ir/entities.rs
+++ b/lib/cretonne/src/ir/entities.rs
@@ -55,6 +55,12 @@ impl Value {
     }
 }
 
+impl Default for Value {
+    fn default() -> Value {
+        Value(u32::MAX)
+    }
+}
+
 /// An opaque reference to an instruction in a function.
 #[derive(Copy, Clone, PartialEq, Eq, Hash, Debug, PartialOrd, Ord)]
 pub struct Inst(u32);
--- a/lib/cretonne/src/ir/entities.rs
+++ b/lib/cretonne/src/ir/entities.rs
@@ -55,12 +55,6 @@ impl Value {
     }
 }
 
-impl Default for Value {
-    fn default() -> Value {
-        Value(u32::MAX)
-    }
-}
-
 /// An opaque reference to an instruction in a function.
 #[derive(Copy, Clone, PartialEq, Eq, Hash, Debug, PartialOrd, Ord)]
 pub struct Inst(u32);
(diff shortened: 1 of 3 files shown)


lib/cretonne/src/ir/entities.rs
```diff
@@ -55,6 +55,12 @@ impl Value {
     }
 }
 
+impl Default for Value {
```
stoklund: This is unfortunate, I would prefer if we could avoid defining `Default` for `Value`.

Note that `EntityMap` doesn't actually require its elements to implement `Default` because it already stores the default element to use in the `default` member.

We should add a `EntityMap::with_default(default: V)` constructor in addition to `EntityMap::new()` which does require `V: Default`. Using such a method, you should be able to provide default values even for your nested mappings, without requiring anything to implement `Default`.

lib/frontend/src/ssa.rs
```diff
@@ -141,6 +140,12 @@ impl ReservedValue for Block {
     }
 }
 
+impl Default for Block {
```
stoklund: Here again, you would be able to use `EntityMap::with_default()` instead.

lib/frontend/src/ssa.rs
```diff
@@ -39,7 +38,7 @@ where
 {
     // Records for every variable and for every revelant block, the last definition of
     // the variable in the block.
-    variables: EntityMap<Variable, HashMap<Block, Value>>,
+    variables: EntityMap<Variable, EntityMap<Block, Value>>,
```
stoklund: Should this be `EntityMap<Block, PackedOption<Value>>`? That gives you a sensible default value.

lib/cretonne/src/ir/entities.rs
sunfishcode: Makes sense. I added a patch to do that now.

lib/frontend/src/ssa.rs
```diff
@@ -228,8 +227,10 @@ where
     ) -> (Value, SideEffects) {
         // First we lookup for the current definition of the variable in this block
         if let Some(var_defs) = self.variables.get(var) {
-            if let Some(val) = var_defs.get(&block) {
-                return (*val, SideEffects::new());
+            if let Some(packed) = var_defs.get(block) {
```
stoklund: I think `var_defs[block].expand()` would work fine too.

PR #155, opened 2017-09-13
Simplify targets and settings for cton-util wasm.

sunfishcode:
This makes licm and simple_gvn tests require a target ISA. Those passes don't inherently require a target ISA, though in any realistic use case they'll have one, and this allows them to have access to the contained Flags.

This also moves the responsibility for asking for cfg and domtree computations the code that needs them, to reduce the number of places that implicitly know about pass dependencies.

And it adds a more complete way to specify flags and isa to cton-wasm via the command-line.

### DIFF
--- a/filetests/licm/basic.cton
+++ b/filetests/licm/basic.cton
@@ -1,5 +1,8 @@
 test licm
 
+; Target chosen arbitrarily
+isa riscv
+
 function %simple_loop(i32) -> i32 {
 
 ebb1(v0: i32):
--- a/filetests/licm/complex.cton
+++ b/filetests/licm/complex.cton
@@ -1,5 +1,8 @@
 test licm
 
+; Target chosen arbitrarily
+isa riscv
+
 function %complex(i32) -> i32 {
 
 ebb0(v0: i32):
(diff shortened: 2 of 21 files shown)


stoklund:
This PR seems to contain a number of unrelated changes that I would prefer to deal with in separate PRs.

- For passes that need to access the `Flags`, I don't understand the benefit of giving them a `TargetIsa` when you can just pass a `&Flags` reference. If these passes really are ISA independent, I don't want the possibility of them sneaking in ISA-dependent behavior that then requires more testing.
- Regarding the "host" ISA in the `lookup` function. This opens the door for file tests that behave differently on different host ISAs I really want to avoid that because we would need much more CI coverage. LLVM has tests like this and it's causing a world of problems. Most realistic ISAs also need further information about CPU features to be useful, and I don't want to do CPU detection in Cretonne.
- I don't think it makes sense to pass the dominator tree to `legalize_function()` just to invalidate it there. If the legalizer doesn't use the dominator tree, it shouldn't be responsible for invalidating it. I wouldn't want to pass in the loop analysis for invalidation either.
- For the command line, I would prefer a syntax that is closer to the `set` and `isa` directives in test files. So something like `--set foo` and `--set foo=bar`. Compare to the standard `-DFOO` and `-DFOO=BAR` flags to a C compiler. For the `--isa` option, I think the argument should be the whole `isa` line, so `--isa "intel haswell"`.

sunfishcode:
Thanks for the feedback! I'm submitting a new series of pull requests.

PR #156, opened 2017-09-13
Make passes assert their dependencies consistently.

sunfishcode:
This avoids ambiguity about whose responsibility it is to run
to compute cfg, domtree, and loop_analysis data.

### DIFF
--- a/lib/cretonne/src/dominator_tree.rs
+++ b/lib/cretonne/src/dominator_tree.rs
@@ -219,10 +219,7 @@ impl DominatorTree {
 
     /// Reset and compute a CFG post-order and dominator tree.
     pub fn compute(&mut self, func: &Function, cfg: &ControlFlowGraph) {
-        debug_assert!(
-            cfg.is_valid(),
-            "DominatorTree computation requires a computed CFG"
-        );
+        debug_assert!(cfg.is_valid());
         self.compute_postorder(func, cfg);
         self.compute_domtree(func, cfg);
         self.valid = true;
--- a/lib/cretonne/src/dominator_tree.rs
+++ b/lib/cretonne/src/dominator_tree.rs
@@ -212,7 +212,6 @@ impl DominatorTree {
 
     /// Allocate and compute a dominator tree.
     pub fn with_function(func: &Function, cfg: &ControlFlowGraph) -> DominatorTree {
-        debug_assert!(cfg.is_valid());
         let mut domtree = DominatorTree::new();
         domtree.compute(func, cfg);
         domtree
--- a/lib/cretonne/src/dominator_tree.rs
+++ b/lib/cretonne/src/dominator_tree.rs
@@ -37,6 +37,8 @@ pub struct DominatorTree {
 
     // Scratch memory used by `compute_postorder()`.
     stack: Vec<Ebb>,
+
+    valid: bool,
 }
 
 /// Methods for querying the dominator tree.
@@ -51,6 +53,7 @@ impl DominatorTree {
     /// Note that this post-order is not updated automatically when the CFG is modified. It is
     /// computed from scratch and cached by `compute()`.
     pub fn cfg_postorder(&self) -> &[Ebb] {
+        debug_assert!(self.is_valid());
         &self.postorder
     }
 
@@ -203,11 +206,13 @@ impl DominatorTree {
             nodes: EntityMap::new(),
             postorder: Vec::new(),
             stack: Vec::new(),
+            valid: false,
         }
     }
 
     /// Allocate and compute a dominator tree.
     pub fn with_function(func: &Function, cfg: &ControlFlowGraph) -> DominatorTree {
+        debug_assert!(cfg.is_valid());
         let mut domtree = DominatorTree::new();
         domtree.compute(func, cfg);
         domtree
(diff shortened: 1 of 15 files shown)


lib/cretonne/src/dominator_tree.rs
```diff
@@ -37,6 +37,8 @@ pub struct DominatorTree {
 
     // Scratch memory used by `compute_postorder()`.
     stack: Vec<Ebb>,
+
+    valid: bool,
```
stoklund: What's the motivation for having a separate valid flag instead of just clearing the `nodes` when invalidating the tree?

This variable doesn't seem to be reset anywhere.

lib/cretonne/src/dominator_tree.rs
```diff
@@ -203,20 +206,27 @@ impl DominatorTree {
             nodes: EntityMap::new(),
             postorder: Vec::new(),
             stack: Vec::new(),
+            valid: false,
         }
     }
 
     /// Allocate and compute a dominator tree.
     pub fn with_function(func: &Function, cfg: &ControlFlowGraph) -> DominatorTree {
+        debug_assert!(cfg.is_valid());
```
stoklund: This is asserted in `compute()`, no need for duplicating it.

lib/cretonne/src/licm.rs
```diff
@@ -16,9 +16,13 @@ pub fn do_licm(
...
 ) {
-    cfg.ensure(func);
-    domtree.ensure(func, cfg);
-    loop_analysis.ensure(func, cfg, domtree);
+    debug_assert!(cfg.is_valid(), "LICM requires the CFG to be computed");
+    debug_assert!(domtree.is_valid(), "LICM requires a computed DominatorTree");
+    debug_assert!(
+        loop_analysis.is_valid(),
+        "LICM requires a computed LoopAnalysis"
+    );
```
stoklund: I think it would be ok to drop the explanatory text so these assertions don't go multi-line. The expression being asserted is quite self-explanatory.

src/filetest/regalloc.rs
```diff
@@ -44,11 +44,12 @@ impl SubTest for TestRegalloc {
         let mut comp_ctx = cretonne::Context::new();
         comp_ctx.func = func.into_owned();
 
-        comp_ctx.flowgraph();
+        comp_ctx.cfg.compute(&comp_ctx.func);
         // TODO: Should we have an option to skip legalization?
         comp_ctx.legalize(isa).map_err(|e| {
             pretty_error(&comp_ctx.func, context.isa, e)
         })?;
+        comp_ctx.domtree.compute(&comp_ctx.func, &comp_ctx.cfg);
```
stoklund: It seems unfortunate to have these direct calls to the passes spread all over the testing code. How about adding argument-less methods like `comp_ctx.ensure_cfg()`, and `comp_ctx.ensure_domtree()`?

stoklund:
I think it is a good idea to make passes assert instead of just recomputing what they need.

Since we're using the `Context` in many contexts (haha), it looks like the information about the requirements and invalidations of a pass gets spread all over. I think it would make sense to try and keep that information localized to the `Context`.

What do you think about an approach like this:

```rust
    /// Run the legalizer for `isa` on the function.
    pub fn legalize(&mut self, isa: &TargetIsa) -> CtonResult {
        self.ensure_cfg();
        // Legalization invalidates the domtree and loops by mutating the CFG.
        self.domtree.clear();
        self.loop_analysis.clear();
        legalize_function(&mut self.func, &mut self.cfg, isa);
        self.verify_if(isa)
    }
```

This means that you can always call `comp_ctx.legalize()` from some test code, and it will revalidate the CFG if needed and invalidate any analysis that gets clobbered. If, say, `legalize_function` starts depending on and preserving the domtree, we only need make the change in one place.

lib/cretonne/src/dominator_tree.rs
sunfishcode: What prompted me was the `empty` tests in dominator_tree.rs, topo_order.rs, and others, which construct empty functions and call various dominator/etc. functions on them. They do call `compute`, but there are no basic blocks, so `nodes` is empty. Would you prefer to change those tests to add an entry block?

The lack of a reset is a bug in my patch. It should be reset in `clear`.

stoklund: OK, that makes sense. Empty functions are a silly corner case, but let's just leave the `valid` flag in.

sunfishcode:
I like that general direction. `Context` has a `compile` which seems to already be heading there, and my patch [takes it a step further](https://github.com/sunfishcode/cretonne/blob/3b308fbbfbf43b7d85055262be1aa1cbcf1914e0/lib/cretonne/src/context.rs). However, if we also make fine-grained operations like `legalize` provide for their dependencies themselves, then it gets harder to look at code that calls `legalize` to see what's getting run and when.

An appealing approach is to have `compile`, or other future similar operations, be the main interface that most users will call. It'll handle all the dependencies of the things it runs. `compile` can be moderately configurable, through `opt_level` and other flags, so it should be usable by most users.

Users that want more control can run the individual passes themselves, though they'll need to set up all the dependencies themselves. This isn't as convenient as having a dependencies scheduled automatically, though there will be assertions to detect mistakes. However in exchange for taking on this inconvenience, they'll get nice upside: they'll have a single list of all the passes that are being run, in plain straight-line code, which is something I've often missed when working under a `PassManager`.

What do you think?

stoklund:
Sure, we can do it that way.

I do think we should provide convenience methods like `ensure_cfg()`, `ensure_domtree()` etc for running those dependency passes. This avoids the clutter of passing in arguments to these passes everywhere.

PR #158, opened 2017-09-14
Add settings and isa command-line options to cton-util wasm.

sunfishcode:
This implements the full parsing for `set` and `isa` commands, following the [feedback](https://github.com/stoklund/cretonne/pull/155#issuecomment-329295426).

### DIFF
--- a/src/wasm.rs
+++ b/src/wasm.rs
@@ -11,7 +11,6 @@ use cretonne::Context;
 use cretonne::verifier;
 use cretonne::settings;
 use cretonne::isa::{self, TargetIsa};
-use std::borrow::Borrow;
 use std::fs::File;
 use std::error::Error;
 use std::io;
@@ -86,7 +85,7 @@ pub fn run(
             path.to_path_buf(),
             name,
             &flags,
-            isa.borrow(),
+            &*isa,
         )?;
     }
     Ok(())
--- a/src/wasm.rs
+++ b/src/wasm.rs
@@ -5,10 +5,13 @@
 //! and tables, then emitting the translated code with hardcoded addresses to memory.
 
 use cton_wasm::{translate_module, DummyRuntime, WasmRuntime};
+use cton_reader::{parse_options, Location};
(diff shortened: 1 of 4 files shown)


src/wasm.rs
```diff
@@ -50,16 +53,32 @@ pub fn run(
...
+        None => return Err(String::from("expected ISA name")),
+        Some(w) => w,
+    };
+    let isa_builder = match isa::lookup(isa_name) {
+        Err(isa::LookupError::Unknown) => return Err(format!("unknown ISA '{}'", isa_name)),
+        Err(isa::LookupError::Unsupported) => {
+            return Err(format!("support for ISA '{}' not enabled", isa_name))
+        }
+        Ok(b) => b,
+    };
```
stoklund: This might work better with `map_err` combined with a `?` operator, but NBD.

src/wasm.rs
```diff
@@ -70,6 +89,7 @@ pub fn run(
             path.to_path_buf(),
             name,
             &flags,
+            isa.borrow(),
```
stoklund: The good old C++ trick `&*isa` works too.

src/wasm.rs
```diff
@@ -50,16 +53,32 @@ pub fn run(
...
+        flag_set.iter().map(|x| x.as_str()),
+        &mut flag_builder,
+        &Location { line_number: 0 },
+    ).map_err(|err| err.to_string())?;
     let flags = settings::Flags::new(&flag_builder);
 
+    let mut words = flag_isa.trim().split_whitespace();
+    // Look for `isa foo`.
+    let isa_name = match words.next() {
+        None => return Err(String::from("expected ISA name")),
```
stoklund: Does this mean that `cton-util wasm` doesn't work without an `--isa` option? Was that intended?

sunfishcode: Yes, for the moment. I'll submit a PR to create a host-detection crate and use that for a default soon.

PR #159, opened 2017-09-15
Add a "host" crate for autodetecting the host.

sunfishcode:
I'm open to suggestions here.

### DIFF
--- a/lib/native/.gitignore
+++ b/lib/native/.gitignore
@@ -1,0 +1,3 @@
+target/
+**/*.rs.bk
+Cargo.lock
--- a/lib/native/.gitignore
+++ b/lib/native/.gitignore
@@ -1,3 +1,0 @@
-target/
-**/*.rs.bk
-Cargo.lock
--- a/lib/native/src/lib.rs
+++ b/lib/native/src/lib.rs
@@ -22,67 +22,67 @@ pub fn builders() -> Result<(settings::Builder, isa::Builder), ()> {
 
     // TODO: Add RISC-V support once Rust supports it.
 
-    let name = if cfg!(target_arch = "x86") {
-        parse_x86_cpuid(&mut flag_builder);
-        "intel"
-    } else if cfg!(target_arch = "x86_64") {
+    if cfg!(target_pointer_width = "64") {
         flag_builder.enable("is_64bit").unwrap();
-        parse_x86_cpuid(&mut flag_builder);
+    }
+
+    let name = if cfg!(any(target_arch = "x86", target_arch = "x86_64")) {
         "intel"
     } else if cfg!(target_arch = "arm") {
         "arm32"
     } else if cfg!(target_arch = "aarch64") {
-        flag_builder.enable("is_64bit").unwrap();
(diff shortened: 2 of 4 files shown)


stoklund:
Just a couple high-level comments:

1. I would prefer a more specific crate name than `cretonne-host`. How about `cretonne-cpuid`?
2. I really want to discourage tests that behave differently on different hosts. Specifically, the `lookup()` function provided here is a little too close to that kind of use by making it easy to write `isa host` in a test file somewhere. Would it make sense to have an API that is more focused on "I want to generate code for running on this host CPU", and that doesn't let you configure a cross compiler?

sunfishcode:
`cretonne-cpuid` sounds somewhat x86-specific. How about `cretonne-native` to reflect that it's focused on native compilation and doesn't support cross compilation?

I've now updated the PR with patches that change the API to be less like `lookup` and that doesn't have a builtin fallback to cross compilation. Does this look right?

lib/native/.gitignore
```diff
@@ -0,0 +1,3 @@
+target/
+**/*.rs.bk
+Cargo.lock
```
stoklund: This should be covered by the top-level `.gitignore`, no?

lib/native/src/lib.rs
```diff
@@ -0,0 +1,88 @@
...
+pub fn builders() -> Result<(settings::Builder, isa::Builder), ()> {
+    let mut flag_builder = settings::builder();
+
+    // TODO: Add RISC-V support once Rust supports it.
+
+    let name = if cfg!(target_arch = "x86") {
+        parse_x86_cpuid(&mut flag_builder);
+        "intel"
+    } else if cfg!(target_arch = "x86_64") {
+        flag_builder.enable("is_64bit").unwrap();
```
stoklund: This probably simpler with `if cfg!(target_pointer_width = "64")`.

lib/native/src/lib.rs
```diff
@@ -0,0 +1,88 @@
...
+            match err {
+                isa::LookupError::Unknown => panic!(),
+                isa::LookupError::Unsupported => (()),
+            }
+        },
+    )
+}
+
+#[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
+fn parse_x86_cpuid(flag_builder: &mut settings::Builder) {
```
stoklund: The flags set by this function are Intel-specific, so they live in the `isa::Builder`. The `settings::Builder` deals with ISA-independent flags.

lib/native/.gitignore
sunfishcode: Ah, that was copied from `lib/wasm/.gitignore`, which is probably an leftover from when that code lived in its own repo. I'll remove both.

lib/native/src/lib.rs
sunfishcode: Fixed.

lib/native/src/lib.rs
sunfishcode: Fixed.

PR #160, opened 2017-09-15
RISC-V compressed instruction encodings [WIP]

angusholder:
How should I go about ensuring the registers supplied aren't 0? Or is that a concern for somewhere else in the code?


stoklund:
The register allocator shold ensure that only valid registers are passed to your GPRC operands.

However, for these compressed instructions, we also have full sized encodings that accept all registers, and that is the.encodings that are active during register allocation.

We are missing a compression pass that runs after register allocation and selects smaller encodings when the allocated registers permit it.

lib/cretonne/src/isa/riscv/binemit.rs
```diff
@@ -236,4 +236,53 @@ fn put_cs<CS: CodeSink + ?Sized>(bits: u16, rs1: RegUnit, rs2: RegUnit, sink: &m
...
+    i |= ((offset >> 5) & 0b1) << 2;
+    i |= ((offset >> 1) & 0b11) << 3;
+    i |= ((offset >> 6) & 0b11) << 5;
+    i |= rs1 << 7;
+    i |= ((offset >> 3) & 0b11) << 10;
+    i |= ((offset >> 8) & 0b1) << 12;
+    i |= funct3 << 13;
+
+    sink.put2(i);
 }
```
bjorn3: Missing trailing newline

angusholder: Ah thanks, I haven't gotten round to running the auto formatter yet

lib/cretonne/meta/isa/riscv/registers.py
```diff
@@ -18,6 +18,8 @@
         units=32, prefix='f')
 
 GPR = RegClass(IntRegs)
+# The popular registers used in some compressed instructions; x8-x15
+GPRC = GPR[8:15]
```
stoklund: This is using Python's slice semantics, so you want `8:16`.

lib/cretonne/meta/isa/riscv/encodings.py
```diff
@@ -153,3 +156,75 @@
 RV32.enc(base.regmove.i32, Irmov, OPIMM(0b000))
 RV64.enc(base.regmove.i64, Irmov, OPIMM(0b000))
 RV64.enc(base.regmove.i32, Irmov, OPIMM32(0b000))
+
+
+def rv32_enc_c(inst, recipe, bits):
+    RV32.enc(inst, recipe, bits, isap=use_c)
```
stoklund: Note that encoding recipes can have their own `instp` and `isap` predicates that are automatically combined with the per-opcode predicates. That is probably easier to use that this function.

angusholder: Ah thanks, I'll get that in the next commit

lib/cretonne/meta/isa/riscv/registers.py
angusholder: Ah of course! I'm a bit rusty on Python

angusholder:
@stoklund I think I've come across what appears to be a bug: in gen_encoding.py, it seems to assume that if there's an isap then instp also exists, so if you run my last commit, it tries to dereference a None

stoklund:
That sounds like a bug in `gen_encoding.py`. Is it easy to fix?

angusholder:
I've worked out what was going on, seems that the isap code path hadn't been used before, because when adding in an ISA predicate it generated an if statement with no open brace, and also ended up not returning a return value on all code paths.

I've now completed the encodings for all the compressed instructions, just have a few questions:

The C.LUI instruction mustn't use x2 because the use of that register indicates an ADDI16SP instruction, but I can't see any obvious way to constrain an instruction from using just 1 particular register, how should I approach that?

Several recipes correspond to just a single instruction so the bits argument feels rather redundant, would you consider it bad style to inline it?

I haven't implemented every constraint described in the RISC-V reference because I can't work out if some are actually important (eg: register must not be x0, or immediate must not be 0), I'll have another look at them later.

stoklund:
About `C.LUI`. We don't have a way of describing complex register class constraints like that. In this case, `%x2` is the stack pointer, so it is reserved and the register allocator will never use it. I think it is OK to add a coment to the recipe definition, and leave it accepting `%x2`.

The same thing for `%x0`: This is the constant zero register which is reserved, so the register allocator will never pick it.

If a recipe only corresponds to a single instruction, it's OK to leave the `bits` field as zero. Or even use it for something else. Just document how you're using it when it differs from the default usage.

lib/cretonne/meta/gen_encoding.py
```diff
@@ -173,11 +173,15 @@ def emit_recipe_predicates(isa, fmt):
...
                     name,
                     'isap' if isap else '_'), '}'):
-            if isap:
+            if isap is not None:
                 n = isa.settings.predicate_number[isap]
-                with fmt.indented('if isap.test({})'.format(n), '}'):
-                    fmt.line('return false;')
-            emit_instp(instp, fmt)
+                if instp is not None:
+                    with fmt.indented('if isap.test({}) {{'.format(n), '}'):
```
stoklund: Should this also be `if !isap.test(...`?

angusholder: Ah that would make more sense!

stoklund:
The verifier fix looks great! would you mind landing that in a separate PR?

lib/cretonne/meta/isa/riscv/encodings.py
```diff
@@ -168,23 +170,20 @@
 RV64.enc(base.regmove.i32, Irmov, OPIMM32(0b000))
 
 
-def rv_enc(inst, recipe, bits, isap=None):
+def rv_enc(inst, recipe, bits, isap=None, instp=None):
     # type: (Instruction, EncRecipe, int, PredNode) -> None
```
bjorn3: Does this need to be updated?

bnjbvr:
Hi @angusholder! It's been a while since you've opened this PR, so I wanted to ping you to know what's the status here, please? Since it says "WIP" in the title, I'm assuming it is not finished yet. In the meanwhile, we have started porting the meta language from Python to the Rust crate, so there might be a few conflicts to solve.

There are a few ways I can think of to move this forward:

- wait for the Rust port to happen, and rewrite this PR in Rust thereafter. (There's no clear deadline on when the Rust port will be landed, so it's the least predictable timeline)
- rebase this PR, adapt to the latest changes, and resolve the WIP status.
- close this PR if you're not interested on working on this anymore (which would be fine, since it's open-source work).

There might be also some changes that could live in their own PRs and be merged sooner than the new RISC-V encodings, if you wanted to do this; I see a mention about a verifier fix, not sure if it's still valid or not.

Any thoughts here?

angusholder:
I'm definitely interested in doing some more work on Cranelift, I'll have a look see if I can split up this PR at all, and see how easy it will be to rebase it up to the latest. I'd probably keep it as the python DSL for now. Doesn't look like too much has happened to the RISC-V backend since I touched it, although the git history isn't too helpful with how much everything has moved around!

bnjbvr:
> Doesn't look like too much has happened to the RISC-V backend since I touched it, although the git history isn't too helpful with how much everything has moved around!

The RISC V backend has not much changed indeed.

Regarding the meta crate changes, which are probably the most interesting changes you need to be aware of, as of now, only the registers (and some of the settings) generation has been moved from the Python code (in cranelift-codegen/meta-python) to the Rust meta crate (in cranelift-codegen/meta/src).

bnjbvr:
Hi @angusholder ! It's been a while and the entire meta code has been ported to Rust last week, so this is now very stale. I've quickly looked at the content of this PR, and I think it's still worth to have, at least in the new meta code. You'll see that recipes/encodings look quite similar to what the Python code had, at least for RISCV. Would you be interested in closing this PR (it's a bit big and contains a merge commit, which we'd like to avoid to keep a linear history), and open smaller PRs that bring the same commits, in smaller groups? Or is there another way to proceed that could work for you and make reviews easier? Thanks!

bnjbvr:
Closing for lack of response; feel free to open a new issue or PR if you want to get back to working on this! Thanks.

PR #162, opened 2017-09-22
Choose compressed instruction encodings where constraints allow

angusholder:
This isn't done yet, I was wondering what the best way is to tell that an instruction needs a REX encoding due to REX.w? Can it be done in a general way, or should this be intel-specific? For context, the tests current fail on:

    FAIL filetests/isa/intel/binary64.cton: binemit(%I64): Bad machine code for inst159: v141 = bint.i64 v121
    Want: 48 0f b6 cb
    Got:  0f b6 cb

The way I'm currently checking constraints does feel like it could be generalised somewhat.

### DIFF
--- a/filetests/isa/intel/binary64.cton
+++ b/filetests/isa/intel/binary64.cton
@@ -220,9 +220,9 @@ ebb0:
     ; asm: movq -50(%r10), %rdx
     [-,%rdx]            v141 = load.i64 v3-50           ; bin: 49 8b 52 ce
     ; asm: movl 50(%rcx), %edi
-    [-,%rdi]            v142 = uload32.i64 v1+50        ; bin: 40 8b 79 32
+    [-,%rdi]            v142 = uload32.i64 v1+50        ; bin: 8b 79 32
     ; asm: movl -50(%rsi), %edx
-    [-,%rdx]            v143 = uload32.i64 v2-50        ; bin: 40 8b 56 ce
+    [-,%rdx]            v143 = uload32.i64 v2-50        ; bin: 8b 56 ce
     ; asm: movslq 50(%rcx), %rdi
     [-,%rdi]            v144 = sload32.i64 v1+50        ; bin: 48 63 79 32
     ; asm: movslq -50(%rsi), %rdx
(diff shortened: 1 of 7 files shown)


stoklund:
You should have enough information to check the constraints in a general way without knowing anything about the target ISA. It is the responsibility of the encoding definitions to make sure that you will never see an invalid encoding such as a REX-less encoding for an instruction that requires a REX.W bit to be set.

The `bint.i64` instruction is a special case. It zero-extends a 0 or 1 value in an 8-bit register to a 4-bit register with a `movzbq %bl, %rcx` instruction which has REX.W=1. However, since 32-bit instructions implicitly zero the high 32 bits of the destination register, it could just as well be implemented with a `movzbl %bl, %ecx` instruction which has REX.W=0. That is what the REX-less encoding is doing here:

```python
I64.enc(base.bint.i64.b1, *r.urm.rex(0x0f, 0xb6, w=1))
I64.enc(base.bint.i64.b1, *r.urm_abcd(0x0f, 0xb6))  # zext to i64 implicit.
```

It would probably be cleaner if we didn't set `w=1` on the first encoding either.

lib/cretonne/meta/isa/intel/recipes.py
```diff
@@ -160,6 +160,11 @@ def __call__(self, *ops, **kwargs):
                 instp=self.instp,
                 isap=self.isap,
                 emit=replace_put_op(self.emit, name))
+
+            recipe.ins = tuple(GPR8 if (reg is GPR) else reg for reg in tuple(recipe.ins))
+            recipe.outs = tuple(GPR8 if (reg is GPR) else reg for reg in tuple(recipe.outs))
```
stoklund: Yes, exactly, but you should also map FPR -> FPR8. And perhaps move the list comprehension into its own function?

lib/cretonne/meta/isa/intel/encodings.py
```diff
@@ -185,21 +191,24 @@ def enc_flt(inst, recipe, *args, **kwargs):
...
 enc_i32_i64_ld_st(base.load, True, r.ld, 0x8b)
 enc_i32_i64_ld_st(base.load, True, r.ldDisp8, 0x8b)
 enc_i32_i64_ld_st(base.load, True, r.ldDisp32, 0x8b)
 
-I64.enc(base.uload32.i64, *r.ld.rex(0x8b))
-I64.enc(base.uload32.i64, *r.ldDisp8.rex(0x8b))
-I64.enc(base.uload32.i64, *r.ldDisp32.rex(0x8b))
+enc_i64(base.uload32.i64, r.ld, 0x8b)
+enc_i64(base.uload32.i64, r.ldDisp8, 0x8b)
+enc_i64(base.uload32.i64, r.ldDisp32, 0x8b)
```
stoklund: Yep. Nice.

filetests/isa/intel/binary64.cton
```diff
@@ -220,9 +220,9 @@ ebb0:
     ; asm: movq -50(%r10), %rdx
     [-,%rdx]            v141 = load.i64 v3-50           ; bin: 49 8b 52 ce
     ; asm: movl 50(%rcx), %edi
-    [-,%rdi]            v142 = uload32.i64 v1+50        ; bin: 40 8b 79 32
+    [-,%rdi]            v142 = uload32.i64 v1+50        ; bin: 8b 79 32
```
stoklund: This is pretty cool. Now the test matches the output from llvm-mc.

src/filetest/binemit.rs
```diff
@@ -104,16 +104,54 @@ impl SubTest for TestBinEmit {
...
+                        legal_encodings
+                            .filter(|e| {
+                                let mut result = true;
+                                let constraints = &encinfo.constraints[e.recipe()];
+
+                                for (&arg, constraint) in
+                                    func.dfg.inst_args(inst).iter().zip(constraints.ins)
+                                {
+                                    let reg = func.locations[arg].unwrap_reg();
+                                    if !constraint.regclass.contains(reg) {
```
stoklund: You have the right idea here, but this code would be better if it were broken down into smaller functions.

There are more kinds of constraints that just register classes, and you can't assume that a value location is a register. Values can also live in stack slots, and there will be instruction encodings that can access stack slots directly.

To begin with, you can just match a `ValueLoc::Stack` against a `ConstraintKind::Stack`. I will be adding some complexity to this type of constraint soon, though.

stoklund:
I would suggest that you move the constraint checking into two methods:

```rust
impl OperandConstraint {
    /// Check if this operand constraint is satisfied by the given value location.
    /// For tied constraints, this only checks the register class, not that the
    /// counterpart operand has the same value location.
    pub fn satisfied(loc: ValueLoc, func: &Function) -> bool;
}

impl RecipeConstraints {
    /// Check that these constraints are satisfied by the operands on `inst`.
    pub fn satisfied(inst: Inst, func: &Function) -> bool;
}
```

angusholder:
I think this is now ready to go

PR #164, opened 2017-10-03
Do a full compile in 'cton-util wasm'.

sunfishcode:
This removes the `optimize` option, as one can do that with
`--set`, eg. `--set opt_level=fastest`. And it adds an option to
print the compilation output.

And, it enables simple_gvn and licm for opt_level=fastest in Context.

### DIFF
--- a/lib/cretonne/src/context.rs
+++ b/lib/cretonne/src/context.rs
@@ -18,7 +18,7 @@ use isa::TargetIsa;
 use legalize_function;
 use regalloc;
 use result::{CtonError, CtonResult};
-use settings::FlagsOrIsa;
+use settings::{FlagsOrIsa, OptLevel};
 use verifier;
 use simple_gvn::do_simple_gvn;
 use licm::do_licm;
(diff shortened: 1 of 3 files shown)


lib/cretonne/src/context.rs
```diff
@@ -68,6 +68,12 @@ impl Context {
 
         self.compute_cfg();
         self.legalize(isa)?;
+        if isa.flags().opt_level() == OptLevel::Best {
+            self.compute_domtree();
+            self.compute_loop_analysis();
+            self.licm(isa)?;
+            self.simple_gvn(isa)?;
```
stoklund: I think I would prefer for these code motion passes to be guarded by their own setting for now, so they won't be enabled by default yet.

There are still some details of the IL semantics that we need to work out before these passes are safe:

- The legalizer inserts `spill` instructions writing function arguments into `OutgoingArg` stack slots before a call. These writes to the outgoing stack frame need to happen after the previous function call which can clobber that part of the stack. It's possible we should just tag `spill` as `can_store` and work out some rules for aliasing.
- I will be adding SSA value types for representing CPU flags soon. These types will require special treatment since they are not allowed to be live across any instruction which can clobber the flags. The `clobbers_flags` flag will be a property of the instruction encoding, not the opcode.

sunfishcode:
Ok, I removed the code for running simple_gvn and licm for now. We can figure out what to do with those later.

PR #170, opened 2017-10-18
Clarify undefined behavior and notrap.

sunfishcode:
This PR proposes that load/store/etc. become plain native unchecked memory accesses, without reliable trapping, and with full UB on OOB.

Verifiable safety may be had by using a heap_addr/stack_addr/global_addr immediately before every raw access. Bounds-check elimination optimizations may be performed by using a single heap_addr/stack_addr/global_addr for multiple accesses. Load DCE and DSE can be performed on normal load/store/etc. in the usual ways.

### DIFF
--- a/docs/langref.rst
+++ b/docs/langref.rst
@@ -441,7 +441,7 @@ When the address is the result of a :inst:`heap_addr`, :inst:`stack_addr`, or
 :inst:`global_addr` with sufficient size to cover the access given its size
 and offset, and there are no intervening memory reallocations, they are *safe*,
 and either succeed or reliably *trap*, depending on whether the computed address
-is *mapped* or not.
+is :term:`mapped` or not.
 
 When these conditions are not met, the behavior of these instructions is
 undefined.
@@ -461,13 +461,13 @@ optimizations.
 ======= =========================================
 Flag    Description
 ======= =========================================
-notrap  Address is known to be *mapped*.
+notrap  Address is known to be :term:`mapped`.
 aligned Trapping allowed for misaligned accesses.
 ======= =========================================
 
 Normal memory accesses are defined to either :ref:`succeed or trap <memory>`
-depending on whether the accesed memory is *mapped*. When the ``notrap`` flag
-is set, the behavior is undefined if the memory is not *mapped*.
+depending on whether the accesed memory is :term:`mapped`. When the ``notrap``
+flag is set, the behavior is undefined if the memory is not mapped.
 
 Loads and stores are *misaligned* if the resultant address is not a multiple of
 the expected alignment. By default, misaligned loads and stores are allowed,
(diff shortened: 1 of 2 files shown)


stoklund:
We need reliable trapping semantics in order to implement a sandbox. The guard pages after a heap are required to cause a trap when accessed. This is part of the load/store semantics when there is no `notrap` flag.

I don't want to copy the C/LLVM model for plain loads and stores where trapping is covered by the undefined behavior, we should explicitly include trapping as part of the load/store semantics.

The question is how does the `notrap` flag change the semantics of loads and stores? What do we actually need?

You don't define what you meant by "OOB". Do you mean in terms of a WebAssembly heap?

stoklund:
To be clear, the language in `README.rst` is quite out of date now, and it makes sense to delete the UB blurb.

My old idea of an easily verifiable subset didn't work out either. Sandboxed code today uses `load` and `store` instructions which are not part of a safe subset. You can still verify safety by looking at where the addresses come from, but it is not a s simple as checking for a safe subset of instructions.

It turns out that safety verification isn't really relevant for Cretonne.

sunfishcode:
That makes sense; we don't want trapping to be covered by UB. I've now updated the PR to include trapping as part of the possible defined behavior. I also reintroduced `notrap`, modified to produce UB if the memory is not mapped. This feels like what we actually need, though I should think about it some more. We don't need to rush this.

docs/langref.rst
```diff
@@ -415,18 +415,30 @@ Indirect function calls use a signature declared in the preamble.
...
 accessing memory, as well as :ref:`extending loads and truncating stores
-<extload-truncstore>`. There are also more restricted operations for accessing
-specific types of memory objects.
+<extload-truncstore>`.
+
+When the address is the result of a :inst:`heap_addr`, :inst:`stack_addr`, or
+:inst:`global_addr` with sufficient size to cover the access given its size
+and offset, and there are no intervening memory reallocations, they are *safe*,
+and either succeed or reliably *trap*, depending on whether the computed address
+is *mapped* or not.
```
stoklund: Instead of using italics, we should define these terms in the glossary and refer to them with `:term:`

sunfishcode: Ok, I've drafted up possible definitions for "mapped", "trap", and "safe". They're still a little vague in some respects, but I think this takes us in the right direction.

docs/langref.rst
```diff
@@ -415,18 +415,30 @@ Indirect function calls use a signature declared in the preamble.
...
+<extload-truncstore>`.
+
+When the address is the result of a :inst:`heap_addr`, :inst:`stack_addr`, or
+:inst:`global_addr` with sufficient size to cover the access given its size
+and offset, and there are no intervening memory reallocations, they are
+:term:`safe`, and either succeed or reliably :term:`trap`, depending on whether
+the computed address is :term:`mapped` or not.
+
+When these conditions are not met, the behavior of these instructions is
+undefined.
```
stoklund: This sentence makes it impossible to map Rust to Cretonne, doesn't it?

stoklund:
I think we should avoid defining "safe" as the opposite of "undefined behavior" or vice versa. I think your glossary definition of "safe" is OK in that respect.

We don't want a Rust program to be undefined behavior just because it doesn't use the heaps and globals.

sunfishcode:
Yes; I was thinking too much about the WebAssembly use cases there. I've now introduced concepts of "addressable" and "accessible" memory, which seem like a useful vocabulary to describe memory outside of just heaps, globals, and the stack.

stoklund:
I think we can land this in the current state. Please squash the commits into one, though.

sunfishcode:
Landed. We can obviously continue to refine these concepts and the documentation as needed.

PR #177, opened 2017-10-25
Avoid dangling references to block params when sealing an unreachable block.

sunfishcode:
This started out as a fairly ordinary bug fix, but the fix exposed another problem: `f32const` and `f64const` are documented as working for either scalar or vector, but they currently only support scalar. The naive fix is to give them an explicit type parameter, which is what this patch currently does, though it does makes things more verbose. And it's an API change.

Another option would be to just remove support for SIMD values from ssa.rs for now.

### DIFF
--- a/docs/example.cton
+++ b/docs/example.cton
@@ -4,7 +4,7 @@ function %average(i32, i32) -> f32 native {
     ss1 = local 8            ; Stack slot for ``sum``.
 
 ebb1(v1: i32, v2: i32):
-    v3 = f64const 0x0.0
+    v3 = f64const.f64 0x0.0
     stack_store v3, ss1
     brz v2, ebb3                  ; Handle count == 0.
     v4 = iconst.i32 0
@@ -28,6 +28,6 @@ ebb2(v5: i32):
     return v17
 
 ebb3:
-    v100 = f32const +NaN
+    v100 = f32const.f32 +NaN
     return v100
 }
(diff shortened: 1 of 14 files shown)


stoklund:
So the `f32const` and `f64const` instructions started out being polymorphic the same way you're doing it here. I decided to switch them to monomorphic scalar-only because the type variable was confusing. "Why do you need to specify the type of an `f32const`? It's obviously `f32`". I also think `f32const.f32` is quite grating lexically. (It seems I didn't update the documentation properly.)

I'm also not sure how big the need for splat constants is (other than the zero-initializer). Aarch64, for example has an *FMOV (scalar, immediate)* instruction that can materialize a few floating point constants, but only in scalar form. There is no vector splatting version of it. The Intel ISAs can only splat all zeros or all one bits.

I would prefer to keep these instructions scalar/monomorphic for now. I think an argument could be made that `iconst` should also be scalar-only, but since it's naturally polymorphic, there's little cost to allowing the vector splat constants there.

For your problem in the `emit_zero()` function, you can just add a `splat` instruction when materializing floating point zero vectors. Alternatively, I also think it could make sense to add a polymorphic `zero` instruction that can create the all-zeros value for all the same types that `load` supports.

sunfishcode:
Works for me. I don't think I need a polymorphic `zero` at this point, though I am considering factoring out the `emit_zero` function into a utility so that it can be shared among the 3 places that now need it.

Closing this PR as the patch is now much smaller.

PR #183, opened 2017-10-27
Add support for symbolic global variables.

### LINKED ISSUE #182: Support for symbolic memory addresses
Cretonne doesn't seem to have a way to support global variables, in a C/C++/Rust sense. It has "globals", but they have to be defined in the VMCtx, or accessible from pointers in the VMCtx. There's no equivalent of `FunctionName` for globals, for example, which could be used for arbitrary addresses to be resolved by a subsequent link step.

Would it make sense to add a field to `GlobalVarData` holding a `FunctionName` (or whatever that may evolve into)?

sunfishcode:
This fixes #182. I'm not attached to the particular syntax or naming here.

### DIFF
--- a/docs/langref.rst
+++ b/docs/langref.rst
@@ -563,6 +563,16 @@ runtime data structures.
                  variable.
     :result GV: Global variable.
 
+.. inst:: GV = globalsym name
+
+    Declare a global variable at a symbolic address.
+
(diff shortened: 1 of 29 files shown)


sunfishcode:
I've now added legalization support which changes a few things, so I'll submit a new PR.

PR #184, opened 2017-10-30
Add support for symbolic global variables.

### LINKED ISSUE #182: Support for symbolic memory addresses

sunfishcode:
This fixes #182. As it currently stands, it also makes several API changes that baldrdash would probably need to be updated for:
 - rename `FunctionName` to `ExternalName`
 - add a `reloc_globalsym` to any `RelocSink` impls (which could probably just panic since the frontend in that case wouldn't ever generate a globalsym)
 - enable the `allones_funcaddrs` setting

### DIFF
--- a/lib/cretonne/meta/base/settings.py
+++ b/lib/cretonne/meta/base/settings.py
@@ -69,5 +69,13 @@ spiderwasm_prologue_words = NumSetting(
         pushed return address on Intel ISAs.
         """)
 
+#
+# BaldrMonkey requires that not-yet-relocated function addresses be encoded
+# as all-ones bitpatterns.
+#
+allones_funcaddrs = BoolSetting(
+        """
+        Emit not-yet-relocated function addresses as all-ones bit patterns.
+        """)
 
 group.close(globals())
--- a/lib/cretonne/meta/isa/intel/encodings.py
+++ b/lib/cretonne/meta/isa/intel/encodings.py
@@ -2,7 +2,7 @@
 Intel Encodings.
 """
 from __future__ import absolute_import
-from cdsl.predicates import IsUnsignedInt
+from cdsl.predicates import IsUnsignedInt, Not
 from base import instructions as base
 from base.formats import UnaryImm
 from .defs import I32, I64
@@ -11,6 +11,7 @@ from . import settings as cfg
 from . import instructions as x86
 from .legalize import intel_expand
 from base.legalize import narrow, expand
+from base.settings import allones_funcaddrs
 from .settings import use_sse41
 
 try:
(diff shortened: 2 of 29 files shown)


lib/cretonne/meta/base/settings.py
```diff
@@ -69,5 +69,13 @@
         pushed return address on Intel ISAs.
         """)
 
+#
+# BaldrMonkey requires that not-yet-relocated function addresses be encoded
+# as all-ones bitpatterns.
+#
+allones_funcaddrs = BoolSetting(
```
stoklund: Yep, works for me.

lib/cretonne/meta/isa/intel/encodings.py
```diff
@@ -260,8 +261,22 @@ def enc_i32_i64_ld_st(inst, w_bit, recipe, *args, **kwargs):
...
-I64.enc(base.func_addr.i64, *r.fnaddr8.rex(0xb8, w=1))
+I32.enc(base.func_addr.i32, *r.fnaddr4(0xb8),
+        isap=Not(allones_funcaddrs))
+I64.enc(base.func_addr.i64, *r.fnaddr8.rex(0xb8, w=1),
+        isap=Not(allones_funcaddrs))
+
+I32.enc(base.func_addr.i32, *r.allones_fnaddr4(0xb8),
+        isap=allones_funcaddrs)
+I64.enc(base.func_addr.i64, *r.allones_fnaddr8.rex(0xb8, w=1),
+        isap=allones_funcaddrs)
```
stoklund: We generate slightly better encoding tables when you can put the predicates on the recipe instead of on the specific instruction. It doesn't make much of a difference.

stoklund:
I was thinking that if we can shrink `ExternalName` to a 12-byte type without any allocation or destructor, we could have a single `external_addr` instruction which simply contains the `ExternalName` it is referencing instead of going through function or global declarations.

We don't need to make that change as part of this PR, though.

lib/cretonne/meta/isa/intel/encodings.py
sunfishcode: I looked into this, and there appear to be some bugs with predicates on recipes. I think I can fix it, but I'll do it in a separate patch because it's a little complex.

sunfishcode:
I looked into creating an `external_addr` but binemit needs to know whether to call `reloc_func_addr` or `reloc_globalsym_addr`. One option would be to merge those, and pass the `ExternalName` rather than the `FuncRef`/`GlobalVar`, but I haven't yet convinced myself that there wouldn't be use cases needing the `FuncRef`/`GlobalVar`.

lib/cretonne/meta/isa/intel/encodings.py
stoklund: I think @angusholder ran into the same problem in #160. ISTR he had a fix in that PR.

PR #190, opened 2017-11-14
Add a "module" crate, for basic linking of functions and data.

sunfishcode:
This factors out the `Module` type from llvm2cretonne so that it can be used by other projects. Right now it's quite simplistic and just holds onto everything in memory at once.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,6 +18,7 @@ cretonne-reader = { path = "lib/reader" }
 cretonne-frontend = { path = "lib/frontend" }
 cretonne-wasm = { path = "lib/wasm" }
 cretonne-native = { path = "lib/native" }
+cretonne-module = { path = "lib/module" }
 filecheck = { path = "lib/filecheck" }
 docopt = "0.8.0"
 serde = "1.0.8"
--- a/lib/module/Cargo.toml
+++ b/lib/module/Cargo.toml
@@ -1,0 +1,15 @@
+[package]
+name = "cretonne-module"
+version = "0.0.0"
+authors = ["The Cretonne Project Developers"]
+publish = false
+description = "Cretonne layer supporting multiple functions and data objects"
+repository = "https://github.com/stoklund/cretonne"
+license = "Apache-2.0"
+
+[lib]
+name = "cton_module"
+
+[dependencies]
+cretonne = { path = "../cretonne" }
+ordermap = "0.3.0"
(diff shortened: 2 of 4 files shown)


stoklund:
This has a very LLVM flavor to it, which makes sense given the origin.

I am a little wary of putting this in the main Cretonne repo because it is not using the data structures as intended. Cretonne tries to support streaming and multi-threading, so the `il::Function` data structure (really, `cretonne::Context`) is meant to be reused for compiling multiple functions. If people are looking for examples, I don't want them to get the wrong idea (and the testing code is already pretty far off from intended usage).

I think the large IR module that can only be accessed by one thread is one of LLVM's biggest weaknesses, so I am not so happy to duplicate that in the Cretonne repo. I know that you are just trying to get things going with the simplest possible design, by I do worry that if we start out in this direction, it will only get harder to fix it and get back to the multi-threaded, streaming compilation.

The module as a "table of contents" for a translation unit would make sense for me. Could it work without the `functions` vector?

The word "module" is almost meaningless because it is used for so many things. Is there a more specific name we could use?

sunfishcode:
Ok. I'll take a look at reforming it along these lines. 

I don't have any great ideas for alternative names though. How about "CompilationGroup"? "SymbolCoterie"? ;-)

stoklund:
Thanks. Is it a translation unit?

sunfishcode:
"Translation unit" is the C/C++ term, where it's traditional to emit a single ".o" file for each C/C++ translation unit, though the correspondence is weaker with the existence of LTO.

stoklund:
So what is the concept we are trying to model here?

These days with ThinLTO etc, I think an LLVM module really is "stuff that we want to compile with a single thread". That's probably *not* a concept we should translate to Cretonne's world?

sunfishcode:
The present need is "stuff that we want to write into a single .o file", which is the original and still quite popular sense of Module in LLVM.

stoklund:
That makes sense. Perhaps we could name it that way: “object file index”

m4b:
Chiming in randomly, feel free to ignore, what about

* AbstractObject
* ObjectPattern
* Tapestry ;)

pchickey:
Chiming in to say I just spent yesterday lifting the Module type into my codebase by hand, and I'm going to go delete that and import this instead :)

sunfishcode:
Closing this; see #288 for ongoing work on this topic.

PR #193, opened 2017-11-17
Replace FunctionBuilder's Drop impl with a finalize function.

sunfishcode:
This has the advantage of not triggering assertion failures in the event
of abandoning a partially-built function. It has the disadvantage of
requiring users to call finalize() explicitly.

This is a followup to https://github.com/stoklund/cretonne/issues/192. Unfortunately, there isn't an obvious thing to return a state value. I looked at using the `Function` for this, but when used with `Context` which owns its `Function`, it's not possible for the `FunctionBuilder` to temporarily own it so that it can be returned. I'd be happy to implement something else if anyone has any better ideas.

### DIFF
--- a/lib/frontend/src/frontend.rs
+++ b/lib/frontend/src/frontend.rs
@@ -52,6 +52,15 @@ struct Position {
     basic_block: Block,
 }
 
+impl Position {
+    fn beginning() -> Self {
+        Self {
+            ebb: Ebb::new(0),
+            basic_block: Block::new(0),
+        }
+    }
+}
+
 impl<Variable> ILBuilder<Variable>
 where
     Variable: EntityRef,
@@ -213,10 +222,7 @@ where
             func: func,
             srcloc: Default::default(),
             builder: builder,
-            position: Position {
-                ebb: Ebb::new(0),
-                basic_block: Block::new(0),
-            },
+            position: Position::beginning(),
         }
     }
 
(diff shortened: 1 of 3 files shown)


lib/frontend/src/frontend.rs
```diff
@@ -52,6 +52,15 @@ struct Position {
     basic_block: Block,
 }
 
+impl Position {
+    fn beginning() -> Self {
```
stoklund: This function name might be a bit dangerous, if it is supposed to imply the entry block of the function. Cretonne doesn't guarantee or require that `ebb0` is the entry block.

IIRC, this was one of those cases where technically there should be an `Option` somewhere because the position briefly needs to be pointing nowhere, but that would require unwraps everywhere.

Cretonne proper has the same problem with the `Layout::entry_block()` function which currently returns an `Option<Ebb>` that is immediately unwrapped at every use.

lib/frontend/src/frontend.rs
```diff
@@ -53,12 +54,16 @@ struct Position {
...
 
 impl Position {
-    fn beginning() -> Self {
+    fn default() -> Self {
         Self {
-            ebb: Ebb::new(0),
-            basic_block: Block::new(0),
+            ebb: Ebb::reserved_value(),
+            basic_block: Block::reserved_value(),
         }
```
stoklund: Nice. That should get you manny assertions if you try to use it.

sunfishcode: Yeah, except I realized that this is essentially just reimplementing `PackedOption`. So I changed it to use `PackedOption` now.

PR #195, opened 2017-11-21
Treat VmContext as positional when using Native CallConv

tyler:
Previously, on x64 the VmContext was always stored in `r14`. This makes it so that is only true if the CallConv is set to SpiderWASM. Otherwise, it treats the VmContext is a standard positional argument.

This requires threading the CallConv through legalize_args and into ArgAssigner.

More than happy to refactor this however you see fit if you'd like this done differently.

Thanks!

### DIFF
--- a/lib/cretonne/src/abi.rs
+++ b/lib/cretonne/src/abi.rs
@@ -3,7 +3,7 @@
 //! This module provides functions and data structures that are useful for implementing the
 //! `TargetIsa::legalize_signature()` method.
 
-use ir::{ArgumentLoc, AbiParam, ArgumentExtension, Type};
+use ir::{ArgumentLoc, AbiParam, ArgumentExtension, CallConv, Type};
 use std::cmp::Ordering;
 
 /// Legalization action to perform on a single argument or return value when converting a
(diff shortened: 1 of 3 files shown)


lib/cretonne/src/abi.rs
```diff
@@ -81,13 +81,13 @@ impl ValueConversion {
 /// This will be implemented by individual ISAs.
 pub trait ArgAssigner {
     /// Pick an assignment action for function argument (or return value) `arg`.
-    fn assign(&mut self, arg: &AbiParam) -> ArgAction;
+    fn assign(&mut self, arg: &AbiParam, call_conv: CallConv) -> ArgAction;
```
stoklund: Instead of adding a `call_conv` parameter to this trait, you can add a `call_conv` member to the `struct Args` type defined in `intel/abi.rs`. This avoids threading this information through the ISA-independent code.

lib/cretonne/src/isa/intel/abi.rs
```diff
@@ -66,8 +66,7 @@ impl ArgAssigner for Args {
         }
 
         // Handle special-purpose arguments.
-        // TODO: The registers below are for `spiderwasm`. Should we check the calling convention?
-        if ty.is_int() {
+        if ty.is_int() && call_conv == ir::CallConv::SpiderWASM {
```
stoklund: With `call_conv` as an `Args` member, this then becomes `if ty.is_int() && self.call_conv == ir::CallConv::SpiderWASM`.

tyler:
In retrospect, that makes a lot of sense. Especially because it's only needed in the Intel codepath. 

I've updated the PR to reflect that.

PR #199, opened 2017-11-27
Fix wat syntax so wasm tests pass

pchickey:
You will need `wat2wasm` installed on your machine to ensure these tests run.

I also added a few lines of code to make the filetests ignore files starting with `.` so that it doesnt try to run vim .swp files, but perhaps the correct behavior is to just ignore files that dont start with wat or wasm, instead of panicking?

We should install wat2wasm on the CI environment so this doesn't get ignored in the future - I'll open an issue so we can discuss how to do that.

### DIFF
--- a/lib/wasm/tests/testsuite.rs
+++ b/lib/wasm/tests/testsuite.rs
@@ -23,6 +23,15 @@ fn testsuite() {
     let mut paths: Vec<_> = fs::read_dir("../../wasmtests")
         .unwrap()
         .map(|r| r.unwrap())
+        .filter(|p| {
+            // Ignore files starting with `.`, which could be editor temporary files
+            if let Some(stem) = p.path().file_stem() {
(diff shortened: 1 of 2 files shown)


lib/wasm/tests/testsuite.rs
```diff
@@ -23,6 +23,14 @@ fn testsuite() {
     let mut paths: Vec<_> = fs::read_dir("../../wasmtests")
         .unwrap()
         .map(|r| r.unwrap())
+        .filter(|p| {
+            if let Some(stem) = p.path().file_stem() {
+                if let Some(stemstr) = stem.to_str() {
+                    return !stemstr.starts_with(".");
```
sunfishcode: Please add a comment mentioning the motivation.

pchickey: Amended that commit with a comment

sunfishcode:
Thanks!

Issue #187, opened 2017-11-08
Prologue/epilogue doesn't adjust stack pointer


sunfishcode:
Prologues and epilogues don't yet include code to adjust the stack pointer.

For example, running `cton-util compile` on this input:
```
function %bar(f32, f32, f32, f32) -> f32 native {
    sig0 = () -> f32 native
    fn0 = sig0 %foo

ebb0(v0: f32, v1: f32, v2: f32, v3: f32):
    v4 = call fn0()
    v5 = fadd v4, v0
    v6 = fadd v5, v1
    v7 = fadd v6, v2
    v8 = fadd v7, v3
    return v8
}
```
Produces this output:
```
function %bar(f32 [%xmm0], f32 [%xmm1], f32 [%xmm2], f32 [%xmm3]) -> f32 [%xmm0] native {
    ss0 = spill_slot 4, offset -4
    ss1 = spill_slot 4, offset -8
    ss2 = spill_slot 4, offset -12
    ss3 = spill_slot 4, offset -16
    sig0 = () -> f32 [%xmm0] native
    fn0 = sig0 %foo

                                ebb0(v9: f32 [%xmm0], v10: f32 [%xmm1], v11: f32 [%xmm2], v12: f32 [%xmm3]):
[RexMp2fspSib32#57e,ss0]            v0 = spill v9
[RexMp2fspSib32#57e,ss1]            v1 = spill v10
[RexMp2fspSib32#57e,ss2]            v2 = spill v11
[RexMp2fspSib32#57e,ss3]            v3 = spill v12
[Op1call_id#e8,%xmm0]               v4 = call fn0()
[RexMp2ffiSib32#56e,%xmm1]          v13 = fill v0
[RexMp2fa#658,%xmm0]                v5 = fadd v4, v13
[RexMp2ffiSib32#56e,%xmm1]          v14 = fill v1
[RexMp2fa#658,%xmm0]                v6 = fadd v5, v14
[RexMp2ffiSib32#56e,%xmm1]          v15 = fill v2
[RexMp2fa#658,%xmm0]                v7 = fadd v6, v15
[RexMp2ffiSib32#56e,%xmm1]          v16 = fill v3
[RexMp2fa#658,%xmm0]                v8 = fadd v7, v16
[Op1ret#c3]                         return v8
}
```
which is this machine code:
```
0000000000000000 <bar>:
[...]

PR #201, opened 2017-11-28
[WIP] Function prologues and epilogues

tyler:
This is the beginning of attempting to implement what is discussed in #187.

So far this is adding the new instructions and inserting full prologues and epilogues for the Intel ISA. It is quite unfinished and is lacking in proper documentation for the new instructions. However, I thought it was useful to start getting feedback early. 

So, this adds two new Intel-specific instructions (`x86_push`, `x86_pop`) and two ISA-independent instructions (`copy_special`, `adjust_sp_imm`). 

For the prologue it inserts the necessary frame pointer finagling, and pushes all callee-saved registers to the stack. It attempts to insert an epilogue for each return statement, doing the reverse.

The code is still quite rough. It's unclear to me whether I'm using the correct Cretonne abstractions for modifying the functions and ebbs. The new code itself also needs to be refactored significantly.

The boundaries between what belongs where in the meta Python code is a bit vague to me, so I'm eager for feedback on that. I'm also a bit unclear on which things need to be polymorphic and how to make that bulletproof.

Anyway, thanks for looking.


filetests/isa/intel/prologue-epilogue.cton
```diff
@@ -0,0 +1,25 @@
...
+
+    ; ... function body ...
+
+
+    adjust_sp_imm +168 ; bin: 48 81 c4 000000a8
+    [-,%r12] v100 = x86_pop.i64 ; bin: 41 5c
+    [-,%rbx] v101 = x86_pop.i64 ; bin: 40 5b
+    [-,%rbp] v102 = x86_pop.i64 ; bin: 40 5d
+    return v102
+}
```
bjorn3: Missing newline

tyler: whoops!

stoklund:
Sorry for the delay in reviewing this. I’ve been sick, I’ll take a look as soon as I get better.

tyler:
@stoklund Feel better.

I just added a bunch of new commits. This actually generates the prologues and epilogues now and at least appears to be functioning correctly. (For x86-64 anyway.) Requires a bunch more work, but I could use feedback on the direction before I start polishing and improving.

tyler:
As of the latest two commits, this should be working on Intel in both 64-bit and 32-bit mode. I hope.

lib/cretonne/meta/isa/intel/encodings.py
```diff
@@ -227,6 +227,18 @@ def enc_i32_i64_ld_st(inst, w_bit, recipe, *args, **kwargs):
 enc_both(base.fill.b1, r.fiSib32, 0x8b)
 enc_both(base.regfill.b1, r.rfi32, 0x8b)
 
+# Push and Pop
+enc_i32_i64(x86.push, r.pushq, 0x50)
+enc_i32_i64(x86.pop, r.popq, 0x58)
+
+# Copy Special
+enc_i64(base.copy_special, r.copysp, 0x89)
+I32.enc(base.copy_special, *r.copysp(0x89))
```
sunfishcode: I gave this patch a quick try, and noticed one issue: on x64, the "mov %rsp,%rbp" is getting encoded as "0x40 0x89 0xe5", which is actually a 32-bit mov. The REX prefix is 0x40, which is missing the 'W' bit, which would make it 0x48, which is what's needed to make the operation 64-bit.

I haven't tested it, but one way to fix this might be adding a `w_bit` parameter like the one `enc_i32_i64_ld_st` has, to `enc_i64`, and then making use of it for the Copy Special encoding, though I don't know if that's the cleanest way to do it.

tyler:
I literally just made a commit to do exactly this. Will push it up when I
get back to my computer.

lib/cretonne/src/isa/intel/mod.rs
```diff
@@ -111,4 +117,145 @@ impl TargetIsa for Isa {
...
+        // Insert prologue
+        {
+            let mut pos = EncCursor::new(func, self).at_first_insertion_point(entry_ebb);
+            pos.ins().x86_push(fp);
+            pos.ins().copy_special(
+                RU::rsp as RegUnit,
+                RU::rbp as RegUnit,
+            );
+            if local_stack_size > 0 {
+                pos.ins().adjust_sp_imm(-(local_stack_size as i32));
```
sunfishcode: This should return a `CtonError::ImplLimitExceeded` if `local_stack_size` is greater than `i32::MAX as u32`, so that we don't overflow.

lib/cretonne/meta/isa/intel/encodings.py
stoklund: The direct syntax for an encoding with a REX prefix is:

```python
I64.enc(base.copy_special, *r.copysp.rex(0x89, w=1))
```

lib/cretonne/src/isa/intel/mod.rs
stoklund: The stack layout algorithm enforces a maximum 2 GB stack frame size for its own sanity. This should probably be better documented.

stoklund:
Thanks for working on this, Tyler, and sorry again for my reviewing this so late. Let me address your high-level questions before I dive into the code.

You are mostly using the Cretonne interfaces for modifying IR as intended. A few tips:
- The `dfg.add_ebb_param()` method returns the value it just added, so you don't need to dig it out after the call.
- It's almost never necessary to pattern-match `InstructionData`. You can use `dfg[inst].opcode() == Opcode::Return`, for example.
- Using `dfg[inst].opcode().is_return()` is more future-proof in case we get an alternative return opcode (think returning by tail-calling another function).
- I tend to reserve the layout iterators (`layout.ebbs()` and `layout.ebb_insts()`) for analysis functions only. Prefer a cursor instead when mutating a function during iteration. See [unreachable_code.rs](https://github.com/stoklund/cretonne/blob/master/lib/cretonne/src/unreachable_code.rs) for an example.

Your changes to the Python meta-code look pretty good. You've put things in the right place. The distinction between which instructions should go in the shared base instruction set and which should go in the Intel-specific instruction set is a bit tricky. The rule I use to put instructions in the shared base set go something like:

- The instruction must be generally useful to more than one ISA, and
- The instruction must be feasible to legalize with a reasonable macro expansion on *all* ISAs.

The push and pop instructions don't meet those criteria because a) most ISAs want to do their own special thing in the prologue, and b) arm64 requires the stack pointer to be 16-byte aligned at all times, so you must push two registers at a time.

You are using polymorphism correctly. For the push and pop instructions, it could make sense to use the existing `iWord` type variable to restrict the range to i32-i64, but it's not a big deal. The `EncCursor` you're using will panic if you try to insert an instruction that doesn't have an encoding.

Generally, in the base instruction set, type variables have the largest range that makes sense for the instruction's semantics, while the ISA-specific instruction sets tend to use a range that matches the ISA capabilities.

The `copy_special` instruction can't be polymorphic because there are no SSA values around, and Cretonne doesn't distinguish between `%ebp` and `%rbp` in its `RegUnit` type either. This means that we don't have a way of distinguishing between 32-bit and 64-bit copies when defining encodings. This isn't a problem in practice since special-purpose registers are almost always pointer-sized integer registers. So we should just define (and document) this instruction as copying between pointer-sized integer registers, and any ISA that has more special needs will have to define a new ISA-specific instruction to handle that.

stoklund:
A note about vectors.

Cretonne works hard to reduce malloc/free churn to improve performance. Even though Rust links with jemalloc which is "fast", not calling malloc/free is still a lot faster than calling malloc/free. Most of the data structures are based on `Vec`s that are ultimately owned by the `cretonne::Context` which means that the memory allocated will be reused for compiling the next function, even when calling `Context::clear()`. This is important because some modules can contain large numbers of tiny functions, so even once-per-function allocations can become quite expensive.

This also means that we try to avoid using transient vectors as much as possible. This can cause some friction with Rust because sticking stuff in a vector is an easy way of resolving a fight with the borrow checker. Here's some techniques I've found useful to avoid temporary vectors:

- When returning multiple values from a function, prefer slices, arrays, tuples, or an iterator when possible. This is usually more work, although your `callee_saved_registers()` function can be changed to return `&'static [RU]` pretty easily.
- When a temporary vector is required for an algorithm, try to stick it in a data structure ultimately owned by the context so the memory allocation can be recycled. The `DominatorTree::stack` member is an example. You can't use this technique in `prologue_epilogue()` because you don't have access to any data structure owned by the context except for the function, and we're not sticking scratch space vectors in there.
- Use functional decomposition. For example, you have code that needs to do the same thing to the CSRs and the frame pointer, so you stick everything in a vector and work on that. Instead, break out the shared operation to a new function and call that function for the frame pointer, then for each CSR. This avoids the allocation.

LLVM has a `SmallVector` type which uses a fixed buffer on the stack for its initial allocation, and only falls back to malloc/free when the initial capacity is exceeded. This is usually all you need to dramatically reduce malloc/free churn. There is a [smallvec crate](https://crates.io/crates/smallvec) which implements the same thing for Rust, but when I benchmarked it, it was slower than `Vec` in every case.

I have been working on an implementation that is faster than `Vec` at the cost of some elegance. It may be part of Cretonne in the future.

lib/cretonne/src/ir/instructions.rs
```diff
@@ -235,6 +236,12 @@ pub enum InstructionData {
         src: RegUnit,
         dst: RegUnit,
     },
+    CopySpecial {
```
stoklund: Whenever I add a new instruction format, I also add a test case to `filetests/parser/tiny.cton` which then covers both the printing and parsing of the new format.

lib/cretonne/src/ir/instructions.rs
```diff
@@ -120,6 +120,7 @@ pub enum InstructionData {
     },
     Ternary { opcode: Opcode, args: [Value; 3] },
     MultiAry { opcode: Opcode, args: ValueList },
+    NullAry { opcode: Opcode },
```
stoklund: Ha, it's no that long ago I deleted this format because there were no uses left, but it's the right thing for a pop.

lib/cretonne/src/isa/intel/abi.rs
```diff
@@ -153,3 +153,11 @@ pub fn allocatable_registers(
 
     regs
 }
+
+pub fn callee_saved_registers(flags: &shared_settings::Flags) -> Vec<RU> {
```
stoklund: As I mentioned, this can pretty easily return `&'static [RU]`.

lib/cretonne/src/isa/intel/mod.rs
```diff
@@ -111,4 +117,149 @@ impl TargetIsa for Isa {
     fn reloc_names(&self) -> &'static [&'static str] {
         &binemit::RELOC_NAMES
     }
+
+    fn prologue_epilogue(&self, func: &mut ir::Function) -> result::CtonResult {
```
stoklund: I think the whole body of this method should be moved into `intel/abi.rs` to keep the top-level module clean. See `legalize_signature()`.

lib/cretonne/src/isa/intel/mod.rs
```diff
@@ -111,4 +117,149 @@ impl TargetIsa for Isa {
...
+            ir::types::I64
+        } else {
+            ir::types::I32
+        };
+        let csrs = abi::callee_saved_registers(&self.shared_flags);
+
+        let mut csr_stack_size = word_size; // Size of RBP to start with
+        for _reg in &csrs {
+            csr_stack_size += word_size;
+        }
```
stoklund: `(csrs.len() + 1) * word_size` ;-)

You could consider including the frame pointer in the set returned by `callee_saved_registers`. We don't support frame pointer elimination at the moment, but when we do, `%rbp` will behave exactly like any other CSR.

lib/cretonne/src/isa/intel/mod.rs
```diff
@@ -111,4 +117,149 @@ impl TargetIsa for Isa {
...
+
+        let stack_offset = -(csr_stack_size as i32);
+        let slot = ir::StackSlotData {
+            kind: ir::StackSlotKind::IncomingArg,
+            size: csr_stack_size,
+            offset: stack_offset,
+        };
+        func.create_stack_slot(slot);
```
stoklund: This is good, but I'd add a comment explaining what is going on.

Stack frame math is hard, but I think the offset is wrong here. It should be relative to the caller's outgoing stack pointer, not your incoming stack pointer. The difference is the return address pushed by the call instruction itself:

```
outgoing sp --> xxxx in caller's frame
incoming sp --> return address
                pushed %rbp
                ...
                last pushed CSR
```

lib/cretonne/src/isa/intel/mod.rs
```diff
@@ -111,4 +117,149 @@ impl TargetIsa for Isa {
...
+
+        let total_stack_size = layout_stack(&mut func.stack_slots, word_size)?;
+        let local_stack_size = total_stack_size - csr_stack_size;
+
+        // Build up list of args, which we'll append forwards to the params and
+        // backwards to the returns.
+        let mut csr_args = Vec::new();
```
stoklund: Most of the code below here, I would extract to a function that does everything you need for one CSR. It avoids the memory allocation.

lib/cretonne/src/isa/intel/mod.rs
```diff
@@ -111,4 +117,149 @@ impl TargetIsa for Isa {
...
+                ir::ArgumentPurpose::CalleeSaved,
+                *reg as RegUnit,
+            ));
+        }
+
+        for csr_arg in &csr_args {
+            func.signature.params.push(*csr_arg);
+        }
+        for csr_arg in csr_args.iter().rev() {
+            func.signature.returns.push(*csr_arg);
```
stoklund: I see what you're doing here, but to be clear, the order of these arguments only matter relative to the order of arguments on the return instruction.

lib/cretonne/src/isa/intel/mod.rs
```diff
@@ -111,4 +117,149 @@ impl TargetIsa for Isa {
...
+        }
+
+        // Append param to entry EBB
+        let entry_ebb = func.layout.entry_block().expect("missing entry block");
+        func.dfg.append_ebb_param(entry_ebb, csr_type);
+
+        // Find our frame pointer parameter Value
+        let fp = func.special_param(ir::ArgumentPurpose::FramePointer)
+            .expect("missing frame pointer");
```
stoklund: As I mentioned, `append_ebb_param()` returns the parameter value it added, so this is not necessary.

lib/cretonne/src/isa/intel/mod.rs
```diff
@@ -111,4 +117,149 @@ impl TargetIsa for Isa {
...
+        func.locations[fp] = ir::ValueLoc::Reg(RU::rbp as RegUnit);
+
+        let mut csr_vals = Vec::new();
+        for reg in &csrs {
+            // Append param to entry EBB
+            func.dfg.append_ebb_param(entry_ebb, csr_type);
+
+            let csr_arg = func.dfg.ebb_params(entry_ebb).last().expect(
+                "no last argument",
+            );
```
stoklund: Same thing

lib/cretonne/src/isa/intel/mod.rs
```diff
@@ -111,4 +117,149 @@ impl TargetIsa for Isa {
...
+
+            for csr_arg in csr_vals {
+                pos.ins().x86_push(csr_arg);
+            }
+        }
+
+        // Find all 'return' instructions
+        let mut return_insts = Vec::new();
+        for ebb in func.layout.ebbs() {
+            for inst in func.layout.ebb_insts(ebb) {
```
stoklund: We don't have any conditional return instructions, and I don't think Intel ever will, so return instructions are always going to be [terminator instructions](http://cretonne.readthedocs.io/en/latest/langref.html#term-terminator-instruction). This means they are guaranteed to be at the end of an EBB, so you can just check the last instruction.

For an algorithm like this, I would use a cursor to navigate the EBBs and  pass the cursor instead of the function into `insert_epilogue` for mutation.

lib/cretonne/src/isa/intel/mod.rs
```diff
@@ -111,4 +117,149 @@ impl TargetIsa for Isa {
...
+    ) {
+        let mut return_values = Vec::new();
+
+        let mut pos = EncCursor::new(func, self).at_inst(inst);
+        if stack_size > 0 {
+            pos.ins().adjust_sp_imm(stack_size);
+        }
+        for reg in csrs.iter().rev() {
+            let csr_ret = pos.ins().x86_pop(csr_type);
+            return_values.push((csr_ret, *reg));
```
stoklund: If you need to, you can insert instructions backwards by using `pos.prev_inst()` after inserting each pop. This might come up if you are trying to get rid of vectors in these algorithms.

tyler:
Great! Thank you for the feedback. I'm going to go one by one and push commits to address each of them.

re: avoiding temporary Vecs, my plan was to get rid of all of them while refactoring. Your description of doing this via functional decomposition closely matches what I was thinking. Will get to that soon. I made a checklist for the feedback for my own benefit:

- [ ] Refactor `prologue_epilogue`
  - [ ] Functionally decompose for clarity and to get rid of Vecs
  - [ ] Move the whole thing into `abi.rs`
  - [ ] Use ret value of calls to `add_ebb_param`
  - [ ] Use `dfg[inst].opcode().is_return()` instead of pattern-matching
  - [ ] Switch to using a cursor to navigate functions
    - [ ] Check the last inst in ebbs rather than iterating
  - [ ] `(csrs.len() + 1) * word_size`
  - [ ] Add frame pointer
  - [ ] Evaluate and fix the stack offset
  - [ ] Use `pos.prev_inst()` to help eliminate temporary Vec
- [ ] `callee_saved_registers` func should return a `&'static [RU]`
- [ ] Fix up new instructions
  - [ ] Use `iWord` type var in instruction definition
  - [ ] Switch `adjust_sp_imm` to use an `imm64`
  - [ ] Eliminate `AdjustSpImm` format
  - [ ] copy_special should specifically mention pointer-sized registers
  - [ ] Document new instructions
- [ ] Testing
  - [ ] Read testing.html and add small "compile" test
  - [ ] Add new instruction formats to `filetests/parser/tiny.cton`

lib/cretonne/src/isa/intel/mod.rs
tyler: oh jeez, this one is embarrassing

stoklund:
I forgot to add: Please add encoding tests for the new instructions to `filetests/isa/intel/binary{32,64}.cton`. I find that with Intel encodings I often get something backwards (the AT&T syntax doesn't help either), so I write the assembly instructions I expect in the `asm:` comments and run it through an assembler with the `sed` command at the top if the file.

If you don't have `llvm-mc`, there is probably some `gas` listing mode that does the same. I haven't tried this.

On macOS with Homebrew, the "llvm" package does install `llvm-mc`, but not in PATH. Mine is in `/usr/local/Cellar/llvm/5.0.0/bin`.

lib/cretonne/meta/base/instructions.py
```diff
@@ -537,6 +537,21 @@
...
         other_side_effects=True)
 
+copy_special = Instruction(
+        'copy_special', r"""
+        Copies a value from one special register to another. e.g. rbp -> rsp.
+        """,
+        ins=(src, dst),
+        other_side_effects=True)
+
+Offset = Operand('Offset', offset32, 'Offset from current stack pointer')
```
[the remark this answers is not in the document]
tyler: I'm slightly confused about this one. I had thought about this and decided against it because the value should never be larger than 32-bit, and Intel's ADD operation doesn't support an imm64. Which would mean the generated code would need to mov the imm64 to a register and then add it, which seemed silly to me.

Should I maybe use an imm32 instead?

stoklund:
> On Dec 1, 2017, at 17:06, Tyler McMullen <notifications@github.com> wrote:
> 
> I'm slightly confused about this one. I had thought about this and decided against it because the value should never be larger than 32-bit, and Intel's ADD operation doesn't support an imm64. Which would mean the generated code would need to mov the imm64 to a register and then add it, which seemed silly to me.

The different kinds of immediate operands are more about choosing the in-memory representation of the instructions, and to some extent the textual representation too. So we tend to use imm64 if it fits because it leads to fewer instruction formats overall. 

You don’t need to provide encodings or expansions for larger offsets than what you need. You can supply an instp=... predicate to your recipe to make sure it only gets immediate that are in range. Look for examples using an IsSignedInt predicate.

If we later need to use the instruction in a more general context, we can add legalization patterns that will help all ISAs deal with large immediates.

tyler:
Ah! Ok that makes sense. I had not noticed the `instp` predicates. Thank you.

filetests/isa/intel/prologue-epilogue.cton
```diff
@@ -0,0 +1,25 @@
+test legalizer
```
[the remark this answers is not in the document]
tyler: As far as I can see the `test compile` directive doesn't actually allow any filecheck directives. Am I missing something? Happy to add them if necessary.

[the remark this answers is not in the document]
stoklund: Oh, you are right. I think it should. I’ll add that.

tyler:
Thanks for adding the new filecheck directives. Just added a test that uses it.

I still need to re-examine the stack math, but otherwise I think I've covered most of the issues you mentioned.

lib/cretonne/src/isa/intel/mod.rs
tyler: I am struggling with this a bit. If I don't put in a stack slot, but I do have local variables, when I run it through Cretonne the offset of that `local` slot is set to 0. Which would seem, to me, to imply that the `incoming_arg` stack slot I make here should have an offset of 0, and I should let `layout_stack` move the locals to `-csr_stack_size`.

But I feel like I'm missing something here, because I see the comment in `stacklayout.rs` which says:
```
    /// Offset of stack slot relative to the stack pointer in the caller.
    ///
    /// On Intel ISAs, the base address is the stack pointer *before* the return address was
    /// pushed. On RISC ISAs, the base address is the value of the stack pointer on entry to the
    /// function.
```

And that would seem to me, to imply that the `locals` slot (assuming no `incoming_arg` slot) should always start at -8. What am I missing?

stoklund: If you don't add any stack slots, the offset of your local slot should be the negative of its size. Is your local slot zero-sized?

The stack layout algorithm does not know that Intel calls automatically save the return address at incoming offset -8, so it is your responsibility to tell it by adding an incoming stack slot at that offset. I think you can add a single incoming stack slot with a size that covers return address + saved fp + saved csrs. It's offset should be the negative of its size, which means it ends exactly at offset 0 where the caller has placed outgoing arguments.

Note that the stack layout algorithm assigns descending offsets (going towards negative infinity), so the size of a local stack slot will affect which offset it gets.

But the offset of a stack slot refers to the lowest address in the stack slot, just like you would do with a pointer to a struct.

stoklund: If you have a single 8-byte local variable, this is what I would expect if you don't add any incoming stack slots:

```
    ss0 = local 8, offset -8
```

This is wrong because that local actually occupies the same bytes as the return address. If you mark these bytes as reserved by adding an incoming stack slot, I would expect you get this:

```
    ss0 = local 8, offset -16
    ss1 = incoming 8, offset -8
```

If you want to reserve space for the return address, a pushed fp and two CSRs, you should get:

```
    ss0 = local 8, offset -40
    ss1 = incoming 32, offset -32
```

tyler: Ok! I think I finally get it. Thanks for your patience. Latest commit should take care of it.

filetests/isa/intel/prologue-epilogue.cton
```diff
@@ -1,25 +1,28 @@
...
+; nextln:     x86_push v4
+; nextln:     x86_push v5
+; nextln:     v11 = x86_pop.i64
+; nextln:     v10 = x86_pop.i64
+; nextln:     v9 = x86_pop.i64
+; nextln:     v8 = x86_pop.i64
+; nextln:     v7 = x86_pop.i64
+; nextln:     v6 = x86_pop.i64
+; nextln:     return v6, v7, v8, v9, v10, v11
+; nextln: }
```
bjorn3: Missing trailing newline

tyler:
> Two details: ...

Done and done. :)

tyler:
I was actually wondering if I should take a stab at restricting the set of callee-saved registers that we actually push to the set that is used in the function in this PR. I can also just open a new PR for that once this merges.

There's also the issue of making the register allocator aware of callee-saved registers. I started looking into that and it seemed, conceptually, fairly straightforward. But I'm not the register allocation expert here. :)

stoklund:
I think we should land this as is, unless you have more pending changes.

The dynamic set of callee-saved registers is involved enough that I think we should use a separate PR.

tyler:
All my changes are in. Ready when you are. 👍

PR #206, opened 2017-12-07
Position-Independent Code

pchickey:
This PR adds support for PIC code generation to the Intel 64-bit isa.

* A boolean setting "is_pic" is added to the base settings.
* The relocation types required for PIC are added to the Reloc enum.
* Added the appropriate recipes and encodings to the intel isa.
* Added a filetest for testing the encoding of the instructions added.

### DIFF
--- a/filetests/isa/intel/binary64-pic.cton
+++ b/filetests/isa/intel/binary64-pic.cton
@@ -1,0 +1,54 @@
+; binary emission of 64-bit code.
+test binemit
+set is_64bit
+set is_compressed
+set is_pic
+isa intel haswell
+
+; The binary encodings can be verified with the command:
+;
+;   sed -ne 's/^ *; asm: *//p' filetests/isa/intel/binary64-pic.cton | llvm-mc -show-encoding -triple=x86_64
+;
+
+; Tests for i64 instructions.
+function %I64() {
+    fn0 = function %foo()
+    sig0 = ()
+
+    gv0 = globalsym %some_gv
+
+    ; Use incoming_arg stack slots because they won't be relocated by the frame
+    ; layout.
+    ss0 = incoming_arg 8, offset 0
+    ss1 = incoming_arg 1024, offset -1024
+    ss2 = incoming_arg 1024, offset -2048
+    ss3 = incoming_arg 8, offset -2056
+
+ebb0:
+
+    ; asm: call foo@PLT
+    call fn0()                                  ; bin: e8 PLTRel4(%foo) 00000000
+
+    ; asm: mov 0x0(%rip), %rax
+    [-,%rax]            v0 = func_addr.i64 fn0        ; bin: 48 8b 05 GotPCRel4(%foo) 00000000
+    ; asm: mov 0x0(%rip), %rsi
+    [-,%rsi]            v1 = func_addr.i64 fn0        ; bin: 48 8b 35 GotPCRel4(%foo) 00000000
+    ; asm: mov 0x0(%rip), %r10
+    [-,%r10]            v2 = func_addr.i64 fn0        ; bin: 4c 8b 15 GotPCRel4(%foo) 00000000
+
+    ; asm: call *%rax
+    call_indirect sig0, v0()                  ; bin: ff d0
+    ; asm: call *%rsi
+    call_indirect sig0, v1()                  ; bin: ff d6
+    ; asm: call *%r10
+    call_indirect sig0, v2()                  ; bin: 41 ff d2
+
+    ; asm: mov 0x0(%rip), %rcx
+    [-,%rcx]            v3 = globalsym_addr.i64 gv0    ; bin: 48 8b 0d GotPCRel4(%some_gv) 00000000
+    ; asm: mov 0x0(%rip), %rsi
+    [-,%rsi]            v4 = globalsym_addr.i64 gv0    ; bin: 48 8b 35 GotPCRel4(%some_gv) 00000000
+    ; asm: mov 0x0(%rip), %r10
+    [-,%r10]            v5 = globalsym_addr.i64 gv0    ; bin: 4c 8b 15 GotPCRel4(%some_gv) 00000000
+
+    return
+}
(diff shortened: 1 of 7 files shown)


lib/cretonne/meta/isa/intel/legalize.py
```diff
@@ -53,6 +53,15 @@
                 (a, dead) << x86.sdivmodx(x, xhi, y)
             ))
 
+
+#intel_expand.legalize(
+#        a << insts.func_addr.bind(i64)(x),
+#        Rtl(
+#            # First, get the pointer-to-x using a 32-bit
+#            # rip-relative mov; and reloc with REX_GOTPCRELX 
+#            # Second, get x from that pointer (ordinary load)
```
sunfishcode: This sounds right, except for the second load. The GOT holds the address of x, which is what `func_addr` needs to compute.

The other thing to watch out for here is the difference between `R_X86_64_GOTPCREL` and `R_X86_64_GOTPCRELX`. The latter is an optimized variant for the case where you know you're not using a REX prefix (and `R_X86_64_REX_GOTPCRELX` is for when you know you *are* using a REX prefix). The easier thing to do here would be to just use plain `R_X86_64_GOTPCREL` and leave a note to revisit this when desired.

tyler: Hey there. I'm not totally sure I follow what you're saying re: the second load. 

```
$ gcc -O0 -fPIC -c pic2.c -o pic2.o
$ cat pic2.c
extern void (*foo)();

int main() {
	void (*bar) = foo;
	return 0;
}

$ objdump -dx pic2.o

... snip ... 

Disassembly of section .text:

0000000000000000 <main>:
   0:	55                   	push   %rbp
   1:	48 89 e5             	mov    %rsp,%rbp
   4:	48 8b 05 00 00 00 00 	mov    0x0(%rip),%rax        # b <main+0xb>
			7: R_X86_64_GOTPCREL	foo-0x4
   b:	48 8b 00             	mov    (%rax),%rax
   e:	48 89 45 f8          	mov    %rax,-0x8(%rbp)
  12:	b8 00 00 00 00       	mov    $0x0,%eax
  17:	5d                   	pop    %rbp
  18:	c3                   	retq
```

It looks to me like these two lines imply a two-step process here when loading a pointer from the GOT:
```
   4:	48 8b 05 00 00 00 00 	mov    0x0(%rip),%rax        # b <main+0xb>
			7: R_X86_64_GOTPCREL	foo-0x4
   b:	48 8b 00             	mov    (%rax),%rax
```

sunfishcode: Compare that with what you get when you change this:
```c
extern void (*foo)();
```
to this:
```
extern void foo();
```
In the first form, the GOT load gets the address of foo, which is a global variable holding the address of some function. In the second, the GOT load gets the address of foo, which is a function. The second is what `func_addr` does.

tyler: Oh jeez. Thank you, @sunfishcode.

pchickey:
We hacked up an encoding:
https://github.com/stoklund/cretonne/pull/206/commits/e778cee616c3562b15bed72879454c03107d4037#diff-a2fd9639f31da67c58bb418a70792ea4R580
To get the rip-relative stuff to encode we put some "magic constants" into the modrm_rm and rex2 functions, there's probably a cleaner way to do that.

With that, we have the filetest passing for just the func_addr portions, which is a start.

stoklund:
This direction looks good to me. 

FWIW, I just fixed a bug in `gen_encoding.py` so you can now use `isap=` predicates on recipes as well as encodings, should you need that.

tyler:
Hey there, @stoklund. This appears to be working in our tests. 

I started looking into how I would implement this for 32-bit mode and it appears to be a major pain. I'm wondering whether PIC for 32-bit is a necessity. We can do it, if necessary, but I'd definitely prefer not to.

We need to clean up the abuses of `modrm_rm` and `rex2` in the recipes. As well as some kind of documentation for the PIC setting. But otherwise, I'm wondering what else you'd like to see in this PR to make it mergeable.

Thanks!

tyler:
Also, I'm not sure if we've missed any instructions that are not PIC-safe. As far as I could see, it was only the gvaddr and fnaddr, but there might be others that I don't know about.

tyler:
Ah, I see that we need to adjust the `call` instruction as well. Although I believe the only difference is the relocation type.

stoklund:
Thanks for working on this! This week is Mozilla's all-hands meeting in Austin, so @sunfishcode and I may be a bit less responsive than usual.

I don't think you need to worry about 32-bit PIC code. It is indeed a pain, and there are other more serious things missing before we can compile for 32-bit targets. We will get to it eventually, but you shouldn't let it block you.

One comment for this PR: I am not so happy about exposing the `isa::{arm32,arm64,intel,riscv}` sub-modules as public API because they contain a lot of details are are likely to change, and they can be enabled and disabled individually with feature flags. OTOH, I agree that the `Reloc` type exposed in the `CodeSink` trait is not very user-friendly, and we should fix it by exposing an enum to clients.

I think that instead of having a per-ISA enum with relocations, we should just have one big enum containing enough relocations for all ISAs. I think that all of the relocations we have defined for the Intel ISA could be useful for other ISAs too, and even if we have relocations that only make sense for one ISA, they can still live in one big enum.

So instead of exposing the `intel::RelocKind`, we should:

- Replace the global `binemit::Reloc` struct type with an enum that contains enough relocations for everybody. Since the other ISAs are not fully developed yet, I think this amounts to adding a `Call` variant to the variants that are currently defined for Intel.
- Drop the `TargetIsa::reloc_names()` method and simply have a `Display` implementation for the new `Reloc` enum. Compare to `impl Display for IntCC` for example.

Does that make sense? It might be better to open a new PR with this change, to keep things separate.

pchickey:
Your suggestion about the relocation names makes sense and is pretty straightforward. I guess I hit friction decoding Reloc because I'm the first user to give it a try :). I will make that change as a separate PR.

pchickey:
Thanks for merging the reloc refactor so quickly!

I did a big cleanup of our branch, squashing it down to a small number of independent commits. I also rewrote the PR description.

I believe the only thing left to do is fixup the magic constants in the recipes, which we'll work on this afternoon.

pchickey:
I think this is ready to go now

PR #210, opened 2017-12-14
Align IntelGOTPCRel4 with R_X86_64_GOTPCREL

sunfishcode:
The new IntelGOTPCRel4 directly corresponds to R_X86_64_GOTPCREL, except that it also implies an implicit `- 4` to account for the difference between the end of an instruction (where the PC is considered to be in PC-relative on intel) and the immediate field. That's the right thing to do for faerie right now, as faerie currently adds the -4 [itself](https://github.com/m4b/faerie/blob/bfb3d90443efec6773aa39a93c366d2e23489bf1/src/elf.rs#L480), however it's subtle and makes Cretonne's code less consistent with x86-64 ABI documentation.

I think it makes sense to add an addend field to `reloc_external`, and use it to move the responsibility for adding addends into Cretonne.

This change breaks cretonne+faerie users at present, so we shouldn't merge this PR yet. Faerie's relocation API is [still evolving](https://github.com/m4b/faerie/issues/25#issuecomment-351601310) and I'm optimistic we can harmonize everything.

### DIFF
--- a/src/filetest/binemit.rs
+++ b/src/filetest/binemit.rs
@@ -72,12 +72,28 @@ impl binemit::CodeSink for TextSink {
         write!(self.text, "{}({}) ", reloc, ebb_offset).unwrap();
     }
 
-    fn reloc_external(&mut self, reloc: binemit::Reloc, name: &ir::ExternalName) {
+    fn reloc_external(
+        &mut self,
+        reloc: binemit::Reloc,
+        name: &ir::ExternalName,
+        addend: binemit::Addend,
+    ) {
         write!(
             self.text,
-            "{}({}) ",
+            "{}({}",
             reloc,
             name,
+        ).unwrap();
+        if addend != 0 {
+            write!(
+                self.text,
+                "{:+}",
+                addend,
+            ).unwrap();
+        }
+        write!(
+            self.text,
+            ") ",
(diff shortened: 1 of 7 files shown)


src/filetest/binemit.rs
```diff
@@ -72,12 +72,18 @@ impl binemit::CodeSink for TextSink {
...
+    fn reloc_external(
+        &mut self,
+        reloc: binemit::Reloc,
+        name: &ir::ExternalName,
+        addend: binemit::Addend,
+    ) {
         write!(
             self.text,
-            "{}({}) ",
+            "{}({}{:+}) ",
```
stoklund: Could we print the `Addend`  the same way we print an `Offset32`? That is, print nothing rather than `+0` for a null addend.

sunfishcode: Sure. I updated the patch.

sunfishcode:
@pchickey Ok, I think I've convinced myself that this PR makes sense on its own. Do you agree? If so, be advised that you'll need to (get to ;-)) remove the `- 4`'s on your side.

pchickey:
Looks great to me! Thanks for asking :)

sunfishcode:
Great! Let's do this.

PR #213, opened 2018-01-08
Fix static heap oob traps

tyler:
During the expansion of a static heap lookup Cretonne checks whether an address can be statically determined to be out of bounds, and emits an unconditional trap in place of the lookup. This is great, but the problem is that `trap` is a terminator instruction and now falls into the middle of an Ebb. And so compilation fails with: 
```
a terminator instruction was encountered before the end of ebb11
```

This patch splits the Ebb after emitting the trap. However, this breaks cfg and the domtree and probably everything else too. So, in addition, we also recompute the cfg, the domtree, and run the unreachable code elimination pass as well.

I'm sure there's a more efficient way to handle this, but it was not clear at first glance. Happy to change this to make it less brute-force.

Thanks!

### DIFF
--- a/filetests/isa/intel/legalize-memory.cton
+++ b/filetests/isa/intel/legalize-memory.cton
@@ -65,6 +65,27 @@ ebb0(v0: i32, v999: i64):
     return v4
 }
 
+function %staticheap_static_oob_sm64(i32, i64 vmctx) -> f32 spiderwasm {
+    gv0 = vmctx+64
+    heap0 = static gv0, min 0x1000, bound 0x1000_0000, guard 0x8000_0000
+
+ebb0(v0: i32, v999: i64):
+    ; Everything after the obviously OOB access should be eliminated, leaving
+    ; the `trap heap_oob` instruction as the terminator of the Ebb and moving
+    ; the remainder of the instructions into an inaccessible Ebb.
+    ; check: $ebb0(
+    ; nextln:     trap heap_oob
+    ; check: ebb1:
+    ; nextln:     v2 = iconst.i64 0
+    ; nextln:     v3 = load.f32 v2+16
+    ; nextln:     return v3
+    ; nextln: }
+    v1 = heap_addr.i64 heap0, v0, 0x1000_0001
+    v2 = load.f32 v1+16
+    return v2
+}
+
+
 ; SpiderMonkey VM-style static 4+2 GB heap.
 ; Offsets >= 2 GB do require a boundscheck.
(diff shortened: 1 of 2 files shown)


stoklund:
Nice catch! It seems I didn't write enough test cases for the heap legalization code.

I think you have the right idea: Split the EBB and let a later optimization delete the unreachable code. As I recall, not having this special case caused a bunch of trouble with integer arithmetic overflow in the bounds checks instead.

The legalizer should maintain a valid control flow graph while it is running. Some other code in `legalizer/split.rs` actually depends on it. This is pretty easy to do: Just call `cfg.recompute_ebb()` for those EBBs that have been changed. The `expand_heap_addr()` function already has a `_cfg` argument with the CFG that needs to be kept up to date.

When you keep the CFG up to date, the changes to `context.rs` should not be necessary since `context::compile()` already will recompute the domtree and eliminate unreachable code before moving on to register allocation.

tyler:
I made those changes, however, removing the `eliminate_unreachable_code()` call lands me back on this error: 
```
uses value arg from non-dominating ebb0
```

Presumably, this is because the second half of the Ebb that was split has no incoming edges. If I add back in the `compute_domtree` and `eliminate_unreachable_code` calls, this works.

sunfishcode:
Off the wall idea: what if, instead of `trap`, static heap oobs were replaced by `trapz` with a constant zero input? That'd still eliminate the tricky arithmetic, but it wouldn't be a terminator, so you wouldn't need to touch the CFG or domtree. It's slightly less tidy, but in theory static heap oobs should be very rare in practice.

stoklund:
> On Jan 8, 2018, at 17:45, Tyler McMullen <notifications@github.com> wrote:
> 
> I made those changes, however, removing the eliminate_unreachable_code() call lands me back on this error:
> 
> uses value arg from non-dominating ebb0
> Presumably, this is because the second half of the Ebb that was split has no incoming edges.
> 
The verifier is not supposed to check dominance in unreachable code because dominator trees and unreachable code together is generally a mess.

Looking at the verifier code, it looks like the branch that handles EBB parameter values is checking `is_reachable` for the wrong EBB. It should check the EBB using the value for reachability, not the one defining the value.

If you fix that verifier bug, does it work then?

stoklund:
I fixed the unreachable code verifier bug since it showed up in another context too.

tyler:
(Sorry, I somehow forgot about this one until just now.)

Yes, your fix for the unreachable code verifier bug seems to have fixed this. I just rebased this branch and removed the lines I'd added to context.rs. :+1:

PR #224, opened 2018-01-15
no_std feature

lachlansneff:
This pull request implements `no_std` features in `lib/cretonne`, `lib/native`, `lib/frontend`, and `lib/wasm`.

Important to note, they still require an environment with liballoc, so any bare-metal applications using this must have an allocator.

My main reason to port cretonne to no_std is for use in my own projects, but I can see quite a number of possible uses for it aside from that.

Since cretonne has very few dependencies, porting wasn't difficult, the main issues being std::collections and std::error. Both we resolved by using external crates (only during no_std builds) that implement apis very close to std::collections and std::error.

Comments and criticism welcome!


lib/cretonne/src/simple_gvn.rs
```diff
@@ -7,6 +7,9 @@ use ir::{InstructionData, Function, Inst, Opcode, Type};
 use scoped_hash_map::ScopedHashMap;
 use timing;
 
+#[cfg(feature = "no_std")]
+use alloc::Vec;
```
sunfishcode: Could this just be `use std::vec::Vec`, so that it doesn't need the `#[cfg(...)]` part?

lib/cretonne/src/topo_order.rs
```diff
@@ -4,6 +4,9 @@ use entity::SparseSet;
 use dominator_tree::DominatorTree;
 use ir::{Ebb, Layout};
 
+#[cfg(feature = "no_std")]
+use alloc::Vec;
```
sunfishcode: Could this also use `std::vec::Vec`?

lib/wasm/src/sections_translator.rs
```diff
@@ -17,6 +17,10 @@ use wasmparser;
 use std::str::from_utf8;
 use environ::ModuleEnvironment;
 
+// this is for no_std builds, but has no affect on regular builds
```
sunfishcode: Here and elsewhere, say "effect" rather than "affect" ("affect" is usually a verb).

sunfishcode: Actually, I just noticed that these comments appear in some places that do this and not others, and it seems like it'd be good to be consistent about it. My instinct is that we could probably just drop these comments, but I don't have a strong feeling either way.

lachlansneff:
Whoops, that was a formatting error. Let me fix that.

lib/wasm/src/sections_translator.rs
sunfishcode: One more consistency nit: in some places, these new `use` lines have a blank line between them and the other `use` lines, and in others they don't. Again, I don't have a strong feeling about which is better, but it'd be nice to be consistent.

lachlansneff: Yeah, I'm not sure how I made that mistake. :/

lachlansneff: The majority of the work I did on porting was a couple of weeks ago. My styles appear to have changes since then, I'll make it consistent.

lib/cretonne/src/isa/mod.rs
```diff
@@ -54,6 +54,8 @@ use result;
 use timing;
 use isa::enc_tables::Encodings;
 
+use std::boxed::Box;
```
sunfishcode: Whitespace nit: Various places like this still have a blank line between the `use` lines.

lib/cretonne/src/lib.rs
```diff
@@ -46,3 +58,22 @@ mod stack_layout;
...
 mod write;
+
+/// This replaces `std` in builds with no_std.
+#[cfg(feature = "no_std")]
+mod std {
+    pub use core::*;
+    #[macro_use]
+    pub use alloc::{boxed, vec, string};
+    pub mod prelude {
+        pub use core::prelude as v1;
```
sunfishcode: I don't see `v1` used anywhere; is this line needed?

lib/cretonne/src/isa/mod.rs
lachlansneff: Sorry about that. I'll remove those.

lib/cretonne/src/lib.rs
lachlansneff: Not sure why I put that there. I think I saw that `nom` (the parser) had that in their version of an impromptu `std` and thought that having a prelude would make sense. Since it isn't automatically included in every file, obviously it's not. 😞

lachlansneff:
The decision is to have this live in a separate fork, at least until we know how often it gets broken.

eternaleye:
Ah, wish I'd seen this sooner. One thing to be aware of - negative cargo features (those which reduce API surface) are _invalid_, and cause Cargo's dependency logic to come to erronous conclusions which will cause spooky build failures at a distance for dependents. To quote my [comment](https://github.com/m4b/goblin/issues/8#issuecomment-253703418) on another project:

> There are two distinct concerns here:
> 1. Opt-in vs. opt-out in terms of "what the default is"
> 2. Opt-in vs. opt-out in terms of "what the polarity of each feature is"
> 
> These are orthogonal, and actually have very different behaviors:
> 
> Opt-in vs. opt-out as far as defaults is a matter of preference - generally speaking, if something has no dependencies and no conflicts or concerns, it's often nicer to downstream to enable it by default,
> 
> Opt-in vs. opt-out regarding feature polarity, however, has exactly one answer: Opt-in. Anything else _breaks_ Cargo. The reason is that Cargo assumes that it is _always_ safe to turn on a feature, and that doing so will _never_ break a dependent. As a result, features with opt-out polarity will get enabled if _any_ dependent asks for them, breaking anyone who uses the now-missing APIs.
> 
> The correct approach:
> 
> `Cargo.toml`:
> 
> ``` TOML
> [features]
> default = ["elf", "elf64", "..."]
> ```
> - Anything in `default` is on by default
> - Anything _not_ in `default` is off by default
> - Crates that don't need all of the default features can opt out of the default set and choose them á la carte
> - Crates that need non-default features can still enable them
> - Nobody's build gets broken.
> 
> But also, beware: Features that change the _behavior_ of an API - like you seem to describe `no_endian_fd` - are _always_ unsafe, because if one dependent flips them, that affects _every_ dependent.
> 
> Those kinds of features need to be separate APIs, under different names or modules, with the feature simply _enabling_ them.

sunfishcode:
@eternaleye That's good to know; thanks!

It turns out that we're not merging the patch into the main tree at this time. That said, I'm interested in `no_std` in Rust in general, so I'm experimenting a little with a fork of this PR [here](https://github.com/sunfishcode/cretonne/tree/no_std). I've now added a `std` feature which is enabled by default, along with a `no_std` feature which can be enabled to bring in the `hashmap_core` and `error_core` replacement libraries. With this setup, one of `std` or `no_std` must be enabled. But at least now they're both additive :-). (You can enable both, but it just acts like a `std` build.)

eternaleye:
@sunfishcode Hm, I'd disagree that it's strictly additive - replacement libraries, in particular, are not additive. If one dependent _explicitly_ uses `hashmap_core` (and enables `no_std`), and another uses the `std::collections` one (and leaves `std` enabled), then a third package depending on both _cannot possibly_ build correctly.

This is exactly the kind of "spooky action at a distance" that is problematic.

The API surface without a feature needs to be a strict subset of the API surface with a feature; substitutions are not a strict subset (Rust has nominal typing, not structural).

sunfishcode:
@eternaleye Ah, interesting. Yes, `HashMap` is only used internally at the moment, but `Error` is indeed exposed in the API. I'm now experimenting with using the experimental `failure` crate to replace `Error`. (It's just a side experiment for now, so I'm ok depending on experimental things.)

pchickey:
for what its worth, one of the big changes I made when I did a (private, for now) rewrite of cton-wasm was to use `failure` throughout. So far I have zero regrets.

Issue #211, opened 2017-12-21
Intel mov encoding bug


pchickey:
We discovered an intel encoding bug for mov.

We found a cretonne instruction that is load.i64 with destination register r13.
morally trying to do `mov (%r13),%r13`. Cretonne uses the `ld` recipe because
the displacement is 0. It chooses the encoding `4d 8b 2d`.

Those bytes actually encode a rip-relative addressed load with a 32-bit
displacement. the ModRM byte has a special case where it acts as a rip-relative
load iff in_reg0's bottom 3 bits == 5.

This means that when cretonne uses register 5 (which never happens because thats %rbp) or 13 (which does when you have enough register pressure) with this recipe, the code
it outputs will be incorrect. We tried to figure out a way to exclude those
registers but couldn't figure out a way to specify a RegClass of non-contigious
registers, and we couldn't find a way to add a predicate that ensures those
registers will not be used.

The correct encoding in this case is to use a ModRM byte that also expects a
disp8 afterwards, which will allow you to use register 13 as the destination.
The disp8 is just 0. We confirmed this encoding using gcc as `4d 8b 6d 00`.

We aren't sure if this issue effects other instruction encodings.

We hit the bug running musl's memcpy implementation, compiled through clang 5.0 wasm32-unknown-unknown and s2wasm. Because that code does uintptr fiddling to try to optimize for aligned loads and stores, is a very bad implementation for wasm, but it is interesting because it creates lots of register pressure, and helped us find this cool bug :) . cton file is uploaded here: https://gist.github.com/pchickey/79dd8f9a87c859233b97145de7b77703

tyler:
@stoklund: We're more than happy to do the work to fix this. We just need some advice on what direction to take in fixing it.

stoklund:
Sorry for missing this, it must have gotten buried in the holiday business.

I didn't know there was a special case for `%r13` too, I thought it was only `%rbp` that had this exception. You are right, of course.

I think the right way of fixing this is to use a RegClass constraint on the encoding that excludes `%rbp` and `%r13`. On the Rust side of things, register classes are already represented as bit masks of units (see `isa::RegClassData`), but you are right that we don't have any Python syntax for expressing non-contiguous register classes.

[...]

PR #225, opened 2018-01-16
Illegalize rbp/r13 for zero-offset loads on Intel x64

tyler:
This PR addresses #211.

The encoding used in the `ld` recipe on x64 is for zero-offsets loads. However, the encoding is invalid for `r13` and `rbp` as it has been reused for RIP-relative loads. This PR makes `r13` and `rbp` invalid registers for that recipe.

In order to do that, the internal representation of RegClass had to be changed from a range of register units to a bitmap of register units, as this becomes a non-contiguous class.

Side Note: I was hoping that this would cause the code generator to generate the longer 8-bit displacement encoding for loads from r13, but it looks like instead it copies the value into another register and then loads from that one. Not sure if it's worth trying to address that here.

### DIFF
--- a/lib/cretonne/meta/cdsl/registers.py
+++ b/lib/cretonne/meta/cdsl/registers.py
@@ -26,7 +26,7 @@ from . import is_power_of_two, next_power_of_two
 
 
 try:
-    from typing import Sequence, Tuple, List, Dict, Any, TYPE_CHECKING  # noqa
+    from typing import Sequence, Tuple, List, Dict, Any, Optional, TYPE_CHECKING  # noqa
     if TYPE_CHECKING:
         from .isa import TargetISA  # noqa
         # A tuple uniquely identifying a register class inside a register bank.
@@ -196,7 +196,7 @@ class RegClass(object):
     """
 
     def __init__(self, bank, count=0, width=1, start=0, bitmask=None):
-        # type: (RegBank, int, int, int, int) -> None
+        # type: (RegBank, int, int, int, Optional[int]) -> None
         self.name = None  # type: str
         self.index = None  # type: int
         self.bank = bank
--- a/lib/cretonne/meta/cdsl/registers.py
+++ b/lib/cretonne/meta/cdsl/registers.py
@@ -30,8 +30,8 @@ try:
     if TYPE_CHECKING:
         from .isa import TargetISA  # noqa
         # A tuple uniquely identifying a register class inside a register bank.
-        # (count, width, start)
-        RCTup = Tuple[int, int, int]
+        # (width, bitmask)
+        RCTup = Tuple[int, int]
 except ImportError:
     pass
 
@@ -195,24 +195,29 @@ class RegClass(object):
     :param start: The first unit to allocate, relative to `bank.first.unit`.
     """
 
-    def __init__(self, bank, count=None, width=1, start=0):
-        # type: (RegBank, int, int, int) -> None
+    def __init__(self, bank, count=0, width=1, start=0, bitmask=None):
+        # type: (RegBank, int, int, int, int) -> None
         self.name = None  # type: str
         self.index = None  # type: int
         self.bank = bank
-        self.start = start
         self.width = width
+        self.bitmask = 0
 
         # This is computed later in `finish_regclasses()`.
         self.subclasses = list()  # type: List[RegClass]
         self.toprc = None  # type: RegClass
 
         assert width > 0
-        assert start >= 0 and start < bank.units
 
-        if count is None:
-            count = bank.units // width
-        self.count = count
+        if bitmask:
+            self.bitmask = bitmask
+        else:
+            assert start >= 0 and start < bank.units
+            if count == 0:
+                count = bank.units // width
+            for a in range(count):
+                u = start + a * self.width
+                self.bitmask |= 1 << u
 
         bank.classes.append(self)
 
@@ -238,7 +243,7 @@ class RegClass(object):
         The tuple can be used as a dictionary key to ensure that there are no
         duplicate register classes.
         """
-        return (self.count, self.width, self.start)
+        return (self.width, self.bitmask)
 
     def intersect(self, other):
         # type: (RegClass) -> RCTup
@@ -249,17 +254,11 @@ class RegClass(object):
         """
         if self.width != other.width:
             return None
-        s_end = self.start + self.count * self.width
-        o_end = other.start + other.count * other.width
-        if self.start >= o_end or other.start >= s_end:
+        intersection = self.bitmask & other.bitmask
+        if intersection == 0:
             return None
 
-        # We have an overlap.
-        start = max(self.start, other.start)
-        end = min(s_end, o_end)
-        count = (end - start) // self.width
-        assert count > 0
-        return (count, self.width, start)
+        return (self.width, intersection)
 
     def __getitem__(self, sliced):
         # type: (slice) -> RegClass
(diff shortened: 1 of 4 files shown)


lib/cretonne/meta/isa/intel/registers.py
```diff
@@ -46,7 +46,9 @@
         names=['eflags'])
 
 GPR = RegClass(IntRegs)
+GPR_NORIP = GPR.without(GPR.rbp, GPR.r13)
```
sunfishcode: Minor bikeshed: How about `GPR_NO_RBP_R13` :-).

Alternatively, could you add a brief comment here explaining the "NORIP" name? Something like "In some contexts, the rbp/r13 encoding is interpreted to mean rip" or so.

lib/cretonne/meta/cdsl/registers.py
```diff
@@ -195,24 +195,29 @@ class RegClass(object):
...
-        self.count = count
+        if bitmask:
+            self.bitmask = bitmask
+        else:
+            assert start >= 0 and start < bank.units
+            if count == 0:
+                count = bank.units // width
+            for a in range(count):
+                u = start + a * self.width
+                self.bitmask |= 1 << u
```
stoklund: Intel register classes all have width 1, but ARM has D-registers that each take up one S-register and Q-registers that each take up 4 S-registers. This is modeled with width=2 and width=4 respectively.

When width > 1, I think the bit mask should have a set bit for all of the units covered by the register, rather than just the first one.

You can do that without a loop like this: `(1 << (count*self.width)) - 1) << start`.

lib/cretonne/meta/cdsl/registers.py
```diff
@@ -195,24 +195,29 @@ class RegClass(object):
     :param start: The first unit to allocate, relative to `bank.first.unit`.
     """
 
-    def __init__(self, bank, count=None, width=1, start=0):
-        # type: (RegBank, int, int, int) -> None
+    def __init__(self, bank, count=0, width=1, start=0, bitmask=None):
+        # type: (RegBank, int, int, int, int) -> None
```
sunfishcode: I'm not very familiar with the mypy type system; is it valid to have a value of `None` if the mypy type is `int`?

lib/cretonne/meta/cdsl/registers.py
```diff
@@ -299,19 +330,13 @@ def mask(self):
...
-            # words. This will only happen with widths that are not powers of
-            # two.
-            assert b + self.width <= 32, 'Register straddles words'
-            mask[u // 32] |= 1 << b
-
-        return mask
+        out_mask = []
+        mask32 = (1 << 32) - 1
+        bitmask = self.bitmask << self.bank.first_unit
+        for i in range(3):
```
sunfishcode: This `3` here is kind of a magic number. Could we avoid it by doing something like this?
```python
    while bitmask != 0:
        ... // do stuff with bitmask & mask32
        bitmask >>= 32
```

lib/cretonne/meta/cdsl/registers.py
```diff
@@ -273,12 +272,44 @@ def __getitem__(self, sliced):
         assert sliced.step is None, 'Subclass striding not supported'
 
         w = self.width
-        s = self.start + sliced.start * w
+        s = self.start() + sliced.start * w
```
stoklund: I suppose we get mysterious results if we attempt to slice a non-contiguous register class. I don't think we need to support that, but perhaps an assertion would be in order?

Testing for a contiguous bit mask:

```python
def is_contiguous_mask(x):
    y = x | (x-1)
    return x != 0 and ((y+1) & y) == 0
```

See LLVM's [`isShiftedMask_32` function](https://llvm.org/doxygen/MathExtras_8h_source.html#l00409).

stoklund:
> Side Note: I was hoping that this would cause the code generator to generate the longer 8-bit displacement encoding for loads from r13, but it looks like instead it copies the value into another register and then loads from that one. Not sure if it's worth trying to address that here.

If the register allocator sees the restricted encoding, it will work around it by inserting copies. This is how it is supposed to work:

1. The legalizer selects the encoding with the least restrictive constraints.
2. Register allocator satisfies those constraints.
3. Instruction compression pass tries to find a better (i.e., smaller) encoding whose constraints are still satisfied by the allocator's choices.

The legalizer doesn't actually search for the least restrictive encoding. It just uses the first one that fits, so we define encodings in a least-restrictive-first order.

The instruction compression pass hasn't been written yet. This is also where we will get rid of all those REX prefixes.

lib/cretonne/meta/cdsl/registers.py
tyler: The problem with that is that the Rust array this code is used to generate is expected to always have a length of 3. (https://github.com/stoklund/cretonne/blob/master/lib/cretonne/src/isa/registers.rs#L15-L21) However, there is a constant for this that I forgot to use: `MASK_LEN`. Fixing that.

lib/cretonne/meta/cdsl/registers.py
tyler: Good point. 👍

lib/cretonne/meta/cdsl/registers.py
tyler: Good point. That should be `Optional[int]`.

lib/cretonne/meta/isa/intel/registers.py
tyler: done!

lib/cretonne/meta/cdsl/registers.py
tyler: This doesn't seem to match the way that RegClass masks are currently generated. For instance:

```
--- good/registers-arm32.rs	2018-01-16 18:35:54.000000000 -0800
+++ bad/registers-arm32.rs	2018-01-16 18:36:15.000000000 -0800
@@ -60,7 +60,7 @@
     toprc: 1,
     first: 0,
     subclasses: 0x2,
-    mask: [0x55555555, 0x55555555, 0x00000000],
+    mask: [0xffffffff, 0xffffffff, 0x00000000],
     info: &INFO,
 };
 #[allow(dead_code)]
@@ -73,7 +73,7 @@
     toprc: 2,
     first: 0,
     subclasses: 0x4,
-    mask: [0x11111111, 0x11111111, 0x00000000],
+    mask: [0xffffffff, 0xffffffff, 0x00000000],
     info: &INFO,
 };
 #[allow(dead_code)]
```

This diff is of the generated registers-arm32.rs file, the former being how it works in master and the latter with this change. The top part of the diff is the `D` register class, and the bottom is `Q`.

This change also causes the `pressure.rs` unit tests to fail.

stoklund: My mistake. Sorry about that!

tyler:
> The legalizer doesn't actually search for the least restrictive encoding. It just uses the first one that fits, so we define encodings in a least-restrictive-first order.

Sorry if I’m being dumb here, but doesn’t that mean that this (in the Intel encodings.py) is backwards:
```
for recipe in [r.ld, r.ldDisp8, r.ldDisp32]:
    enc_i32_i64_ld_st(base.load, True, recipe, 0x8b)
    enc_i64(base.uload32.i64, recipe, 0x8b)
    I64.enc(base.sload32.i64, *recipe.rex(0x63, w=1))
    enc_i32_i64_ld_st(base.uload16, True, recipe, 0x0f, 0xb7)
    enc_i32_i64_ld_st(base.sload16, True, recipe, 0x0f, 0xbf)
    enc_i32_i64_ld_st(base.uload8, True, recipe, 0x0f, 0xb6)
    enc_i32_i64_ld_st(base.sload8, True, recipe, 0x0f, 0xbe)
```

Given that `ld` is the most restrictive, and `ldDisp32` the least?

tyler:
Thanks for the merge! :)

stoklund:
It is least restrictive first for register constraints, but smallest encoding first for constraints on the immediate operands. This order is less important once we get a compression pass.

PR #228, opened 2018-01-18
Fix the relocation addend for PIC calls on Intel

tyler:
Hello!

Currently the addend for PLTRel4 relocation that we output for position-independent calls has an incorrect addend. After linking this causes it to jump 4 bytes past the beginning of the PLT thunk, leading to "fun".

This PR corrects the addend and the test for it. Thanks!


tyler:
Thanks!

PR #230, opened 2018-01-25
derive Clone for cretonne::Context and constituients

pchickey:
I believe this is harmless.


stoklund:
Everything of substance is stored in the `Function`, the rest of the context is just analyses that can be recomputed and data structures holding on to reusable memory allocations. Would this work for you instead?

```rust
new_ctx = Context::new();
new_ctx.func.clone_from(&old_ctx.func);
```

This just clones the function into a blank context which should then be able to do what you need.

pchickey:
Ah, I wasn't aware of `clone_from`, so I couldn't figure out how to do that.

Can I expose a `Context::for_function(f: Function)` constructor that does basically that, minus the clone?

stoklund:
Sure!

PR #232, opened 2018-01-26
Change InsertBuilder's reuse array from [Option<Value>] to [Value].

sunfishcode:
My goal here is to either simplify the code or learn why it isn't that simple :-).

### DIFF
--- a/lib/cretonne/meta/gen_legalizer.py
+++ b/lib/cretonne/meta/gen_legalizer.py
@@ -299,9 +299,8 @@ def emit_dst_inst(node, fmt):
                 elif any(d.is_output() for d in node.defs):
                     # We have some output values to be reused.
                     array = ', '.join(
-                            ('Some({})'.format(d) if d.is_output()
-                                else 'None')
-                            for d in node.defs)
+                            '{}'.format(d)
+                            for d in node.defs if d.is_output())
                     builder += '.with_results([{}])'.format(array)
 
         fmt.line('{}.{};'.format(builder, node.expr.rust_builder(node.defs)))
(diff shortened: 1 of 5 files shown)


stoklund:
Look at the generated code in `out/legalizer.rs` for the `iadd_carry` pattern:

```
                // Unwrap (a, c) << iadd_carry(x, y, c_in)

...
                    pos.func.dfg.clear_results(inst);
                    let (a1, c1) = pos.ins().iadd_cout(x, y);
                    let c_int = pos.ins().bint(typeof_x, c_in);
                    let (a, c2) = pos.ins().with_results([Some(a), None]).iadd_cout(a1, c_int);
                    let c = pos.ins().with_result(c).bor(c1, c2);
```

The inserted `iadd_cout` instruction defines the `a` sum value directly, but the carry-out variable `c2` needs to be allocated a new value.

This is generated from

```python
expand.legalize(
        (a, c) << iadd_carry(x, y, c_in),
        Rtl(
            (a1, c1) << iadd_cout(x, y),
            c_int << bint(c_in),
            (a, c2) << iadd_cout(a1, c_int),
            c << bor(c1, c2)
        ))
```

This is all because we have no RAUW, of course.

stoklund:
Oh, now I see why your code compiles ;-)

The slice you pass in to `with_results` has positional result values. It works in the `iadd_carry` case because the `None` value happens to be last. But there is nothing preventing us from writing patterns that would require `with_results([None, Some(a)])` in the future.

IIRC, the reason there's no assertion on the slice length is to allow code sharing with the single-value `with_result` which is more common.

sunfishcode:
Ok, now I see. Thanks!

PR #238, opened 2018-02-01
Adds support for legalizing CLZ, CTZ and POPCOUNT on baseline

julian-seward1:
x86_64 targets (that is, devices that support SSE2 only).  Changes:

* Adds a new generic instruction, SELECTIF, that does value selection (a la
  conditional move) similarly to existing SELECT, except that it is
  controlled by condition code input and flags-register inputs.

* Adds a new Intel x86_64 variant, 'baseline', that supports SSE2 and
  nothing else.

* Adds new Intel x86_64 instructions BSR and BSF.

* Implements generic CLZ, CTZ and POPCOUNT on x86_64 'baseline' targets
  using the new BSR, BSF and SELECTIF instructions.

* Implements SELECTIF on x86_64 targets using conditional-moves.

* new test filetests/isa/intel/baseline_clz_ctz_popcount.cton
  (for legalization)

* new test filetests/isa/intel/baseline_clz_ctz_popcount_encoding.cton
  (for encoding)

* Allow lib/cretonne/meta/gen_legalizer.py to generate non-snake-caseified
  Rust without rustc complaining.

### DIFF
--- a/filetests/isa/intel/baseline_clz_ctz_popcount.cton
+++ b/filetests/isa/intel/baseline_clz_ctz_popcount.cton
@@ -1,0 +1,51 @@
+
+test compile
+set is_64bit
+isa intel baseline
+
+
+; clz/ctz on 64 bit operands
+
+function %i64_clz(i64) -> i64 {
+ebb0(v10: i64):
+  v11 = clz v10
+  ; check: x86_bsr
+  ; check: selectif.i64
+  return v11
+}
+
+function %i64_ctz(i64) -> i64 {
+ebb1(v20: i64):
+  v21 = ctz v20
+  ; check: x86_bsf
+  ; check: selectif.i64
+  return v21
+}
+
+
+; clz/ctz on 32 bit operands
+
+function %i32_clz(i32) -> i32 {
+ebb0(v10: i32):
+  v11 = clz v10
+  ; check: x86_bsr
+  ; check: selectif.i32
+  return v11
+}
+
+function %i32_ctz(i32) -> i32 {
+ebb1(v20: i32):
+  v21 = ctz v20
+  ; check: x86_bsf
+  ; check: selectif.i32
+  return v21
+}
+
+
+; popcount on 64 bit operands
+
+function %i64_popcount(i64) -> i64 {
+ebb0(v30: i64):
+  v31 = popcnt v30;
+  return v31;
+}
--- a/lib/cretonne/meta/isa/intel/encodings.py
+++ b/lib/cretonne/meta/isa/intel/encodings.py
@@ -368,6 +368,20 @@ enc_both(base.trueif, r.seti_abcd, 0x0f, 0x90)
 enc_both(base.trueff, r.setf_abcd, 0x0f, 0x90)
 
 #
+# Conditional move (a.k.a integer select)
+#
+I64.enc(base.selectif.i64, *r.cmov.rex(0x0F, 0x40, w=1))
+I64.enc(base.selectif.i32, *r.cmov.rex(0x0F, 0x40, w=0))
+
+#
+# Bit scan forwards and reverse
+#
+I64.enc(x86.bsf.i64, *r.bsf_and_bsr.rex(0x0F, 0xBC, w=1))
+I64.enc(x86.bsf.i32, *r.bsf_and_bsr.rex(0x0F, 0xBC, w=0))
+I64.enc(x86.bsr.i64, *r.bsf_and_bsr.rex(0x0F, 0xBD, w=1))
+I64.enc(x86.bsr.i32, *r.bsf_and_bsr.rex(0x0F, 0xBD, w=0))
+
+#
 # Convert bool to int.
 #
 # This assumes that b1 is represented as an 8-bit low register with the value 0
--- a/lib/cretonne/meta/isa/intel/legalize.py
+++ b/lib/cretonne/meta/isa/intel/legalize.py
@@ -4,7 +4,7 @@ Custom legalization patterns for Intel.
 from __future__ import absolute_import
 from cdsl.ast import Var
 from cdsl.xform import Rtl, XFormGroup
-from base.immediates import imm64, floatcc
+from base.immediates import imm64, intcc, floatcc
 from base.types import i32, i64
 from base import legalize as shared
 from base import instructions as insts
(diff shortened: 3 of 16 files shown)


lib/cretonne/meta/isa/intel/encodings.py
```diff
@@ -367,6 +367,20 @@ def enc_i32_i64_ld_st(inst, w_bit, recipe, *args, **kwargs):
...
+I64.enc(base.selectif.i64, *r.cmov.rex(0x0F, 0x40, w=1))
+I64.enc(base.selectif.i32, *r.cmov.rex(0x0F, 0x40, w=0))
+
+#
+# Bit scan forwards and reverse
+#
+I64.enc(x86.bsf.i64, *r.bsf_and_bsr.rex(0x0F, 0xBC, w=1))
+I64.enc(x86.bsf.i32, *r.bsf_and_bsr.rex(0x0F, 0xBC, w=0))
+I64.enc(x86.bsr.i64, *r.bsf_and_bsr.rex(0x0F, 0xBD, w=1))
+I64.enc(x86.bsr.i32, *r.bsf_and_bsr.rex(0x0F, 0xBD, w=0))
```
stoklund: Same thing here:

```python
enc_i32_i64(x86.bsf, r.bsf_and_bsr, 0x0F, 0xBC)
```

lib/cretonne/meta/isa/intel/legalize.py
```diff
@@ -100,3 +100,93 @@
...
+    ))
+
+intel_expand.legalize(
+    a << insts.ctz.i64(x),
+    Rtl(
+        c_minus_one << insts.iconst(imm64(-1)),
+        c_sixty_three << insts.iconst(imm64(63)),
+        (index1, r2flags) << x86.bsf(x),
+        index2 << insts.selectif(intcc.eq, r2flags, c_minus_one, index1),
+        a << insts.isub(c_sixty_three, index2),
```
stoklund: I think `BSF` returns the right value already for non-zero inputs, so no subtraction is required. Just force a 64  result when the input is zero.

(I agree we need executable tests for this stuff).

lib/cretonne/meta/isa/intel/legalize.py
```diff
@@ -100,3 +100,93 @@
...
+        v10 << insts.band     (v9, c77),
+        v11 << insts.isub     (v8, v10),
+        v12 << insts.ushr_imm (v11, 4),
+        v13 << insts.iadd     (v11, v12),
+        c0F << insts.iconst   (imm64(0x0F0F0F0F0F0F0F0F)),
+        v14 << insts.band     (v13, c0F),
+        c01 << insts.iconst   (imm64(0x0101010101010101)),
+        v15 << insts.imul     (v14, c01),
+        v16 << insts.ushr_imm (v15, 56)
+    ))
```
stoklund: Interesting, this is not the algorithm I know. It's a great test case for @d1m0's SMT-solver verification of patterns.

Does this generalize to a `popcnt.i32` version?

lib/cretonne/meta/isa/intel/recipes.py
```diff
@@ -1021,6 +1022,32 @@ def floatccs(iform):
...
+# 1 byte, modrm(r,r), is after the opcode
+#
+cmov = TailRecipe(
+        'cmov', IntSelect, size=1, ins=(FLAG.eflags, GPR, GPR), outs=1,
+        requires_prefix=False,
+        clobbers_flags=False,
+        emit='''
+        PUT_OP(bits | icc2opc(cond), rex2(in_reg2, in_reg1), sink);
+        modrm_rr(in_reg2, in_reg1, sink);
+        ''')
```
stoklund: This is where I think you'll need `outs=2` instead of `outs=1` so the output register is the same as the when-false input register.

filetests/isa/intel/baseline_clz_ctz_popcount.cton
```diff
@@ -0,0 +1,51 @@
...
+  ; check: selectif.i32
+  return v21
+}
+
+
+; popcount on 64 bit operands
+
+function %i64_popcount(i64) -> i64 {
+ebb0(v30: i64):
+  v31 = popcnt v30;
```
sunfishcode: This test is lacking `check:` lines. Also, should there be an `i32_popcount` test?

lib/cretonne/meta/isa/intel/settings.py
```diff
@@ -40,6 +40,7 @@
 
 # Presets corresponding to Intel CPUs.
 
+baseline = Preset(has_sse2)
```
sunfishcode: Yay for being able to assume we have SSE2! :-)

julian-seward1:
(with apologies if this appears twice)

Thanks for the feedback!  I think I fixed everything.  There is still
one bit of confusion though, regarding register ordering in 'cmov'.
You wrote:

> +    ; 64-bit wide cmov
> +
> +    ; asm: cmoveq %rdx, %r11
> +    [-,%r11]     v51 = selectif.i64 eq v48, v30, v34   ; bin: 4c 0f 44 da
>
> The existing `select` instruction takes operands in the following order:
> (condition, when-true, when-false). This matches the C ternary `c ? x : y`
> operator. I think the `selectif` instruction should follow the same pattern,
> and that is how you defined it too.
>
> So I think this encoding is backwards. We have v30 in %r11 and v34 in %rdx.
> This means that when the flags have an `eq` condition, we get v34 in the
> output register, and when they don't we get v30 (which just stays in %r11).
>
> Using the (when-true, when-false) argument order, the test should be:
>
> ```
> ; asm: cmoveq %r11, %rdx
> [-,%rdx]     v51 = selectif.i64 eq v48, v30, v34 ; bin: ...
> ```
>
> I think this comes down to using `outs=2` on the `cmov` `TailRecipe` so the
> output register is tied to the last input operand, the when-false value.

If I change both the test and the TailRecipe as you describe above, and update
the expected output, then the test still fails:

  Want: 49 0f 44 d3
  Got:  4c 0f 44 da

In order to get it to pass, I *also* have to swap the roles of in_reg1 and
in_reg2 in the recipe -- that is, change it from

  PUT_OP(bits | icc2opc(cond), rex2(in_reg2, in_reg1), sink);
  modrm_rr(in_reg2, in_reg1, sink);

to

  PUT_OP(bits | icc2opc(cond), rex2(in_reg1, in_reg2), sink);
  modrm_rr(in_reg1, in_reg2, sink);

Given that this swaps the roles of the two registers a *second* time relative
to your fix, I can't infer whether it's now correct, or whether it's just as
wrong as it was to start with, and there's something else I need to fix.  Do
you know?

----------------

> +intel_expand.legalize(
> +    a << insts.ctz.i64(x),
> +    Rtl(
> +        c_minus_one << insts.iconst(imm64(-1)),
> +        c_sixty_three << insts.iconst(imm64(63)),
> +        (index1, r2flags) << x86.bsf(x),
> +        index2 << insts.selectif(intcc.eq, r2flags, c_minus_one, index1),
> +        a << insts.isub(c_sixty_three, index2),
>
> I think `BSF` returns the right value already for non-zero inputs, so no
> subtraction is required. Just force a 64  result when the input is zero.

Ah, well spotted.  I had assumed the CTZ case was just a mirror-image swap
of the CLZ case, around the middle point of the input word.  But as you say,
BSF uses the same bit-indexing scheme as BSR, so the symmetry I assumed
doesn't actually exist.

----------------

> [legalization for popcount64]
> Interesting, this is not the algorithm I know. It's a great test case for
> @d1m0's SMT-solver verification of patterns.

This is from Hacker's Delight (the 2003 printing), Sec 5.1, page 68.  It is
a variant of the "classic" divide-and-conquer algorithm, in which the
recursive bit-gathering operation is stopped when each byte lane contains
the number of 1s originally in that byte lane.  All subsequent iterations
are removed and replaced by a single multiply, which is probably a win if
it can typically be done in 3 cycles.  I also liked this variant because it
requires fewer large constants than some other variants.

> Does this generalize to a `popcnt.i32` version?

Yes, that is exactly the same, at 32-bit width, with the final multiply and
shift changed appropriately.  I added an implementation for popcount.i32 now.

stoklund:
I would run that cmov instruction through an assembler and use the reg1/reg2 argument order that matches. Some Intel instructions come in both MR and RM forms where the order is reversed, and I think I got sub backwards to begin with.

+1 for your popcount algorithm

Thanks again.

PR #241, opened 2018-02-14
gen_settings: dont try to display a Preset descriptor in Flags

pchickey:
Trying to display a preset as part of the flags does not make sense, and before this commit it does not display anything meaningful - the printout just says e.g. "haswell =\n".

The offset byte a preset descriptor isnt a valid offset into the flag bytes, it is actually an offset into the PRESETS table. It will cause a panic when the offset is out of bounds for the flag bytes, which happens in the intel isa as of this commit.

Perhaps there should be two different DESCRIPTORS tables and two different Descriptor types for these two different "kinds" of things, but this is the least invasive change I could make to fix the panic.

Also, this panic (and this fix) doesn't get covered by Cretonne's CI. I'm not sure how to resolve that but if you give me direction on how to, I'll add tests that at least ensure that the display trait for each ISA flags does not panic for each of the available presets.


pchickey:
The unsightly `if let Detail::Preset = thing {} else { actual body }` construct is because Detail doesn't derive Eq, so that was the most economical way to pattern match on everything but Preset.

stoklund:
In the top-level `lib/cretonne/src/settings.rs` there is a `display_default()` unit test. We could do something similar to that in `.../intel/settings.rs` to catch such a panic.

To get rid of the unsightly `if let...`, you could add a `Detail::is_preset()` method to the enum. I've used that trick in other places such as `Affinity::is_none()`.

stoklund:
BTW, `pip3 install flake8` lets you catch the Python formatting issues on your own machine.

pchickey:
ok, ill take care of all of those things and update the PR.

pchickey:
should be fixed now

PR #242, opened 2018-02-16
Don't renumber entities in the parser.

sunfishcode:
This is still a work in progress, but I'm interested in feedback about whether this is a desirable approach. All the tests pass, though there's more cleanup to do.

This makes it easier to debug testcases:
 - the entity numbers in a .cton file match the entity numbers used
   within Cretonne.
 - serializing and deserializing doesn't cause indices to change.

One disadvantage is that if a .cton file uses sparse entity numbers, deserializing to the in-memory form doesn't compact it. However, the text format is not intended to be performance-critical, so this isn't expected to be a big burden.

### DIFF
--- a/docs/testing.rst
+++ b/docs/testing.rst
@@ -160,20 +160,6 @@ directives in the test file. LLVM's :command:`FileCheck` command has a
 ``CHECK-LABEL:`` directive to help separate the output from different functions.
 Cretonne's tests don't need this.
 
-Filecheck variables
-~~~~~~~~~~~~~~~~~~~
-
-Cretonne's IL parser causes entities like values and EBBs to be renumbered. It
-maintains a source mapping to resolve references in the text, but when a
-function is written out as text as part of a test, all of the entities have the
-new numbers. This can complicate the filecheck directives since they need to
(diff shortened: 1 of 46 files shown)


stoklund:
I think this is a good direction. Some high-level comments:

- It would be good to clearly separate the DFG methods that are only for use by the parser. They can really mess things up if they're used from normal compiler code. Perhaps they should all have a specific prefix, or they should belong to some trait implemented by the DFG that only the parser uses.
- It would be nice if we could avoid duplicating the `make_inst_results_reusing` method.
- The parser's source map should probably be renamed since its not a map any longer.
- The filecheck variables, `$v4`, `$ebb6`, etc should probably be removed too. It doesn't make a lot of sense if they are always identity mappings.

sunfishcode:
Patches updated. I made several cleanups and addressed the review comments, except:
 - I left the name `SourceMap` in place as its now a map from entities to their source locations. 
 - I didn't move the DFG's parser routines into a separate trait, because that seems to require maintaining two copies of the interface. Instead, I just moved them into their own impl and gave them distinctive names.

PR #249, opened 2018-02-24
Adds support to transform integer div and rem by constants into cheap…

julian-seward1:
…er equivalents.

Adds support for transforming integer division and remainder by constants
into sequences that do not involve division instructions.

* div/rem by constant powers of two are turned into right shifts, plus some
  fixups for the signed cases.

* div/rem by constant non-powers of two are turned into double length
  multiplies by a magic constant, plus some fixups involving shifts,
  addition and subtraction, that depends on the constant, the word size and
  the signedness involved.

* The following cases are transformed: div and rem, signed or unsigned, 32
  or 64 bit.  The only un-transformed cases are: unsigned div and rem by
  zero, signed div and rem by zero or -1.

* This is all incorporated within a new transformation pass, "preopt", in
  lib/cretonne/src/preopt.rs.

* In preopt.rs, fn do_preopt() is the main driver.  It is designed to be
  extensible to transformations of other kinds of instructions.  Currently
  it merely uses a helper to identify div/rem transformation candidates and
  another helper to perform the transformation.

* In preopt.rs, fn get_div_info() pattern matches to find candidates, both
  cases where the second arg is an immediate, and cases where the second
  arg is an identifier bound to an immediate at its definition point.

* In preopt.rs, fn do_divrem_transformation() does the heavy lifting of the
  transformation proper.  It in turn uses magic{S,U}{32,64} to calculate the
  magic numbers required for the transformations.

* There are many test cases for the transformation proper:
    filetests/preopt/div_by_const_non_power_of_2.cton
    filetests/preopt/div_by_const_power_of_2.cton
    filetests/preopt/rem_by_const_non_power_of_2.cton
    filetests/preopt/rem_by_const_power_of_2.cton
    filetests/preopt/div_by_const_indirect.cton
  preopt.rs also contains a set of tests for magic number generation.

* The main (non-power-of-2) transformation requires instructions that return
  the high word of a double-length multiply.  For this, instructions umulhi
  and smulhi have been added to the core instruction set.  These will map
  directly to single instructions on most non-intel targets.

* intel does not have an instruction exactly like that.  For intel,
  instructions x86_umulx and x86_smulx have been added.  These map to real
  instructions and return both result words.  The intel legaliser will
  rewrite {s,u}mulhi into x86_{s,u}mulx uses that throw away the lower half
  word.  Tests:
    filetests/isa/intel/legalize-mulhi.cton (new file)
    filetests/isa/intel/binary64.cton (added x86_{s,u}mulx encoding tests)


lib/cretonne/src/preopt.rs
```diff
@@ -0,0 +1,1051 @@
...
+//----------------------------------------------------------------------
+//
+// General pattern-match helpers.
+
+// Find out if |value| actually resolves to a constant, and if so what its
+// value is.
+fn get_const(value: Value, dfg: &DataFlowGraph) -> Option<i64> {
+    match dfg.value_def(value) {
+        ValueDef::Result(definingInst, resultNo) => {
+            // FIXME: do we need to range-check definingInst here?
```
sunfishcode: The answer to this question is no, we don't need a separate range-check (besides implicit one that Rust does). The verifier pass can diagnose problems like this in an orderly way, so if such a problem somehow persists to this point, a panic is fine.

lib/cretonne/src/preopt.rs
```diff
@@ -0,0 +1,1051 @@
...
+
+    }
+}
+
+
+//----------------------------------------------------------------------
+//
+// General pattern-match helpers.
+
+// Find out if |value| actually resolves to a constant, and if so what its
```
sunfishcode: Style nit: Cretonne follows Rust style in general, so comments like this should use \`value\` rather than |value|.

lib/cretonne/src/preopt.rs
```diff
@@ -0,0 +1,1051 @@
...
+            let qf; // final quotient
+            let q0 = pos.ins().iconst(I32, mulBy as i64);
+            let q1 = pos.ins().umulhi(n1, q0);
+            if doAdd {
+                assert!(shiftBy >= 1 && shiftBy <= 32);
+                let t1 = pos.ins().isub(n1, q1);
+                let t2 = pos.ins().ushr_imm(t1, 1);
+                let t3 = pos.ins().iadd(t2, q1);
+                // I never found any case where shiftBy == 1 here.
+                // So there's no attempt to fold out a zero shift.
```
sunfishcode: I think it makes sense to add a `debug_assert!(shiftBy != 1)` or so, to check this assumption (here and in the 64-bit version).

filetests/preopt/div_by_const_non_power_of_2.cton
```diff
@@ -0,0 +1,273 @@
...
+ebb0(v0: i32):
+    v1 = udiv_imm v0, 641
+    ; check: iconst.i32 0x0066_3d81
+    ; check: umulhi v0, v2
+    ; check: copy v3
+    return v1
+}
+
+; AFAIK, the complex case never occurs with zero shift, or at least
+; I never found an example.
```
sunfishcode: My preference here is to add an assert for this assumption (see below), and then we don't need the comment here (or similar comments in the other tests). If the assumption ever changes, we'll add tests for it.

lib/cretonne/src/preopt.rs
```diff
@@ -0,0 +1,1051 @@
...
+            if isRem {
+                pos.func.dfg.replace(inst).iconst(I64, 0);
+            } else {
+                pos.func.dfg.replace(inst).copy(n1);
+            }
+        }
+
+        // U64 div, rem by a power-of-2
+        &DivRemByConstInfo::DivU64(n1, d) |
+        &DivRemByConstInfo::RemU64(n1, d) if d.count_ones() == 1 => {
```
sunfishcode: Rust has a `.is_power_of_two()` which can do this.

lib/cretonne/src/preopt.rs
```diff
@@ -0,0 +1,1051 @@
...
+            Opcode::SdivImm => (true, false),
+            Opcode::SremImm => (true, true),
+            _other => return None,
+        };
+        // Pull the operation size (type) from the left arg
+        let argL_ty = dfg.value_type(arg);
+        return package_up_divrem_info(arg, argL_ty, imm.into(), isSigned, isRem);
+    }
+
+    if let &InstructionData::Binary { opcode, args } = idata {
```
sunfishcode: There's an interesting question here: should we check for the `iconst` case, as the code here does, or just expect some previous optimization should have already combined a `udiv` with an `iconst` to make a `udiv_imm`, and only check for that. Since we don't have such optimizations yet, I think it makes sense to keep this code for now. But I would like to have a TODO comment here mentioning that we should continue to think about this.

filetests/preopt/div_by_const_indirect.cton
```diff
@@ -0,0 +1,33 @@
...
+    v2 = udiv v0, v1
+    return v2
+}
+
+function %indir_sdiv64(i64) -> i64 {
+ebb0(v0: i64):
+    v1 = iconst.i64 -90210
+    v2 = sdiv v0, v1
+    return v2
+}
```
sunfishcode: This test will need `check:` lines.

lib/cretonne/src/preopt.rs
```diff
@@ -0,0 +1,1051 @@
...
+                assert!(shiftBy >= 1 && shiftBy <= 32);
+                let t1 = pos.ins().isub(n1, q1);
+                let t2 = pos.ins().ushr_imm(t1, 1);
+                let t3 = pos.ins().iadd(t2, q1);
+                // I never found any case where shiftBy == 1 here.
+                // So there's no attempt to fold out a zero shift.
+                qf = pos.ins().ushr_imm(t3, (shiftBy - 1) as i64);
+            } else {
+                assert!(shiftBy >= 0 && shiftBy <= 31);
+                // Whereas shiftBy == 0 has been observed here.
```
sunfishcode: Wording nit: rather than talk about observations, just say that there are known cases where shiftBy == 0.

lib/cretonne/src/preopt.rs
```diff
@@ -0,0 +1,1051 @@
...
+                    // Curiously, we don't care here what the sign of d is.
+                    pos.func.dfg.replace(inst).isub(n1, t4);
+                } else {
+                    // S32 div by a power-of-2
+                    let t4 = pos.ins().sshr_imm(t3, k as i64);
+                    if isNeg {
+                        pos.func.dfg.replace(inst).irsub_imm(t4, 0);
+                    } else {
+                        pos.func.dfg.replace(inst).copy(t4);
+                    };
```
sunfishcode: Unneeded semicolon.

sunfishcode:
Thanks for working on this, this looks really good!

julian-seward1:
Dan, thanks for the review.  I just pushed a new version here with fixes
for all review comments.

sunfishcode:
Looks good!

PR #255, opened 2018-03-05
I64Extend16S should operate on 16 bits, not 8.

waywardmonkeys:
This showed up in clippy warnings as the code for this case was
the same as for `I64Extend8S`.


sunfishcode:
Nice find! I expected the sign-extension-ops proposal tests to cover these, but it seems something there isn't working yet.

PR #257, opened 2018-03-07
Elaborate on Cretonne's rustc backend goals.

sunfishcode:
This is a start in addressing the feedback [here](https://users.rust-lang.org/t/proposal-rllvm-rust-implementation-of-llvm/16021): it spells out some rustc goals for Cretonne in more detail. I've made an attempt to clarify that the project has some well-scoped goals to keep things realistic in the present, while also having a long-term vision to inspire the future.

### DIFF
--- a/rustc.rst
+++ b/rustc.rst
@@ -37,7 +37,7 @@ including:
 - Design for superoptimization. Traditionally, compiler development teams have
   spent many years of manual effort to identify patterns of code that can be
   matched and replaced. Superoptimizers have been contributing some to this
-  effort, but in the future, I expect we'll be able to reverse roles.
+  effort, but in the future, we may be able to reverse roles.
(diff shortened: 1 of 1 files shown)


rustc.rst
```diff
@@ -2,15 +2,65 @@
...
+optimization, but for some users, completely unoptimized code is too slow to test
+with, so a "pretty good" mode might be good enough.
+
+There's plenty of work to do to achieve these goals, and if achieve them, we'll have
+enabled a Rust compiler written entirely in Rust, and enabled faster Rust compile
+times for important use cases.
+
+
+
+With all that said, there is a potential goal behond that, which is to build a
```
pepyakin: beyond that?

sunfishcode: Fixed, thanks!

PR #271, opened 2018-03-17
Rename `I32` -> `X86_32` and `I64` -> `X86_64`

afnanenayet:
Addressing [issue 265](https://github.com/Cretonne/cretonne/issues/265)

I'm not sure if there are any other references to `I32` and `I64` that I missed, if there are please let me know!


sunfishcode:
Cool. While reviewing the patch here, it occurs to me that `enc_i64` is different than the 64-bit part of `enc_i32_i64`, and it's sometimes used for things that don't involve the type `i64`, such as `movaps` (`enc_both` calls `enc_i64`), so I think it makes sense to rename `enc_i64` to `enc_x86_64` too.

sunfishcode:
Looks good!

PR #272, opened 2018-03-18
Use the wabt crate

### LINKED ISSUE #270: Use the wabt crate
Currently running the in-tree wasm tests requires the `wat2wasm` command be installed. There's now a [wabt crate](https://crates.io/crates/wabt), so we should switch to using that, with an explicit dependency, instead.

pepyakin:
Fixes #270 

I'm not sure about adding wabt in cton-util since it'll make cton-util transitively dependent on C/C++ compiler.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,6 +25,7 @@ serde = "1.0.8"
 serde_derive = "1.0.8"
 tempdir = "0.3.5"
 term = "0.5.1"
+wabt = "0.2.2"
 
 [workspace]
 
--- a/README.rst
+++ b/README.rst
@@ -56,11 +56,6 @@ You can then run tests with::
 
     ./test-all.sh
 
-You may need to install the *wat2wasm* tool from the `wabt
-<https://github.com/WebAssembly/wabt>`_ project in order to run all of the
-WebAssembly tests. Tests requiring wat2wasm are ignored if the tool is not
-installed.
-
 Building the documentation
 --------------------------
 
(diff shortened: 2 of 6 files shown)


sunfishcode:
Hmm, that's a good point.

Thinking about this some more, for the tests in the tree, maybe we should just check the .wasm files into the repo. That way we won't need a C++ compiler or a wabt install just to run the tests.

However, it would still be convenient for development to have `cton-util wasm` be able to read .wat files directly. So, what if we made the `wabt-rs` code here a [feature](https://doc.rust-lang.org/cargo/reference/manifest.html#the-features-section) with an optional dependency?

pepyakin:
> However, it would still be convenient for development to have cton-util wasm be able to read .wat files directly. So, what if we made the wabt-rs code here a feature with an optional dependency?

Opt-in or opt-out? I'm not sure, since they both have pros and cons.

sunfishcode:
How about opt-in? I see the pros and cons, so maybe the thing that decides it is that not everyone using Cretonne is doing so with wasm. And not everyone doing so with wasm is using the in-tree implementation of `FuncEnvironment`, which is a significant part of what's being tested by those tests.

quadrupleslap:
Here's a version where the entire wasm subcommand is opt-out:

https://github.com/quadrupleslap-forks-things/cretonne/tree/use-wabt

sunfishcode:
@quadrupleslap I hadn't thought of making the whole wasm subcommand opt-out, but at first glance it seems to work out well. Would you mind submitting a new PR?

Eventually, once the APIs stablize more, we may end up splitting cretonne-wasm out into its own repo, but for now it's convenient to keep them together because they're still evolving together (eg. https://github.com/cretonne/cretonne/pull/346).

pepyakin:
@quadrupleslap Thanks for working on this! 

Sorry about that this slipped from my radars : <

PR #276, opened 2018-03-22
Use clippy

### LINKED ISSUE #264: Use clippy
It'd be great to make more use of clippy. One obstacle right now is that enabling clippy on Cretonne produces a large number of warnings, making it difficult to know where to start.

One possible way to get started would be to add a bunch of `#[cfg_attr(feature = "cargo-clippy", allow(...))]` lines to the lib.rs files to suppress all the lints that clippy currently gives. That way, we'd be able to run `cargo clippy` and get useful reports for the rest of the lints. And then, we could work on re-enabling individual lints one at a time.

However, I'm open to other suggestions too.

pchickey:
This is a start at using clippy throughout Cretonne.

Current issues:
* Only `cargo +nightly clippy` works - stable cargo and clippy has a bug that seems to be related to the way arguments are passed when used in the root directory (works in all the other crates)
* Clippy and the old rustfmt we are using disagree about whether braces belong in some places - rustfmt insists on them, clippy flags them as a problem. (All of the problems clippy flags outside of cretonne were fixed until I ran rustfmt.)

Currently, there are a ton of lints turned off in the cretonne crate. A significant portion of them are fixable without much effort, but others would require doing work on the code generator.

Unfortunately, because generated code is made part of other source files via `include!()`, I wasn't put top level clippy pragmas in the generated sources only, because it wouldn't be at the top of the source file.

Fixes #264

### DIFF
--- a/check-clippy.sh
+++ b/check-clippy.sh
@@ -5,16 +5,6 @@ set -euo pipefail
 
 if cargo install --list | tee /dev/null | grep -q "^clippy v0"; then
     exit 0
-fi
-
-if [[ ${1:-""} != "--install" ]]; then
-    echo "********************************************************************"
-    echo "*  Please install clippy to lint rust code.                        *"
-    echo "********************************************************************"
-    echo "$0 --install"
-    sleep 1
+else
     exit 1
 fi
-
-echo "Installing clippy"
-cargo +nightly install clippy
--- a/check-clippy.sh
+++ b/check-clippy.sh
@@ -1,0 +1,20 @@
+#!/bin/bash
+set -euo pipefail
+
+# Usage: check-clippy.sh [--install]
+
+if cargo install --list | tee /dev/null | grep -q "^clippy v0"; then
+    exit 0
+fi
+
+if [[ ${1:-""} != "--install" ]]; then
+    echo "********************************************************************"
+    echo "*  Please install clippy to lint rust code.                        *"
+    echo "********************************************************************"
+    echo "$0 --install"
+    sleep 1
+    exit 1
+fi
+
+echo "Installing clippy"
+cargo +nightly install clippy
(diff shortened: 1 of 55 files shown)


lib/cretonne/src/binemit/relaxation.rs
```diff
@@ -76,14 +76,13 @@ pub fn relax_branches(func: &mut Function, isa: &TargetIsa) -> Result<CodeOffset
...
                         let dest_offset = cur.func.offsets[dest];
-                        if !range.contains(offset, dest_offset) {
-                            // This is an out-of-range branch.
-                            // Relax it unless the destination offset has not been computed yet.
-                            if dest_offset != 0 || Some(dest) == cur.func.layout.entry_block() {
-                                offset +=
-                                    relax_branch(&mut cur, offset, dest_offset, &encinfo, isa);
-                                continue;
-                            }
+                        // This coubd be an out-of-range branch.
```
tyler: coubd -> could

lib/cretonne/src/ir/function.rs
```diff
@@ -141,6 +141,8 @@ impl Function {
     }
 
     /// Return an object that can display this function with correct ISA-specific annotations.
+    // Needless lifetimes lint false positive
+    #[cfg_attr(feature = "cargo-clippy", allow(needless_lifetimes))]
```
sunfishcode: It feels like needless_lifetimes has enough false positives that it'd be better to disable this lint globally than clutter up the code disabling it locally every time.

check-clippy.sh
```diff
@@ -0,0 +1,20 @@
...
+if cargo install --list | tee /dev/null | grep -q "^clippy v0"; then
+    exit 0
+fi
+
+if [[ ${1:-""} != "--install" ]]; then
+    echo "********************************************************************"
+    echo "*  Please install clippy to lint rust code.                        *"
+    echo "********************************************************************"
+    echo "$0 --install"
+    sleep 1
```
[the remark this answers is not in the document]
pchickey: Reduced this to one-line echo.

lib/cretonne/src/ir/function.rs
pchickey: agreed

pchickey:
I fixed up the remaining trivial lints, documented the reasons for each lint that is currently `allow`ed, and took care of the various review comments.

sunfishcode:
Great, thanks for getting this infrastructure set up!

PR #279, opened 2018-03-23
[WIP] Add a Trap sink to code generation

tyler:
This is a start to implementing a mechanism to allow users to track Trap sites. I'm not totally sure what else folks would want for this, so I figured I'd push early. Feedback welcome.

Let me know if this is a reasonable direction, and I'll clean it up, add tests and so on.

Thanks!

### DIFF
--- a/filetests/isa/intel/binary32-float.cton
+++ b/filetests/isa/intel/binary32-float.cton
@@ -148,30 +148,30 @@ ebb0:
     ; Load/Store
 
     ; asm: movss (%ecx), %xmm5
-    [-,%xmm5]           v100 = load.f32 v0                      ; bin: f3 0f 10 29
+    [-,%xmm5]           v100 = load.f32 v0                      ; bin: heap_oob f3 0f 10 29
     ; asm: movss (%esi), %xmm2
-    [-,%xmm2]           v101 = load.f32 v1                      ; bin: f3 0f 10 16
+    [-,%xmm2]           v101 = load.f32 v1                      ; bin: heap_oob f3 0f 10 16
     ; asm: movss 50(%ecx), %xmm5
-    [-,%xmm5]           v110 = load.f32 v0+50                   ; bin: f3 0f 10 69 32
+    [-,%xmm5]           v110 = load.f32 v0+50                   ; bin: heap_oob f3 0f 10 69 32
     ; asm: movss -50(%esi), %xmm2
-    [-,%xmm2]           v111 = load.f32 v1-50                   ; bin: f3 0f 10 56 ce
+    [-,%xmm2]           v111 = load.f32 v1-50                   ; bin: heap_oob f3 0f 10 56 ce
     ; asm: movss 10000(%ecx), %xmm5
-    [-,%xmm5]           v120 = load.f32 v0+10000                ; bin: f3 0f 10 a9 00002710
+    [-,%xmm5]           v120 = load.f32 v0+10000                ; bin: heap_oob f3 0f 10 a9 00002710
     ; asm: movss -10000(%esi), %xmm2
-    [-,%xmm2]           v121 = load.f32 v1-10000                ; bin: f3 0f 10 96 ffffd8f0
+    [-,%xmm2]           v121 = load.f32 v1-10000                ; bin: heap_oob f3 0f 10 96 ffffd8f0
 
     ; asm: movss %xmm5, (%ecx)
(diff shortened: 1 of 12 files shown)


sunfishcode:
Neat! In the SpiderMonkey embedding, we have a similar need, and currently we just walk the instructions in the `Function` to find the traps, but it's pretty awkward. This `TrapSink` interface looks like it would work better.

At a quick glance, it looks like the one thing we'd need that's not in this patch yet is the `ir::SourceLoc` of of the trap instruction. Do you think it would make sense to add a `SourceLoc` argument to the `trap` function and pass it  `func.srclocs[inst]`?

sunfishcode:
Actually, looking at this more, SpiderMonkey also needs to know the locations of other instructions which can trap too, including heap loads and stores, integer div/rem, and float-to-int conversions. Do you think it'd make sense to add TrapSink calls for those instructions too?

tyler:
Yep! All those things make sense. I had only been thinking about it from the perspective of ops that *will* trap, instead of ops that *can* trap. Will add that and the srcloc suggestion.

tyler:
Actually, I guess this might be a little trickier. Would you want to add TrapSink calls for heap loads/stores even if they've been explicitly bounds-checked? It seems to me that if they were to trap, that would be unexpected behavior that the embedder would want to handle differently (i.e. panic). But I don't think one would be able to tell the difference between an expected out-of-bounds access, and an unexpected one in that case? (Other than, I guess, the target address.)

sunfishcode:
Hmm. Ok, if we can tell the `TrapSink` about all instructions that are *expected* to possibly trap, then embedders should be able to reliably distinguish between expected and unexpected traps. That way, if an embedder ever gets an unexpected trap from wasm code, it would mean the sandbox may have been breached, so it should probably just halt the application immediately.

That would mean, if we're doing bounds checking with no guard pages, the explicit trap generated by the bounds checking code will be reported to the TrapSink, but the actual load/store won't be. If we're using guard pages, then the load/store should be reported to the TrapSink, as a possible expected trap. Does that make sense?

tyler:
@sunfishcode That totally makes sense. That's what I was trying to say, but you said it more clearly. :)

I'm wondering how I can distinguish between those two cases though. If these are emitted directly from the recipes code, I think the only way would be to have two sets of load/stores instructions. One for unchecked loads/stores and another for bounds-checked ones.

Perhaps I'm missing a clear shortcut here.

sunfishcode:
I wonder if we can use the [`notrap` flag](https://cretonne.readthedocs.io/en/latest/langref.html#memory-operation-flags) on loads and stores for this. That is, only report a load/store to `TrapSink` if it doesn't have `notrap` set. We'd also want to have something somewhere automatically set `notrap` on bounds-checked loads/stores. But maybe we can figure that part out later.

tyler:
Ah! Ok. That makes sense to me.

tyler:
Ok, I think this is good to go, assuming I haven't missed anything. (Which I probably have.)

I didn't touch the ISAs outside of "intel". I can take a stab at that if you want, but it's outside my area of knowledge so it'll take some time.

tyler:
One note for posterity, I suppose: I'm not super happy about emitting these from recipes. In theory, the recipes could be reused across some instructions that trap and some that don't. Or they could have different trapcodes.

Would it make sense, either now or in the longer term, to attach the expected trapcodes to the instructions themselves instead? I guess that could be complicated because whether certain instructions trap is ISA dependent.

sunfishcode:
Attaching the trapcodes to the instructions is also complicated by the fact that sometimes the trap happens in the middle of a recipe, and trap handlers want to know the specific machine instruction where the trap will be generated.

My rough idea for how we can reduce the redundancy is to introduce helper functions. Eg. instead of having recipes do
```rust
         // ud2.
         sink.trap(code, func.srclocs[inst]);
         sink.put1(0x0f);
         sink.put1(0x0b);
````
they'd do more things like
```rust
        put_ud2(sink, func.srclocs[inst]);
```
It's harder for loads and stores, because of PUT_OP, but I find PUT_OP makes it hard to read the code, so I'm hoping to find a better way to handle that anyway. But of course, I'm open to any ideas here.

At a quick glance, the patch looks good. I like how this shows up in bitemit tests :-). I'll do a more detailed review soon.

sunfishcode:
Looks good! This may evolve as we get more experience using this feature in practice, but it looks like we have a good base to build on here.

PR #280, opened 2018-03-23
[WIP] Initialize static-style heap bases in the entry block.

sunfishcode:
Static-style heaps are allocated at a fixed address, so there's no need
to load their base address more than once per function. Instead of
load a base address before every heap access, load the bases of all
addresses once, in the function entry block.

### DIFF
--- a/lib/cretonne/src/ir/function.rs
+++ b/lib/cretonne/src/ir/function.rs
@@ -11,6 +11,7 @@ use ir::{InstEncodings, ValueLocations, JumpTables, StackSlots, EbbOffsets, Sour
 use ir::{Ebb, JumpTableData, JumpTable, StackSlotData, StackSlot, SigRef, ExtFuncData, FuncRef,
          GlobalVarData, GlobalVar, HeapData, Heap};
 use isa::{TargetIsa, EncInfo};
+use packed_option::PackedOption;
 use std::fmt;
 use write::write_function;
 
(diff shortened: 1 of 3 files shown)


sunfishcode:
It's not possible right now, but that's something we should fix. Right now, Cretonne interprets the heap base global as the address of a pointer to the heap. So, it always does a load from that address to get the heap address. I think what we really want here is to get rid of this implicit indirection, and just have the heap base global mean the heap address. Users that need indirection could still get it, using a `deref`.

That way, we could have the heap base be a `globalsym`, which would embed the heap address in the code, exactly as you describe.

It'd also let us do things like have the heap be allocated at an offset from the `vmctx` pointer, rather than being pointed to by a pointer some offset from `vmctx`, which @pchickey may be interested in.

lachlansneff:
Ah, good. That should help with code quality and increase performance significantly.

sunfishcode:
I filed https://github.com/cretonne/cretonne/issues/358 about removing the implicit indirection on globals.

lachlansneff:
I'd like to get this merged in as well, no that the implicit indirection is soon to be removed.

sunfishcode:
My gut feeling is that #366 is going to be a better approach than this PR, though it'll be a bit more work. So, I'm not currently expecting to merge this PR, unless someone needs this sooner.

sunfishcode:
Closing, as we're not currently planning to pursue this direction. #366 is where future work will happen.

PR #281, opened 2018-03-23
Mark `emit_to_memory` as unsafe, and introduce a safe `emit`.

sunfishcode:
Some users have found `emit_to_memory` awkward when writing code into a `Vec<u8>`, and @eddyb pointed out that it should actually be an unsafe function, given what it does. The patch here makes `emit_to_memory` unsafe, and adds an `emit` function which provides a safe interface for the case of writing to a `Vec<u8>`.

### DIFF
--- a/lib/codegen/src/binemit/memorysink.rs
+++ b/lib/codegen/src/binemit/memorysink.rs
@@ -38,7 +38,10 @@ pub struct MemoryCodeSink<'a> {
 
 impl<'a> MemoryCodeSink<'a> {
     /// Create a new memory code sink that writes a function to the memory pointed to by `data`.
-    pub fn new<'sink>(
+    ///
+    /// This function is unsafe since `MemoryCodeSink` does not perform bounds checking on the
+    /// memory buffer.
+    pub unsafe fn new<'sink>(
         data: *mut u8,
         relocs: &'sink mut RelocSink,
         traps: &'sink mut TrapSink,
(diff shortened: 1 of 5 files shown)


eddyb:
Can you make `compile` return a different type which has the `emit` methods? I found it a bit odd that it specifically works when you call `compile` first. Or `emit` could run `compile` instead of panicking, if it hasn't been ran?

sunfishcode:
I like the idea of making `emit` just run `compile`, so that you can do everything in a single call. I'll look into that.

lachlansneff:
What about having another call that takes an instance of a trait as the parameter to make custom code outputs easier?

For example:
```rust
pub trait CodeEmittee {
    fn get_write_buffer(&mut self, len: usize) -> &mut [u8];
}

...

fn emit_code<T: CodeEmittee>(&mut self, T) {
    ...
}
```

sunfishcode:
I've now updated the PR to add a `compile_and_emit` function which provides an interface which does everything in one step, and is safe. And I've marked the underlying `emit_to_memory` and `MemoryCodeSink` as unsafe. 

Also relevant is the new [Module](https://docs.rs/cretonne-module/0.5.0/cretonne_module/struct.Module.html) API, which also encapsulates the unsafe interfaces, and provides an even higher level API.

lib/codegen/src/binemit/memorysink.rs
```diff
@@ -38,7 +38,10 @@ pub struct MemoryCodeSink<'a> {
 
 impl<'a> MemoryCodeSink<'a> {
     /// Create a new memory code sink that writes a function to the memory pointed to by `data`.
-    pub fn new<'sink>(
+    ///
+    /// This function is unsafe since `MemoryCodeSink` does not perform bounds checking on the
+    /// memory buffer.
```
bjorn3: "And cant guarrantee that the data pointer is valid."

sunfishcode: Added!

PR #283, opened 2018-03-24
Implement shift-immediate encodings for x86

### LINKED ISSUE #274: x86-64 encoding for ushr_imm etc.
We should define x86-64 encodings for ushr_imm, sshr_imm, and ishl_imm, as these are fairly frequent instructions.

pchickey:
Closes #274

I think the tests are correct. I wasn't able to put the shift-immediate encoding test right next to the shift tests because of value naming (at least, without renaming all subsequent values), which is slightly annoying.

### DIFF
--- a/filetests/isa/intel/binary64.cton
+++ b/filetests/isa/intel/binary64.cton
@@ -529,6 +529,21 @@ ebb0:
     ; asm: addq $-2147483648, %rsp
     adjust_sp_imm -2147483648                   ; bin: 48 81 c4 80000000
 
+    ; Shift immediates
+    ; asm: shlq $12, %rsi
+    [-,%rsi]             v515 = ishl_imm v2, 12   ; bin: 48 c1 e6 0c
+    ; asm: shlq $13, %r8
+    [-,%r8]              v516 = ishl_imm v4, 13   ; bin: 49 c1 e0 0d
+    ; asm: sarq $32, %rsi
+    [-,%rsi]             v517 = sshr_imm v2, 32   ; bin: 48 c1 fe 20
+    ; asm: sarq $33, %r8
+    [-,%r8]              v518 = sshr_imm v4, 33   ; bin: 49 c1 f8 21
+    ; asm: shrl $62, %rsi
+    [-,%rsi]             v519 = ushr_imm v2, 62   ; bin: 48 c1 ee 3e
+    ; asm: shrl $63, %r8
+    [-,%r8]              v520 = ushr_imm v4, 63   ; bin: 49 c1 e8 3f
+
+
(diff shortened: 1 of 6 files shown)


filetests/isa/intel/binary64.cton
```diff
@@ -1053,6 +1053,19 @@ ebb0:
...
+    ; asm: shll $12, %r10d
+    [-,%r10]             v527 = ishl_imm v3, 12   ; bin: 41 c1 e2 0c
+    ; asm: sarl $5, %esi
+    [-,%rsi]             v528 = sshr_imm v2, 5    ; bin: c1 fe 05
+    ; asm: sarl $32, %r10d
+    [-,%r10]             v529 = sshr_imm v3, 32   ; bin: 41 c1 fa 20
+    ; asm: shrl $8, %esi
+    [-,%rsi]             v530 = ushr_imm v2, 8    ; bin: c1 ee 08
+    ; asm: shrl $63, %r10d
+    [-,%r10]             v531 = ushr_imm v3, 63   ; bin: 41 c1 ea 3f
```
sunfishcode: This covers 32-bit shifts on x86-64, with and without rex prefixes. We should also test 32-bit shifts on x86-32, and 64-bit shifts on x86-64.

lib/cretonne/meta/isa/intel/encodings.py
```diff
@@ -169,6 +169,15 @@ def enc_i32_i64_ld_st(inst, w_bit, recipe, *args, **kwargs):
...
     X86_64.enc(inst.i32.any, *r.rc(0xd3, rrr=rrr))
 
+for inst,           rrr in [
+        (base.ishl_imm, 4),
+        (base.ushr_imm, 5),
+        (base.sshr_imm, 7)]:
+    X86_32.enc(inst.i32, *r.rib(0xc1, rrr=rrr))
+    X86_64.enc(inst.i64, *r.rib.rex(0xc1, rrr=rrr, w=1))
+    X86_64.enc(inst.i32, *r.rib.rex(0xc1, rrr=rrr))
+    X86_64.enc(inst.i32, *r.rib(0xc1, rrr=rrr))
```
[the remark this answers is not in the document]
pchickey: Thats correct. I was fooled because you cant use that helper above, because the instructions require the `.any` suffix. (I added a comment there to note why)

filetests/isa/intel/binary64.cton
pchickey: got them in https://github.com/Cretonne/cretonne/pull/283/commits/171348f2822d7b43fa44e35a030ee899b4c00882

sunfishcode:
Looks good. Thanks for adding that comment!

PR #293, opened 2018-03-31
Only save callee-saved registers that are used

tyler:
Currently we save all callee-saved registers in the prologue of every function, and restore them in the epilogue. With this PR, we now only save and restore those that are actually used in the function.

I'm not particularly happy with the code in this one. I feel like I'm abusing AllocatableSet. But it appears to work and can be refactored.

Test fixes/changes forthcoming.

### DIFF
--- a/filetests/isa/intel/prologue-epilogue.cton
+++ b/filetests/isa/intel/prologue-epilogue.cton
@@ -9,24 +9,14 @@ ebb0:
     return
 }
 
-; check: function %foo(i64 fp [%rbp], i64 csr [%rbx], i64 csr [%r12], i64 csr [%r13], i64 csr [%r14], i64 csr [%r15]) -> i64 fp [%rbp], i64 csr [%rbx], i64 csr [%r12], i64 csr [%r13], i64 csr [%r14], i64 csr [%r15] system_v {
-; nextln:     ss0 = explicit_slot 168, offset -224
-; nextln:     ss1 = incoming_arg 56, offset -56
-; check: ebb0(v0: i64 [%rbp], v1: i64 [%rbx], v2: i64 [%r12], v3: i64 [%r13], v4: i64 [%r14], v5: i64 [%r15]):
+; check: function %foo(i64 fp [%rbp]) -> i64 fp [%rbp] system_v {
+; nextln:     ss0 = explicit_slot 168, offset -184
+; nextln:     ss1 = incoming_arg 16, offset -16
+; check: ebb0(v0: i64 [%rbp]):
 ; nextln:     x86_push v0
 ; nextln:     copy_special %rsp -> %rbp
-; nextln:     x86_push v1
-; nextln:     x86_push v2
-; nextln:     x86_push v3
-; nextln:     x86_push v4
-; nextln:     x86_push v5
-; nextln:     adjust_sp_imm -168
-; nextln:     adjust_sp_imm 168
-; nextln:     v11 = x86_pop.i64
-; nextln:     v10 = x86_pop.i64
-; nextln:     v9 = x86_pop.i64
-; nextln:     v8 = x86_pop.i64
-; nextln:     v7 = x86_pop.i64
-; nextln:     v6 = x86_pop.i64
-; nextln:     return v6, v7, v8, v9, v10, v11
+; nextln:     adjust_sp_imm -176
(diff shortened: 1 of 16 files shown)


lib/cretonne/src/isa/intel/abi.rs
```diff
@@ -168,6 +169,31 @@ pub fn callee_saved_registers(flags: &shared_settings::Flags) -> &'static [RU] {
     }
 }
 
+fn callee_saved_registers_used(
+    flags: &shared_settings::Flags,
+    func: &ir::Function,
+) -> AllocatableSet {
+    let mut all_callee_saved = AllocatableSet::empty();
+    for reg in callee_saved_registers(flags) {
+        all_callee_saved.free(GPR, *reg as RegUnit);
```
sunfishcode: While we don't support the Windows x64 calling convention yet, it has some XMM registers as callee-saved. What would you think about renaming `callee_saved_registers` to `callee_saved_gprs`, to remind us to do something else when we need to support callee-saved xmm registers?

Also, it's not new in your patch here, but `callee_saved_registers` doesn't need to be `pub`.

lib/cretonne/src/isa/intel/abi.rs
```diff
@@ -168,6 +169,31 @@ pub fn callee_saved_registers(flags: &shared_settings::Flags) -> &'static [RU] {
...
+    }
+
+    let mut used = AllocatableSet::empty();
+    for value_loc in func.locations.values() {
+        match value_loc {
+            &ValueLoc::Reg(ru) => {
+                if !used.is_avail(GPR, ru) {
+                    used.free(GPR, ru);
```
sunfishcode: If we can think of a good name, it'd be nice to add a function to `AllocatableSet` aka `RegisterSet` that adds a register to the set without asserting that it's not already set.

sunfishcode: A subtlety here is that the ValueLoc here only contains one register unit of a potentially multi-unit register. This doesn't occur on x86, because we've decided to ignore the H registers in the register allocator. But it wouldn't necessarily be ok for other architectures, so we should put comments here advising people not to blindly reuse this code without thinking about it.

filetests/isa/intel/prologue-epilogue.cton
```diff
@@ -9,24 +9,14 @@ ebb0:
     return
 }
 
-; check: function %foo(i64 fp [%rbp], i64 csr [%rbx], i64 csr [%r12], i64 csr [%r13], i64 csr [%r14], i64 csr [%r15]) -> i64 fp [%rbp], i64 csr [%rbx], i64 csr [%r12], i64 csr [%r13], i64 csr [%r14], i64 csr [%r15] system_v {
```
tyler: I've adjusted this test to match what's being output now, but I'm not totally convinced that it's correct. @sunfishcode, can you look at this and make sure it's sane? I struggled for a bit to convince it to use a callee-saved register in a synthetic test here, but with your new pre and post opt passes, that's increasingly hard to do. :)

tyler:
I think I've addressed all the points brought up. Save for the one about adding an unchecked `free` function to `RegisterSet`... basically just because I couldn't think of a clear name for it.

filetests/isa/intel/prologue-epilogue.cton
sunfishcode: The code looks fine now, and I've also run some non-trivial tests on it and it seems to be working as intended so far. I'll take a closer look at the test here, and I can also help out and add some tests that use callee-saved registers.

sunfishcode:
Looking at testcases, we're no longer adding unused callee-saved registers as arguments and return values. This works for now, however if we add post-regalloc optimizations or analyses in the future, they may need to special-case callee-saved registers to know that they're live. On the other hand, that might not be a big problem, and it reduces clutter to avoid adding arguments that aren't used. So I'm not sure what's best here yet.

sunfishcode:
Thinking about this more, we don't currently add callee-saved registers as arguments/returns on calls, so as things stand right now post-regalloc code would have to special-case them anyway. That leads me to think it's fine that we're not adding them all all the time in the prologue/epilogue code.

sunfishcode:
Looks good! This likely won't be the last word on callee-saved registers ever ;-), but I think it's a good step right now.

PR #294, opened 2018-04-03
[WIP] Module API

sunfishcode:
This is an early sketch of a Module API, along with sketches of implementations for Faerie and "SimpleJIT", which aims to be an easy-to-use JIT interface. See https://github.com/Cretonne/cretonne/issues/288 for details.


sunfishcode:
Now with basic faerie support. It took a fair amount of juggling to support the `RelocSink` visitor for faerie, but I think I've found on an approach that works well, and hides most of the juggling from the external API.

sunfishcode:
Now with basic JIT support too, enough to compile functions into memory which can be called. The `Module` API is starting to settle down, though some things could stand to be streamlined still. And it needs tests.

Next up will be a simple demo program that uses the JIT support.

sunfishcode:
This branch now has roughly all the functionality I expect we'll need for the first iteration here, though I expect it'll evolve a little more with experience. I'll be writing some tests next.

sunfishcode:
The faerie backend is now complete enough to support [a version of the demo](https://github.com/sunfishcode/simplejit-demo/tree/faerie) that writes out a .o file rather than JITing to memory. The Module API is starting to settle down.

I'm considering merging this branch to master soon, so that we can publish crates on crates.io, so that the demo doesn't have to use a `git` dependency. Of course, feedback remains welcome in all areas.

sunfishcode:
This is far enough along that I'm confident the overall approach is a good general direction, so I'm going to merge it into trunk now, and continue development there.

PR #297, opened 2018-04-10
Move entity into a separate crate

Amanieu:
I would like to use `EntityMap` in my own projects, so I split all the entity-related code (including `PackedOption`) into a separate crate.


sunfishcode:
This seems reasonable, so let's try it. Entity's API has been stable for a while, and I'm not aware of any major upcoming changes. The `__core` export drew my attention, but I don't see any major practical problems, and don't have any better suggestions myself :-}.

PR #298, opened 2018-04-11
Add a "colocated" flag to symbol references.

sunfishcode:
This adds a "colocated" flag to function and symbolic global variables which
indicates that they are defined along with the current function, so they can
use PC-relative addressing.

This also changes the function decl syntax; the name now always precedes the
signature, and the "function" keyword is no longer included.


sunfishcode:
This is also convenient for the upcoming faerie backend, as faerie doesn't yet support non-colocated data.

PR #301, opened 2018-04-18
Update no_std branch to 0.5.0

gmorenz:
For the most part this is nice and simple.

I had to change the wasmparser dependency to point to the git version, since the published version doesn't support no_std yet.

I noticed that there are a few very minor places where error handling differs between master and no_std, nothing substantial though.


sunfishcode:
wasmparser 0.16.0, just published, should include the no_std support, so let's switch to that.

gmorenz:
Done :+1:

sunfishcode:
Looks good!

PR #306, opened 2018-04-23
cargo fuzz integration

jfoote:
Hello,

This PR defines a simple [cargo-fuzz](https://github.com/rust-fuzz/cargo-fuzz) fuzz target for Cretonne's `translate_module` using a `DummyEnvironment`. It uses [binaryen-rs](https://github.com/pepyakin/binaryen-rs) to generate valid WebAssembly modules from the random input that cargo-fuzz/libfuzzer generates.

Is this something the project would be interested in using for periodic or continuous fuzzing? This approach is not perfect, so I thought it would be best to ask/converse here first.

_Notes_

- I ran this against a snapshot of `master` (as of Friday afternoon EST) on ~8 cores x 2 days and didn't turn up any crashes (only slow compiles)
    - Unfortunately it seems there is not a simple way to generate a line-level coverage report for the generated test corpus yet (neither by way of cargo-fuzz nor directly via rustc)
- This approach should find panics/crashes in Cretonne, but not validate the correctness of the code (a separate and interesting problem)
- I tweaked [binaryen-rs](https://github.com/pepyakin/binaryen-rs/compare/master...jfoote:foote/wasm-mvp-fuzz) so that it only produces Cretonne-compatible WebAssembly (i.e. MVP only, not Atomics) to avoid false positives
    - If we'd like to move forward I'll work with the maintainers there to create a more reasonable patch to support this functionality
- Input to the fuzzer is a random string (not valid WebAssembly), so we can't readily distill the universe of WebAssembly to create a seed corpus

Thanks,
Jon


sunfishcode:
> This PR defines a simple cargo-fuzz fuzz target for Cretonne's translate_module using a DummyEnvironment. It uses binaryen-rs to generate valid WebAssembly modules from the random input that cargo-fuzz/libfuzzer generates.

> Is this something the project would be interested in using for periodic or continuous fuzzing?

Yes, I think this is great, both on its own, and as a step towards more fuzzing in the future!

> This approach should find panics/crashes in Cretonne, but not validate the correctness of the code (a separate and interesting problem)

Right; the `DummyEnvironment` isn't complete enough for general correctness testing, though in the future we may have more complete options in the tree.

>    I tweaked binaryen-rs so that it only produces Cretonne-compatible WebAssembly (i.e. MVP only, not Atomics) to avoid false positives
        If we'd like to move forward I'll work with the maintainers there to create a more reasonable patch to support this functionality

That'd be great.

> Input to the fuzzer is a random string (not valid WebAssembly), so we can't readily distill the universe of WebAssembly to create a seed corpus

I have a collection of some interesting wasm modules here: https://github.com/sunfishcode/wasm-collection which may be of interest. Beware though that some of those are large.

jfoote:
> I have a collection of some interesting wasm modules here: https://github.com/sunfishcode/wasm-collection which may be of interest. Beware though that some of those are large.

Thanks. I think it may be worth supplanting or supplementing this fuzz target with one that consumes/produces WebAssembly directly (using a feedback-driven fuzzer like cargo fuzz, a template-based/generational fuzzer like [dharma](https://github.com/MozillaSecurity/dharma), or something else); this corpus will be helpful in that case.

For posterity, the reason this fuzz target can't leverage existing WebAssembly modules is that it consumes a "seed" input string that is used to generate a valid WebAssembly module via binaryen, and (as far as I know) binaryen cannot generate a "seed" input string from a WebAssembly module (the `--translate-to-fuzz` function is one-way, essentially).

>> create a more reasonable patch to support this functionality

> That'd be great.

PR/discussion: https://github.com/pepyakin/binaryen-rs/pull/4. I'll update here once it is sorted out.

jfoote:
This is still blocked on https://github.com/pepyakin/binaryen-rs/pull/4, but in the meantime I made a few updates:

- Added a build/integration test to run the fuzz target with a single input
    - This prevents accidental breakage of the fuzzer integration, a best-practice motivated by [google/oss-fuzz](https://github.com/google/oss-fuzz)
- Uses the jfoote/binaryen-rs https://github.com/pepyakin/binaryen-rs/pull/4 PR branch
    - Required to support integration testing until the PR is merged

jfoote:
https://github.com/pepyakin/binaryen-rs/pull/4 has been merged; I think this PR is ready to go. Note that the fuzzer integration test will only run if the rust nightly toolchain is available (since cargo-fuzz requires it).

PR #307, opened 2018-04-23
Improvements to Modules API

pchickey:
Defines an `enum ModuleError`, which derives `Fail`, to be used throughout `Module`. Implement `ModuleError` results at all the places that used to panic, or were otherwise unimplemented.

Exposes `Module::get_name(&self, &str) -> Option<FuncOrDataId>`, and the `FuncOrDataId` type. This is convenient for users so we don't have to do our own declaration bookkeeping.

This was spurred by a discussion about possibly adding string-name method variants (using declaration string names instead of FuncId or DataId). The `get_name` method allows the user to implement those themselves, if they are required.

### DIFF
--- a/lib/faerie/src/backend.rs
+++ b/lib/faerie/src/backend.rs
@@ -12,13 +12,24 @@ use std::fs::File;
 use target;
 use traps::{FaerieTrapManifest, FaerieTrapSink};
 
+#[derive(Debug)]
+/// Setting to enable collection of traps. Setting this to `Enabled` in
+/// `FaerieBuilder` means that a `FaerieTrapManifest` will be present
+/// in the `FaerieProduct`.
+pub enum FaerieTrapCollection {
+    /// `FaerieProduct::trap_manifest` will be `None`
+    Disabled,
+    /// `FaerieProduct::trap_manifest` will be `Some`
+    Enabled,
+}
+
 /// A builder for `FaerieBackend`.
 pub struct FaerieBuilder {
     isa: Box<TargetIsa>,
     name: String,
     format: container::Format,
     faerie_target: faerie::Target,
-    collect_traps: bool,
+    collect_traps: FaerieTrapCollection,
 }
 
 impl FaerieBuilder {
@@ -26,14 +37,15 @@ impl FaerieBuilder {
     /// can be passed to
     /// [`Module::new`](cretonne_module/struct.Module.html#method.new].
     ///
-    /// Note: To support calls JIT'd functions from Rust or other compiled
-    /// code, it's necessary for the `call_conv` setting in `isa`'s flags
-    /// to match the host platform.
+    /// Faerie output requires that TargetIsa have PIC (Position Independent Code) enabled.
+    ///
+    /// `collect_traps` setting determines whether trap information is collected in a
+    /// `FaerieTrapManifest` available in the `FaerieProduct`.
     pub fn new(
         isa: Box<TargetIsa>,
         name: String,
         format: container::Format,
-        collect_traps: bool,
+        collect_traps: FaerieTrapCollection,
     ) -> Result<Self, ModuleError> {
         debug_assert!(isa.flags().is_pic(), "faerie requires PIC");
         let faerie_target = target::translate(&*isa)?;
@@ -80,10 +92,9 @@ impl Backend for FaerieBackend {
             isa: builder.isa,
             artifact: faerie::Artifact::new(builder.faerie_target, builder.name),
             format: builder.format,
-            trap_manifest: if builder.collect_traps {
-                Some(FaerieTrapManifest::new())
-            } else {
-                None
+            trap_manifest: match builder.collect_traps {
+                FaerieTrapCollection::Enabled => Some(FaerieTrapManifest::new()),
+                FaerieTrapCollection::Disabled => None,
             },
         }
     }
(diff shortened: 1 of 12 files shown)


lib/module/src/module.rs
```diff
@@ -86,6 +86,26 @@ pub struct FunctionDeclaration {
...
+pub enum ModuleError {
+    /// Indicates an identifier was used before it was declared
+    #[fail(display = "Undeclared identifier {}", _0)]
+    Undeclared(String),
+    /// Indicates an identifier was used contrary to the way it was declared
+    #[fail(display = "Incompatible declaration of identifier {}", _0)]
+    IncompatibleDeclaration(String),
+    /// Indicates that an identifier was already declared
+    #[fail(display = "Duplicate declaration of identifier {}", _0)]
+    DuplicateDeclaration(String),
```
sunfishcode: Right now, duplicate declarations are fine as long as they're compatible. Is there a use for the `DuplicateDeclaration` error code?

lib/module/src/module.rs
```diff
@@ -362,6 +382,17 @@ where
...
     }
+    /// Like `declare_func_in_func` but the declared func is referred to by name.
+    pub fn declare_func_in_func_by_name(
+        &self,
+        funcname: &str,
+        in_func: &mut ir::Function,
+    ) -> Result<ir::FuncRef, ModuleError> {
+        match self.names.get(funcname) {
+            Some(&FuncOrDataId::Func(funcid)) => Ok(self.declare_func_in_func(funcid, in_func)),
+            _ => unimplemented!(),
```
sunfishcode: Should this return an `IncompatibleDeclaration`, or some other error? (and ditto for the other cases below)

lib/module/src/module.rs
```diff
@@ -7,7 +7,7 @@
 
 use Backend;
 use cretonne_codegen::entity::{EntityRef, PrimaryMap};
-use cretonne_codegen::result::{CtonError, CtonResult};
+use cretonne_codegen::result::CtonError;
```
sunfishcode: It occurs to me that we may want to rename `CtonError` to `CodegenError` or now, though we don't necessarily need to do that in this PR.

lib/module/src/module.rs
```diff
@@ -73,19 +94,56 @@ impl Linkage {
...
+    Undeclared(String),
+    /// Indicates an identifier was used contrary to the way it was declared
+    #[fail(display = "Incompatible declaration of identifier {}", _0)]
+    IncompatibleDeclaration(String),
+    /// Indicates an identifier was defined more than once
+    #[fail(display = "Duplicate definition of identifier {}", _0)]
+    DuplicateDefinition(String),
+    /// Indicates an identifier was defined, but was declared as an import
+    #[fail(display = "Invalid definition of identifier {}", _0)]
+    InvalidDefinition(String),
```
sunfishcode: Can the error message and maybe also the enum arm for this one mention "imported", to be more specific about the problem? Something like `ImportGivenDefinition` and "Imported identifier assigned a definition"?

lib/module/src/module.rs
```diff
@@ -73,19 +94,56 @@ impl Linkage {
     }
 }
 
-#[derive(Copy, Clone, PartialEq, Eq, Hash, PartialOrd, Ord)]
-enum FuncOrDataId {
+
+/// A declared name may refer to either a function or data declaration
+#[derive(Copy, Clone, PartialEq, Eq, Hash, PartialOrd, Ord, Debug)]
+pub enum FuncOrDataId {
+    /// When its a FuncId
```
sunfishcode: "its" -> "it's"; here and below

pchickey:
I've added an (optional at run-time) trap manifest to cretonne-faerie in 3c30fe3.

lib/faerie/src/backend.rs
```diff
@@ -238,7 +258,11 @@ impl Backend for FaerieBackend {
 /// [`finish`](../cretonne_module/struct.Module.html#method.finish) function.
 /// It provides functions for writing out the object file to memory or a file.
 pub struct FaerieProduct {
-    artifact: faerie::Artifact,
+    /// Faerie artifact with all functions, data, and links from the module defined
+    pub artifact: faerie::Artifact,
+    /// Optional trap manifest
+    pub trap_manifest: Option<FaerieTrapManifest>,
```
sunfishcode: Could you add a comment mentioning how to enable this (using the `FaerieBuilder`)?

lib/faerie/src/backend.rs
```diff
@@ -3,20 +3,22 @@
...
 use target;
+use traps::{FaerieTrapManifest, FaerieTrapSink};
 
 /// A builder for `FaerieBackend`.
 pub struct FaerieBuilder {
     isa: Box<TargetIsa>,
     name: String,
     format: container::Format,
     faerie_target: faerie::Target,
+    collect_traps: bool,
```
sunfishcode: Would you mind introducing a custom enum here, rather than using bool, so make end user code easier to read?

lib/faerie/src/backend.rs
```diff
@@ -31,14 +33,16 @@ impl FaerieBuilder {
         isa: Box<TargetIsa>,
         name: String,
         format: container::Format,
-    ) -> Result<Self, Error> {
+        collect_traps: bool,
```
sunfishcode: Could you add some documentation mentioning what the traps argument does? Just clarify that this enables the `trap_manifest` member from the `FaerieProduct`.

pchickey:
Changed that to an enum and elaborated the docs.

I also made two minor changes to error handling.

Issue #285, opened 2018-03-27
Register+register addressing modes


sunfishcode:
Cretonne should have a way to use register+register addressing modes, as they are present on x86, ARM, and many other architectures.

This is particularly desirable for WebAssembly-style code, where most loads and stores are relative to a heap base in a register, since it allows folding the add of the heap base with the offset value into the load/store.

sunfishcode:
#### New instructions

Adding variants of `load` and `store` with additional operands would require an increase to the size of the `InstructionData` struct, which would increase the size of all instructions. To avoid this, we can introduce `load_complex` and `store_complex` instructions which use `VARIABLE_ARGS` instead of their fixed number of arguments (see `x_return` in instructions.py for another example of this). This has the effect of storing the operands as `ValueList`s, which are held outside the `InstructionData`.

A `load_complex` with 1 operand value would be a load from a simple base register. With 2, it'd be a base register plus an index register. `store_complex` is similar, but its list would start with an additional input value operand for the value to be stored.

#### Text format
In the text format, simple addresses currently look like `v0+16`; it'd be nice to print complex addresses as `v0+v1+16` or so with the `+16` being recognized by `optional_offset32()`. So for parsing, where `load` and `store` do `optional_offset32()`, it should check for an optional index operand also. The operands can be held in a `VariableArgs` while parsing, which is just a `Vec` underneath, and converted into a value list with `into_value_list` like the parsing for `InstructionData::Call` does.

#### Encoding
To define encodings, it will be necessary to have predicates that test for single-register and two-register loads and stores, which can be simple functions that test the length of the value list. These predicates can be specified with the `instp` instruction predicate mechanism.

sunfishcode:
A few more details:

#### Optimization

`load_complex` and `store_complex` can be formed in `postopt`. That way they'll see all the low-level arithmetic and raw loads and stores expanded from heap accesses and other operations.

Not all targets will want this optimization, so we control whether this optimization is done by adding a query function similar to `uses_cpu_flags()`. This will allow targets to enable or disable the optimization. In the future, this could evolve into a way for targets to specify specific addressing constraints, however for now it's sufficient to just add `uses_complex_addresses(&self) -> bool`, where x86 can just return true.

sunfishcode:
This is fixed by #309! :raised_hands:

PR #309, opened 2018-04-24
load_complex and store_complex instructions

tyler:
This is a work-in-progress. I'm opening it now so I have something to refer to for feedback and help. :)

This PR is intended to address #285.

Things to do:
- [x] Add base instructions
- [x] Add 8-bit, 16-bit + sign/zero-extended variants
- [x] length check the value lists given to the recipes
- [x] Adjust lexer and parser to handle `v0+v1+...` as value lists
- [x] Make it possible to emit correct ModRM, REX, and SIB bytes for these instructions
- [x] Add target encodings
- [x] Encoding tests
  - [x] x86 64
  - [x] x86 32
- [x] Post-opt pass to generate load/store_complex
- [x] Post-opt tests
- [x] Documentation

### DIFF
--- a/filetests/isa/x86/binary64.cton
+++ b/filetests/isa/x86/binary64.cton
@@ -607,32 +607,56 @@ ebb0:
     [-,%rcx]            v527 = load_complex.i64 v521+v522+1       ; bin: heap_oob 48 8b 4c 18 01
     ; asm: movl 1(%rax,%rbx,1), %ecx
     [-,%rcx]            v528 = load_complex.i32 v521+v522+1       ; bin: heap_oob 8b 4c 18 01
+
+    ; asm: mov    0x100000(%rax,%rbx,1),%rcx
     [-,%rcx]            v529 = load_complex.i64 v521+v522+0x1000  ; bin: heap_oob 48 8b 8c 18 00001000
+    ; asm: mov    0x100000(%rax,%rbx,1),%ecx
     [-,%rcx]            v530 = load_complex.i32 v521+v522+0x1000  ; bin: heap_oob 8b 8c 18 00001000
+    ; asm: movzbq (%rax,%rbx,1),%rcx
     [-,%rcx]            v531 = uload8_complex.i64 v521+v522         ; bin: heap_oob 48 0f b6 0c 18
+    ; asm: movzbl (%rax,%rbx,1),%ecx
     [-,%rcx]            v532 = uload8_complex.i32 v521+v522         ; bin: heap_oob 0f b6 0c 18
+    ; asm: movsbq (%rax,%rbx,1),%rcx
     [-,%rcx]            v533 = sload8_complex.i64 v521+v522         ; bin: heap_oob 48 0f be 0c 18
+    ; asm: movsbl (%rax,%rbx,1),%ecx
     [-,%rcx]            v534 = sload8_complex.i32 v521+v522         ; bin: heap_oob 0f be 0c 18
+    ; asm: movzwq (%rax,%rbx,1),%rcx
     [-,%rcx]            v535 = uload16_complex.i64 v521+v522         ; bin: heap_oob 48 0f b7 0c 18
+    ; asm: movzwl (%rax,%rbx,1),%ecx
     [-,%rcx]            v536 = uload16_complex.i32 v521+v522         ; bin: heap_oob 0f b7 0c 18
+    ; asm: movswq (%rax,%rbx,1),%rcx
     [-,%rcx]            v537 = sload16_complex.i64 v521+v522         ; bin: heap_oob 48 0f bf 0c 18
+    ; asm: movswl (%rax,%rbx,1),%ecx
     [-,%rcx]            v538 = sload16_complex.i32 v521+v522         ; bin: heap_oob 0f bf 0c 18
+    ; asm: mov    (%rax,%rbx,1),%ecx
     [-,%rcx]            v539 = uload32_complex v521+v522             ; bin: heap_oob 8b 0c 18
+    ; asm: movslq (%rax,%rbx,1),%rcx
     [-,%rcx]            v540 = sload32_complex v521+v522             ; bin: heap_oob 48 63 0c 18
 
     ; Store Complex
     [-,%rcx]            v600 = iconst.i64 1
     [-,%rcx]            v601 = iconst.i32 1
+    ; asm: mov    %rcx,(%rax,%rbx,1)
     store_complex v600, v521+v522        ; bin: heap_oob 48 89 0c 18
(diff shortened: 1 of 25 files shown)


filetests/isa/x86/binary64.cton
```diff
@@ -594,6 +594,75 @@ ebb0:
...
+    ; asm: movzbq (%rax,%rbx,1),%rcx
+    [-,%rcx]            v531 = uload8_complex.i64 v521+v522         ; bin: heap_oob 48 0f b6 0c 18
+    ; asm: movzbl (%rax,%rbx,1),%ecx
+    [-,%rcx]            v532 = uload8_complex.i32 v521+v522         ; bin: heap_oob 0f b6 0c 18
+    ; asm: movsbq (%rax,%rbx,1),%rcx
+    [-,%rcx]            v533 = sload8_complex.i64 v521+v522         ; bin: heap_oob 48 0f be 0c 18
+    ; asm: movsbl (%rax,%rbx,1),%ecx
+    [-,%rcx]            v534 = sload8_complex.i32 v521+v522         ; bin: heap_oob 0f be 0c 18
+    ; asm: movzwq (%rax,%rbx,1),%rcx
+    [-,%rcx]            v535 = uload16_complex.i64 v521+v522         ; bin: heap_oob 48 0f b7 0c 18
```
sunfishcode: Please vertically line up the comments in the lines above. It doesn't have to match the rest of the file or anything, just make it so that everything grouped together here is consistent.

filetests/parser/tiny.cton
```diff
@@ -158,9 +158,12 @@ ebb0(v1: i32):
     v6 = load.i64 aligned notrap v1
     v7 = load.i64 v1-12
     v8 = load.i64 notrap v1+0x1_0000
+    v9 = load_complex.i64 v1+v2
+    v10 = load_complex.i64 v1+v2+0x1
     store v2, v1
     store aligned v3, v1+12
     store notrap aligned v3, v1-12
+    store_complex v3, v1+v2+0x1
```
sunfishcode: For completeness, can we also test a store_complex with no constant offset?

filetests/parser/tiny.cton
```diff
@@ -158,9 +158,12 @@ ebb0(v1: i32):
     v6 = load.i64 aligned notrap v1
     v7 = load.i64 v1-12
     v8 = load.i64 notrap v1+0x1_0000
+    v9 = load_complex.i64 v1+v2
+    v10 = load_complex.i64 v1+v2+0x1
```
sunfishcode: In the current patch, is it valid to have a load_complex with just one register? If so, we should test that too. But if not, it's not important (for now).

filetests/parser/tiny.cton
```diff
@@ -171,9 +174,12 @@ ebb0(v1: i32):
 ; nextln:     v6 = load.i64 notrap aligned v1
 ; nextln:     v7 = load.i64 v1-12
 ; nextln:     v8 = load.i64 notrap v1+0x0001_0000
+; nextln:     v9 = load_complex.i64 v1+v2
+; nextln:     v10 = load_complex.i64 v1+v2+1
 ; nextln:     store v2, v1
 ; nextln:     store aligned v3, v1+12
 ; nextln:     store notrap aligned v3, v1-12
+; nextln:     store_complex v3, v1+v2+1
```
sunfishcode: As above, test a store_complex with no constant offset.

lib/codegen/meta/cdsl/instructions.py
```diff
@@ -209,7 +209,7 @@ def _verify_polymorphic(self):
                     self.other_typevars = self._verify_ctrl_typevar(tv)
                     self.ctrl_typevar = tv
                     self.use_typevar_operand = True
-            except RuntimeError as e:
+            except Exception as e:
```
sunfishcode: Is there some new exception being raised that needs this change?

lib/reader/src/lexer.rs
```diff
@@ -169,6 +170,25 @@ impl<'a> Lexer<'a> {
...
+            if c.is_digit(10) {
+                return true;
+            }
+            match c {
+                '-' => return true,
+                '+' => return true,
+                '.' => return true,
+                _ => {}
+            }
+            if self.looking_at("NaN") || self.looking_at("Inf") || self.looking_at("sNan") {
```
sunfishcode: "sNan" should be "sNaN".

filetests/parser/tiny.cton
tyler: I didn't implement single value load_complex in this patch, unfortunately.

lib/codegen/meta/cdsl/instructions.py
tyler: I need to fix this differently; thank you for pointing it out. I took a shortcut here and forgot to go back.

What's going on is this:
```
  File "/Users/tyler/.devly/library/repos/isolation/lib/cretonne/lib/codegen/meta/cdsl/instructions.py", line 209, in _verify_polymorphic
    opnum = self.value_opnums[self.format.typevar_operand]
IndexError: tuple index out of range
```
What is happening is that, unlike every other polymorphic instruction (I think), `load_complex` has no "value opnums". All the values are in a `ValueList`. So, when we index the empty `value_opnums` with the `typevar_operand` in the format, which is set to 0, we get an `IndexError`.

As I look at this again, I think it's probably a bug somewhere in the `InstructionFormat` class. I'll poke at that and try to figure out what the problem is.

filetests/isa/x86/binary64.cton
```diff
@@ -594,6 +594,74 @@ ebb0:
...
+    ; asm: mov    %ecx,(%rax,%rbx,1)
+    store_complex v601, v521+v522               ; bin: heap_oob 89 0c 18
+    ; asm: mov    %ecx,0x1(%rax,%rbx,1)
+    store_complex v601, v521+v522+1             ; bin: heap_oob 89 4c 18 01
+    ; asm: mov    %ecx,0x100000(%rax,%rbx,1)
+    store_complex v601, v521+v522+0x1000        ; bin: heap_oob 89 8c 18 00001000
+    ; asm: mov    %ecx,(%rax,%rbx,1)
+    istore32_complex v600, v521+v522            ; bin: heap_oob 89 0c 18
+    ; asm: data16 mov %rcx,(%rax,%rbx,1)
+    istore16_complex v600, v521+v522            ; bin: heap_oob 66 48 89 0c 18
```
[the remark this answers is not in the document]
tyler: You were totally right here. I was confused. I think the latest push fixes this.

sunfishcode:
Great! I believe this is now ready to land!

PR #312, opened 2018-04-27
[WIP] Make ir::ExternalName::LibCall work with cretonne-modules

pchickey:
Right now this PR is a sketch describing how LibCalls should be represented as FuncId, and the lines of responsibilities between Modules and Backends.

I really am just guessing on most of this.

Based on top of #307

### DIFF
--- a/lib/codegen/src/ir/libcall.rs
+++ b/lib/codegen/src/ir/libcall.rs
@@ -103,6 +103,50 @@ impl LibCall {
             _ => return None,
         })
     }
+
+    /// The `LibCall` enum can be converted to an integer representation using the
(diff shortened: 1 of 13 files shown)


sunfishcode:
That's a good idea, but I think we can look at this inside-out too.

The main source of `ExternalName::LibCall`s is codegen. So the main place they show up in the Module code is in the `RelocSink`'s `reloc_external`. For faerie, what `reloc_external` needs to do with a `LibCall` is map it to a string, for example map `LibCall::CeilF32` to "ceilf", so that it can declare a function with that name to faerie and then use it.

If we just do that, then frontends using the Module library wouldn't be able to reference `LibCall` names directly, but I that might be ok. We generally want frontends using instructions like `ceil` rather than calling the libcalls themselves.

PR #316, opened 2018-04-30
Get simplejit building in Windows

jonathandturner:
Wasn't able to test this yet, but it should at least unblock building Cretonne in Windows.


sunfishcode:
Thanks for the patch! It turns out that this overlaps with https://github.com/cretonne/cretonne/pull/316, which was also recently opened. I'm considering waiting for #316 here, even though it's a larger patch, because it seems Cretonne won't really be meaningfully usable on Windows until the calling convention part of #316 is ready too. Does that make sense?

jonathandturner:
Totally.  No problem 😄

Issue #292, opened 2018-03-29
Document that i8 and i16 arithmetic support is incomplete


sunfishcode:
Because WebAssembly doesn't need them, support for i8 and i16 arithmetic is fairly incomplete, leading to bugs like https://github.com/Cretonne/cretonne/issues/176. For now, we should add a notice to the documentation advising users this.

sunfishcode:
This is fixed in #299 and #320.

PR #320, opened 2018-05-02
document that low bitsize integers don't have complete arithmetic support

TyOverby:
I couldn't figure out how to get docs on the arithmetic functions, but I got it on the type!

### DIFF
--- a/lib/codegen/meta/cdsl/types.py
+++ b/lib/codegen/meta/cdsl/types.py
@@ -202,11 +202,12 @@ class IntType(LaneType):
         assert bits > 0, 'IntType must have positive number of bits'
         warning = ""
         if bits < 32:
-            warning = "WARNING: arithmetic on {}-bit integers is incomplete".format(bits)
+            warning += "\nWARNING: "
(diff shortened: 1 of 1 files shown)


TyOverby:
closes https://github.com/cretonne/cretonne/issues/292

TyOverby:
Gaah, I really should add that to my global git ignore.  Thanks @phase!

sunfishcode:
#292 was fixed by #299, which it seems didn't get closed, however your patch here also adds text on the same lines as the types, which seems nice to have in addition, so I think it'd be good to merge too.

Could you update this PR to remove the spurious file, and fix the 80-column line so that pyflake is happy? Thanks!

TyOverby:
Done (I think)

sunfishcode:
These new failures are due to changes on the CI machine; see https://github.com/cretonne/cretonne/issues/323 for details. https://github.com/cretonne/cretonne/pull/321 should land with a fix soon.

sunfishcode:
Looks good; thanks!

PR #321, opened 2018-05-02
cretonne-faerie: add a translation mechanism for LibCalls

pchickey:
Exposes a new field `libcall_names: Vec<String>` in `FaerieBuilder`, and a new method `FaerieBuilder::default_libcall_names` to provide sensible default values for those libcalls.

The `FaerieTrapSink` looks up any libcall in the names provided, imports that function, and makes a link to it, instead of panicking on encountering a libcall.

### DIFF
--- a/lib/faerie/src/backend.rs
+++ b/lib/faerie/src/backend.rs
@@ -42,6 +42,11 @@ impl FaerieBuilder {
     ///
     /// `collect_traps` setting determines whether trap information is collected in a
     /// `FaerieTrapManifest` available in the `FaerieProduct`.
+    ///
+    /// `libcall_names` vector provides a way to translate `cretonne_codegen`'s `ir::LibCall` enum
+    /// to symbols. LibCalls are inserted in the IR as part of the legalization for certain
+    /// floating point instructions, and for stack probes. If you don't know what to use for this
+    /// argument, use `FaerieBuilder::default_libcall_names()`.
     pub fn new(
         isa: Box<TargetIsa>,
         name: String,
@@ -82,7 +87,6 @@ impl FaerieBuilder {
             "round".to_owned(),
         ]
     }
-
 }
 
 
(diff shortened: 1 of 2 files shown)


lib/faerie/src/backend.rs
```diff
@@ -59,8 +66,49 @@ impl FaerieBuilder {
...
             collect_traps,
+            libcall_names,
         })
     }
+
+    /// A default set of names for `ir::LibCall`. Indexed by `ir::LibCall` cast to an integer
+    /// representation. A function by this name is imported into the object as part of the
+    /// translation of a `ir::ExternalName::LibCall` variant. Calls to a LibCall should only be
+    /// inserted into the IR by the `cretonne_codegen` legalizer pass.
+    pub fn default_libcall_names() -> Vec<String> {
```
sunfishcode: I know using an array was my suggestion, but looking at it here, I'm worried that it's somewhat fragile, if we ever reorder the libcalls. Would you mind going with your idea, to pass in a closure, instead?

lib/faerie/src/backend.rs
```diff
@@ -59,8 +66,49 @@ impl FaerieBuilder {
...
+        vec![
+            "__cretonne_probestack".to_owned(),
+            "ceilf".to_owned(),
+            "ceil".to_owned(),
+            "floorf".to_owned(),
+            "floor".to_owned(),
+            "truncf".to_owned(),
+            "trunc".to_owned(),
+            "roundf".to_owned(),
+            "round".to_owned(),
```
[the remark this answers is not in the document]
pchickey: Thanks, I didn't know that.

lib/faerie/src/backend.rs
pchickey: I made this change. Its the first time I dealt with closure lifetimes like this so I'm not 100% sure I got it right, please let me know.

pchickey:
The installation of `mypy` was pinned to version 0.521. The latest version, 0.600, encounters a bunch of errors at the moment. Detailed in bug #323

lib/faerie/src/backend.rs
```diff
@@ -59,16 +70,36 @@ impl FaerieBuilder {
...
+        Box::new(move |libcall| match libcall {
+            ir::LibCall::Probestack => "__cretonne_probestack".to_owned(),
+            ir::LibCall::CeilF32 => "ceilf".to_owned(),
+            ir::LibCall::CeilF64 => "ceil".to_owned(),
+            ir::LibCall::FloorF32 => "floorf".to_owned(),
+            ir::LibCall::FloorF64 => "floor".to_owned(),
+            ir::LibCall::TruncF32 => "truncf".to_owned(),
+            ir::LibCall::TruncF64 => "trunc".to_owned(),
+            ir::LibCall::NearestF32 => "nearbyint".to_owned(),
+            ir::LibCall::NearestF64 => "nearbyintf".to_owned(),
```
sunfishcode: Thanks for making this change. As we get more experience with this, it may want to evolve, but this looks good for now.

PR #322, opened 2018-05-02
Issue 311 - Add a pass to make NaN bits deterministic.

### LINKED ISSUE #311: Add a flag to make NaN bits deterministic
Floating-point instructions on x86, ARM, and others produce NaN values with differing bit patterns. To support users that require fully deterministic computation, we should add a mode which enables automatic canonicalization of NaN values.

In particular, for WebAssembly, NaN bit patterns are the only non-determinism in the language. So if we had an option to make NaN bit patterns deterministic for users that want that, we could run WebAssembly entirely deterministically (other than resource limitations, and eventually threads and SIMD which are expected to come with their own nondeterminism). This isn't required by the WebAssembly spec, but it's not prohibited either.

#### Background: NaN bit patterns

Floating-point values are represented as a sign bit, an exponent, and a significand (sometimes called the mantissa field or the fraction field). See [here](https://en.wikipedia.org/wiki/Single-precision_floating-point_format#IEEE_754_single-precision_binary_floating-point_format:_binary32) for an illustration of what this looks like for `f32`; `f64` is the same but with bigger fields.

NaN is defined as a special case: when the exponent field is all ones, and the significand has at least one non-zero bit, it's special-case defined to mean NaN. This means that there are many possible bit patterns which mean NaN.

IEEE 754 doesn't fully specify what the bits of NaN values should be when the hardware generates them. And, it turns out, popular hardware platforms ended up doing different things: when a NaN is generated on x86, it has a different bit pattern than when a NaN is generated on ARM, and so on.

And, the WebAssembly spec chose to expose this nondeterminsim, rather than take on extra overhead to hide it, so WebAssembly programs may see different NaN bit patterns depending on the underlying platform they're running on.

#### Background: Where NaN bit patterns matter

For most arithmetic instructions, all that matters is whether or not a value is *some* NaN, and not *which* specific bit pattern it is. So most of the time, the bit patterns don't matter. However, they are observable in a few specific circumstances: if a NaN is stored to memory, and the bytes of memory are later inspected, if a floating-point value is [bitcasted](https://cretonne.readthedocs.io/en/latest/langref.html#inst-bitcast) to integer, if it is passed to a call or returned from a call to another function (which we assume might store the value), if is the operand of a [copysign](https://cretonne.readthedocs.io/en/latest/langref.html#inst-fcopysign), or if it's the operand of certain target-specific instructions.

#### Canonicalizing NaNs

Canonicalizing here means checking to see if a value is *some* NaN, and if so, replacing it with a specific hand-crafted "canonical" or "default" NaN, so that the result has a deterministic bit pattern. This typically adds some overhead, and the benefit is determinism.

To test whether a value is NaN, we can use floating-point equality comparisons. NaN is defined to be not equal to itself, while all other values (including Infinity) are defined to be equal to themselves, so `x != x` is an efficient way to test whether `x` is *some* NaN.

A canonicalization sequence can consist of such a [comparison](https://cretonne.readthedocs.io/en/latest/langref.html#inst-fcmp), followed by a [select](https://cretonne.readthedocs.io/en/latest/langref.html#inst-select), with an [f32const](https://cretonne.readthedocs.io/en/latest/langref.html#inst-f32const) or [f64const](https://cretonne.readthedocs.io/en/latest/langref.html#inst-f64const) to provide the canonical alternative.

#### Cretonne implementation

The first step can be to insert canonicalizing code after every floating-point arithmetic instruction, to eliminate the nondeterminism immediately whenever it appears. That would be a good way to start, and though it would have a fair amount of overhead, it would likely still be much faster than using software floating-point libraries.

Not all users will want this though, so we should create a new pass (see [preopt.rs](https://github.com/cretonne/cretonne/blob/master/lib/codegen/src/preopt.rs) for example of how to write a pass) for this, which can be easily enabled or disabled. To start with, the pass can just visit all instructions in a function, and replace each floating-point arithmetic instruction with a sequence that does that operation followed by a canonicalization step.

For these purposes, `fneg`, `fabs`, and `fcopysign` are not considered to count as "arithmetic" instructions, because they're defined to just operate in terms of the sign bit and leave all other bits alone.

In follow-up steps, we can make the pass more clever and avoid inserting canonicalization code in places where bit patterns don't "escape" computations.

data-pup:
This PR intends to fix #311. I have marked some lines that I was unsure about with 'FIXUP' comments, which I would not mind fixing if you have opinions about them. Otherwise, I can simply remove the comments and squash the commits down :)

This passes all of the existing test cases in the `codegen` crate, as well as the top-level crate. I can add more tests if you think that is warranted, but might need some guidance on that front.

Thanks again to @sunfishcode for the guidance, I had a lot of fun doing this!

### DIFF
--- a/lib/codegen/meta/base/settings.py
+++ b/lib/codegen/meta/base/settings.py
@@ -91,6 +91,16 @@ enable_float = BoolSetting(
         """,
         default=True)
 
+enable_nan_canonicalization = BoolSetting(
+        """
+        Enable NaN canonicalization
+
+        This replaces NaNs with a single canonical value, for users requiring
+        entirely deterministic WebAssembly computation. This is not required
+        by the WebAssembly spec, so it is not enabled by default.
+        """,
+        default=False)
+
 enable_simd = BoolSetting(
         """Enable the use of SIMD instructions.""",
         default=True)
--- a/lib/codegen/src/context.rs
+++ b/lib/codegen/src/context.rs
@@ -18,6 +18,7 @@ use isa::TargetIsa;
 use legalize_function;
 use licm::do_licm;
 use loop_analysis::LoopAnalysis;
+use nan_canonicalization::do_nan_canonicalization;
 use postopt::do_postopt;
 use preopt::do_preopt;
 use regalloc;
@@ -124,6 +125,9 @@ impl Context {
         if isa.flags().opt_level() != OptLevel::Fastest {
             self.preopt(isa)?;
         }
+        if isa.flags().enable_nan_canonicalization() {
+            self.canonicalize_nans(isa)?;
+        }
         self.legalize(isa)?;
         if isa.flags().opt_level() != OptLevel::Fastest {
             self.postopt(isa)?;
@@ -210,6 +214,12 @@ impl Context {
         do_preopt(&mut self.func);
         self.verify_if(isa)?;
         Ok(())
+    }
+
+    /// Perform NaN canonicalizing rewrites on the function.
+    pub fn canonicalize_nans(&mut self, isa: &TargetIsa) -> CtonResult {
+        do_nan_canonicalization(&mut self.func);
+        self.verify_if(isa)
     }
 
     /// Run the legalizer for `isa` on the function.
--- a/lib/codegen/src/nan_canonicalization.rs
+++ b/lib/codegen/src/nan_canonicalization.rs
@@ -1,0 +1,81 @@
+//! A NaN-canonicalizing rewriting pass. Patch floating point arithmetic
+//! instructions that may return a NaN result with a sequence of operations
+//! that will replace nondeterministic NaN's with a single canonical NaN value.
+
+use cursor::{Cursor, FuncCursor};
+use ir::{Function, Inst, InstBuilder, InstructionData, Opcode, Value};
+use ir::condcodes::FloatCC;
+use ir::immediates::{Ieee32, Ieee64};
+use ir::types;
+use ir::types::Type;
+use timing;
+
+// Canonical 32-bit and 64-bit NaN values.
+static CANON_32BIT_NAN: u32 = 0b01111111110000000000000000000000;
+static CANON_64BIT_NAN: u64 = 0b0111111111111000000000000000000000000000000000000000000000000000;
+
+/// Perform the NaN canonicalization pass.
+pub fn do_nan_canonicalization(func: &mut Function) {
+    let _tt = timing::canonicalize_nans();
+    let mut pos = FuncCursor::new(func);
+    while let Some(_ebb) = pos.next_ebb() {
+        while let Some(inst) = pos.next_inst() {
+            if is_fp_arith(&mut pos, inst) {
+                add_nan_canon_seq(&mut pos, inst);
+            }
+        }
+    }
+}
+
+/// Returns true/false based on whether the instruction is a floating-point
+/// arithmetic operation. This ignores operations like `fneg`, `fabs`, or
+/// `fcopysign` that only operate on the sign bit of a floating point value.
+fn is_fp_arith(pos: &mut FuncCursor, inst: Inst) -> bool {
+    match pos.func.dfg[inst] {
+        InstructionData::Unary { opcode, .. } => {
+            opcode == Opcode::Ceil || opcode == Opcode::Floor || opcode == Opcode::Nearest ||
+                opcode == Opcode::Sqrt || opcode == Opcode::Trunc
+        }
+        InstructionData::Binary { opcode, .. } => {
+            opcode == Opcode::Fadd || opcode == Opcode::Fdiv || opcode == Opcode::Fmax ||
+                opcode == Opcode::Fmin || opcode == Opcode::Fmul ||
+                opcode == Opcode::Fsub
+        }
+        InstructionData::Ternary { opcode, .. } => opcode == Opcode::Fma,
+        _ => false,
+    }
+}
+
+/// Append a sequence of canonicalizing instructions after the given instruction.
+fn add_nan_canon_seq(pos: &mut FuncCursor, inst: Inst) {
+    // Select the instruction result, result type. Replace the instruction
+    // result and step forward before inserting the canonicalization sequence.
+    let val = pos.func.dfg.first_result(inst);
+    let val_type = pos.func.dfg.value_type(val);
+    let new_res = pos.func.dfg.replace_result(val, val_type);
+    let _next_inst = pos.next_inst().expect("EBB missing terminator!");
+
+    // Insert a comparison instruction, to check if `inst_res` is NaN. Select
+    // the canonical NaN value if `val` is NaN, assign the result to `inst`.
+    let is_nan = pos.ins().fcmp(FloatCC::NotEqual, new_res, new_res);
+    let canon_nan = insert_nan_const(pos, val_type);
+    pos.ins().with_result(val).select(
+        is_nan,
+        canon_nan,
+        new_res,
+    );
+
+    pos.prev_inst(); // Step backwards so the pass does not skip instructions.
+}
+
+/// Insert a canonical 32-bit or 64-bit NaN constant at the current position.
+fn insert_nan_const(pos: &mut FuncCursor, nan_type: Type) -> Value {
+    match nan_type {
+        types::F32 => pos.ins().f32const(Ieee32::with_bits(CANON_32BIT_NAN)),
+        types::F64 => pos.ins().f64const(Ieee64::with_bits(CANON_64BIT_NAN)),
+        _ => {
+            // Panic if the type given was not an IEEE floating point type.
+            panic!("Could not canonicalize NaN: Unexpected result type found.");
+        }
+    }
+}
--- a/lib/codegen/src/timing.rs
+++ b/lib/codegen/src/timing.rs
@@ -73,6 +73,8 @@ define_passes!{
     prologue_epilogue: "Prologue/epilogue insertion",
     binemit: "Binary machine code emission",
     layout_renumber: "Layout full renumbering",
+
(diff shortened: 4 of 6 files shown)


lib/codegen/src/context.rs
```diff
@@ -124,6 +125,7 @@ impl Context {
         if isa.flags().opt_level() != OptLevel::Fastest {
             self.preopt(isa)?;
         }
+        self.canonicalize_nans(isa)?; // FIXUP: Should this be wrapped in a conditional?
```
sunfishcode: Yes, this should be guarded by `if isa.flags().enable_nan_canonicalization`, so that we don't do the canonicalization unless we're asked to.

data-pup:
I can take care of the linting issues, pardon that. If there are any issues you see regarding the code overall though, let me know!

lib/codegen/src/nan_canonicalization.rs
```diff
@@ -0,0 +1,120 @@
...
+use ir::condcodes::FloatCC;
+use ir::immediates::{Ieee32, Ieee64};
+use ir::types;
+use ir::types::Type;
+use timing;
+
+// Canonical 32-bit and 64-bit NaN values.
+static CANON_32BIT_NAN: u32 = 0b01111111100000000000000000000001;
+static CANON_64BIT_NAN: u64 =
+    0b0111111111110000000000000000000000000000000000000000000000000001;
```
sunfishcode: We need to set the most significant bit of the significand field to 1, so that these are quiet NaNs rather than signaling NaNs. (For completeness, I'm ignoring pre-NAN2008 MIPS here.) Even though IEEE 754 doesn't say much here, it does say that a NaN returned from arithmetic is a quiet NaN.

It would also be good to set the least significant bit to 0, so that we're consistent with common hardware implementations such as [ARM's "default NaN" mode](http://infocenter.arm.com/help/index.jsp?topic=/com.arm.doc.ddi0274h/Babffifj.html).

lib/codegen/src/nan_canonicalization.rs
```diff
@@ -0,0 +1,120 @@
...
+        }
+    }
+}
+
+/// Returns true/false based on whether the instruction is a floating-point
+/// arithmetic operation. This ignores operations like `fneg`, `fabs`, or
+/// `fcopysign` that only operate on the sign bit of a floating point value.
+fn is_fp_arith(pos: &mut FuncCursor, inst: Inst) -> bool {
+    // FIXUP: Should `ceil`, `floor`, `trunc`, `nearest`, and
+    // immediate constants be considered as well?
```
sunfishcode: Yes for `ceil`, `floor`, `trunc`, and `nearest`, and no for immediate constants.

lib/codegen/meta/base/settings.py
```diff
@@ -91,6 +91,12 @@
         """,
         default=True)
 
+enable_nan_canonicalization = BoolSetting(
+        """
+        Enable NaN canonicalization
+        """,
+        default=True)
```
sunfishcode: I think the default should be `False`, since wasm doesn't require this behavior.

lib/codegen/meta/base/settings.py
```diff
@@ -91,6 +91,12 @@
         """,
         default=True)
 
+enable_nan_canonicalization = BoolSetting(
+        """
+        Enable NaN canonicalization
```
sunfishcode: It'd be good to briefly explain what canonicalization means here, and mention the word "deterministic", to help people who need that find it.

lib/codegen/src/nan_canonicalization.rs
data-pup: Awesome, can do :)

lib/codegen/src/nan_canonicalization.rs
```diff
@@ -0,0 +1,110 @@
...
+        InstructionData::Ternary { opcode, .. } => opcode == Opcode::Fma,
+        _ => false,
+    }
+}
+
+/// Patch instructions that may result in a NaN result with operations to
+/// identify and replace NaN's with a single canonical NaN value.
+fn add_nan_canon_instrs(pos: &mut FuncCursor, inst: Inst) {
+    // FIXUP: Are these unwraps safe? An arithmetic operation should never be
+    // found at the end of an EBB, if I understand this correctly.
```
sunfishcode: Yeah, the unwraps are safe, though this would be a place where using `.expect("...")` would be a little friendlier than `.unwrap()` because forgetting to add a terminator to an ebb is a bug that does sometimes happen.

data-pup:
Thanks for the review by the way! I'll add those changes, the commit that was just pushed is fixing the formatting problems. I'll take care of the other points you mentioned tonight.

lib/codegen/src/timing.rs
```diff
@@ -73,6 +73,8 @@ define_passes!{
     prologue_epilogue: "Prologue/epilogue insertion",
     binemit: "Binary machine code emission",
     layout_renumber: "Layout full renumbering",
+
+    canonicalize_nans: "Canonicalization of NaN's",
```
sunfishcode: Style nit: no apostrophe. "NaNs" is the plural of "NaN".

lib/codegen/src/nan_canonicalization.rs
```diff
@@ -0,0 +1,110 @@
...
+            let canon_nan = Ieee64::with_bits(CANON_64BIT_NAN);
+            pos.ins().f64const(canon_nan);
+        }
+        _ => {} // FIXUP: Should this panic or throw some sort of Error?
+    }
+}
+
+/// Given some instruction that could potentially return a nondeterministic
+/// NaN value, determine if the operation is using 32-bit or 64-bit floating
+/// point numbers, and return the corresponding NaN value.
```
sunfishcode: You can get this type from the instruction's result type, rather than digging into the operands. `func.dfg.value_type(func.dfg.first_result(inst))` or so. Which is simpler, but also, that's the value that you'll be canonicalizing.

lib/codegen/src/nan_canonicalization.rs
```diff
@@ -0,0 +1,110 @@
...
+    // Insert a comparison function, and a canonical NaN constant. Select
+    // the constant value, and move forward to the next instruction.
+    let is_nan: Value = pos.ins().fcmp(FloatCC::NotEqual, inst_res, inst_res);
+    insert_nan_const(pos, inst);
+    let canon_nan_instr: Inst = pos.prev_inst().unwrap();
+    let canon_nan_res: Value = pos.func.dfg.first_result(canon_nan_instr);
+    pos.goto_inst(next_inst);
+
+    // Insert a select instruction to canonicalize the NaN value.
+    pos.ins().select(is_nan, canon_nan_res, inst_res);
```
sunfishcode: Ok, so now let's talk about wiring in the new result value :-). The tricky thing is that we don't want to rewire *all* uses of the old value, because we need the fcmp and select to continue to use the old value.

As I mentioned earlier, there are a few ways to do it, but here's what I think is best: You can use `dfg`'s `replace_result` to replace the result value of the arithmetic instruction with a new `Value`. Then you can use `pos.ins()`'s `with_result` to when inserting the `select` using the original `Value`.

data-pup:
Thanks! Wrapping up edits for the previous review notes right now, I will take a look into wiring in the new result once I am done with that :) if I have any questions I’ll follow up with you here or on irc

data-pup:
Took a swing at this, having a little bit of trouble figuring out why the CI build is failing. When I run the `test-all.sh` script locally, everything seems to work. I've installed flake8 and mypy with pip3, but the check script in `lib/codegen/meta` does not seem to find them successfully. Aside from that though, I didn't alter any of the files that seem to be causing the build to fail? Not sure how to proceed from here with regards to that :(

I also wanted to check that I had wired up the `replace_result` and `with_result` parameters correctly, I was a little worried I might have these mixed up.

lib/codegen/src/nan_canonicalization.rs
```diff
@@ -0,0 +1,81 @@
...
+    let val = pos.func.dfg.first_result(inst);
+    let val_type = pos.func.dfg.value_type(val);
+    let new_res = pos.func.dfg.replace_result(val, val_type);
+    let _next_inst = pos.next_inst().expect("EBB missing terminator!");
+
+    // Insert a comparison instruction, to check if `inst_res` is NaN. Select
+    // the canonical NaN value if `val` is NaN, assign the result to `inst`.
+    let is_nan = pos.ins().fcmp(FloatCC::NotEqual, val, val);
+    let canon_nan = insert_nan_const(pos, val_type);
+    pos.ins().with_result(new_res).select(
```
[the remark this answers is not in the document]
data-pup: Ah, thanks. So, `fcmp` and `select` should use `new_res` as a parameter, and I should pass `val` to `with_result`? I can do that right now.

lib/codegen/src/nan_canonicalization.rs
```diff
@@ -0,0 +1,81 @@
...
+use ir::{Function, Inst, InstBuilder, InstructionData, Opcode, Value};
+use ir::condcodes::FloatCC;
+use ir::immediates::{Ieee32, Ieee64};
+use ir::types;
+use ir::types::Type;
+use timing;
+
+// Canonical 32-bit and 64-bit NaN values.
+static CANON_32BIT_NAN: u32 = 0b01111111100000000000000000000001;
+static CANON_64BIT_NAN: u64 = 0b0111111111110000000000000000000000000000000000000000000000000001;
```
sunfishcode: In case github swallowed my comment here, we need to use quiet NaN bit-patterns here, so the most significant bit of the significand field needs to be 1. And then, we should make the last significant bit 0, because that matches what several other systems consider to be the canonical NaN bit pattern.

data-pup:
Neat! I can apply that change today, thank you for the help :)

sunfishcode:
Looks good!

PR #325, opened 2018-05-03
x86 recipes: emit StackOverflow trap for all sp-relative loads and stores

pchickey:
This is the analog of the HeapOutOfBounds traps emitted on other loads and stores.

I'm not 100% sure that I caught every possible case.


sunfishcode:
Looks good; just one test needs updating.

PR #331, opened 2018-05-12
Use SVG version of Gitter icon in README.md

jdanford:


sunfishcode:
Is there a quick summary of the advantage of using SVG for something like this?

sunfishcode:
Oh, I see now. This fixes the issue with the png badge being blurry. Thanks!

Issue #317, opened 2018-04-30
Optimize `0.0` constants


sunfishcode:
Cretonne currently lowers all floating-point constant instructions by emitting integer constants with the desired bit patterns and then moving them into floating-point registers. The code that does this is [here](https://github.com/cretonne/cretonne/blob/master/lib/codegen/src/legalizer/mod.rs#L245).

For reference, the `expand_fconst` function used here is registered [here](https://github.com/cretonne/cretonne/blob/master/lib/codegen/meta/base/legalize.py#L78).

However, most CPUs have faster ways to set registers to zero. For example, on x86, one can do `xorps %xmm0, %xmm0`. Xor'ing a register with itself seems roundabout, but it turns out to be common enough that CPUs recognize it specially, and often optimize it in register renaming rather than actually executing it as an instruction.

A simple way to implement this for x86 would be to define an encoding rule for x86 similar to the rules for [integer constants](https://github.com/cretonne/cretonne/blob/master/lib/codegen/meta/isa/x86/encodings.py#L189), using an `instp` predicate to test for the value of 0. Be aware that in IEEE 754 floating-point, negative zero compares equal to zero, but is a distinct bit pattern; the [`is_sign_positive`](https://doc.rust-lang.org/std/primitive.f64.html#method.is_sign_positive) function may be useful in distinguishing between the two.

Cretonne has encodings for xorps:
https://github.com/cretonne/cretonne/blob/master/lib/codegen/meta/isa/x86/encodings.py#L595
https://github.com/cretonne/cretonne/blob/master/lib/codegen/meta/isa/x86/recipes.py#L332
which may be useful examples, though the recipe as written is only usable for `Binary` operators, so I think the best way to proceed here is to add a new recipe.

Here's a simple testcase for f32:
```
function %foo() -> f32 fast {
ebb0:
    v0 = f32const 0.0
    return v0
}
```

data-pup:
This issue looks really neat, I would love to try and take care of this. If you have any other advice, feel free to send it my way :)

sunfishcode:
Sounds good!

The first step here is to define a recipe in lib/codegen/meta/isa/x86/recipes.py. Above I linked to another recipe that uses xorps but does so for `Binary` operators. So you'll need to make a similar recipe, but for `UnaryImm` operators, with no declared register inputs, that uses its output register for the x86 instruction encoding inputs. And of course, please ask questions if anything is unclear, from the x86 encoding bits to the python DSL bits.

[...]

PR #334, opened 2018-05-16
Optimizing 0.0 Constants (WIP/Question)

data-pup:
Hi! This is still a little bit of a work in progress, but I wanted to open up a PR and ask some questions about next steps towards solving #317. As of now, I have added a new predicate, and the scaffold of a recipe.

I'm having a little bit of trouble understanding how I should get this registered, and how to notate the `fconst` line in the new test file. Should I be expecting to add some new logic to the `expand_fconst` function, or is this going to involve adding a new function to `legalize/mod.rs`?

Thank you in advance for your assistance! :)

### DIFF
--- a/filetests/isa/x86/optimized-zero-constants.cton
+++ b/filetests/isa/x86/optimized-zero-constants.cton
@@ -1,0 +1,9 @@
+test binemit
+isa x86
+
+function %foo() -> f32 fast {
+ebb0:
+  ; asm: xorps %xmm0 %xmm0
+  [-,%xmm0]    v0 = f32const 0.0     ; bin: 0f 57 c0
+  return v0
+}
--- a/lib/codegen/meta/cdsl/predicates.py
+++ b/lib/codegen/meta/cdsl/predicates.py
@@ -262,6 +262,21 @@ class IsEqual(FieldPredicate):
         self.value = value
 
 
+class IsZero(FieldPredicate):
+    """
+    Instruction predicate that checks if an immediate instruction format field
+    is equal to zero.
+
+    :param field: `FormatField` to be checked.
+    :param value: The constant value to check.
+    """
+
+    def __init__(self, field):
+        # type: (FormatFiend) -> None
+        # FIXUP: Is this calling `super` correctly?
+        super(IsZero, self).__init__(field, 'is_zero', ())
+
+
 class IsSignedInt(FieldPredicate):
     """
     Instruction predicate that checks if an immediate instruction format field
--- a/lib/codegen/meta/cdsl/predicates.py
+++ b/lib/codegen/meta/cdsl/predicates.py
@@ -262,7 +262,7 @@ class IsEqual(FieldPredicate):
         self.value = value
 
 
-class IsZero(FieldPredicate):
+class IsZero32BitFloat(FieldPredicate):
     """
     Instruction predicate that checks if an immediate instruction format field
     is equal to zero.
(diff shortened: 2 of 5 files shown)


lib/codegen/meta/cdsl/predicates.py
```diff
@@ -262,6 +262,21 @@ def __init__(self, field, value):
...
+    Instruction predicate that checks if an immediate instruction format field
+    is equal to zero.
+
+    :param field: `FormatField` to be checked.
+    :param value: The constant value to check.
+    """
+
+    def __init__(self, field):
+        # type: (FormatFiend) -> None
+        # FIXUP: Is this calling `super` correctly?
```
data-pup: Wasn't totally sure about the type comment here, and wanted to make sure this looked alright to you.

lib/codegen/meta/isa/x86/recipes.py
```diff
@@ -544,6 +544,16 @@ def floatccs(iform):
...
         ''')
 
+# XX /n Unary with floating point immediate.
+# FIXUP: ^ I am unusure about the notation above.
+fimm_z = TailRecipe(
+    'fimm_z', UnaryImm, size=4, ins=(), outs=FPR,
+    instp=IsZero(UnaryImm.imm),
+    emit='''
+        unimplemented!();
+    ''')
```
data-pup: I was a little unsure about where to go from here. Should I be using a `PUT_OP` statement here?

lib/codegen/meta/isa/x86/recipes.py
```diff
@@ -544,6 +544,16 @@ def floatccs(iform):
         sink.put8(imm as u64);
         ''')
 
+# XX /n Unary with floating point immediate.
+# FIXUP: ^ I am unusure about the notation above.
```
sunfishcode: The notation above is right, but the comment should be more specific, because this recipe will only be usable for 0.0 and not other floating-point immediates.

lib/codegen/meta/isa/x86/recipes.py
sunfishcode: Yeah, I think the simplest approach here would be to stay as close to the `fa` recipe as possible.

data-pup:
Thanks for the review notes @sunfishcode! :D

filetests/isa/x86/optimized-zero-constants.cton
```diff
@@ -0,0 +1,9 @@
+test binemit
+isa x86
```
sunfishcode: Currently in cretonne, `x86` defaults to 32-bit x86. Since the encoding.py code below only handles 64-bit x86 at the moment, it'd be good to add "set is_64bit=1" here so that we can get closer to plumbing this feature all the way through, and we can go back and add 32-bit support later.

lib/codegen/meta/isa/x86/encodings.py
```diff
@@ -204,6 +205,14 @@ def enc_i32_i64_ld_st(inst, w_bit, recipe, *args, **kwargs):
 # bool constants.
 enc_both(base.bconst.b1, r.pu_id_bool, 0xb8)
 
+# floating-point constants equal to 0.0 can be encoded using `xorps`.
+# X86_32.enc(base.iconst.i32, *r.fimm_z(0x57),
+#             instp=IsZero(UnaryImm.imm))
+X86_64.enc(base.f32const, *r.f32imm_z(0x57),
+            instp=IsZero32BitFloat(UnaryIeee32.imm))
+X86_64.enc(base.f64const, *r.f64imm_z(0x57),
+            instp=IsZero64BitFloat(UnaryIeee64.imm))
```
sunfishcode: This looks good. Two notes:
 - It looks like this is missing the `0x0f` byte of the xorps instruction.
 - We'll also need to add `rex` forms for `X86_64`, to support xmm registers xmm8 through xmm15. There are other examples of adding `.rex(...)` calls in this file. Note that the `rex` form needs to be declared before the non-`rex` form.

lib/codegen/meta/isa/x86/recipes.py
```diff
@@ -544,6 +545,24 @@ def floatccs(iform):
         sink.put8(imm as u64);
         ''')
 
+# XX /n Unary with floating point 32-bit immediate equal to zero.
+f32imm_z = TailRecipe(
+    'f32imm_z', UnaryIeee32, size=4, ins=(), outs=FPR,
```
sunfishcode: The declared size will need to be updated too. Note that this is the size of the "tail" portion of the encoding, which is just the one modrm byte.

lib/codegen/meta/isa/x86/encodings.py
sunfishcode: The recipe uses a `*args`, so you can pass the 0x0f as a separate argument, like `*r.f32imm_z(0x0f, 0x57)`

data-pup:
I am going to go ahead and close this, since the branch was named for asking questions, and I have a branch will all of my changes cleaned up and squashed together :) Thanks again so much for the help @sunfishcode!

PR #345, opened 2018-05-20
Allow wasm embedders to reject wasm modules with unsupported features.

sunfishcode:
Define `WasmError` (and an accompanying `WasmResult`) to represent
errors translating WebAssembly functions. Make `translate_call` and
related functions return `WasmResult`s so that embedders have the
flexibility to reject features they don't support.

Move `InvalidInput` out of `CtonError` and into `WasmError`, where it's
now named `InvalidWebAssembly`, as it's a WebAssembly-specific error
condition. Also extend it to preserve the original error message and
bytecode offset.

@lachlansneff Does this look like it would work for your use case?

### DIFF
--- a/lib/codegen/src/result.rs
+++ b/lib/codegen/src/result.rs
@@ -7,13 +7,6 @@ use verifier;
 /// When Cretonne fails to compile a function, it will return one of these error codes.
 #[derive(Fail, Debug, PartialEq, Eq)]
 pub enum CtonError {
-    /// The input is invalid.
-    ///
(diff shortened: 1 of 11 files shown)


lachlansneff:
This looks great! For now, I'll be using it to statically reject wasm binaries that try to reserve access to resources that they don't have the necessary permissions for.

sunfishcode:
Ok, great! I'll proceed with this for now, and we can adjust it as needed.

PR #346, opened 2018-05-21
Define a "table" concept.

sunfishcode:
"Table" is to WebAssembly tables as "Heap" is to WebAssembly linear
memories.

This adds a `table_addr` instruction, which is the table analog to `heap_addr` for linear memories, and adds builtin bounds checking, allowing cretonne to provide more of the WebAssembly `call_indirect` implementation in the core codegen crate.


sunfishcode:
That's a good question.

Some engines implement signature checking by passing an extra argument, which is an integer value uniquely identifying a signature, and checking it in the callee. That requires the callee is prepared for an extra argument, and it requires cross-module coordination to ensure that the signature values match. That keeps per-call code size low, and doesn't require putting signatures in the tables, but requires some coordination that cretonne-codegen and cretonne-wasm themselves aren't in a position to do.

I'm not sure how to proceed here yet.

Techcable:
Is there a reason you removed the `OutOfBounds` error? I use it in my compiler and it seems to work great. Is it considered bad practice or something like that?

sunfishcode:
It's renamed to `TableOutOfBounds`, for consistency with `HeapOutOfBounds`.

Are you using `OutOfBounds` for something other than table bounds checking?

sunfishcode:
@Techcable I've now added a patch which restores the `OutOfBounds` trap code.

sunfishcode:
I've now significantly revised and updated this patch.

I decided to defer modelling the multiple-entrypoint optimization for call_indirect signature checking for now. Users that need that can still do it themselves, but teaching Cranelift itself about multiple-entry functions is more complex.

PR #348, opened 2018-05-25
Update to rustfmt-preview

### LINKED ISSUE #277: Update to rustfmt-preview
"Old" rustfmt 0.9.0 is getting long in the tooth, and is at odds with some of clippy's linting rules. Additionally, as implemented in #276, Clippy must be used via a nightly cargo.

So, the real issue is, do require a nightly toolchain be present on the system so that an up-to-date rustfmt (and clippy) can be used?

sunfishcode:
This updates to the current stable rustfmt-preview. I fixed the problems in the CI scripts, and we now get the right version. 

This also removes the check-rustfmt.sh script, and just relies on `cargo +stable fmt --all` to select the right version. Let's see how that goes.

This fixes #277.


sunfishcode:
It took several more tweaks to get the CI script to work reliable, but it's working now. So let's do this!

PR #356, opened 2018-05-29
Use "nightly" feature of raw-cpuid when possible.

lachlansneff:
The raw-cpuid library links in a c-library to call the `cpuid` instruction unless the nightly feature is enabled. When the `core` feature is enabled, enable the "nightly" feature for `raw-cpuid` to avoid that.


sunfishcode:
Thanks!

Issue #349, opened 2018-05-25
Stack limit checking


sunfishcode:
Cretonne is currently capable of detecting stack overflow when the stack is allocated with a guard page at the end, and it performs stack probes as needed when allocating large frames. Most popular platforms have guard pages, and this allows them to handle stack overflow with very little overhead in the common case. However, in environments where there is no guard page, or it's desirable to enforce a stack usage quota before the guard page is reached, a simple technique is to add explicit checks to ensure that the stack pointer does not exceed a specific value.

Implementing this in Cretonne will require the following steps:
 - First, remove the `stack_check` instruction. This instruction was intended to provide similar functionality, however as an instruction, it gets executed after the function prologue, where it's too late. The function prologue may have already used the stack space for saving and spilling registers. So, it's not a useful instruction, and we should just remove it.
 - Next, we should add a `stack_limit` variable to `Function`, which should be an `Option<GlobalRef>`. When it's not `None`, this will represent a `GlobalVar` address that the stack pointer should be checked against.
 - We'll also need text syntax for the `stack_limit` variable. Since this isn't an instruction, it should go in the [function preamble](https://cretonne.readthedocs.io/en/latest/langref.html#term-function-preamble), before the first basic block. I'm picturing something like `stack_limit = gv0`.
    - For parsing support, start in [`parse_preamble`](https://github.com/cretonne/cretonne/blob/master/lib/reader/src/parser.rs#L989); when the token is an `Identifier`, check to see if the string is `stack_limit`, and if so, proceed to consume a '=' and parse a global ref, using [`match_gv`](https://github.com/cretonne/cretonne/blob/master/lib/reader/src/parser.rs#L392).
    - For printing, add the approrpriate support in [`write_preamble`](https://github.com/cretonne/cretonne/blob/master/lib/codegen/src/write.rs#L41).
    - For documentation, add a brief subsection to the [memory section](https://github.com/cretonne/cretonne/blob/master/docs/langref.rst#memory) describing `stack_limit`.
    - And for testing, add a simple test under filetests/parser.
 - Codegen support.
    - To start with, we should make all calling convention prologues assert that `stack_limt` is None, so that we don't silently ignore it when it isn't implemented.
[...]

PR #359, opened 2018-06-06
Stack limit checking

pepyakin:
First PR on #349

### DIFF
--- a/lib/codegen/src/context.rs
+++ b/lib/codegen/src/context.rs
@@ -295,6 +295,10 @@ impl Context {
 
     /// Insert prologue and epilogues after computing the stack frame layout.
     pub fn prologue_epilogue(&mut self, isa: &TargetIsa) -> CtonResult {
+        assert!(
+            self.func.stack_limit.is_none(),
+            "stack_limit isn't implemented yet"
+        );
         isa.prologue_epilogue(&mut self.func)?;
         self.verify_if(isa)?;
         self.verify_locations_if(isa)?;
--- a/lib/reader/src/parser.rs
+++ b/lib/reader/src/parser.rs
@@ -244,6 +244,15 @@ impl<'a> Context<'a> {
         }
     }
 
+    // Assign the global for the stack limit.
+    fn set_stack_limit(&mut self, gv: GlobalVar, loc: &Location) -> Result<()> {
+        if let Some(_) = self.function.set_stack_limit(Some(gv)) {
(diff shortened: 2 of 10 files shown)


lib/reader/src/parser.rs
```diff
@@ -244,6 +244,13 @@ impl<'a> Context<'a> {
         }
     }
 
+    // Set the global for the stack limit.
+    fn set_stack_limit(&mut self, gv: GlobalVar) -> Result<()> {
+        // TODO: What do we do if there already is a stack_limit?
```
pepyakin: Do we need to do something special here?

lib/reader/src/parser.rs
```diff
@@ -1291,6 +1304,20 @@ impl<'a> Parser<'a> {
...
+    /// stack-limit-decl ::= "stack_limit" "=" GlobalVar(gv)
+    fn parse_stack_limit_decl(&mut self) -> Result<GlobalVar> {
+        self.consume();
+        self.match_token(Token::Equal, "expected '=' in stack limit declaration")?;
+        let gv = self.match_gv("expected global variable")?;
+
+        // Collect any trailing comments.
+        self.token();
+        self.claim_gathered_comments(gv);
+        // TODO: Where should I attach claimed comments?
```
pepyakin: Where should I attach gathered comments?

lib/reader/src/parser.rs
sunfishcode: In the parser here, we can report an error if there are multiple stack_limit declarations.

lib/reader/src/parser.rs
sunfishcode: As things currently stand, comments can only be associated with *entities* (like `gv0` or `ss0` etc.), and the stack_limit isn't an entity. So I think the right thing to do here is to avoid calling `start_gathering_comments` for `stack_limit`, and then you shouldn't have comments to worry about.

lib/codegen/src/context.rs
```diff
@@ -295,6 +295,7 @@ impl Context {
 
     /// Insert prologue and epilogues after computing the stack frame layout.
     pub fn prologue_epilogue(&mut self, isa: &TargetIsa) -> CtonResult {
+        assert!(self.func.stack_limit.is_none());
```
sunfishcode: Can you add a brief string here too, like "stack_limit not yet implemented" or so?

sunfishcode:
Looks good!

PR #361, opened 2018-06-13
Change Cretonne's license to "Apache-2.0 WITH LLVM-exception".

sunfishcode:
We are planning to make a minor change to Cretonne's license, changing it from "Apache-2.0" to "Apache-2.0 WITH LLVM-exception". This is the license the LLVM project is planning to transition to [0], and it is registered with SPDX [1].

[0] http://llvm.org/foundation/relicensing/
[1] https://spdx.org/licenses/LLVM-exception.html

LLVM-exception adds a clause providing that the output of compilation isn't subject to the attribution requirements, and a clause providing explicit GPLv2 compatibility. The full text of the license can be found on the LLVM relicensing page [2].

[2] http://llvm.org/foundation/relicensing/LICENSE.txt

If you have made significant contributions to Cretonne and do not wish your changes to be relicensed, please reply here, or contact me privately. If there are no objections by one month from now, we will proceed with the change.


sunfishcode:
(As a note to anyone looking for a general example of how to change a license, please be aware that this is just one step of a larger process.)

Issue #358, opened 2018-06-06
Remove implicit indirection in globals


sunfishcode:
As discussed in https://github.com/cretonne/cretonne/pull/280#issuecomment-395191702, we should remove the implicit indirection in globals.

The implicit indirection gets added when `heap_addr` and `stack_limit` get legalized, which are currently the two things in cretonne that use globals:
https://github.com/cretonne/cretonne/blob/master/lib/codegen/src/legalizer/heap.rs#L65
https://github.com/cretonne/cretonne/blob/master/lib/codegen/src/legalizer/heap.rs#L171
https://github.com/cretonne/cretonne/blob/master/lib/codegen/src/legalizer/mod.rs#L296

In theory, this should just be a matter of removing those loads, and the bits of code that prepare for them, and then updating the documentation.

This will create silent API breakage, so we should also start thinking of how best to communicate this to users. Cretonne doesn't claim to have a stable API, but it would be good if we could help people avoid having to debug something subtle like a missing dereference.

lachlansneff:
I'll work on this.

lachlansneff:
This can probably be closed now.

PR #363, opened 2018-06-13
Remove implicit indirection when computing heap base.

lachlansneff:
This removes the indirect when computing the heap bases for static and dynamic heaps. Previously, the cretonne codegen assumed that the supplied vmctx would contain an array of pointers to heaps. You can still do that, but you have to implement the necessary derefs explicitly.

This is a silent, breaking change and needs to be communicated to the users of cretonne.

This pr solves #358.


sunfishcode:
This looks good. It's just waiting to coordinate with #364.

sunfishcode:
Now this just needs a merge conflict resolved.

lachlansneff:
Huh, I thought I fixed the conflict last night.

PR #364, opened 2018-06-14
Change the concept of a global variable to global value.

lachlansneff:
It's more accurate to refer to global variables as global values in cretonne. This pr changes all instances of "global variable" to "global value." It also renames the `global_addr` instruction to `global_value`.


sunfishcode:
This looks good! I expect we'll have more "digesting" to do, to account for this conceptual change, but the PR here is a good base.

PR #365, opened 2018-06-15
Pop and fill don't trigger StackOverflow.

sunfishcode:
In theory, pop and fill instructions should only be reading stack memory that has already been successfully written to, so they themselves should never be able to trigger `StackOverflow` traps. This reduces the amount of bookkeeping needed.


sunfishcode:
I've convinced myself that these are ok to do. If the stack size changes out from underneath a running function, we wouldn't be able to rerun the stack probes or recheck the stack limit, so this isn't something that would work in the general case.

PR #372, opened 2018-06-23
Stack Limit as an Argument Purpose

### LINKED ISSUE #349: Stack limit checking

pepyakin:
Another approach to close #349


lib/codegen/src/isa/x86/abi.rs
```diff
@@ -488,9 +488,29 @@ fn insert_common_prologue(
             // Simply decrement the stack pointer.
             pos.ins().adjust_sp_down_imm(Imm64::new(stack_size));
         }
+
+        // Check if there is a special stack limit parameter. If there is
+        // then insert stack check.
+        if let Some(stack_limit) = pos.func.special_param(ArgumentPurpose::StackLimit) {
+            insert_stack_check(pos, stack_limit);
```
pepyakin: This is essentially same as `stack_check` instruction in the sense that it also does the check after the stack frame has been already opened.

How should I proceed here? Will Something like this

```
a = sp + imm(stack_size)
f = ifcmp a, stack_limit
trapif gte f, stack_ovf
```

do?

sunfishcode: It's `sp - imm(stack_size)` and `lte` because the stack grows down, but otherwise, yes, that's the idea. But you may need to reorganize the surrounding code a little, so that if we're also doing probes, we do the limit check before the probes.

pepyakin: Ah, yeah, you're entirely right. The thing that the [`stack_check` used the-stack-grows-up](https://github.com/pepyakin/cretonne/blob/ff11982b9c923a8423b1f71bbcf1d4e9806d980c/lib/codegen/src/legalizer/mod.rs#L297) confused me.

On the second thought, we can't use `stack_size` alone (At least if we want to guarantee that we never modify the stack below the given `StackLimit`). Because it doesn't account for the size required for RBP, all CSRs and potentially stack space needed for `probestack`.

lib/codegen/src/isa/x86/abi.rs
```diff
@@ -425,6 +425,15 @@ fn insert_common_prologue(
     csrs: &RegisterSet,
     isa: &TargetIsa,
 ) {
+    if stack_size > 0 {
+        // Check if there is a special stack limit parameter. If so insert stack check.
+        if let Some(stack_limit_arg) = pos.func.special_param(ArgumentPurpose::StackLimit) {
+            // TODO: Instead of the `stack_size` we need to pass `stack_frame_size` which
```
pepyakin: Reviving the comment about the stack size here:

```
On the second thought, we can't use `stack_size` alone (At least if we want to guarantee that we never modify the stack below the given `StackLimit`). Because it doesn't account for the size required for RBP, all CSRs and potentially stack space needed for `probestack`.
```

filetests/isa/x86/prologue-epilogue.cton
```diff
@@ -228,3 +228,11 @@ ebb4:
...
 ; check: regmove v5, %rcx -> %rbx
 ; check: [Op1popq#58,%rbx]                   v15 = x86_pop.i64
+
+; Stack limit checking
+
+function %stack_limit(i64 stack_limit) {
+    ss0 = explicit_slot 168
+ebb0(v0: i64):
+    return
+}
```
pepyakin:
```
function %stack_limit(i64 stack_limit [%rdi], i64 fp [%rbp]) -> i64 fp [%rbp] fast {
    ss0 = explicit_slot 168, offset -184
    ss1 = incoming_arg 16, offset -16

                                ebb0(v0: i64 [%rdi], v4: i64 [%rbp]):
[RexOp1umr#8089,%rax]               v1 = copy v0
[RexOp1r_id#8081,%rax]              v2 = iadd_imm v1, 176
[RexOp1rcmp_sp#8039,%rflags]        v3 = ifcmp_sp v2
[trapif#00]                         trapif uge v3, stk_ovf
[Op1pushq#50]                       x86_push v4
[RexOp1copysp#8089]                 copy_special %rsp -> %rbp
[RexOp1adjustsp_id#d081]            adjust_sp_down_imm 176
[RexOp1adjustsp_id#8081]            adjust_sp_up_imm 176
[Op1popq#58,%rbp]                   v5 = x86_pop.i64
[Op1ret#c3]                         return v5
}

.byte 72, 137, 248, 72, 129, 192, 176, 0, 0, 0, 72, 57, 224, 114, 2, 15, 11, 85, 72, 137, 229, 72, 129, 236, 176, 0, 0, 0, 72, 129, 196, 176, 0, 0, 0, 93, 195

Disassembly:
0x0: mov rax, rdi
0x3: add rax, 0xb0
0xa: cmp rax, rsp
0xd: jb 0x11
0xf: ud2
0x11: push rbp
0x12: mov rbp, rsp
0x15: sub rsp, 0xb0
0x1c: add rsp, 0xb0
0x23: pop rbp
0x24: ret
```

does this look correct?

sunfishcode: Broadly yes. One issue I noticed with doing the check before the `push rbp` is that the frame size value doesn't include the 8 bytes for that push. A related issue is that if there are any `call` instructions in the body, they'll implicitly do a push of the return address, which also isn't accounted for, so you should add those to the frame size for the check.

lib/codegen/src/isa/x86/abi.rs
```diff
@@ -425,6 +425,21 @@ fn insert_common_prologue(
...
+    if stack_size > 0 {
+        // Check if there is a special stack limit parameter. If so insert stack check.
+        if let Some(stack_limit_arg) = pos.func.special_param(ArgumentPurpose::StackLimit) {
+            // Total stack size is the size of all stack area used by the function, including
+            // pushed CSRs, frame pointer. 
+            // Also, the size of a return address, implicitly pushed by a x86 `call` instruction, also 
+            // should be accounted for.
+            // TODO: Check if the function body actually contains a `call` instruction.
+            let word_size = isa.pointer_bytes();
+            let total_stack_size = (csrs.iter(GPR).len() + 1 + 1) as i64 * word_size as i64;
```
pepyakin: For now, it pessimistically accounts for an implicit return address pushed by a call instruction.

What is the best way to find out if the function contains a call instruction? Iterate over all instructions here?

bjorn3: Why is `+ 2` written as `+ 1 + 1`?

pepyakin: It is just simply because I wanted to make it clear that one is for frame pointer register and another one for a return address. But apparently I've failed : D

sunfishcode: The absolute simplest way would be to just assume there's always a call.

Beyond that, iterating through the instructions is the best we can do for now.

pepyakin:
Hey, @sunfishcode ! I removed the code for the old approach (the one with `stack_limit` attr) and updated the test. 

Do we need something else to merge it? Are we ok with the assumption that there is always a call?

sunfishcode:
Yes, the assumption there's always a call is ok for now. So yes, I think we can merge this now.

Issue #373, opened 2018-06-23
Please show ebb number in verifier errors


bjorn3:
Currently I get stuff like `location: inst3423`

sunfishcode:
This is fixed in #375!

PR #375, opened 2018-06-25
Adds decoration to the verifier errors.

### LINKED ISSUE #68: Prettier verifier errors
The verifier (#3) is beginning to find real bugs now. Its method for displaying verifier errors could be improved. Here's an example of the current output:

```
$ target/debug/cton-util test filetests/isa/riscv/expand-i32.cton 
FAIL filetests/isa/riscv/expand-i32.cton: inst1: arg 0 (v0) has type i16, must match function signature of i32
inst1: return v0

function large_imm(i16) -> i32 {
ebb0(vx0: i16):
    v0 = iadd_imm vx0, 0x3b9a_ca00
    return v0
}
```

The reference to `inst1` is cryptic because instruction references don't appear in the textual IL, so we print the instruction itself in the first line. This is not a great solution.

It would be much better to underline the offending instruction, like the filecheck tests do:

```
FAIL filetests/isa/riscv/expand-i32.cton: inst1: arg 0 (v0) has type i16, must match function signature of i32
function large_imm(i16) -> i32 {
ebb0(vx0: i16):
    v0 = iadd_imm vx0, 0x3b9a_ca00
    return v0
    ^~~~~~~~~
verifier: arg 0 (v0) has type i16, must match function signature of i32
}
```

# Implementation suggestion

One possible implementation is to move the guts of `write_function` into a more general

```rust
pub fn decorate_function<WL: FnMut(AnyEntity, usize, &str) -> fmt::Result>(w: WF, func: &Function, isa: Option<&TargetIsa>) -> fmt::Result;
```

The `WL` closure would be called to write out each line with the arguments: The entity being defined on the line, the indentation to insert before the line, and the contents of the line itself.

carolinecullen:
example:

function %bad(i32) fast {
ebb0(v0: i32):
    brnz.i32 v0, ebb1
    return
    ^~~~~~

verifier inst1: Internal return not allowed with return_at_end=1

ebb1:
    trapz.i32 v0, user6
    return
}

Fixes #68

### DIFF
--- a/lib/codegen/src/print_errors.rs
+++ b/lib/codegen/src/print_errors.rs
@@ -1,11 +1,15 @@
 //! Utility routines for pretty-printing error messages.
 
 use ir;
+use ir::entities::Inst;
+use ir::function::Function;
 use isa::TargetIsa;
 use result::CodegenError;
+use std::fmt;
 use std::fmt::Write;
 use std::string::{String, ToString};
 use verifier::VerifierError;
+use write::decorate_function;
 
 /// Pretty-print a verifier error.
 pub fn pretty_verifier_error(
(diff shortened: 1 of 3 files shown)


bjorn3:
Also fixes #373

lib/codegen/src/print_errors.rs
```diff
@@ -1,27 +1,68 @@
...
+        &mut |w, func, isa, inst, indent| pretty_function_error(w, func, isa, inst, indent, err),
+        &mut w,
+        func,
+        isa,
+    ).unwrap();
+    w
+}
+
+/// Pretty-print a function verifier error.
+pub fn pretty_function_error(
```
sunfishcode: Would this function be useful outside of this file? If not, it'd be good to remove the `pub`.

lib/codegen/src/write.rs
```diff
@@ -9,9 +9,36 @@ use packed_option::ReservedValue;
 use std::fmt::{self, Write};
 use std::string::String;
 
+pub fn write_function_plain(
```
sunfishcode: Same with the `pub` here.

lib/codegen/src/write.rs
```diff
@@ -9,9 +9,36 @@ use packed_option::ReservedValue;
...
 pub fn write_function(w: &mut Write, func: &Function, isa: Option<&TargetIsa>) -> fmt::Result {
+    decorate_function(
+        &mut |w, func, isa, inst, indent| write_function_plain(w, func, isa, inst, indent),
+        w,
+        func,
+        isa,
+    )
+}
+
+pub fn decorate_function<
```
sunfishcode: It'd be good to have a brief documentation comment here describing what `decorate_function` does.

lib/codegen/src/write.rs
```diff
@@ -141,7 +168,15 @@ pub fn write_ebb_header(
     writeln!(w, "):")
 }
 
-pub fn write_ebb(w: &mut Write, func: &Function, isa: Option<&TargetIsa>, ebb: Ebb) -> fmt::Result {
+pub fn write_ebb<
```
sunfishcode: For consistency with `decorate_function`, I think we should rename `write_ebb` to `decorate_ebb`.

(In the future we could add a `write_ebb` which would do what `write_function` does, passing in `write_function_plain`, if we were to need such a thing.)

carolinecullen:
Added the changes from your comments @sunfishcode !

sunfishcode:
Looks good!

PR #379, opened 2018-06-30
Add comment support

bjorn3:
This allows users to add comments to the text emitted by `write_function`


sunfishcode:
This looks cool, I like the example! Would it make sense to indent the comments to align them with the instructions?

On the patch itself, I'm concerned about adding these comments to an ir data structure. Optimization passes may reorder or modify instructions, and if these comments stick around, they could be confusing.

As a possible.alternative, would it be frasible to use the new `decorate_function` that @carolinecullen added, for printing the ir right after it's produced? The frontend would build the comments EntityMap itself, and then pass it into `decorate_function` in the closure. 

It may still be useful to add the nop instruction, though we should clarify that nops don't codegen to machine code nops.

bjorn3:
> It may still be useful to add the nop instruction, though we should clarify that nops don't codegen to machine code nops.

Done, also added test

sunfishcode:
Looks good!

sunfishcode:
There are some merge conflicts; would you mind resolving those?

bjorn3:
Rebased

PR #383, opened 2018-07-01
Made Capstone an optional dependency (fixes #382)

### LINKED ISSUE #382: Make Capstone an optional dependency
I think it would be easy to make `capstone` an optional dependency introduced using a Cargo feature, since it is only [used in a single file](https://github.com/cretonne/cretonne/blob/88fc2303e77e6109cbef58d90acd7ccac3feb5f4/src/compile.rs), for a single purpose throughout cretonne. Furthermore, right now showing the disassembly of what we built is more of a "cool feature" than something we actively advertise or support, and thus I think it makes sense to make it optional, thus dropping a dependency on an external library.

**Edit**: This one is easy. Say the word and I'll do it myself.

6A:
The `disas` Cargo feature was also introduced, to optionally enable disassembly.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,12 +29,13 @@ docopt = "1"
 serde = "1.0.8"
 serde_derive = "1.0.8"
 term = "0.5.1"
-capstone = "0.4"
+capstone = { version = "0.4", optional = true }
 wabt = { version = "0.3", optional = true }
 target-lexicon = "0.0.2"
 
 [features]
 default = ["wasm"]
+disas = ["capstone"]
 wasm = ["wabt", "cretonne-wasm"]
 
 [workspace]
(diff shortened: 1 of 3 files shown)


sunfishcode:
I think this is a good idea. As mentioned in #382, the Capstone dependeny is just on the top-level "utility" crate, and not on any the library crates, so this doesn't affect people using Cretonne as a library, however Capstone is a C library, and I could imagine situations where someone might have trouble building it.

I would prefer it to be built by default though, since it's quite nice for Cretonne developers to be able to see the actual machine code emitted by Cretonne.

sunfishcode:
The CI is failing here now with some errors on the Capstone import.

sunfishcode:
Thanks for seeing this through!

PR #384, opened 2018-07-02
Now diagnosing missing vmctx arguments (fixes #376)

### LINKED ISSUE #376: Verifier: diagnose a missing vmctx argument
In this testcase, there's a vmctx global value, but no vmctx argument, which is not valid:
```cton
target x86_64

function %vmglobal(i64) -> i64 {
    gv4 = vmctx+0
ebb0(v0: i64):
    v1 = global_value.i64 gv4
    return v1
}
```
Currently compiling with `cton-util compile test.cton` gets a panic in codegen. It would be better to catch this in the verifier.

6A:
Here's an example of what is returned by the compiler:

```cretonne
target x86_64

function %vmglobal(i64) -> i64 {
    gv4 = vmctx+0
ebb0(v0: i64):
    v1 = global_value.i64 gv4
    return v1
}
```

Changing the parameter to `i64 vmctx` causes no reported error or panic.

### DIFF
--- a/lib/codegen/src/verifier/mod.rs
+++ b/lib/codegen/src/verifier/mod.rs
@@ -418,6 +418,12 @@ impl<'a> Verifier<'a> {
     fn verify_global_value(&self, inst: Inst, gv: GlobalValue) -> VerifierResult<()> {
         if !self.func.global_values.is_valid(gv) {
             err!(inst, "invalid global value {}", gv)
+        } else if let Some(&ir::GlobalValueData::VMContext { .. }) = self.func.global_values.get(gv) {
+            if self.func.special_param(ir::ArgumentPurpose::VMContext).is_none() {
+                err!(inst, "undeclared vmctx reference {}", gv)
+            } else {
+                Ok(())
+            }
         } else {
             Ok(())
         }
--- a/lib/codegen/src/verifier/mod.rs
+++ b/lib/codegen/src/verifier/mod.rs
@@ -44,6 +44,7 @@
 //! Global variables
 //!
 //! - Detect cycles in deref(base) declarations.
+//! - Detect use of 'vmctx' global value when no corresponding parameter is defined.
 //!
 //! TODO:
 //! Ad hoc checking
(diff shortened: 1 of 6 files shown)


bjorn3:
Could you please add a verifier test?

6A:
@bjorn3 I'm on it.

bnjbvr:
For what it's worth, I was working on the same issue last Friday and wrote a very similar patch, so this looks good to me at least; there was also another issue coming in, maybe the same you found. I'll happy withdraw, since I didn't open a PR on time. Thanks for fixing this!

6A:
@bnjbvr I'm very new on the Cretonne codebase, so I'm glad you had a similar fix. Now if you think your code is for any reason better, you're welcome to submit a PR to replace mine.

bnjbvr:
I'm new too, and the wording in my PR was slightly off, plus you've done the work of opening the PR and investigating extra test cases, so I'm all very fine with the maintainers taking your patch :slightly_smiling_face:

lib/codegen/src/verifier/mod.rs
```diff
@@ -186,6 +186,12 @@ impl<'a> Verifier<'a> {
 
                 cur = base;
             }
+            
+            if let ir::GlobalValueData::VMContext { .. } = self.func.global_values[cur] {
+                if self.func.special_param(ir::ArgumentPurpose::VMContext).is_none() {
```
bnjbvr: CI complains that this line should be formatted differently. You can run `./format-all.sh` from Cretonne's top level dir to have the expected formatting done for you.

6A:
Alright, I'll take care of it right away! Thanks again for the help

6A:
When I do add the corresponding parameter in the failing tests, the tests fail yet again because Cretonne panics in `function.rs`, since for some reason no parameters are added for the `vmctx`, even though the `Signature` has the right `special_param_index`.

6A:
Alright, I've fixed everything. `./test-all.sh` still fails, but because of an error in the documentation of `env_logger`.

sunfishcode:
Yeah, that `env_logger` breakage is only on nightly Rust, and we have `allow_failures: nightly` in the .travis.yml file, so this PR is green now!

sunfishcode:
Looks good. It's interesting how many tests needed to be updated here. It's good because the tests also serve as examples of how to use Cretonne IR, and they make better examples when they're valid :-).

PR #386, opened 2018-07-02
Support systems which don't have a "python" command.

### LINKED ISSUE #381: Do not hardcode "python" in scripts
When trying to build `cretonne-codegen`, the command `python` cannot be found (my setup has `python3` and `python3.6`, but not `python` in the PATH) in `lib/codegen/build.rs`. It seems the files `lib/codegen/meta/check.sh` and `test-all.sh` also use hardcoded versions of Python.

I think it would help to use an environment variable such as `PYTHON` or `PY` instead of hardcoding the name `python` everywhere.

sunfishcode:
Add support for finding an appropriate python command on systems which
don't have "python". Try "python3" and "python2.7".

Fixed #381.


sunfishcode:
This fixes the reported issue in #381, so let's go with that.

PR #388, opened 2018-07-05
Make InstructionData comparable and hashable

Amanieu:
Currently, `EntityList` implements the `Eq` and `Hash` trait but simply panics in the implementation. The original purpose of this hack was to allow some instructions (the ones that don't use `ValueList`) to be stored in a hash table for GVN.

This PR removes the `Eq` and `Hash` implementations on `EntityList` and instead generates custom `eq` and `hash` functions for `InstructionData` which take a `ValueListPool` parameter. These functions are then used in the GVN pass when inserting the `InstructionData` into a `HashMap`.

Note: The use of a `RefCell` makes the code a bit ugly, but there is no other way to do this.

This solves the problem described in https://github.com/cretonne/cretonne/issues/366#issuecomment-398558755.

### DIFF
--- a/lib/codegen/src/simple_gvn.rs
+++ b/lib/codegen/src/simple_gvn.rs
@@ -4,6 +4,8 @@ use cursor::{Cursor, FuncCursor};
 use dominator_tree::DominatorTree;
 use ir::{Function, Inst, InstructionData, Opcode, Type};
 use scoped_hash_map::ScopedHashMap;
+use std::cell::{Ref, RefCell};
+use std::hash::{Hash, Hasher};
 use std::vec::Vec;
 use timing;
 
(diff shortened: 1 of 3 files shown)


lib/codegen/src/simple_gvn.rs
```diff
@@ -20,64 +22,88 @@ fn trivially_unsafe_for_gvn(opcode: Opcode) -> bool {
...
+    fn hash<H: Hasher>(&self, state: &mut H) {
+        let pool = &self.2.borrow().func.dfg.value_lists;
+        self.0.hash(state, pool);
+        self.1.hash(state);
+    }
+}
+impl<'a, 'f: 'a> PartialEq for HashEntry<'a, 'f> {
+    fn eq(&self, other: &Self) -> bool {
+        let pool = &self.2.borrow().func.dfg.value_lists;
+        self.0.eq(&other.0, pool) && self.1 == self.1
```
sunfishcode: `self.1 == self.1` should be `self.1 == other.1`

lib/codegen/src/simple_gvn.rs
```diff
@@ -20,64 +22,88 @@ fn trivially_unsafe_for_gvn(opcode: Opcode) -> bool {
         || opcode.writes_cpu_flags()
 }
 
+/// Wrapper around `InstructionData` which implements `Eq` and `Hash`
+#[derive(Clone)]
+struct HashEntry<'a, 'f: 'a>(InstructionData, Type, &'a RefCell<FuncCursor<'f>>);
```
sunfishcode: Should we call this `HashKey` rather than `HashEntry`, to avoid confusion with `HashMap`'s `entry` concept?

Amanieu:
Updated with comments addressed and formatting fixed.

Note that this is a breaking change for `cretonne-entity`, so we may need a version bump there.

sunfishcode:
Looks good! And thanks for pointing out the breaking change. I'll bump the version next time I publish new packages.

PR #389, opened 2018-07-06
Update wabt to 0.4

Amanieu:
This fixes compilation on system with GCC 8.

See https://github.com/pepyakin/wabt-rs/issues/15


sunfishcode:
Thanks!

PR #391, opened 2018-07-06
Reject with a plain text error when no ISA is defined with wasm

bnjbvr:
If you run the `wasm` command on a file, without a specific ISA defined, it will panic. This fixes it, and restructure the code (adds `\n`, inverts control flow to reduce indent, rename `data` into `module_binary`) in this file to make it easier to understand.


sunfishcode:
Nice cleanup!

PR #395, opened 2018-07-09
RecipeConstraints now has a name for better error diagnosis (fixes #394).

### LINKED ISSUE #394: Add a name field to constraints
Right now, when a constraint is not satisfied, the error `constraints not satisfied` is logged, without more informations. Since [constraints at the Python-level have a name](https://github.com/cretonne/cretonne/blob/309fe26c0cb75a22595cc6ae1c81e0b51e887fd9/lib/codegen/meta/gen_encoding.py#L759), I think we could add a `&'static str` name to the generated Rust code as well, in order to _at least_ report the name of the constraint. If possible, adding a description and/or details to the Python "AST" could be useful as well, and later added to Rust.

6A:


sunfishcode:
In theory, the verifier is already calling `self.encinfo.display(enc)` for this message, which should return a `DisplayEncoding`, which should include the recipe name. Is the name not showing up?

6A:
I'll have to make sure, but didn't get to test it yet

sunfishcode:
In light of your edit, should we close this PR, or are you still thinking about adding new details to `OperandConstraint`s here?

6A:
Sorry, I did not see that notification. I'm closing this.

PR #397, opened 2018-07-10
Moar Clippy fixes

bnjbvr:
I misread the Clippy report yesterday, and there are actually many more issues to fix. I've got a few ones done, and I'll switch to something else now, to let easy work for other beginners :slightly_smiling_face: .

### DIFF
--- a/lib/codegen/meta/gen_instr.py
+++ b/lib/codegen/meta/gen_instr.py
@@ -266,7 +266,7 @@ def gen_instruction_data_impl(fmt):
                     n = '&InstructionData::' + f.name
                     members = ['opcode']
                     if f.typevar_operand is None:
-                        args_eq = 'true'
+                        args_eq = None
                     elif f.has_value_list:
                         members.append('args')
                         args_eq = 'args1.as_slice(pool) == ' \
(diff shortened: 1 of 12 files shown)


bjorn3:
Misspelled title.

kinggoesgaming:
MOAR misspelled titles are needed

lib/codegen/meta/gen_instr.py
```diff
@@ -279,17 +279,29 @@ def gen_instruction_data_impl(fmt):
...
+                        for (idx, field) in enumerate(f.imm_fields):
+                            line = ('{}1 == {}2'
+                                    .format(field.member, field.member))
+                            if (idx != len(f.imm_fields) - 1 or args_eq is not
+                                    None):
+                                line += ' &&'
+                            fmt.line(line)
+
+                        if args_eq is not None:
+                            fmt.line(args_eq)
```
sunfishcode: Could this code be simplified by moving the '&&'s to the following lines? Eg.:
```
     opcode1 == opcode2
     && field == field
     && field == field
     && /* args_eq */
```
That way, each additional line could add its own '&&', rather than having to check whether one will be needed.

Also, this follows recent versions of rustfmt which have been [formatting code this way](https://github.com/cretonne/cretonne/blob/master/lib/codegen/src/isa/x86/abi.rs#L468).

bnjbvr:
Indeed, much simpler!

PR #398, opened 2018-07-11
Add EntityList::from_slice to build a list from an existing slice

Amanieu:
This allocates the entire list at once, rather than pushing one element at a time as is done in `extend`.


sunfishcode:
Looks useful!

PR #399, opened 2018-07-13
Fix broken build on no_std.

lachlansneff:
This does two things. First, updates the use of the alloc crate to the latest nightly version. Secondly, it downgrades the use of the cpuid crate to a version that doesn't have issues compiling in a custom environment.


sunfishcode:
Is https://github.com/gz/rust-cpuid/issues/27 the raw-cpuid issue you're hitting? If so, can you add a comment to the code noting that that's the issue, so that if someone wants to update it again in the future, they'll know what to check?

lachlansneff:
Yep, that's the issue. I'll add a comment when I get to a computer.

Issue #342, opened 2018-05-18
Port the DSL to Rust


sunfishcode:
Cretonne's [Python-based DSL](https://cretonne.readthedocs.io/en/latest/metaref.html) allows target ISAs to be described in a concise manner, and can generate tables, code, and documentation from them. It's common for compilers to use DSLs for such purposes; see also LLVM's [tablegen](https://llvm.org/docs/TableGen/), GCC's [machine description language](https://gcc.gnu.org/onlinedocs/gccint/Machine-Desc.html), libfirm's perl ([example](https://github.com/libfirm/libfirm/blob/2d4f19de8e504013d51ab054cb969118a8fbf91b/ir/be/ia32/ia32_spec.pl)) and python ([example](https://github.com/libfirm/libfirm/blob/master/scripts/ir_spec.py)), and others.

Python compares pretty favorably to obvious alternatives, in particular by being a real programming language, and a fairly friendly one at that. However, using Python in Cretonne has some downsides too:
 - There's extra overhead for developers that don't work with Python on a regular basis. They may not have mypy, flake8, or python-aware or even mypy-aware IDE plugins installed, they may not remember how variable scoping works, and so on.
 - The current Python code makes use of some "magical" features, such as `__getitem__`, `isinstance`, etc. which may be fine for people who work with Python regularly, but which make it harder to follow the code for people who aren't.
 - The current Python code also behaves in "magical" ways sometimes, such as `PUT_OP` in the x86 target. It's used in Rust code, but it's not a Rust construct; it's edited out by `replace_put_op`. In some cases, it's replaced by `put_mp2`, but it's not easy to determine when that name is used, as there are no other occurrences of `put_mp2` or `mp2` in the repository. As another example, the DSL infers instruction formats from instruction fields implicitly, in a way that makes it confusing when one is adding a new format and accidentally colliding with an existing format.
 - Cretonne has ambitions of being able to compile Rust code in the future, so it would be cooler if it could compile all of itself.

None of these are show-stoppers, and one option is to just keep the current code and refactor it to help out with some of these issues. But we should also consider where we want to be in the long term; it'd be easier to do a large architectural change now, while the code is (relatively) simple, as it'll grow more complex in the future as we add more features. One option is Rust's procedural macros, except that they're not yet stable. A possible intermediate step could be to use a build.rs.

[...]

PR #403, opened 2018-07-16
Added initial Rust codegen-meta implementation.

data-pup:
This PR contains the first steps towards implementing the DSL in Rust, as previously discussed in #342. This intends to replace the `gen_types.py` file, and will emit a file named `new_types.rs` in the same directory as the existing `types.rs` file, when the `codegen` crate is built.

Thanks for the mentoring @sunfishcode! Had a lot of fun putting this together.


lib/codegen-meta/src/base/mod.rs
```diff
@@ -0,0 +1,3 @@
+//! Definitions for the base Cretonne language.
```
data-pup: Just noticed a few instances of the name "Cretonne" snuck through, I can rectify these.

lib/codegen/build.rs
```diff
@@ -66,12 +68,35 @@ fn main() {
...
+    // Identify the directory of the Rust codegen-meta external crate.
+    let rust_meta_dir = crate_dir
+        .parent()
+        .map(|d| d.join("codegen-meta"))
+        .unwrap_or_else(|| {
+            eprintln!("Error: Could not find path to lib/codegen-meta crate.");
+            process::exit(1);
+        });
+
+    if let Err(err) = meta::gen_types::generate("new_types.rs", &out_dir) {
```
data-pup: This specifically would be the line that changes when we want to have the codegen-meta crate emit a `types.rs` file instead, and replace the version currently being emitted by the Python code in `codegen/meta/` :smile:

data-pup:
Build isn't quite done yet, but I noticed these problems in the 1.25 build. I can fix these today, should be fairly straightforward.

`error[E0658]: non-reference pattern used to match a reference (see issue #42640)`

`no method named `try_for_each` found for type ...`

bjorn3:
syntex_syntax is deprectated. you.should use proc macros

lib/codegen-meta/src/base/types.rs
```diff
@@ -0,0 +1,262 @@
...
+            Int::I32 => 2,
+            Int::I64 => 3,
+        };
+
+        LANE_BASE + offset
+    }
+}
+
+pub struct IntIterator {
+    index: usize,
```
bjorn3: This could just be an u8.

lib/codegen-meta/src/base/types.rs
```diff
@@ -0,0 +1,262 @@
...
+    type Item = Int;
+    fn next(&mut self) -> Option<Self::Item> {
+        let res = match self.index {
+            0 => Some(Int::I8),
+            1 => Some(Int::I16),
+            2 => Some(Int::I32),
+            3 => Some(Int::I64),
+            _ => None,
+        };
+        self.index += 1;
```
bjorn3: What about release mode overflow? (Call `.next()` `usize::MAX` times and you get `I8` again) You could replace the `_ => None,` with `_ => return None,` to prevent this.

data-pup:
Thanks @bjorn3! Didn't know about that early return trick for iterators, but that makes a lot of sense. That's cleaned up now :+1: 

Added a few other fixes, including fmt::Debug implementations for the different ValueType variants, to give the same output as the `__repr__` methods in the original Python, as well as making changes for compatibility with v1.25.

lib/codegen-meta/src/cdsl/types.rs
```diff
@@ -0,0 +1,467 @@
...
+    /// Get the name of this vector type.
+    pub fn name(&self) -> String {
+        format!("{}X{}", self.base.name(), self.lanes,)
+    }
+
+    /// Find the unique number associated with this vector type.
+    pub fn number(&self) -> u8 {
+        let b = f64::from(self.base.number());
+        let l = (self.lanes as f64).log2();
+        let num = 16_f64 * l + b;
```
sunfishcode: Can we do this computation without using floating point? I guess it's a little more work because we need a helper:

```rust
fn floor_log2(x: u64) -> u32 {
  63 - x.leading_zeros()
}
```
but it would mean that we won't need to add an exception for the `float_arithmetic` clippy lint that we enable in some configurations.

Also, we should give the "16" constant a name.

lib/codegen-meta/src/gen_types.rs
```diff
@@ -0,0 +1,72 @@
+//! Generate sources with type info.
+//!
+//! This generates a `types.rs` file which is included in
+//! `lib/codegen/ir/types.rs`. The file provides constant definitions for the
+//! most commonly used types, including all of the scalar types.
+//!
+//! This ensures that Python and Rust use the same type numbering.
```
sunfishcode: This ensures that the metaprogram and the generated program see the same type numbering :-).

lib/codegen-meta/src/srcgen.rs
```diff
@@ -0,0 +1,315 @@
...
+    } else {
+        let t = s.trim_left();
+        Some(s.len() - t.len())
+    }
+}
+
+/// Given a multi-line string, split it into a sequence of lines after
+/// stripping a common indentation. This is useful for strings defined with
+/// doc strings.
+fn parse_multiline(s: &str) -> Vec<String> {
```
sunfishcode: I was initially concerned that this is too Python-specific, but looking at the way Rust handles multi-line literals, we may indeed want to write Rust code for this in much the way we write Python:
```rust
    fmt.doc_comment("
                    An instruction format

                    Every opcode has a corresponding instruction format
                    which is represented by both the `InstructionFormat`
                    and the `InstructionData` enums.
                    ")
```
which would then want the same logic of stripping common leading whitespace, and so on. Does that sound right?

lib/codegen-meta/src/cdsl/types.rs
```diff
@@ -0,0 +1,467 @@
...
+            float_iter: base_types::FloatIterator::new(),
+        }
+    }
+}
+
+impl Iterator for LaneTypeIterator {
+    type Item = LaneType;
+    fn next(&mut self) -> Option<Self::Item> {
+        if let b @ Some(_) = self.bool_iter.next() {
+            b.map(LaneType::from)
```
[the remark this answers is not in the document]
data-pup: Since the method returns an option, this could also rewritten as:

```rust
      if let Some(b) = self.bool_iter.next() {
            Some(LaneType::from(b))
```

I can write it that way, so it isn't quite as opaque :)

lib/codegen-meta/src/cdsl/mod.rs
```diff
@@ -0,0 +1,77 @@
...
+        }
+    }
+
+    output_chars
+}
+
+/// Check if `x` is a power of two.
+fn _is_power_of_two(x: u8) -> bool {
+    x > 0 && x & (x - 1) == 0
+}
```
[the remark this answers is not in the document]
data-pup: In that case, should I just go ahead and remove both this and the `next_power_of_two` functions from this file?

lib/codegen-meta/src/base/types.rs
```diff
@@ -0,0 +1,247 @@
...
+    /// CPU flags from a floating point comparison.
+    FFlags,
+}
+
+impl Flag {
+    /// Get the number of a flag variant.
+    pub fn number(self) -> u8 {
+        match self {
+            Flag::IFlags => 1,
+            Flag::FFlags => 2,
```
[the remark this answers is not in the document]
data-pup: The `number` methods are responsible for the value inside of the `Type(..)` call in the generated file, like so:

```rust
pub const IFLAGS: Type = Type(0x1);
```

I'm not totally content with this approach to be honest, and had been thinking about moving these up into the `cdsl/types.rs` file instead, so it's a little clearer what's happening. Do you think that sounds like a good idea?

lib/codegen-meta/Cargo.toml
```diff
@@ -0,0 +1,18 @@
+[package]
+name = "cranelift-codegen-meta"
+authors = ["The Cranelift Project Developers"]
+version = "0.15.0"
+description = "DSL for cranelift-codegen code generator library"
+license = "Apache-2.0"
```
sunfishcode: Also, please add a copy of the LICENSE file under lib/codegen-meta (this is a recent change).

lib/codegen-meta/src/cdsl/mod.rs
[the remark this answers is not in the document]
sunfishcode: Yes.

lib/codegen-meta/src/base/types.rs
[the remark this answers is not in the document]
sunfishcode: Yeah. I didn't notice this before, but a lot of this logic is in cdsl/types.py in the Python version, so yeah, that makes sense.

lib/codegen-meta/src/base/types.rs
```diff
@@ -0,0 +1,247 @@
...
+// 0x80-0xff: Vector types
+//
+// Vector types are encoded with the lane type in the low 4 bits and log2(lanes)
+// in the high 4 bits, giving a range of 2-256 lanes.
+static LANE_BASE: u8 = 0x70;
+
+#[derive(Debug, Clone, Copy, Eq, PartialEq)]
+pub enum Bool {
+    /// 1-bit bool.
+    B1 = 1,
```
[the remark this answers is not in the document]
data-pup: Lane types such as `Bool` or `Int` use enums with explicit discriminators, so that the `bits` value  can be set without a match statement. One example of where this happens is in the `impl From<base_types::Bool> for LaneType` block in `cdsl/types.rs` :)

lib/codegen-meta/src/srcgen.rs
```diff
@@ -0,0 +1,315 @@
...
+/// Match formatting class.
+///
+/// Match objects collect all the information needed to emit a Rust `match`
+/// expression, automatically deduplicating overlapping identical arms.
+///
+/// Note that this class is ignorant of Rust types, and considers two fields
+/// with the same name to be equivalent.
+struct _Match<'a> {
+    _expr: &'a str,
+    arms: BTreeMap<(Vec<&'a str>, &'a str), HashSet<&'a str>>,
```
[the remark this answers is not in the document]
data-pup: Yeah! I was trying to recreate the original Python as close as I could, but the collections available in Rust meant that this was the closest I could get.

data-pup:
Took care of the style issue mentioned :) Sounds good! I'll leave the generated file as 'new_types.rs`.

sunfishcode:
Thanks! I'm excited about where we'll go with this :-)

data-pup:
Me too! It was a lot of fun to work on this. If you have any ideas for next steps, I'd love to try and help out with some of that as well! We can talk about that back in #342 :smile:

PR #405, opened 2018-07-18
Misc refactorings when looking at the wasm code

bnjbvr:
Sorry this is in a form of potpourri. These are small fixes to comments or small refactorings which make the code easier to understand, in my opinion:

- fixed typo in dominator_tree.
- simplify some control flow in preopt by playing with block expression to avoid mutable borrow of an immutably borrowed value (which justified the previous code).
- simplify to the module translator to avoid duplicate code for data sections.
- reuse and clear the Context in wasm.rs. After all, that's the first thing the comment of Context recommends to do :) And it improves the testing of the re-usability of Context.

### DIFF
--- a/lib/wasm/src/module_translator.rs
+++ b/lib/wasm/src/module_translator.rs
@@ -3,9 +3,9 @@
 use cranelift_codegen::timing;
 use environ::{ModuleEnvironment, WasmError, WasmResult};
 use sections_translator::{
-    parse_data_section, parse_elements_section, parse_export_section, parse_function_section,
-    parse_function_signatures, parse_global_section, parse_import_section, parse_memory_section,
-    parse_start_section, parse_table_section,
+    parse_code_section, parse_data_section, parse_elements_section, parse_export_section,
+    parse_function_section, parse_function_signatures, parse_global_section, parse_import_section,
+    parse_memory_section, parse_start_section, parse_table_section,
 };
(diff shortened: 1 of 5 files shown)


lib/wasm/src/module_translator.rs
```diff
@@ -94,10 +94,19 @@ pub fn translate_module<'data>(
...
+                    ParserState::EndSection => break,
+                    ParserState::Error(e) => return Err(WasmError::from_binary_reader_error(e)),
+                    ref s => panic!("wrong content in code section: {:?}", s),
+                }
+                let mut reader = parser.create_binary_reader();
+                let size = reader.bytes_remaining();
+                environ.define_function_body(reader
+                    .read_bytes(size)
+                    .map_err(WasmError::from_binary_reader_error)?)?;
+            },
```
sunfishcode: This looks good, and seeing it structured this way, it occurs to me that it'd be good to factor this out into a `parse_code_section` function, in sections_translator.rs, similar to the other `parse_*_section` functions. That way, it'd complete the separation of the module-level parsing logic from the section parsing.

(We may eventually want to get rid of `translate_module`'s top-level loop, and just parse the sections in their standard order, but that's a somewhat bigger change.)

bnjbvr:
Here we go! I like how it's now more consistent with the other section translators.

bnjbvr:
Duh, fixed!

sunfishcode:
Thanks!

PR #406, opened 2018-07-19
Promote BasicBlock to a named struct

bnjbvr:
This promotes the BasicBlock from a tuple to a struct:
- this makes it easier to remember code that uses its fields thanks to the names, which carry more meaning than the field numbers.
- this helps also in destructuring patterns to know which kind of data structure is being referenced

I did the same for the (Block, Inst) that appears in the `predecessors` vector in `EbbHeaderBlockData`, but I'm less sure it provides a lot of value: the tuple is only used within the file, so it's much more contained. It's still a small readability win in my opinion, though. (Happy to drop this commit if it's not considered useful)

I haven't applied this to all tuples in the code; this could be an opportunity for easy contributions for newcomers. `BasicBlock` really was the only one which bothered me and had me check the definition back and forth a few times.


bnjbvr:
(It's finally passing the CI testing, yayyy!)

lib/codegen/src/flowgraph.rs
```diff
@@ -31,7 +31,25 @@ use std::mem;
...
+    pub fn new(ebb: Ebb, inst: Inst) -> Self {
+        Self { ebb, inst }
+    }
+}
+
+impl PartialEq for BasicBlock {
+    fn eq(&self, rhs: &Self) -> bool {
+        self.ebb == rhs.ebb && self.inst == rhs.inst
+    }
+}
```
sunfishcode: Instead of manally implementing `eq`, we can just add `#[derive(PartialEq, Eq)]` to the `BasicBlock` struct and we'll get an `eq` function automatically :-).

lib/frontend/src/ssa.rs
```diff
@@ -104,17 +104,28 @@ impl<Variable> BlockData<Variable> {
...
+                    .position(|&PredBlock { branch, .. }| branch == inst)
                     .expect("the predecessor you are trying to remove is not declared");
-                data.predecessors.swap_remove(pred).0
+                data.predecessors.swap_remove(pred).block
             }
         }
     }
 }
 
+pub struct PredBlock {
```
sunfishcode: This is local to this file, so it doesn't need to be `pub`.

bnjbvr: Can't actually be done! The public fn `predecessors` returns a slice of `PredBlock`, so this must be public too.

sunfishcode: Ah, you're right. It's used in frontend.rs, which doesn't really use `PredBlock`, it just tests if there are *any* predecessors. What if we add a `has_any_predecessors()` function in ssa.rs? Then `predecessors()` could be private too.

bnjbvr:
Done!

PR #408, opened 2018-07-20
Clippy improvements

waywardmonkeys:


lib/filetests/src/runner.rs
```diff
@@ -44,7 +44,7 @@ impl Display for QueueEntry {
                 f,
                 "{}.{:03} {}",
                 dur.as_secs(),
-                dur.subsec_nanos() / 1_000_000,
+                dur.subsec_millis(),
```
sunfishcode: `subsec_millis` is [new in Rust 1.27](https://doc.rust-lang.org/std/time/struct.Duration.html#method.subsec_millis) and we currently support Rust 1.25. For now, we should add that clippy lint to the `allow` list in the containing crate's lib.rs file, with a comment explaining why.

waywardmonkeys:
I've updated this PR as you requested.

sunfishcode:
Thanks!

PR #410, opened 2018-07-20
Rename native_pointer() to pointer_type()

bnjbvr:
... to make it more apparent it's a type that's being returned.


sunfishcode:
Cool, this is also more consistent with `TargetIsa`'s `pointer_type()`.

PR #412, opened 2018-07-22
Remove gen_build_deps module.

### LINKED ISSUE #409: Can we omit rerun-if-changed lines for codegen-meta?
Currently, codegen-meta has a gen_build_deps module which emits `rerun-if-changed` lines for all .rs files in the codegen-meta crate. However, according to [the cargo documentation](https://doc.rust-lang.org/cargo/reference/build-scripts.html), changes in build dependencies always trigger build.rs rebuilds, and codegen-meta is a build dependency of codegen. So it seems like we may not need codegen-meta's gen_build_deps module.

I also did a quick experiment where I commented out the call to meta::gen_build_deps, and it does appear that everything is rebuilt if I just touch one of the codegen-meta source files, which seems to confirm it.

@data-pup Does this sound right to you? If so, it would simplify the code a little.

data-pup:
Remove gen_build_deps module from codegen-meta crate. Fixes #409.


sunfishcode:
> +0 −62

Woo! ;-)

PR #414, opened 2018-07-23
Use log.rs and implement a file-per-thread logging implementation

bnjbvr:
Still a WIP, but let's start discussing. The main difference with the current design is that new threads need to call `cranelift_logger::setup_logger()` before logging anything. (In the current design, the `dbg!` macro ensures it's initialized each time it's called)

I think the cranelift logger could even be named differently and be made available to everyone as a crate on crates.io, since it looks generic enough to be usable for other projects.

Every user of the `dbg!` macro uses the `log` macros behind the curtain, so they need to contain `log` in their dependencies, and declare macro_use, which is a bit invasive. Fortunately, the same version of `log` can be used in all the cranelift crates.

I'd be happy to get some feedback to know if this is worth continuing!


Amanieu:
You can use `thread_local!` to have the logger initialized when it is first used in the current thread. This avoids the need to call `setup_logger`.

bnjbvr:
> You can use thread_local! to have the logger initialized when it is first used in the current thread. This avoids the need to call setup_logger.

That's right, and it's how the implementation started, because the filename prefix was known at compile time. In the library I'm proposing, the prefix is not known, so the thread_local! variable type is an Option, so we can defer initialization to later, when we're passed the filename prefix we want to use.

Amanieu:
An alternative would be to set up a callback function once at startup, which is called the first time the logger is invoked in each thread.

lib/logger/src/lib.rs
```diff
@@ -0,0 +1,102 @@
...
+    info!(
+        "Set up logging to {:?}; filename prefix is {}",
+        enable, filename_prefix
+    );
+}
+
+#[inline]
+fn enabled() -> bool {
+    STATE.load(atomic::Ordering::Relaxed) > 0
+}
```
sunfishcode: In the old code, this had a check for `cfg!(debug_assertions)` and just returned false if they weren't enabled. Is there a reason for not doing that here?

lib/codegen/Cargo.toml
```diff
@@ -11,11 +11,13 @@ keywords = ["compile", "compiler", "jit"]
 build = "build.rs"
 
 [dependencies]
+cranelift-logger = { path = "../logger", version = "0.1.0", default-features = false }
 cranelift-entity = { path = "../entity", version = "0.16.1", default-features = false }
 failure = { version = "0.1.1", default-features = false, features = ["derive"] }
 failure_derive = { version = "0.1.1", default-features = false }
 hashmap_core = { version = "0.1.9", optional = true }
 target-lexicon = { version = "0.0.3", default-features = false }
+log = "0.4.3"
```
sunfishcode: Would it work to add `default-features = false` to the log dependencies (here and in the other Cargo.toml files)? That will allow us to continue to support `no_std` in these crates.

lib/logger/src/lib.rs
sunfishcode: Also, if we added that here, do you know if that would be enable to allow debugging code to be optimized out in release builds?

lib/logger/src/lib.rs
```diff
@@ -0,0 +1,102 @@
...
+
+struct ThreadPerFileLogger;
+
+impl log::Log for ThreadPerFileLogger {
+    fn enabled(&self, metadata: &Metadata) -> bool {
+        metadata.level() <= Level::Info && enabled()
```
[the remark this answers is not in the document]
bnjbvr: I have to admit this is just a copy/paste of the `log` example. In fact, this will allow up to `Info` but the order is inverted, so we'd have `error!`, `warn!` and `info!` show up, but not `debug!` and `trace!`. Will set to `trace` to allow every type of log commands at the moment, maybe we can add a way to control via an environment variable.

lib/logger/src/lib.rs
bnjbvr: First question: I moved this in the code that initializes the lib in codegen, because it seemed like a Cranelift choice, not something any user of this library would want in general.

Second question: it seems not, but there's a macro for doing so, provided by `log`: [log_enabled](https://docs.rs/log/0.4.3/log/macro.log_enabled.html)

bnjbvr:
I've pushed a few new things:

- no more `CRANELIFT_DBG` environment variable. We now control the level of debugging with the `RUST_LOG` variable, as most loggers do. It's very minimal in this implementation, which means that we can only allow filter levels (off, error, warn, info, debug, trace), but it could reuse the same parsing as `env_logger` has, for instance, in the future. Just setting the env variable enables logging into files.
- no more `dbg!` macro. We can simply use the `log` macros. I've changed most things so they use `debug!`, because a lot of output seems to be debug info in e.g. the register allocation, but we can change this easily.
- I've added another logger implementation on the top level crate (as an example I've chosen `pretty_env_logger`, but any other one would work too). We can now run Cranelift with `--debug` and it will print all the debug output to stderr/stdout, instead of using one file per thread. As a bonus of using the `RUST_LOG` variable (which seems pretty standard), selecting the filter level works the same with this new crate.

lib/codegen/src/dbg.rs
```diff
@@ -1,132 +1,21 @@
...
-}
-
-/// `dbg!` isn't supported in `no_std` mode, so expand it into nothing.
-#[cfg(not(feature = "std"))]
-#[macro_export]
-macro_rules! dbg {
-    ($($arg:tt)+) => {};
+pub fn setup_logging() {
+    cranelift_logger::initialize("cranelift.dbg.");
 }
```
sunfishcode: If I understand what's going on here, this is the only place where cranelift_codegen depends on cranelift_logger explicitly. Is there a way we can move this, to avoid that dependency?

lib/logger/src/lib.rs
```diff
@@ -0,0 +1,120 @@
...
+            "warn" => LevelFilter::Warn,
+            "info" => LevelFilter::Info,
+            "debug" => LevelFilter::Debug,
+            "trace" => LevelFilter::Trace,
+            _ => {
+                eprintln!("unknown RUST_LOG value, defaulting to Trace");
+                LevelFilter::Trace
+            }
+        }
+    });
```
sunfishcode: This led me to peruse various docs, and I happened to notice the comment in here which gave me an idea:

https://docs.rs/env_logger/0.5.11/env_logger/struct.Builder.html#method.build

Would it work and make sense to nest env_logger or pretty_env_logger inside cranelift_logger? Then they could parse RUST_LOG for us, provide their filtering features, and so on.

If that doesn't work out well, that's ok, I just wanted to put the idea out there.

lib/codegen/src/dbg.rs
bnjbvr: That's correct. This function is used in 3 external locations (2 when creating new threads, in particular, which need to set up their TLS state), so that was one way to avoid leaking the filename prefix `cranelift.dbg`. We could just not care too much about uniformity in naming and have the callers directly call `cranelift_logger::initialize`, if you'd prefer. Alternatively, if there's an initialization function anywhere in any cranelift crate, we could use that instead.

lib/logger/src/lib.rs
bnjbvr: I wondered about this too, I'll try tomorrow.

lib/codegen/src/dbg.rs
sunfishcode: What if we define the string `cranelift.dbg` in cranelift-codegen and export it, so callers can call `cranelift_logger::initialize` themselves without hard-coding the string?

bnjbvr:
We could indeed use the `env_logger` crate to parse the `RUST_LOG` variable, which brings in more improvements, including the capability to set a filter level per module (file), for any number of file. This is even nicer.

lib/codegen/Cargo.toml
```diff
@@ -16,6 +16,7 @@ failure = { version = "0.1.1", default-features = false, features = ["derive"] }
 failure_derive = { version = "0.1.1", default-features = false }
 hashmap_core = { version = "0.1.9", optional = true }
 target-lexicon = { version = "0.0.3", default-features = false }
+log = { version = "0.4.3", default-features = false}
```
sunfishcode: The `dbg!` macro had a way to disable itself in release builds, which was controlled by `cfg!(debug_assertions)`. Fortunately, it looks like the `log` crate can do this to, and is more flexible: https://docs.rs/log/0.4.3/log/#compile-time-filters . What would you think about setting either `release_max_level_info` or maybe `release_max_level_warn` here, so that most log messages are disabled in release builds?

Whatever we do here, we should do the same thing for the wasm and module crates too.

lib/module/src/module.rs
```diff
@@ -463,7 +463,7 @@ where
     pub fn define_function(&mut self, func: FuncId, ctx: &mut Context) -> ModuleResult<()> {
         let compiled = {
             let code_size = ctx.compile(self.backend.isa()).map_err(|e| {
-                dbg!(
+                warn!(
```
sunfishcode: Bikeshed: warn "Designates hazardous situations.", so this message seems more like an info.

bnjbvr:
I've now address last comments and published the `file-per-thread-logger` crate so this merge request becomes much simpler now. ([source](https://github.com/bnjbvr/file-per-thread-logger) and [crate](https://crates.io/crates/file-per-thread-logger))

sunfishcode:
Thanks for seeing this all the way through!

PR #417, opened 2018-07-24
Add fuzz target for cranelift_reader::parse_test.

frewsxcv:


sunfishcode:
Cool! It's fun how simple it is to fuzz things like that.

The parser hasn't been hardened at all and probably has a lot of bugs like this. Perhaps this can be the start of making it more robust!

PR #421, opened 2018-07-25
Fixed remaining clippy warnings

Aaronepower:
This will fix all current clippy warnings except for the ones fixed in #408.


sunfishcode:
Cool, thanks!

PR #423, opened 2018-07-26
lib/codegen-meta moved into lib/codegen.

### LINKED ISSUE #415: Move codegen-meta into codegen
We should move `cranelift-codegen-meta` into `cranelift-codegen`.

@sunfishcode feel free to chip in with any advice about doing this, or if publishing the meta crate separately does make more sense. Otherwise, I can try and take care of this sometime this week : )

data-pup:
Fixes #415.

To better encapsulate the `codegen-meta` crate, it has been moved into the codegen crate itself, directly alongside the existing Python metaprogramming code.

This is still in a separate crate, for a few reasons. In #342, we mentioned that using the build.rs would be an intermediate step along the way towards eventually using procedural macros. If we do end up going this route in the future, these will need to be separate, because of this caveat mentioned [here](https://doc.rust-lang.org/book/second-edition/appendix-04-macros.html#procedural-macros-for-custom-derive).

> The next step is to define the procedural macro. At the time of this writing, **procedural macros need to be in their own crate.**
>
> The two crates will need to be published separately, and programmers using these crates will need to add both as dependencies and bring them both into scope. We could **instead have the hello_macro crate use hello_macro_derive as a dependency and reexport the procedural macro code.**

For our case, I think we can do what is mentioned at the end there. If the meta crate is used as a dependency in the codegen crate, we can reexport things that may be needed elsewhere if need be. Then, when we do decide to move towards a procedural macro implementation, we shouldn't need to change very much structurally.


sunfishcode:
Looking ahead, I think a really natural path for this crate to live would be lib/codegen/meta. Except that that's where the python scripts are :o). So what if we renamed the existing meta directory to "meta-python" or something, and then renamed the Rust codegen-meta to just meta?

I wonder if it would even make sense to even name the crate "meta", and let the fact that it's (a) not published, and (b) under the codegen directory, be sufficient to avoid confusion.

data-pup:
Awesome! I think that makes a lot of sense. I can take care of that this weekend :smile:

sunfishcode:
Thanks!

PR #429, opened 2018-07-27
Implement wasm saturating conversions

bnjbvr:
I initially thought I could factor the code more, since the behavior doesn't seem to change much with respect to the non-saturating ops. In fact, this is different enough that a different function is justified for the signed comparison; for the unsigned one, we could maybe use a boolean toggle to decide if we want a saturating operation or not. Will investigate more.

Still a work in progress; I need to test it against Baldrdash to make sure it is correct.

### DIFF
--- a/filetests/wasm/conversions.clif
+++ b/filetests/wasm/conversions.clif
@@ -69,6 +69,54 @@ ebb0(v0: f64):
     return v1
 }
 
+function %i32_trunc_s_sat_f32(f32) -> i32 {
+ebb0(v0: f32):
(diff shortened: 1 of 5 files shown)


bnjbvr:
Tested with Baldrdash and this passes all the Spidermonkey saturating conversion tests \o/

bnjbvr:
Finally got it working and passing tests in Spidermonkey (after updating the baldrdash patch), so it is ready for review.

bnjbvr:
Updated, thanks.

sunfishcode:
Thanks!

Issue #343, opened 2018-05-18
Use types to distinguish between wasm function body indices and wasm function indices


sunfishcode:
The function index space of a WebAssembly module includes the imported functions, followed by the functions defined in the module. Unfortunately, this is often error prone, because if one has indices into eg. the Function or Code sections, those index spaces don't include the imported functions, and it's easy to get such indices confused with indices into spaces that do include the imported functions.

It'd be nice if we could have a distinct types for each of the two kinds of indices, to catch common errors where one performs indexing with the wrong type. For example, `DummyModuleInfo`'s `functions` includes imports, while `function_bodies` doesn't, which would be a good place to start. One way to do this would be to convert them from `Vec` to `PrimaryMap`, which is essentially a `Vec`, but which has infrastructure for declaring specialized index types. lib/codegen/src/ir/entities.rs has examples of declaring dedicated index types for use with `PrimaryMap`.

lachlansneff:
Yeah, this has bitten me a couple of times.

mlea13:
I only have a little bit of experience with Rust but the CraneLift project looks really cool and I'd love to try to help! I was thinking this might be a good place to start?

Just took a quick look at the code. Is there anything special about the `EntityRef` trait defined in lib/entity/src/lib.rs or would it be just as good to use two types that map to a usize similar to the Index types in lib/wasm/src/translation_utils.rs and then have a enum type that could be either of those?

sunfishcode:
Yes, I think this would be a good place to start!

The purpose of EntityRef and PrimaryMap is to have indicies that can't be acxidentally used in the wrong space, using types. If one mistakenly uses a `FunctionIndex` in the `function_bodies` array today, it's a bug. If we used PrimaryMap and EntityRef, they would make this a type error -- at compile time.

An enum would keep track of the kind dynamically. That would make such a bug a runtime error, which would be nicer than letting the bug propagate, but not as nice as a compile-time error :-).

mlea13:
Ah okay, yeah makes sense!

PR #436, opened 2018-08-02
Use types to distinguish between wasm function body indices and wasm function indices.

mlea13:
This fixes https://github.com/CraneStation/cranelift/issues/343.

### DIFF
--- a/lib/wasm/src/environ/dummy.rs
+++ b/lib/wasm/src/environ/dummy.rs
@@ -71,7 +71,7 @@ pub struct DummyModuleInfo {
     pub globals: Vec<Exportable<Global>>,
 
     /// The start function.
-    pub start_func: Option<DefinedFuncIndex>,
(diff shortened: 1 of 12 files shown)


lib/wasm/src/environ/dummy.rs
```diff
@@ -69,7 +70,7 @@ pub struct DummyModuleInfo {
     pub globals: Vec<Exportable<Global>>,
 
     /// The start function.
-    pub start_func: Option<FunctionIndex>,
+    pub start_func: Option<DefinedFuncIndex>,
```
sunfishcode: The wasm start function is in the `FuncIndex` space.

sunfishcode:
Also, the CI failure is because Rust 1.28.0 was just released and rustfmt changed. This is now fixed on master, so you can rebase/merge to fix that.

sunfishcode:
Looks good!

PR #437, opened 2018-08-02
FunctionBuilder no longer Variable-parametric (and updated doc)

denismerigoux:
Following some reports of confusion over how to translate SSA or non-SSA "variables", improved the doc so explain better how the builder API handles them.

### DIFF
--- a/lib/frontend/src/lib.rs
+++ b/lib/frontend/src/lib.rs
@@ -80,10 +80,10 @@
 //!     let mut sig = Signature::new(CallConv::SystemV);
 //!     sig.returns.push(AbiParam::new(I32));
 //!     sig.params.push(AbiParam::new(I32));
-//!     let mut fn_builder_ctx = FunctionBuilderContext::<Variable>::new();
+//!     let mut fn_builder_ctx = FunctionBuilderContext::new();
 //!     let mut func = Function::with_name_signature(ExternalName::user(0, 0), sig);
 //!     {
-//!         let mut builder = FunctionBuilder::<Variable>::new(&mut func, &mut fn_builder_ctx);
+//!         let mut builder = FunctionBuilder::new(&mut func, &mut fn_builder_ctx);
 //!
 //!         let block0 = builder.create_ebb();
 //!         let block1 = builder.create_ebb();
--- a/lib/frontend/src/lib.rs
+++ b/lib/frontend/src/lib.rs
@@ -6,35 +6,38 @@
 //! To get started, create an [`FunctionBuilderContext`](struct.FunctionBuilderContext.html) and
 //! pass it as an argument to a [`FunctionBuilder`](struct.FunctionBuilder.html).
 //!
-//! # Source Language Variables and Cranelift IR values
+//! # Mutable variables and Cranelift IR values
 //!
 //! The most interesting feature of this API is that it provides a single way to deal with all your
 //! variable problems. Indeed, the [`FunctionBuilder`](struct.FunctionBuilder.html) struct has a
-//! type parameter `Variable` that should be instantiated with the type of your Source Language
-//! Variables (SLV). Then, through calling the functions
+//! type parameter `Variable` that should be instantiated with the type of your source language
+//! variables. Then, through calling the functions
 //! [`declare_var`](struct.FunctionBuilder.html#method.declare_var),
 //! [`def_var`](struct.FunctionBuilder.html#method.def_var) and
 //! [`use_var`](struct.FunctionBuilder.html#method.use_var), the
 //! [`FunctionBuilder`](struct.FunctionBuilder.html) will create for you all the Cranelift IR
-//! values corresponding to your SLVs.
+//! values corresponding to your variables.
 //!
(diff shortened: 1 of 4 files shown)


lib/frontend/src/lib.rs
```diff
@@ -1,12 +1,43 @@
...
+//! values corresponding to your SLVs.
+//!
+//! If a SLV is immutable (defined only once), then it will get mapped to one and only Cranelift IR
+//! value. If a SLV is mutable ([`def_var`](struct.FunctionBuilder.html#method.def_var) multiple
+//! times), then internally a [`SSA`](https://en.wikipedia.org/wiki/Static_single_assignment_form)
+//! construction algorithm will automatically create multiple Cranelift IR values that will be
+//! returned to you by [`use_var`](struct.FunctionBuilder.html#method.use_var) depending on where
+//! you want to use your SLV.
+//!
+//! The morality is that you should use these three functions to handle all your SLVs, even those
```
sunfishcode: The idiom you want here is "The moral is ...".

lib/frontend/src/lib.rs
```diff
@@ -1,12 +1,43 @@
...
+//! that are not present in the source code but artefacts of the translation. For instance, if your
+//! source language is expression-based, then you will need to introduce artifical SLVs to store
+//! intermediate results of the computation of your expressions. Hence The `Variable` type that you
```
sunfishcode: This isn't clear to me. Intermediate results of expressions are naturally in SSA form, so they don't need to go through `Variable`s. And, it has worked well in frontends I've worked with to have the main expression translation function return a `Value`. For arithmetic expressions, it's the `Value` of the instruction, and for variable references, it's the `Value` returned by `use_var. This approach avoids creating lots of unnecessary `Variable`s for `SSABuilder` to track.

lib/frontend/src/lib.rs
```diff
@@ -1,12 +1,43 @@
...
 //!
 //! To get started, create an [`FunctionBuilderContext`](struct.FunctionBuilderContext.html) and
 //! pass it as an argument to a [`FunctionBuilder`](struct.FunctionBuilder.html).
 //!
+//! # Source Language Variables and Cranelift IR values
+//!
+//! The most interesting feature of this API is that it provides a single way to deal with all your
+//! variable problems. Indeed, the [`FunctionBuilder`](struct.FunctionBuilder.html) struct has a
+//! type parameter `Variable` that should be instantiated with the type of your Source Language
+//! Variables (SLV). Then, through calling the functions
```
sunfishcode: It feels a bit awkward to introduce an acronym here, especially since as you mention below, this also applies to artifacts of translation, so "Source Level Variables" isn't quite the right term. What would you think about introducing these as "mutable" variables, and say that these can represent mutable variables either in the source or arising from the translation?

lib/frontend/src/lib.rs
denismerigoux: Indeed, you're right. I had forgotten how functions to create instructions returned the value corresponding to their result. I'll change that.

lib/frontend/src/lib.rs
denismerigoux: Thanks :)

lib/frontend/src/lib.rs
```diff
@@ -1,12 +1,46 @@
...
+//! The moral is that you should use these three functions to handle all your mutable variables, even those
+//! that are not present in the source code but artefacts of the translation. Hence The `Variable` type that you
+//! would pass to [`FunctionBuilder`](struct.FunctionBuilder.html) could look like this
+//!
+//! ```
+//! enum Variable {
+//!     OriginalSourceVariable(String),
+//!     TranslationArtefact(u32)
+//! }
+//! ```
```
sunfishcode: `FunctionBuilder` requires the `Variable` parameter to implement [`EntityRef`](https://docs.rs/cranelift-entity/0.18.1/cranelift_entity/trait.EntityRef.html), so it doesn't work well to put `String`s or other data in it. And the advice to use source language variables with it only works for source language variable types which are easy to implement `EntityRef` on.

I'm actually pondering whether we should remove the `Variable` type parameter and just use a concrete `Variable` type instead, and require API users to convert their variables into and out of `Variable` manually. That could be a little tedious, but because `Variable`s were already required by the `EntityRef` trait to basically be `usize` wrappers, so it should be straightforward to convert such types into a concrete `Variable` type (which would itself be a trivial `usize` wrapper). I'm interested in what you think about this approach.

denismerigoux: You're right about the `String` not being a good example. Concerning whether we should make the `Variable` type concrete or not, I agree with you. Looking at the `EntityRef` trait closer, I now realize only aliases of `usize` (or another integer type) can implement it. So it makes no sense to introduce parametricity for something that can only be one type. It will be up to the user to keep a mapping between their program variables and the `usize` index that correspond to it. In that sense I suggest renaming it to `VariableIndex`.

sunfishcode:
Looks good. Thanks!

PR #440, opened 2018-08-03
Stackmaps

mirahmed753:
Added stackmap instruction for reference types. Created clif tests for stackmap instruction.


bjorn3:
Those merge commits cause unrelated changes like CraneLift -> Cretonnne. Could you please rebase?

lib/codegen/src/regalloc/stackmaps.rs
```diff
@@ -0,0 +1,99 @@
...
+            if opcode.is_branch() {
+                // Find what the branch destination is
+                let branch_dest = pos.func.dfg[inst].branch_destination();
+                ebbs_for_stackmap.insert(branch_dest);
+            }
+
+            // Check if it's a call instruction
+            if opcode.is_call() {
+                // insert stackmap instruction
+                let mut value_list = Vec::new();
```
sunfishcode: We should move the `live_value_list` computation below up here, so that we can add the live values to the actual stackmap instruction.

lib/codegen/src/regalloc/stackmaps.rs
```diff
@@ -0,0 +1,99 @@
...
+            // Get opcode of instruction
+            let opcode = pos.func.dfg[inst].opcode();
+
+            println!("Instruction Data: {}", pos.func.dfg.display_inst(inst, None));
+
+            // Check if it's a branch instruction
+            if opcode.is_branch() {
+                // Find what the branch destination is
+                let branch_dest = pos.func.dfg[inst].branch_destination();
+                ebbs_for_stackmap.insert(branch_dest);
```
sunfishcode: Can you add a TODO comment here mentioning that we should only add a stackmap if the branch_dest is above the branch in the layout?

lib/codegen/src/regalloc/stackmaps.rs
```diff
@@ -0,0 +1,99 @@
...
+    } // end for loop for ebb
+
+    // Note: Using a HashSet to insert the stackmap instructions is a temporary method
+    // loop through items in HashSet to insert stackmap instruction
+    for ebb in ebbs_for_stackmap {
+        pos.goto_first_insertion_point(ebb.unwrap());
+
+        // insert stackmap instruction
+        let mut value_list = Vec::new();
+        pos.ins().stackmap(&value_list);
```
sunfishcode: Can you add a TODO comment here mentioning that we need to record the live values at the tops of these ebbs, so that we can populate the stackmap's operands?

lib/codegen/src/ir/builder.rs
```diff
@@ -123,6 +123,8 @@ impl<'f, IIB: InstInserterBase<'f>> InstBuilderBase<'f> for InsertBuilder<'f, II
     }
 
     fn build(mut self, data: InstructionData, ctrl_typevar: Type) -> (Inst, &'f mut DataFlowGraph) {
+        println!("Type: {}", ctrl_typevar);
```
lachlansneff: Just fyi, seems like you missed removing this line.

lachlansneff:
Awesome.

sunfishcode:
This is now proceeding in #781.

PR #441, opened 2018-08-03
Updating Sphinx link and install

carolinecullen:
Trying to build the documentation and needed to install python 3 so I updated it and the old https link was not working because Sphinx's settings were not correct so the link had to be changed to http.

### DIFF
--- a/README.md
+++ b/README.md
@@ -126,7 +126,7 @@ Building the documentation
 --------------------------
 
 To build the Cranelift documentation, you need the [Sphinx documentation
-generator](https://www.sphinx-doc.org/):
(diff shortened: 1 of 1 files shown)


sunfishcode:
According to [this](http://www.sphinx-doc.org/en/master/usage/installation.html), Sphinx supports Python2. Can you say more about the problems you encountered?

carolinecullen:
Error message when trying to pip install sphinx with python 2: 

Could not install packages due to an EnvironmentError: [Errno 13] Permission denied: '/Library/Python/2.7/site-packages/Sphinx-1.7.6.dist-info'

sunfishcode:
Thanks!

PR #452, opened 2018-08-10
Verifier now accepts multiple errors (fixes #387).

### LINKED ISSUE #387: Rework verifier
The `Verifier` has two main flaws:
- Error messages are arbitrary strings, and do not carry much information (see #284).
- Error messages cannot be reported in batch; if multiple errors are present, one error will only be diagnosed once the previous one is fixed.

## Carry more information
The first issue can be fixed by having specific codes for each diagnostic. I'm sorry to bring in my own project, but [here is how I personally handled diagnostics in my compiler](https://github.com/6A/styx-history/blob/c1bff1983e2faf5ce7deadb281ee0a936dafadd1/styx.rs/src/diagnostics.rs#L348-L607).

Diagnostics have a severity (optional here in Cretonne), a code, a message and a span (the two latter are already in Cretonne). Additionally, the `DiagnosticCodes` struct exists, and lists all existing diagnostic codes.

Finally, diagnostics are defined using a macro that adds a function for each diagnostic, as well as a const in the `DiagnosticCodes` struct.

For instance,
```rust
def_error!(0x20, undefined_symbol, symbol: &str => "Undefined symbol '{}'.");
```

would produce
```rust
impl DiagnosticCodes {
    /// Identifiying code of the `undefined_symbol` `Diagnostic`.
    #[allow(non_upper_case_globals)]
    pub const undefined_symbol: usize = 0x20;
}

impl Diagnostic {
    /// (Error) `Diagnostic` with message "Undefined symbol '{}'.".
    #[inline]
    pub fn undefined_symbol(span: Span, symbol: &str) -> Self {
        Self::error(DiagnosticCodes::undefined_symbol, format!("Undefined symbol '{}'.", symbol), span)
    }
}
```

## Report in batch
The verifier has steps that are executed sequentially. Instead of returning `Result<(), VerifierError>`, we could pass a `&mut Vec<VerifierError>` to each sub function, which would then report errors. Furthermore, if a function returns `false`, we consider the error fatal and stop the process right away, since further checks may expect this one check to succeed.

Finally, depending on whether our vector is empty, we consider the function either valid or invalid.

6A:
Tests all pass successfully, and have been updated accordingly.

Besides the following change, everything behaves like before, meaning that every single error is reported using `return fatal!(...)`, which is equivalent to the previous `return err!(...)`.

However, from now on we can define errors as non fatal, and keep verifying. For example, instead of reporting the first error we encounter when ensuring a function's parameters are valid, we can report all errors, and stop verifying the function after checking the parameters.

The mentioned change is that now, error reporting in `pretty_error` reports **every single** reported error, instead of the first one.

I'll probably work on making most errors `nonfatal` after this.

**Edit**: I did not change the version... Maybe I should have?

**Edit 2**: Here are more details about the changes:

1. Added `VerifierErrors` type, which is a just a wrapper around a `Vec<VerifierError>`. It can also be converted to `CodegenError` easily.
2. Now, every function that returns a `VerifierResult` also takes a `&mut VerifierErrors` as parameter. Errors can be added via the macros `report!(errors, "{}", args)`, `fatal!(errors, "{}", args)` (calls `report!` and then returns `Err`), and `nonfatal!(errors, "{}", args)` (calls `report!` and returns `Ok`).
3. `VerifierResult<T>` is now defined as `Result<T, ()>`, where `Ok` represents a lack or error or an error that does not stop the verification process, and `Err` represents an error that does stop the verification process (therefore using `?` can shortcurcuit a `verify_foo` function).
4. The macro `ok_if_no_error` returns `Ok(())` if no error was reported in a `VerifierErrors`, and `Err(())` otherwise. Can be useful for methods that do not shortcircuit, but stop the verification process if they fail.
5. The macro `verify!` takes a function (ie `verify_function`, `ctx.verify`), and an arbitrary number of arguments, and will take care of creating the `VerifierErrors`, passing it to the `verify` function along with the other arguments, and will either return `Ok` if no error was reported, or `Err(VerifierErrors)` otherwise.


lib/codegen/src/regalloc/context.rs
```diff
@@ -134,21 +162,48 @@ impl Context {
...
+                    &self.virtregs,
+                    &mut errors,
+                ).is_ok();
+
+            if !ok {
+                return Err(errors.into());
+            }
+        }
+
+        if errors.0.is_empty() {
```
sunfishcode: Can this condition ever be false? It seems like if we have errors, one of the `if !ok` checks above should have tripped. Can this be an assert instead?

src/wasm.rs
```diff
@@ -148,9 +149,12 @@ fn handle_module(
 
         let func_index = num_func_imports + def_index.index();
         if flag_check_translation {
-            context
-                .verify(fisa)
-                .map_err(|err| pretty_verifier_error(&context.func, fisa.isa, None, &err))?;
+            let mut errors = VerifierErrors::default();
+            let _ = context.verify(fisa, &mut errors);
+
+            if !errors.0.is_empty() {
```
sunfishcode: Instead of testing `errors.0.is_empty()` here, it feels cleaner to test whether the return value of `verify` `.is_err()`.

lib/frontend/src/lib.rs
```diff
@@ -118,11 +118,11 @@
...
 //!
 //!     let flags = settings::Flags::new(settings::builder());
-//!     let res = verify_function(&func, &flags);
+//!     let mut errors = VerifierErrors::default();
+//!     let res = verify_function(&func, &flags, &mut errors);
 //!     println!("{}", func.display(None));
-//!     match res {
-//!         Ok(_) => {}
-//!         Err(err) => panic!("{}", err),
+//!     if !errors.0.is_empty() {
```
sunfishcode: Instead of testing errors.0.is_empty() here, it feels cleaner to test whether the return value of `verify_function` .is_err().

lib/codegen/src/verifier/mod.rs
```diff
@@ -82,23 +82,77 @@ pub use self::cssa::verify_cssa;
...
+    ( $verif: ident . $fun: ident, $( $arg: expr ),+ ) => ({
+        let mut errors = $crate::verifier::VerifierErrors::default();
+        let _ = $verif.$fun( $( $arg ),+, &mut errors);
+
+        if errors.0.is_empty() {
+            Ok(())
+        } else {
+            Err(errors)
+        }
+    });
```
sunfishcode: Could this be simplified to something like
```rust
$verif.$fun( $( $arg ),+, &mut errors).map_err(|_| errors)
```
?

lib/codegen/src/verifier/mod.rs
```diff
@@ -82,23 +82,77 @@ pub use self::cssa::verify_cssa;
...
+
+    ( $fun: path, $( $arg: expr ),+ ) => ({
+        let mut errors = $crate::verifier::VerifierErrors::default();
+        let _ = $fun( $( $arg ),+, &mut errors);
+
+        if errors.0.is_empty() {
+            Ok(())
+        } else {
+            Err(errors)
+        }
```
sunfishcode: Ditto :-).

lib/codegen/src/regalloc/context.rs
6A: It will be `false` if none of the reported errors were fatal. `ok` is `true` iff no fatal error was reported; non-fatal may have been reported, which would make the verification process stop now.

src/wasm.rs
6A: Ditto, the `Result` may be `Ok` but a (non-fatal) error *might* have been logged anyway.

lib/frontend/src/lib.rs
6A: Ditto. I understand that it feels a little hacky to allow an `Ok` return value without it being necessarily fine, but I thought it worked well with `?`. Furthermore, right now my objective was to introduce the new VerifierErrors type, and to use it everywhere. For now I consider (most) errors as fatal, even though some of them shouldn't halt the process completely.

lib/codegen/src/verifier/mod.rs
6A: Indeed, I'm still paranoid about using `.map_err` in macros when raw syntax could be used, but I suppose the compiler will inline it correctly.

6A: Nevermind, I actually want to return an error if an error was reported (`$verif.$fun` might return `Ok` and report errors anyway).

sunfishcode:
To answer your question above, we don't need to bump versions manually; cranelift's API isn't stable yet so we're typically bumping the version on every release right now.

lib/codegen/src/verifier/mod.rs
```diff
@@ -120,15 +174,58 @@ impl Display for VerifierError {
 }
 
 /// Verifier result.
-pub type VerifierResult<T> = Result<T, VerifierError>;
+pub type VerifierResult<T> = Result<T, ()>;
```
sunfishcode: Thought experiment: What if we split `VerifierResult` into two, like
```rust
type VerifierResult<T> = Result<T, VerifierErrors>;
type VerifierPartialResult<T> = Result<T, ()>;
```
and then made every function that currently uses `VerifierResult` *either*:
 - use `VerifierPartialResult` and take an &mut `VerifierResults`, or
 - use `VerifierResult` and don't take an &mut `VerifierResults`
?

6A:
Okay, I modified all ``verify`` functions that weren't used "in groups" so that they all return `VerifierResult<()>` (aka `Result<(), VerifierErrors>`) and take not `&mut VerifierErrors`. All other functions now return `VerifierStepResult<()>` (aka `Result<(), ()>`).

I also added more details on how both of them should be used in their doc comments.

lib/codegen/src/verifier/mod.rs
```diff
@@ -82,23 +82,86 @@ pub use self::cssa::verify_cssa;
...
+
+/// Diagnose a non-fatal error, and return `Ok`.
+macro_rules! nonfatal {
+    ( $( $arg: expr ),+ ) => ({
+        report!( $( $arg ),+ );
+        Ok(())
+    });
+}
+
+/// Faster syntax for calling `verify_*` functions.
```
sunfishcode: "Shorthand" would convey the purpose here better than "faster". Also, a good one-line summary of this macro is that it adapts a function returning a `VerifierStepResult` to return a `VerifierResult`.

lib/codegen/src/verifier/mod.rs
```diff
@@ -82,23 +82,86 @@ pub use self::cssa::verify_cssa;
...
+///
+/// # Example
+/// ```rust,ignore
+/// // Instead of doing all this manually
+/// let mut errors = VerifierErrors::default();
+/// let result = verify_function(func, &mut errors);
```
sunfishcode: I know this is only an example, but it should use `verify_context` or so which returns a `VerifierStepResult` to avoid confusion.

lib/codegen/src/verifier/mod.rs
```diff
@@ -82,23 +82,86 @@ pub use self::cssa::verify_cssa;
...
+///
+/// if errors.is_empty() {
+///     Ok(())
+/// } else {
+///     Err(errors)
+/// }
+///
+/// // This macro allows you to write
+/// verify!(verify_func, func)
```
sunfishcode: Update the name here too.

lib/codegen/src/context.rs
```diff
@@ -177,31 +177,43 @@ impl Context {
...
     pub fn verify<'a, FOI: Into<FlagsOrIsa<'a>>>(&self, fisa: FOI) -> VerifierResult<()> {
-        verify_context(&self.func, &self.cfg, &self.domtree, fisa)
+        let mut errors = VerifierErrors::default();
+        let _ = verify_context(&self.func, &self.cfg, &self.domtree, fisa, &mut errors);
+
+        if errors.is_empty() {
+            Ok(())
+        } else {
+            Err(errors)
+        }
```
sunfishcode: Can this use `verify!`?

lib/codegen/src/context.rs
```diff
@@ -177,31 +177,43 @@ impl Context {
...
     pub fn verify_locations(&self, isa: &TargetIsa) -> VerifierResult<()> {
-        verify_locations(isa, &self.func, None)
+        let mut errors = VerifierErrors::default();
+        let _ = verify_locations(isa, &self.func, None, &mut errors);
+
+        if errors.is_empty() {
+            Ok(())
+        } else {
+            Err(errors)
+        }
```
sunfishcode: Can this use `verify!`?

lib/codegen/src/verifier/mod.rs
6A: Good catch! I wrote the docs *before* changing the signature of `verify_function` so I went over this too quickly...

lib/codegen/src/context.rs
6A: `context` is imported before `verifier` therefore the `verify!` macro doesn't exist yet when this is parsed, so either we change the order at which the modules are loaded or we keep this.

lib/codegen/src/context.rs
6A: Ditto.

sunfishcode:
Looks good, thanks for seeing this through!

Issue #318, opened 2018-05-01
Jump Tables


sunfishcode:
Codegen currently legalizes `br_table` instructions by [lowering them to sequences of branches](https://github.com/cretonne/cretonne/blob/master/lib/codegen/src/legalizer/mod.rs#L187). What we want is to use actual jump tables.

This will require some new `binemit` capabilities, for emitting the actual jump table data, outside of the function body.

In theory, the `RelocSink` trait has support for relocating jump tables (`reloc_jt`), however as this hasn't been tested, it may need to be modified or extended.

A jump table is an array, and the elements can be either absolute addresses or offsets relative to some base, often the base of the jump table itself. PIC prefers relative offsets, so we should start with that, and do absolute offsets later if extra speed for non-PIC is needed. Using a jump table typically involves a bounds-check to see if the `br_table` index is in bounds, and then an indexed load into the table using the `br_table` index. In the case of a relative offset, the base is then added to the result. Then, the final result can be used as the operand to an indirect jump.

Amanieu:
A form of branch relaxation (#73) also applies to PIC jump tables: the offset from the table base can be encoded as `i8`, `i16` or `i32` depending on how far the targets are from the table.

tyler:
I've started working a bit on this and wanted to put some thoughts here and gather feedback.

Alright, so sementically there are few steps to implementing a jump_table with Cranelift.

1. Bounds check the the value indexing the jump table.
2. Load the address from the jump table.
3. Check that the destination address is not "empty", as our jump tables are sparse.
4. Jump indirectly using that address.

So, an expansion of this:

```
function u0:0(i64) -> i64 system_v {
    jt0 = jump_table ebb1, 0, ebb2

ebb0(v0: i64):
    br_table v0, jt0
    ...

ebb1: ...
ebb2: ...
}
```

Might look something like this:

```
function u0:0(i64) -> i64 system_v {
    jt0 = jump_table ebb1, 0, ebb2

ebb0(v0: i64):
    v1 = ifcmp_imm v0, 3
    brif uge v7, ebb3

    v2 = load_address_from_jump_table v0, jt0   # v2 is an integer type
    v3 = ifcmp_imm v2, 0
    brif eq v3, ebb3

    indirect_jump v2


ebb3: (fallthrough target)

ebb1: ...
ebb2: ...
}
```

[...]

PR #453, opened 2018-08-10
Implement jump tables

tyler:
I'm starting to implement jump tables in this pull request. It is extremely sketchy at the moment. I'm just trying to cut a happy path through the code to make at least one case work, and then will iterate from there.

This is intended to address #318.


tyler:
@sunfishcode When you have time, would you have a look at this? It's messy and woefully incomplete, but I want to make sure I'm on the right track here.

lib/codegen/meta-python/base/instructions.py
```diff
@@ -51,6 +51,7 @@
 EBB = Operand('EBB', entities.ebb, doc='Destination extended basic block')
 args = Operand('args', VARIABLE_ARGS, doc='EBB arguments')
 
+
```
bjorn3: Why this extra newline?

lib/codegen/meta-python/base/instructions.py
```diff
@@ -144,6 +145,34 @@
         """,
         ins=(x, JT), is_branch=True)
 
+jump_table_entry = Instruction(
+    'jump_table_entry', r"""
+    Get an entry from a jump table.
+
+    Load an address from a specific entry in a jump table.
+    """,
+    ins=(x, JT), outs=addr)
```
sunfishcode: Wording nit: the value being loaded isn't an "address"; in the current patch it's a relative offset from the start of the table (in the future it could be other things too).

lib/codegen/meta-python/base/instructions.py
```diff
@@ -144,6 +145,34 @@
...
+    the relative base back to it.
+    """,
+    ins=(addr, JT), outs=addr)
+
+indirect_jump_table_br = Instruction(
+    'indirect_jump_table_br', r"""
+    Branch indirectly via a jump table entry.
+
+    Unconditionally jump via a jump table entry that was previously loaded
+    with the ``jump_table_entry`` instruction.
```
sunfishcode: ... and adjusted with the `add_jump_table_base` instruction (or whatever it ends up being called) :-).

lib/codegen/meta-python/base/instructions.py
```diff
@@ -144,6 +145,34 @@
...
+add_jump_table_base = Instruction(
+    'add_jump_table_base', r"""
+    Add the base address of a jump table to a value.
+
+    This is used for jump tables wherein the entries are stored relative to
+    the base of jump table. In order to use these, generated code should first
+    load an entry using ``jump_table_entry``, then use this instruction to add
+    the relative base back to it.
+    """,
+    ins=(addr, JT), outs=addr)
```
sunfishcode: I wonder if we could name this something like `decode_jump_table_entry`? It'd still be an add in the code as we know it today, but I think this name would help express that jump tables could be encoded in other ways too.

lib/codegen/src/binemit/mod.rs
```diff
@@ -123,4 +123,18 @@ where
...
+        for idx in 0..jt_data.len() {
+            match jt_data.get_entry(idx) {
+                Some(ebb) => {
+                    let rel_offset: i32 = func.offsets[ebb] as i32 - jt_offset as i32;
+                    sink.put4(rel_offset as u32)
+                }
+                None => sink.put4(0),
+            }
+        }
+    }
```
sunfishcode: I think this is a fine approach, although we should add a function to the `CodeSink` to notify it that the code has finished and have started on some read-only data. It may not mean a lot right now, although it will mean that we can make clif-util's disassembly view smart and only disassemble the bytes that are actually instructions :).

lib/codegen/meta-python/base/instructions.py
tyler: This is a WIP. I’ll clean up all those things toward the end.

lib/codegen/meta-python/base/instructions.py
tyler: Yeah this makes sense. In the future we’d  need to have this expand to different instructions depending on the jump table implementation? How would you want that to work? Wouldn’t we still need a specific `add_jump_table_base` instruction, or is there some way around that?

tyler:
I appreciate that “nitpicking” honestly, because it’s the abstractions that I’m having a bit of trouble with here.

lib/codegen/meta-python/base/instructions.py
sunfishcode: That's a good question, and thinking about it makes me think that I may have been overthinking this ;-). Let's stick with `add_jump_table_base` for now.

It helps that this instruction isn't something that frontends should be producing directly, so it won't be too painful if we need to make changes.

tyler:
This now at least works in some limited test cases locally. So, it should be slightly easier to have a conversation about it.

Test cases, clean up, settings, etc are coming next.

lib/codegen/src/legalizer/mod.rs
```diff
@@ -170,6 +171,74 @@ fn expand_cond_trap(
...
+    // If the table isn't fully dense, zero-check the entry.
+    if !table_is_fully_dense {
+        pos.ins().brz(entry, fallthrough_ebb, &[]);
+    }
+
+    let addr = pos.ins().iadd(base_addr, entry);
+    pos.ins().indirect_jump_table_br(addr, table);
+
+    let ebb = pos.current_ebb().unwrap();
+    pos.insert_ebb(fallthrough_ebb);
```
sunfishcode: It looks like the fallthrough_ebb doesn't have any instructions in it. And, since Cranelift's `br_table` falls through in the default case, in theory there ought to be code here for splitting the current block at that point.

One option which might simplify things here would be to change Cranelift's `br_table` to be more like wasm's `br_table` and have an explicit "default" destination ebb, rather than just falling through. That way, it could have `is_terminator` set, and then you wouldn't need to create any new basic blocks.

src/compile.rs
```diff
@@ -171,7 +172,25 @@ cfg_if! {
...
+
+                if let Some(s) = i.mnemonic() {
+                    write!(&mut line, "{}\t", s).unwrap();
+                }
+
+                if let Some(s) = i.op_str() {
+                    write!(&mut line, "{}", s).unwrap();
+                }
+
+                println!("{}", line);
```
sunfishcode: Neat! One additional thing that would be nice to do here is pass `print_disassembly` a subslice of the memory which is only `code_size` bytes long, so that we don't attempt to disassemble the jump table entries :-).

lib/codegen/src/binemit/relaxation.rs
```diff
@@ -95,6 +95,11 @@ pub fn relax_branches(func: &mut Function, isa: &TargetIsa) -> CodegenResult<Cod
         }
     }
 
+    for (jt, jt_data) in func.jump_tables.iter() {
+        func.jt_offsets[jt] = offset;
+        offset += jt_data.len() as u32 * 4; // TODO: size should be based on settings
```
sunfishcode: Ideally, rather than settings, we'd want the relaxation pass to compute the minimum size that can hold all the offsets, and use that. For now it's sufficient to update the comment :-).

lib/codegen/src/ir/instructions.rs
```diff
@@ -194,7 +194,13 @@ impl InstructionData {
                 ref args,
                 ..
             } => BranchInfo::SingleDest(destination, &args.as_slice(pool)[2..]),
-            InstructionData::BranchTable { table, .. } => BranchInfo::Table(table),
+            InstructionData::BranchTable { table, .. } => {
+                if !self.opcode().is_branch() {
+                    BranchInfo::NotABranch
```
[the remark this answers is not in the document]
tyler: Nice catch. In a previous iteration, `branch_table_entry` was sharing the same `InstructionFormat`, which led to CFG corruption without this. No longer the case though.

lib/codegen/meta-python/isa/x86/recipes.py
```diff
@@ -1473,6 +1486,36 @@ def floatccs(iform):
         disp4(destination, func, sink);
         ''')
 
+indirect_jmp = TailRecipe(
+        'indirect_jmp', BranchTable, size=1, ins=GPR, outs=(),
+        branch_range=32,
```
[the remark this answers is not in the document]
tyler: When I try to remove the `branch_range` I get the following:
```
Traceback (most recent call last):
...
  File "/Users/tyler/Code/cranelift/lib/codegen/meta-python/cdsl/isa.py", line 475, in __init__
    .format(recipe, self.inst.name))
AssertionError: Recipe Op1indirect_jmp for indirect_jump_table_br must have a branch_range
```
I poked around for a bit, trying to see if there was some way to change that code in `isa.py` to automatically determine whether the branch needs a range. There doesn't seem to be a clean way to do it.

So the options as I see it:
1. Add a new flag to `Instruction` that says "sure, I'm a branch, but I don't need a range".
2. Remove the check in `isa.py`, and try to avoid shooting ourselves in the foot.
3. Something I haven't thought of.

Any ideas?

lib/codegen/src/legalizer/mod.rs
tyler: I don't have a strong opinion on this one. I'd probably default to just splitting the ebb rather than changing the IR.

But I'm in here changing things now, so if we're gonna do that this seems like a good time. Do you have an opinion?

src/compile.rs
tyler: Done and done. It also prints the readonly data afterward now. A simple `br_table` test file I have looks like this:

```
Disassembly of 76 bytes:
   0:	40 55                	push	rbp
   2:	48 89 e5             	mov	rbp, rsp
   5:	40 83 fe 0c          	cmp	esi, 0xc
   9:	40 73 10             	jae	0x1c
   c:	48 8d 05 39 00 00 00 	lea	rax, qword ptr [rip + 0x39]
  13:	48 63 0c b0          	movsxd	rcx, dword ptr [rax + rsi*4]
  17:	48 01 c8             	add	rax, rcx
  1a:	ff e0                	jmp	rax
  1c:	40 eb 24             	jmp	0x43
  1f:	40 b8 64 00 00 00    	mov	eax, 0x64
  25:	40 eb 21             	jmp	0x49
  28:	40 b8 65 00 00 00    	mov	eax, 0x65
  2e:	40 eb 18             	jmp	0x49
  31:	40 b8 66 00 00 00    	mov	eax, 0x66
  37:	40 eb 0f             	jmp	0x49
  3a:	40 b8 67 00 00 00    	mov	eax, 0x67
  40:	40 eb 06             	jmp	0x49
  43:	40 b8 68 00 00 00    	mov	eax, 0x68
  49:	40 5d                	pop	rbp
  4b:	c3                   	ret	

Followed by 48 bytes of read-only data:
   0:  ee ff ff ff  e5 ff ff ff  dc ff ff ff  d3 ff ff ff 
  16:  ee ff ff ff  e5 ff ff ff  dc ff ff ff  d3 ff ff ff 
  32:  ee ff ff ff  e5 ff ff ff  dc ff ff ff  d3 ff ff ff 
```

lib/codegen/src/legalizer/mod.rs
sunfishcode: I think I'd prefer to change the IR, assuming that doesn't turn out to be too much of a detour here.

We'll have to add an Ebb to `BranchInfo::Table` and teach a few things how to handle that. And update lib/wasm/src/code_translator.rs, though that looks straightforward because currently it always does a jump after doing the br_table, so we can just get rid of that jump and add the ebb as the default on the br_table. And in lib/frontend/src/frontend.rs, declare the extra control edge only if it isn't already declared. At a quick scan, that should be the interesting parts. If you want, I can help out with some of this too.

lib/codegen/meta-python/isa/x86/recipes.py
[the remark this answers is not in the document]
sunfishcode: Hmm, yeah. I don't have any other ideas here. So how about option 1? We can call it `is_indirect_branch`. So it'll look a little redundant `is_branch=True, is_indirect_branch=True`, but fortunately there aren't that many of those :-).

lib/codegen/src/legalizer/mod.rs
tyler: No problem! I'll take a stab at it and shout if I need help.

tyler:
@sunfishcode I need to rebase this, but can you take a quick look here? I *think* I have the `br_table` with default destination working.

lib/wasm/src/code_translator.rs
```diff
@@ -309,14 +308,13 @@ pub fn translate_operator<FE: FuncEnvironment + ?Sized>(
                     data.push_entry(branch_ebb);
                 }
```
sunfishcode: The default edge can also be critical, so we should also add a `dest_ebb_sequence` entry for the default edge here too.

lib/codegen/src/dominator_tree.rs
```diff
@@ -359,6 +359,18 @@ impl DominatorTree {
...
+                    for (_, succ) in func.jump_tables[jt].entries() {
+                        if self.nodes[succ].rpo_number == 0 {
+                            self.nodes[succ].rpo_number = SEEN;
+                            self.stack.push(succ);
+                        }
+                    }
+                    if self.nodes[dest].rpo_number == 0 {
+                        self.nodes[dest].rpo_number = SEEN;
+                        self.stack.push(dest);
+                    }
```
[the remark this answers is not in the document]
tyler: I ended up killing `TableWithDefault` as it became painful to factor out the code that applied to it and `Table`. Still factored this one out though.

lib/frontend/src/ssa.rs
```diff
@@ -656,6 +656,32 @@ impl SSABuilder {
...
+                let middle_ebb = func.dfg.make_ebb();
+                func.layout.append_ebb(middle_ebb);
+                let middle_block = self.declare_ebb_header_block(middle_ebb);
+                self.blocks[middle_block].add_predecessor(jump_inst_block, jump_inst);
+                self.mark_ebb_header_block_sealed(middle_block);
+
+                if dest_ebb == default_ebb {
+                    // TODO: Not sure if this case is possible.
+                    unimplemented!();
+                }
```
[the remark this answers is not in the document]
tyler: Done!

lib/wasm/src/code_translator.rs
tyler: And done!

lib/codegen/meta-python/isa/x86/recipes.py
```diff
@@ -1473,6 +1486,36 @@ def floatccs(iform):
...
+        modrm_sib(out_reg0, sink);
+        sib(imm.trailing_zeros() as u8, in_reg0, in_reg1, sink);
+        ''')
+
+jt_base = TailRecipe(
+        'jt_base', BranchTableBase, size=5, ins=(), outs=(GPR),
+        clobbers_flags=False,
+        emit='''
+        PUT_OP(bits, rex2(0, out_reg0), sink);
+        modrm_riprel(out_reg0, sink);
```
sunfishcode: Can you add a comment here mentioning that we don't need a reloc because we put the jump table immediately after the function body so we can compute its offset ourselves?

src/compile.rs
```diff
@@ -97,12 +97,20 @@ fn handle_module(
...
+
         let mut relocs = PrintRelocs { flag_print };
         let mut traps = PrintTraps { flag_print };
-        context
-            .compile_and_emit(isa, &mut mem, &mut relocs, &mut traps)
-            .map_err(|err| pretty_error(&context.func, Some(isa), err))?;
+        let mut code_sink: binemit::MemoryCodeSink;
+        unsafe {
+            code_sink = binemit::MemoryCodeSink::new(mem.as_mut_ptr(), &mut relocs, &mut traps);
+        }
```
sunfishcode: Why is it needed to call `compile` and `emit_function_to_memory` directly rather than using `compile_and_emit`? It's unfortunate to have `unsafe` here in user code, rather than having it be an implementation detail of the library.

tyler: The reason I did this is because I needed access to the `MemoryCodeSink` for the `code_size` member, so I know how much to disassemble. In the `emit_to_memory` function of `context.rs` the `MemoryCodeSink` is created and thrown away. Since this was something only necessary for the CLI, I thought this code unlikely to be called by a user embedding and so not terribly important re: unsafe behavior. 

But if you think it should change, I can certainly make it work somehow.

PR #454, opened 2018-08-12
Split bforest into a separate crate

Amanieu:
Basically the same rationale as #297: this would allow `bforest` to be used outside Cranelift.

I had to make one significant API change: the `Comparator` type parameter was removed from `Map` and `MapForest` and instead the methods accept any type which implements `Comparator<K>`. This is necessary because `impl<PO: ProgramOrder> bforest::Comparator<Ebb> for PO { .. }` is not allowed outside of the crate which defines `Comparator`.


sunfishcode:
The changes here look good to me! I wonder if this makes us more dependent on LTO and whether it's worth adding some `#[inline]`s to help in the non-LTO case. I'm thinking of things like `new`/`is_empty`/`default` in `Map`/`Set`, for example, however I don't have much experience with this kind of thing in Rust. Do you know if this is something we should be concerned about?

bjorn3:
All methods on `Map`/`Set` are generic, because the Type is generic and as such do not need `#[inline]` to be codegen'ed in `cranelift-codegen`.

lib/bforest/README.md
```diff
@@ -0,0 +1,2 @@
+This crate contains array-based data structures used by the core Cranelift code
+generator which represent a set of small ordered sets or maps.
```
sunfishcode: It'd be good to add some of the content from the top-level comment in the current mod.rs file about how this code is specialized.

lib/bforest/Cargo.toml
```diff
@@ -0,0 +1,21 @@
+[package]
+authors = ["The Cranelift Project Developers"]
+name = "cranelift-bforest"
+version = "0.18.1"
+description = "A forest of B+-trees"
+license = "Apache-2.0"
```
sunfishcode: Cranelift code uses "Apache-2.0 WITH LLVM-exception".

sunfishcode: Please also add a copy of the LICENSE file in lib/bforest (so that people pulling it in as a dependency get a copy).

sunfishcode:
Thanks!

sunfishcode:
The cranelift-bforest package is now [published on crates.io](https://crates.io/crates/cranelift-bforest).

Amanieu:
Great!

Issue #434, opened 2018-07-31
Change command line parsing and validation to clap


carolinecullen:
Docopt is currently being used for Cranelifts command line argument parsing, comparatively clap is much more flexible and customizable. Clap also has the added benefit of being a more lightweight approach to parsing and validation compared to the current docopt implementation.

cc @sunfishcode

PR #463, opened 2018-08-13
 Changing from docopt to clap for the clif-util #434

carolinecullen:
Modifies clif-util to change command line parser from docopt to clap in order to make the util more lightweight and readable.


sunfishcode:
The CI failure is because we support Rust 1.25.0, which doesn't have `'_`. Would you mind changing `get_vec` to use a named lifetime?

src/clif-util.rs
```diff
@@ -1,5 +1,5 @@
-#![deny(trivial_numeric_casts, unused_extern_crates, unstable_features)]
-#![warn(unused_import_braces)]
+#![deny(trivial_numeric_casts)]
+#![warn(unused_import_braces, unstable_features, unused_extern_crates)]
```
sunfishcode: This may have been an unintentional merge; please move `unstable_features` and `unused_extern_crates` back to `deny`.

src/clif-util.rs
```diff
@@ -45,125 +41,179 @@ mod print_cfg;
...
-                .version(Some(format!("Cranelift {}", VERSION)))
-                .deserialize()
-        })
-        .unwrap_or_else(|e| e.exit());
-
-    if args.flag_debug {
-        pretty_env_logger::init();
-    } else {
-        file_per_thread_logger::initialize(LOG_FILENAME_PREFIX);
-    }
```
sunfishcode: Please restore these 5 lines (and the things they depend on), which initialize the logger implementation (this was added in #414).

src/clif-util.rs
```diff
@@ -45,125 +41,179 @@ mod print_cfg;
...
+}
+
+fn add_print_flag<'a>() -> clap::Arg<'a, 'a> {
+    Arg::with_name("print")
+        .short("p")
+        .help("Print the resulting Cranelift IR")
+}
+
+/// Takes vector of clap values and converts the values to strings and puts values in the mut vector
+fn get_vec(mut_vec: &mut Vec<String>, argument_vec: Option<clap::Values<'_>>) {
```
sunfishcode: It appears all uses of `get_vec` are creating a new `Vec` to pass in for `mut_vec`. Consequently, it would simplify the code to make this function return a `Vec` (that it creates itself) rather than taking one as `&mut`.

src/clif-util.rs
```diff
@@ -45,125 +41,179 @@ mod print_cfg;
...
+        }
     }
+}
+
+fn add_wasm_or_compile<'a>(cmd: &str) -> clap::App<'a, 'a> {
+    let about_str = match cmd {
+        "wasm" => "Compiles Cranelift IR into target language",
+        "compile" => "Compiles Cranelift IR into target language",
+        // Will not get here, function only called if cmd is wasm or compile
+        _ => "Invalid command",
```
sunfishcode: You can use `panic!("Invalid command")` to indicate that it's not expected to reach here, which would make the code clear without needing a comment.

src/clif-util.rs
```diff
@@ -45,125 +41,179 @@ mod print_cfg;
...
+                    target_val = val;
+                }
+            }
+
+            #[cfg(feature = "wasm")]
+            let result = wasm::run(
+                file_vec,
+                rest_cmd.is_present("verbose"),
+                rest_cmd.is_present("just-decode"),
+                rest_cmd.is_present("check-translation"),
```
sunfishcode: Above, the arguments are named `just_decode` and `check_translation`, with underscores, but the names here have dashes. Are these supposed to match those?

src/clif-util.rs
```diff
@@ -45,125 +41,179 @@ mod print_cfg;
...
+        }
+        ("compile", Some(rest_cmd)) => {
+            let mut file_vec: Vec<String> = Vec::new();
+            get_vec(&mut file_vec, rest_cmd.values_of("file"));
+
+            let mut set_vec: Vec<String> = Vec::new();
+            get_vec(&mut set_vec, rest_cmd.values_of("set"));
+
+            let mut target_val: &str = "";
+            if let Some(clap_target_vec) = rest_cmd.values_of("target") {
```
sunfishcode: Here and below, can this use `value_of` rather than `values_of`, so that you don't need the subsequent loop?

carolinecullen:
Hey @sunfishcode! This is ready for final review :)

sunfishcode:
Hi @carolinecullen! It looks like you may have missed some of the review comments above. Would you mind taking another look?

carolinecullen:
Sorry @sunfishcode , for some reason my brain decided to not see the vector comment. Should be good now!

carolinecullen:
Wasm cmd fixed and target variable name changed. :D  

@sunfishcode

PR #467, opened 2018-08-14
Prevent finalize being called more than once per func / data

bjorn3:


sunfishcode:
Yes indeed, that's a good catch.

PR #470, opened 2018-08-15
Improvements to error reporting

6A:
When I implemented the new verifier result, I accidentally switched two conditions in `pretty_verifier_error`, which caused errors to be incorrectly reported at the start of the body. Furthermore, valid instructions weren't printed. This fixes those issues.

PS: Also realized that disabling the "wasm" feature led to an error in "clif-util.rs", so I fixed it as well.

**Update**: I also made multiple checks non-fatal in the verifier, which means that now multiple instructions can have errors at the same time! Unless we make some strong assertions between `verify` functions, this shouldn't cause any issue in the future.

### DIFF
--- a/lib/codegen/src/print_errors.rs
+++ b/lib/codegen/src/print_errors.rs
@@ -28,11 +28,11 @@ pub fn pretty_verifier_error<'a>(
 
     while i != errors.len() {
         if let ir::entities::AnyEntity::Inst(_) = errors[i].location {
+            i += 1;
+        } else {
             let err = errors.remove(i);
 
-            writeln!(w, "Miscellaneous error: {}\n", err).unwrap()
-        } else {
-            i += 1;
+            writeln!(w, "misc verifier {}\n", err).unwrap()
         }
     }
 
--- a/lib/codegen/src/print_errors.rs
+++ b/lib/codegen/src/print_errors.rs
@@ -25,6 +25,7 @@ pub fn pretty_verifier_error<'a>(
 
     // TODO: Use drain_filter here when it gets stabilized
     let mut i = 0;
+    let mut wrote_error = false;
 
     while i != errors.len() {
         if let ir::entities::AnyEntity::Inst(_) = errors[i].location {
@@ -32,8 +33,13 @@ pub fn pretty_verifier_error<'a>(
         } else {
             let err = errors.remove(i);
 
-            writeln!(w, "misc verifier {}\n", err).unwrap()
+            writeln!(w, "verifier at {}", err).unwrap();
+            wrote_error = true;
         }
+    }
+
+    if wrote_error {
+        w.push('\n');
     }
 
     decorate_function(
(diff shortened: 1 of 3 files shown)


lib/codegen/src/print_errors.rs
```diff
@@ -28,11 +28,11 @@ pub fn pretty_verifier_error<'a>(
...
     while i != errors.len() {
         if let ir::entities::AnyEntity::Inst(_) = errors[i].location {
+            i += 1;
+        } else {
             let err = errors.remove(i);
 
-            writeln!(w, "Miscellaneous error: {}\n", err).unwrap()
-        } else {
-            i += 1;
+            writeln!(w, "misc verifier {}\n", err).unwrap()
```
sunfishcode: Can you comment on this wording change? "Miscellaneous error: {}" isn't the most descriptive, but "misc verifier {}" without a colon seems like it could be confusing.

6A: Before this change, "verifier error summary: {}" was shown at the beginning of the function body when an error was not related to an instruction, but now I think the wording should be changed, since more than one error can be shown this way. Furthermore, with "Miscellaneous error: {}", there are two colons in a row separated by a word, since error messages start with "{entity}:" when displayed.

So in the end I actually changed it in order to look more like what we have for instructions, where the format is "^~~~ verifier {}". 

If you have any other suggestion though, I'll gladly take it!

sunfishcode:
Could you post an example of what this looks like? Or I'll try it myself when I'm at my laptop. I'm just looking for a sense of how this looks.

6A:
I forgot to add a new line at the beginning so it looks weird, but right now here's what the following code...

```
test compile

target x86_64

function %vmglobal_err(i64, i64) -> i64 {
    gv1 = vmctx+0
    gv2 = vmctx+0

ebb0(v0: i64, v1: i64):
    v2 = global_value.i64 gv1
    v3 = global_value.i64 gv2
    return v1
}
```

produces...

```
FAIL .\test.clift: misc verifier gv1: undeclared vmctx reference gv1

misc verifier gv2: undeclared vmctx reference gv2

function %vmglobal_err(i64, i64) -> i64 fast {
    gv0 = globalsym %
    gv1 = vmctx
    gv2 = vmctx

ebb0(v0: i64, v1: i64):
    v2 = global_value.i64 gv1
    v3 = global_value.i64 gv2
    return v1
}
```

If I change two things it will look like

```
FAIL .\test.clift:
misc verifier gv1: undeclared vmctx reference gv1
misc verifier gv2: undeclared vmctx reference gv2

function %vmglobal_err(i64, i64) -> i64 fast {
    gv0 = globalsym %
    gv1 = vmctx
    gv2 = vmctx

ebb0(v0: i64, v1: i64):
    v2 = global_value.i64 gv1
    v3 = global_value.i64 gv2
    return v1
}
```

sunfishcode:
Ah, ok. How about "verifier at {}"?

6A:
Alright, I changed it to `verifier at {}`.

sunfishcode:
Thanks, this looks good. If we think of a better way to word this, we can always change it again :-). Long term, we should ideally have source locations more often; for example, in the examples above, we should ideally be able to point at the actual `gv1` reference. But that's a separate issue. The PR here looks good, thanks!

6A:
I don't mind taking care of the ability to decorate the preamble, but how should I do it? I have a few different ideas.

- Change the `FuncWriter` trait so that it has different methods for `StackSlots`, `GlobalValues`, `Heaps`, etc.
  * Pros: Future friendly and easier to use for other users.
  * Needs the addition of many methods (to which we could add default implementations).
- Copy paste the default `write_preamble` implementation into `PrettyVerifierError` and modify it at will.
  * Pros: Simpler.
  * Cons: Not very future-friendly.

PR #473, opened 2018-08-16
Document tables, and special parameters

sunfishcode:
This adds documentation to ir.rst for the newly added table feature, and for special parameters.

### DIFF
--- a/docs/ir.rst
+++ b/docs/ir.rst
@@ -746,6 +746,40 @@ bounds checking is required for each access:
     :lines: 2-
 
 
+Tables
+------
+
+Code compiled from WebAssembly often needs access to objects outside of its
+linear memory. WebAssembly uses *tables* to allow programs to refer to opaque
+values through integer indices.
+
+A table is declared in the function preamble and can be accessed with the
+:inst:`table_addr` instruction that :term:`traps` on out-of-bounds accesses.
+Table addresses can be smaller than the native pointer size, for example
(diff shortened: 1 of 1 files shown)


docs/ir.rst
```diff
@@ -400,7 +400,7 @@ convention:
     retlist      : paramlist
     param        : type [paramext] [paramspecial]
     paramext     : "uext" | "sext"
-    paramspecial : "sret" | "link" | "fp" | "csr" | "vmctx"
+    paramspecial : "sret" | "link" | "fp" | "csr" | "vmctx" | "sigid" | "stack_limit"
```
bnjbvr: Is it expected that the new `sigid` special parameter is added to the list, but never used in any definition? I'd expect to see it shown in the definition of a call instruction which can call into a table_addr.

sunfishcode: I'm unclear on what you mean here. Are you looking for an example of how to use `sigid`?

bnjbvr: Yes indeed, thanks for checking; other special parameters I've checked in the docs have a usage example, this one does not.

sunfishcode: Ok, I opened https://github.com/CraneStation/cranelift/issues/497 to track the lack of an example for how to use `sigid`.

sunfishcode:
Merging this, as it adds useful documentation, though of course there's more to do :-).

PR #475, opened 2018-08-17
[Module] Remove DataDescription's writable field.

### LINKED ISSUE #456: Cranelift-module requires you to specify mutability of data twice
Both in `declare_data` and `DataContext::define{,_zeroinit}`. This makes it possible to say different things at both places.

sunfishcode:
It was redundant, as data object declarations also have a writable
field, so just use that, avoiding the need for users to declare the
same thing twice.

Fixes #456.


bnjbvr:
It seems there are a few compiler errors in this PR: https://travis-ci.org/CraneStation/cranelift/jobs/417073426#L1016-L1040

Issue #451, opened 2018-08-10
Using table_addr.i32 leads to a verifier error


pepyakin:
```
test compile
target x86_64 haswell

function u0:0(i64 vmctx) system_v {
    gv0 = vmctx+12
    gv1 = vmctx+14
    table0 = dynamic gv0, min 20, bound gv1, element_size 4
    sig0 = (i64 vmctx)

ebb0(v0: i64):
    v1 = iconst.i32 0
    v2 = table_addr.i32 table0, v1, +0
    call_indirect sig0, v2(v0)
    return
}
```

yields this error

```
function u0:0(i64 vmctx [%rdi]) system_v {
    gv0 = vmctx+12
    gv1 = vmctx+14
    sig0 = (i64 vmctx [%rdi]) fast

                                ebb0(v0: i64):
                                    v1 = iconst.i32 0
[RexOp1r_ib#83]                     v3 = iadd_imm v0, 14
                                    ^~~~~~ verifier inst4: arg 0 (v0) has type i64, expected i32

                                    v4 = icmp.i32 uge v1, v3
                                    brz.b1 v4, ebb1
                                    trap table_oob

                                ebb1:
                                    v5 = iadd_imm.i32 v0, 12
                                    v6 = ishl_imm.i32 v1, 2
                                    v2 = iadd.i32 v5, v6
                                    call_indirect.i32 sig0, v2(v0)
                                    return
}
```

With i64 though it works fine

pepyakin:
Known limitation: table_addr works only with `pointer_type`, closing.

sunfishcode:
FWIW, this limitation is now lifted, in #480.

PR #480, opened 2018-08-23
Fix legalization of heap_addrs with 32-bit indices.

sunfishcode:
This makes several changes:

 - It adds an index_type to heap declarations, allowing heaps to specify the
   type for indexing. This also anticipates 64-bit heap support.

 - It adds a memory_type to deref global values, allowing deref globals to
   have types other than pointers. This is used to allow the bound variable
   in dynamic heaps to have type i32, to match the index type in heaps
   with i32 index type.

 - And, it fixes heap legalization to do the bounds check in the heap's
   index type.

This is a minor yak-shave prompted in part by https://github.com/CraneStation/cranelift/issues/451, and will prepare the path for `table_addr` to take i32 indices instead of just pointer-sized indices.

### DIFF
--- a/lib/codegen/src/verifier/mod.rs
+++ b/lib/codegen/src/verifier/mod.rs
@@ -344,18 +344,88 @@ impl<'a> Verifier<'a> {
                 cur = base;
             }
 
-            if let ir::GlobalValueData::VMContext { .. } = self.func.global_values[cur] {
-                if self
-                    .func
-                    .special_param(ir::ArgumentPurpose::VMContext)
-                    .is_none()
-                {
-                    report!(errors, cur, "undeclared vmctx reference {}", cur);
+            match self.func.global_values[gv] {
+                ir::GlobalValueData::VMContext { .. } => {
+                    if self
+                        .func
+                        .special_param(ir::ArgumentPurpose::VMContext)
+                        .is_none()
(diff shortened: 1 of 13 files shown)


lib/codegen/src/verifier/mod.rs
```diff
@@ -1256,51 +1326,82 @@ impl<'a> Verifier<'a> {
...
-                            "input {} must be larger than output {}",
-                            arg_type,
-                            ctrl_type
-                        );
+                        errors,
+                        inst,
+                        "global_value instruction with type {} references global value with type {}",
+                        inst_type,
+                        global_type
+                    );
```
pepyakin: Identation is a bit off

lib/codegen/src/verifier/mod.rs
```diff
@@ -344,21 +344,91 @@ impl<'a> Verifier<'a> {
...
+                            report!(
+                                errors,
+                                heap,
+                                "heap index type {} differs from the type of its bound, {}",
+                                index_type,
+                                bound_type
+                            );
+                        }
+                    }
+                    _ => {}
```
pepyakin: I wonder if the default arm is ok for this case. Worrying about that it might be missed if a new heap style were added.

PR #484, opened 2018-08-25
Verifier now ensures there is no division by 0imm.

6A:


sunfishcode:
The patch looks good. [The instruction descriptions](https://github.com/CraneStation/cranelift/blob/master/lib/codegen/meta-python/base/instructions.py#L1058) should be kept in sync, and we should keep urem_imm/srem_imm consistent with udiv_imm/sdiv_imm. This change also needs to update [preopt](https://github.com/CraneStation/cranelift/blob/master/lib/codegen/src/preopt.rs#L460) to avoid optimzing things like iconst+sdiv into sdiv_imm in the case of zero.

That actually raises an interesting question though. Prohibiting zero for these instructions is a classic case of making analysis easier at the expense of making transforming harder. Is it the right tradeoff here? I'm not not sure yet; here are some things I'm thinking about: One of the advantages of prohibiting zero here is that analysis code that sees a udiv_imm or sdiv_imm doesn't have to worry about the trapping division-by-zero case. However, sdiv_imm can still trap (INT_MIN//-1), so analysis would still need to check the denominator to see if a trap is possible, at least in the signed case.

6A:
Since the verification step seems to be optional, we can still have cases where a division by an immediate 0 happens, can't we? Then in any case we can't really depend on that change in later stages. To me, this is more of a frontend lint than a fully-fledged analysis feature.

sunfishcode:
The verifier currently does a lot of checks malformed code, like referencing values that aren't defined, ebbs that aren't included in the layout, and so on. If you look in [`relax_branches` in context.rs](https://github.com/CraneStation/cranelift/blob/master/lib/codegen/src/context.rs#L327), which is one of the last things to happen, we're prepare to run the verifier after it.

Your idea of a frontend lint feature is interesting though. I wonder if it makes sense to split "lint"-like diagnostics into a separate framework that could be invoked separately, or if it would make sense to have a concept of diagnostic levels in the verifier, so that frontend writers could request all diagnostics, and the backend could request just the required parts.

6A:
> we're prepare to run the verifier after it

It's still optional though, so many errors might go through.

> split "lint"-like diagnostics into a separate framework that could be invoked separately

I don't see anything keeping us from having "external" hints, the likes of clippy for Rust. On one hand it wouldn't make the verification process as seamless, and optimizations couldn't really be added based on it, but it would also make the codegen part lighter and likely easier to work with. Having two different error-reporting systems would be a bit annoying though, IMO.

> concept of diagnostic levels in the verifier

This second approach is really nice (and easy to implement after the recent changes to `verifier/mod.rs`), and if we don't care too much about performances, it wouldn't be too hard to implement.

sunfishcode:
> It's still optional though, so many errors might go through.

We allow disabling the verifier so users who are confident that it won't find any bugs can choose to skip it and speed up compilation. Everything is still expected to uphold all the invariants though.

> >    split "lint"-like diagnostics into a separate framework that could be invoked separately

> I don't see anything keeping us from having "external" hints, the likes of clippy for Rust. On one hand it wouldn't make the verification process as seamless, and optimizations couldn't really be added based on it, but it would also make the codegen part lighter and likely easier to work with. Having two different error-reporting systems would be a bit annoying though, IMO.

I don't have a strong opinion either way yet. Maybe we could make the lint-reporting system also invoke the verifier, so frontend users would just talk to a single interface?

And actually, if these lints are going to be at all opinionated, what would you think about keeping them separate? The urem-by-zero check could plausibly be considered opinionated, if we're thinking of it as a frontend lint rather than an IR invariant.

> >    concept of diagnostic levels in the verifier

> This second approach is really nice (and easy to implement after the recent changes to verifier/mod.rs), and if we don't care too much about performances, it wouldn't be too hard to implement.

I don't think verifier performance is super critical (users who *really* need fast compile time can disable it), but we don't want it to be too slow either. I imagine adding a level wouldn't add that much overhead. So my bigger concerns are just the ones above about what's the right place to check things.

sunfishcode:
To summarize here, I think I do prefer to say that because [this doesn't eliminate traps for sdiv_imm](https://github.com/CraneStation/cranelift/pull/484#issuecomment-415976269), it's not that important to enforce the lack of division by zero in the verifier. However, it would be useful to have either as in optional separate diagnostic level in the verifier, or in a separate linting framework.

So I'm closing this now, however feel free to repoen if you want to reuse this to add it under a diagnostic level, or open a new issue or PR as needed.

PR #485, opened 2018-08-25
Fix error not reported if at least one other error expected.

6A:
In filetests, if multiple errors were reported in a function but only one of them was expected, no fail was reported. This fixes it.

Additionally, now that errors are correctly checked, some tests had to be fixed because more than one error were reported.

### DIFF
--- a/lib/codegen/src/verifier/mod.rs
+++ b/lib/codegen/src/verifier/mod.rs
@@ -328,7 +328,7 @@ impl<'a> Verifier<'a> {
     //  - cycles in the global value declarations.
     //  - use of 'vmctx' when no special parameter declares it.
     fn verify_global_values(&self, errors: &mut VerifierErrors) -> VerifierStepResult<()> {
-        let mut report_cycle = true;
+        let mut cycle_seen = false;
         let mut seen = SparseSet::new();
 
         'gvs: for gv in self.func.global_values.keys() {
@@ -338,9 +338,9 @@ impl<'a> Verifier<'a> {
             let mut cur = gv;
             while let ir::GlobalValueData::Deref { base, .. } = self.func.global_values[cur] {
                 if seen.insert(base).is_some() {
-                    if report_cycle {
+                    if !cycle_seen {
                         report!(errors, gv, "deref cycle: {}", DisplayList(seen.as_slice()));
-                        report_cycle = false; // ensures we don't report the cycle multiple times
+                        cycle_seen = true; // ensures we don't report the cycle multiple times
                     }
                     continue 'gvs;
                 }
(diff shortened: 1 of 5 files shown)


sunfishcode:
Just stopping by to say this looks like a good catch! I'll review it more closely soon.

lib/codegen/src/verifier/mod.rs
```diff
@@ -1511,12 +1528,18 @@ impl<'a> Verifier<'a> {
     pub fn run(&self, errors: &mut VerifierErrors) -> VerifierStepResult<()> {
         self.verify_global_values(errors)?;
         self.typecheck_entry_block_params(errors)?;
+
         for ebb in self.func.layout.ebbs() {
+            // If an encoding error is found in this block, we want
+            // to perform fewer checks; otherwise many errors may be reported
+            // for a single problem.
```
sunfishcode: Can you explain this a little more? It's not clear why remembering this state between instructions is important, since encoding verification only looks at one instruction at a time.

lib/codegen/src/verifier/mod.rs
```diff
@@ -328,6 +328,7 @@ impl<'a> Verifier<'a> {
     //  - cycles in the global value declarations.
     //  - use of 'vmctx' when no special parameter declares it.
     fn verify_global_values(&self, errors: &mut VerifierErrors) -> VerifierStepResult<()> {
+        let mut report_cycle = true;
```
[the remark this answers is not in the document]
6A: Yup, good for me!

lib/codegen/src/verifier/mod.rs
6A: For some existing test `.clif` files, any instruction that wouldn't be encoded would cause subsequent instructions to not be either. Therefore, as soon as an instruction wasn't encoded, an error would be reported and you would be sure that for all subsequent statements, an error would be reported as well, even though the root cause for this was already reported.  If that boolean is `false`, we don't perform checks that would've been influenced by what I just explained, and thus we only report the root cause.

6A: I'll try to find a better name for the variable and to explain this better though. If you have any suggestions, I'll gladly take them!

sunfishcode: I'm curious about this, because verifier errors shouldn't cause cascading errors like this. I just tried an experiment setting `verify_enc_need` to false between instructions, and filetests/isa/riscv/verify-encoding.clif failed. It looks like the problem there is that the `return` statement doesn't have an encoding. If I add `[Iret#19]` to the front of the return, the test passes. So I think the thing to do here is to fix the test, so that we don't need to keep this state between instructions.

(Surprisingly, the verifier only requires that [some instructions](https://github.com/CraneStation/cranelift/blob/master/lib/codegen/src/verifier/mod.rs#L1457) be encoded, though that's another matter.)

6A: The fail in `filetests/isa/riscv/verify-encoding.clif` is why I added this change; I must've misunderstood the reason why it failed.

sunfishcode:
Looks good!

PR #486, opened 2018-08-27
Added bitrev instruction for 32 and 64 bit integers

Aaronepower:


lib/codegen/meta-python/base/legalize.py
```diff
@@ -382,6 +383,29 @@
...
+            d << band_imm(c2, imm64(0x10101)),
+            a << ushr_imm(d, imm64(16)),
+        ))
+
+expand.legalize(
+        a << bitrev64(x),
+        Rtl(
+            a1 << imul_imm(x, imm64(0x0202020202)),
+            a2 << band_imm(a1, imm64(0x010884422010)),
+            a << urem_imm(a2, imm64(1023)),
```
sunfishcode: It looks like the StackOverflow article presents these in a confusing way. I think the algorithms in the patch here are meant to be 8-bit `reverseBits` implementations, for 32-bit and 64-bit processors, which I think is meant to be in the code that follows to implement a reverse for a wider integer.

I recommend looking at the `reverse` version under the heading "Low Memory". That seems to be a pretty good approach, and we can generalize it to 64-bit as well as 16-bit and 8-bit.

And if we want to further specialize for x86, we can specialize that version to use the `bswap` instruction to replace the parts that are swapping bytes. But we can figure that part out later.

lib/codegen/meta-python/base/instructions.py
```diff
@@ -1409,6 +1409,22 @@
...
+        """,
+        ins=x, outs=a)
+
+bitrev64 = Instruction(
+        'bitrev64', r"""
+        Reverse bits of a 64 bit integer.
+
+        Reverses the bits in ``x``.
+        """,
+        ins=x, outs=a)
```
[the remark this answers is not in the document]
Aaronepower: @sunfishcode What should I change it to?

Aaronepower:
Alright, I've updated it with proper implementations for 8, 16, 32, 64 bit implementations. They are adapted from this [playground](https://play.rust-lang.org/?gist=7d2901efb76d0ce00ab8cfb03795e721&version=stable&mode=debug&edition=2015) I used to check that they were working.  Now however for bitrev8 it won't seem to generate any code for the filetests.

lib/codegen/meta-python/base/instructions.py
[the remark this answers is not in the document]
sunfishcode: Just have a single `bitrev` instruction here, rather than one per bitwidth. It'll automatically be available as `bitrev.i8`, `bitrev.i16`, etc..

lib/codegen/meta-python/base/legalize.py
```diff
@@ -382,6 +401,115 @@
             a << bxor(x, y)
         ))
 
+# Expand bitrev
+# Adapted from Stack Overflow.
+# https://stackoverflow.com/questions/746171/most-efficient-algorithm-for-bit-reversal-from-msb-lsb-to-lsb-msb-in-c
+expand.legalize(
+        a << bitrev8(x),
+        Rtl(
+            a1 << band_imm(x, 0xaa),
```
sunfishcode: The DSL needs these immediates to be wrapped in `imm64`, like `imm64(0xaa)`.

Aaronepower: Whoops, I knew that and forgot. That must be what's wrong. 😅

Aaronepower:
Hmm apparently that's not what's wrong in the filetests. This is the output. Showing no expansion.

```
FAIL filetests/legalizer/bitrev.clif: legalizer(%reverse_bits_8):
filecheck failed:
#0 check: v16 = iconst.i32 0xaa
#1 check: v2 = band v0, v16
#2 check: v3 = ushr_imm v2, 1
#3 check: v4 = band_imm v0, 0x55
#4 check: v5 = ishl_imm v4, 1
#5 check: v6 = bor v3, v5
#6 check: v17 = iconst.i32 0xcc
#7 check: v7 = band v6, v17
#8 check: v8 = ushr_imm v7, 2
#9 check: v9 = band_imm v6, 0x33
#10 check: v10 = ushr_imm v9, 2
#11 check: v11 = bor v8, v10
#12 check: v18 = iconst.i32 0xf0
#13 check: v12 = band v11, v18
#14 check: v13 = ushr_imm v12, 4
#15 check: v14 = band_imm v11, 0x0f
#16 check: v15 = ishl_imm v14, 4
#17 check: v1 = bor v13, v15
> function %reverse_bits_8(i8 [%rdi]) -> i8 [%rax] fast {
Missed #0: \bv16 = iconst\.i32 0xaa\b
>                                 ebb0(v0: i8):
> [-]                                 v1 = bitrev v0
> [Op1ret#c3]                         return v1
> }

slow: 0.445 filetests/parser/tiny.clif
slow: 0.393 filetests/preopt/rem_by_const_power_of_2.clif
slow: 0.459 filetests/preopt/div_by_const_non_power_of_2.clif
slow: 0.536 filetests/preopt/rem_by_const_non_power_of_2.clif
slow: 0.495 filetests/isa/x86/prologue-epilogue.clif
139 tests
```

Aaronepower:
@sunfishcode So I just have to change the 8bit & 16bit versions to call widen instead of expand, and then then add a shit instruction to the end of their filetest, or is there more to it?

sunfishcode:
Oh, actually, I was thinking of widening 8-bit bitreverse into a 32-bit bitreverse with a shift at the end, which would work, but be slow. That's why I was thinking of the shift at the end. But I think you're right, we can do the 8-bit bitreverse algorithm in your patch, and just adapt it to work as a widen legalization. No shift needed!

Aaronepower:
@sunfishcode So I've added the literal, but now im getting this error for imm instructions. How do I change it so they can accept uimm64? I'm looking at `parse_inst_operands` and I'm wondering how to change from `match_imm64` to `match_uimm64`. Since really it needs to accept either.
```
--- stderr
Python 2.7.15
Traceback (most recent call last):
  File "/Users/aep/src/rust/cranelift/lib/codegen/meta-python/build.py", line 7, in <module>
    import isa
  File "/Users/aep/src/rust/cranelift/lib/codegen/meta-python/isa/__init__.py", line 10, in <module>
    from . import riscv, x86, arm32, arm64
  File "/Users/aep/src/rust/cranelift/lib/codegen/meta-python/isa/riscv/__init__.py", line 29, in <module>
    from . import encodings, settings, registers  # noqa
  File "/Users/aep/src/rust/cranelift/lib/codegen/meta-python/isa/riscv/encodings.py", line 14, in <module>
    from base.legalize import narrow, expand
  File "/Users/aep/src/rust/cranelift/lib/codegen/meta-python/base/legalize.py", line 483, in <module>
    a1 << band_imm(x, uimm64(0xaaaaaaaaaaaaaaaa)),
  File "/Users/aep/src/rust/cranelift/lib/codegen/meta-python/cdsl/instructions.py", line 357, in __call__
    return Apply(self, args)
  File "/Users/aep/src/rust/cranelift/lib/codegen/meta-python/cdsl/ast.py", line 322, in __init__
    .format(arg, op.kind)
AssertionError: Passing literal 12297829382473034410 to field of wrong kind imm64.
```

sunfishcode:
Ah, I see what's happening. Ok, my advice above to use a uimm64 isn't going to work.

I think what we can do instead is make the `ConstantInt` class be smarter about values in the unsigned 64-bit integer space, like this:
```diff
diff --git a/lib/codegen/meta-python/cdsl/ast.py b/lib/codegen/meta-python/cdsl/ast.py
index 09d4ad25..017ad310 100644
--- a/lib/codegen/meta-python/cdsl/ast.py
+++ b/lib/codegen/meta-python/cdsl/ast.py
@@ -526,7 +526,14 @@ class ConstantInt(Literal):
         """
         Get the Rust expression form of this constant.
         """
-        return str(self.value)
+        # If the value is in the signed imm64 range, print it as-is.
+        if self.value >= -(2**63) and self.value < (2**63):
+            return str(self.value)
+        # Otherwise if the value is in the unsigned imm64 range, print its
+        # bitwise counterpart in the signed imm64 range.
+        if self.value >= (2**63) and self.value < (2**64):
+            return str(self.value - (2**64))
+        assert False, "immediate value not in signed or unsigned imm64 range"
 
 
 class ConstantBits(Literal):
```
That should allow plain `imm64` to be good enough here.

Aaronepower:
@sunfishcode Alright I've updated the PR with those changes.

lib/codegen/meta-python/base/legalize.py
```diff
@@ -382,6 +401,115 @@
...
+            c3 << band_imm(c, imm64(0x0f)),
+            c4 << ishl_imm(c3, 4),
+            a << bor(c2, c4),
+        ))
+
+widen.legalize(
+        a << bitrev.i16(x),
+        Rtl(
+            a1 << band_imm(x, imm64(0xaaaa)),
+            a2 << ushr_imm(a1, 1),
```
sunfishcode: Ah, mypy is still failing because the shift immediates `1` etc. also need to be wrapped in imm64.

Aaronepower: @sunfishcode Doh, I've updated it now with all of them wrapped.

PR #487, opened 2018-08-28
Adds pass command to clif-util.

carolinecullen:
Can now specify passes to be run on files from the command line.
This pass command tosses tests and targets specified at the top of the file.

### DIFF
--- a/lib/filetests/src/lib.rs
+++ b/lib/filetests/src/lib.rs
@@ -76,7 +76,30 @@ pub fn run(verbose: bool, files: &[String]) -> TestResult {
     }
 
     runner.start_threads();
-    runner.run()
+    runner.run(None, None)
+}
+
+/// Used for 'pass' subcommand.
+/// Commands are interpreted as test and executed.
+///
+/// Directories are scanned recursively for test cases ending in `.clif`.
+///
+pub fn run_pass(
+    verbose: bool,
+    passes: &Vec<String>,
+    target: Option<&str>,
+    file: &String,
+) -> TestResult {
+    let mut runner = TestRunner::new(verbose);
+
+    let path = Path::new(file);
+    if path.is_file() {
+        runner.push_test(path);
+    } else {
+        runner.push_dir(path);
+    }
+
+    runner.run(Some(passes), target)
 }
 
 /// Create a new subcommand trait object to match `parsed.command`.
--- a/lib/filetests/src/lib.rs
+++ b/lib/filetests/src/lib.rs
@@ -76,7 +76,7 @@ pub fn run(verbose: bool, files: &[String]) -> TestResult {
     }
 
     runner.start_threads();
-    runner.run(None, None)
+    runner.run()
 }
 
 /// Used for 'pass' subcommand.
@@ -84,12 +84,7 @@ pub fn run(verbose: bool, files: &[String]) -> TestResult {
 ///
 /// Directories are scanned recursively for test cases ending in `.clif`.
 ///
-pub fn run_pass(
-    verbose: bool,
-    passes: &Vec<String>,
-    target: Option<&str>,
-    file: &String,
-) -> TestResult {
+pub fn run_passes(verbose: bool, passes: &[String], target: &str, file: &String) -> TestResult {
     let mut runner = TestRunner::new(verbose);
 
     let path = Path::new(file);
@@ -99,7 +94,7 @@ pub fn run_pass(
         runner.push_dir(path);
     }
 
-    runner.run(Some(passes), target)
+    runner.run_passes(passes, target)
 }
 
 /// Create a new subcommand trait object to match `parsed.command`.
--- a/lib/filetests/src/runner.rs
+++ b/lib/filetests/src/runner.rs
@@ -30,6 +30,12 @@ enum State {
     Done(TestResult),
 }
 
+#[derive(PartialEq, Eq, Debug)]
+pub enum IsPass {
+    Pass,
+    NotPass,
+}
+
 impl QueueEntry {
     pub fn path(&self) -> &Path {
         self.path.as_path()
@@ -118,7 +124,7 @@ impl TestRunner {
 
     /// Scan any directories pushed so far.
     /// Push any potential test cases found.
-    pub fn scan_dirs(&mut self, is_pass: bool) {
+    pub fn scan_dirs(&mut self, pass_status: IsPass) {
         // This recursive search tries to minimize statting in a directory hierarchy containing
         // mostly test cases.
         //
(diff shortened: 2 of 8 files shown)


src/clif-util.rs
```diff
@@ -176,6 +200,21 @@ fn main() {
                 &get_vec(rest_cmd.values_of("file")),
             ).map(|_time| ())
         }
+        ("pass", Some(rest_cmd)) => {
+            handle_debug_flag(rest_cmd.is_present("debug"));
+
+            let mut file_val: &str = "";
+            if let Some(clap_file) = rest_cmd.value_of("single-file") {
```
bnjbvr: Since single-file is required, can we just safely `unwrap` here?

lib/filetests/src/lib.rs
```diff
@@ -76,7 +76,30 @@ pub fn run(verbose: bool, files: &[String]) -> TestResult {
...
+///
+/// Directories are scanned recursively for test cases ending in `.clif`.
+///
+pub fn run_pass(
+    verbose: bool,
+    passes: &Vec<String>,
+    target: Option<&str>,
+    file: &String,
+) -> TestResult {
+    let mut runner = TestRunner::new(verbose);
```
bnjbvr: Is `verbose` the right name here? The caller names it "time-passes", which had me think that it was for timing purposes. So maybe either the caller's argument or the parameter here should be fixed?

lib/filetests/src/lib.rs
```diff
@@ -76,7 +76,30 @@ pub fn run(verbose: bool, files: &[String]) -> TestResult {
...
-    runner.run()
+    runner.run(None, None)
+}
+
+/// Used for 'pass' subcommand.
+/// Commands are interpreted as test and executed.
+///
+/// Directories are scanned recursively for test cases ending in `.clif`.
+///
+pub fn run_pass(
```
bnjbvr: I wonder if `run_passes` would be a better name, since it can actually run a vector of them.

lib/filetests/src/runner.rs
```diff
@@ -329,12 +341,24 @@ impl TestRunner {
...
+    pub fn run(&mut self, passes: Option<&Vec<String>>, target: Option<&str>) -> TestResult {
         let started = time::Instant::now();
-        self.scan_dirs();
-        self.schedule_jobs();
-        self.drain_threads();
-        self.report_slow_tests();
+
+        // If passes are present from command line, then no threads.
+        match passes {
+            Some(pass_vec) => {
```
bnjbvr: Since the two cases are different enough, and there's not much before/after, plus it makes all the other `run()` call sites look a bit weird (because of the `None` argument, which doesn't provide much information about what type it is, what should be provided, etc.), maybe it could be split into a different function `run_passes(&mut self, passes: &Vec<String>, target: &str)`?

lib/filetests/src/runner.rs
```diff
@@ -118,7 +118,7 @@ impl TestRunner {
 
     /// Scan any directories pushed so far.
     /// Push any potential test cases found.
-    pub fn scan_dirs(&mut self) {
+    pub fn scan_dirs(&mut self, is_pass: bool) {
```
bnjbvr: This boolean is not very talkative on call sites; should we use an enum here?

lib/filetests/src/lib.rs
```diff
@@ -76,7 +76,30 @@ pub fn run(verbose: bool, files: &[String]) -> TestResult {
...
+    verbose: bool,
+    passes: &Vec<String>,
```
bjorn3: Please use `&[String]` here instead.

lib/filetests/src/lib.rs
```diff
@@ -76,7 +76,30 @@ pub fn run(verbose: bool, files: &[String]) -> TestResult {
...
+    target: Option<&str>,
+    file: &String,
```
bjorn3: Please use `&str` here instead.

lib/filetests/src/runner.rs
```diff
@@ -215,6 +219,14 @@ impl TestRunner {
         }
     }
 
+    /// Schedule any new job to run for the pass command.
+    fn schedule_pass_job(&mut self, passes: &Vec<String>, target: Option<&str>) {
```
bjorn3: Please use `&[String]` instead of `&Vec<String>`.

lib/filetests/src/runner.rs
```diff
@@ -329,12 +341,24 @@ impl TestRunner {
     }
 
     /// Scan pushed directories for tests and run them.
-    pub fn run(&mut self) -> TestResult {
+    pub fn run(&mut self, passes: Option<&Vec<String>>, target: Option<&str>) -> TestResult {
```
bjorn3: Please use `Option<&[String]>` instead.

lib/filetests/src/runone.rs
```diff
@@ -27,12 +27,12 @@ fn read_to_string<P: AsRef<Path>>(path: P) -> io::Result<String> {
 /// Load `path` and run the test in it.
 ///
 /// If running this test causes a panic, it will propagate as normal.
-pub fn run(path: &Path) -> TestResult {
+pub fn run(path: &Path, passes: Option<&Vec<String>>, target: Option<&str>) -> TestResult {
```
bjorn3: Same here

lib/reader/src/parser.rs
```diff
@@ -32,20 +32,39 @@ use testfile::{Comment, Details, TestFile};
...
+        .map(|file| file.functions.into_iter().map(|(func, _)| func).collect())
 }
 
 /// Parse the entire `text` as a test case file.
 ///
 /// The returned `TestFile` contains direct references to substrings of `text`.
-pub fn parse_test(text: &str) -> ParseResult<TestFile> {
+pub fn parse_test<'a>(
+    text: &'a str,
+    passes: Option<&'a Vec<String>>,
```
bjorn3: Same here

lib/reader/src/parser.rs
```diff
@@ -705,6 +724,15 @@ impl<'a> Parser<'a> {
         }
     }
 
+    /// Parse a list of test command passes specified in command line.
+    pub fn parse_cmdline_passes(&mut self, passes: &'a Vec<String>) -> Vec<TestCommand<'a>> {
+        let mut list = Vec::new();
```
bjorn3: Same here

carolinecullen:
Fixes made based on comments and suggestions.

@bnjbvr @bjorn3

sunfishcode:
Cool, this looks good.

Issue #366, opened 2018-06-16
Add a `readonly` flag for loads


sunfishcode:
See the discussion in https://github.com/Cretonne/cretonne/issues/278#issuecomment-375734141. That comment also imagined a "MutableGlobal" flag, but that's more complex and not presently needed, so let's defer it. 

The steps are roughly:
 - Add a readonly flag to MemFlags, similar to notrap, and document it in ir.rst. The behavior is undefined if the dereferenced memory is mutated at any time between when the function is called and when it is exited.
 - Add a verifier check to reject stores marked readonly. This doesn't quite fall under `instruction_integrity` or others, so lets add a new function in the loop in lib/codegen/src/verifier/mod.rs's `run`, called `immediate_constraints` or so and check it there.
 - Make the legalization for `deref` global values add the `readonly` flag, in `deref_addr`, and mention this in the comments/documentation for `GlobalValueData::Deref`/`deref`.
 - Remove `can_load` from simple_gvn's `trivially_unsafe_for_gvn`, and make gvn check that if the instruction `can_load`, it's valid to consider only if it's one of the load instructions with `readonly`. In theory, gvn should Just Work, because it's hashing the `InstructionData`, so two loads with the same address and flags will automatically look the same.
 - Write a `legalizer` test that tests multiple accesses to a heap with a deref base, and the redundant base computation should be eliminated.

lachlansneff:
I'll do this.

sunfishcode:
One other issue that comes up here: Instructions like `load_complex` have a `ValueList` in their `InstructionData`, so that they can hold more operands than may fit in an `InstructionData`. `Valuelist` is `entity::EntityList<Value>`, which [doesn't really](https://github.com/cretonne/cretonne/blob/master/lib/entity/src/list.rs#L81) implement `Hash`. This is because `EntityList` doesn't actually hold a pointer to the operands, instead it just has an index into a pool, which must be provided separately whenever accessing elements. I don't yet know how best to fix this.

sunfishcode:
#388 is now merged, so the issue I brought up here should no longer be a concern!

lachlansneff:
Just noting here that I am planning to actually do this within the next few days. Had completely forgotten about it until @sunfishcode politely reminded me.

PR #490, opened 2018-08-29
Reorganize the global value kinds.

sunfishcode:
This:
 - renames "deref" global values to "load"
 - adds an explicit "iadd_imm" global value kind, which replaces the
   builtin iadd in "vmctx" and "deref" global values.
 - also renames "globalsym" to "symbol"

I kept finding myself confused about deref's offset, since it applied to
the loaded value, rather than the address of the load.

So here's a patch which makes the global value definitions look more
like instructions, with "load.i32 gv0" replacing "deref(gv0): i32" and
so on, and offsets which require separate iadd_imm instructions made
into separate iadd_imm global values.

This also leads to syntax "iadd_imm.i64 gv0, 48", which is more verbose
than before, but consistent with the corresponding instruction syntax.

I'm still trying to decide if I like the approach and names in this patch. It feels like global values are an evolving concept, and this takes them a step further along a certain path, but I'm not yet clear on where the path leads.

This patch kind of makes global values look like macros, expanding out into sequences of instructions for computing heap bases and whatnot every time we need those values. There have been [some discussions](https://github.com/CraneStation/cranelift/issues/366) about how simple_gvn could eliminate redundant expansions, and that probably would work, though that could potentially be slower than a more clever scheme that avoids emitting redundant code in the first place.


sunfishcode:
I'm going head and merging this, since deref's adding the offset *after* the load is very confusing, and I want to fix this and update various docs.

Issue #488, opened 2018-08-28
Avoid printing redundant copies of value aliases


sunfishcode:
Value aliases are a way to have more than one value index refer to the same conceptual value. For example, when we write:
```
        v2 -> v3
```
we mean that v2 is an alias which refers to the same value as v3. This isn't an instruction, so it doesn't have a specific place to exist in the text format, so currently cranelift prints out aliases each time they're used. The parser accepts redundant alias declarations as long as they're identical, however it gets cluttered and can be confusing:
```
       v2 -> v3
       v2 -> v3
       v4 = iadd v2, v2
```
It would be tidier if we could print aliases immediately after the definition of the value they alias. For example, like this:
```
       v3 = iconst.i32 0
       v2 -> v3
```
That way, they'd only appear once, and in a pretty obvious location.

This isn't completely trivial though, because cranelift doesn't maintain lists of aliases for each value. So it'll require write.rs to first make a pass through all values and collect up all the value aliases. This in turn will require adding a way to iterate over the `values.iter()` of the `DataFlowGraph`. The writer can use an `EntityList<Value, Vec<Value>>` to map from values to their aliases. Then, when the writer is printing instructions, it can check if any of the values of an instruction have aliases, and if so, print them after the instruction.

PR #492, opened 2018-08-30
Print value aliases at referrent definition

kazcw:
This is a WIP implementation for #488--AFAIK it works, but all tests pass with alias printing deleted entirely so I need to add some tests that cover this behavior.

### DIFF
--- a/lib/codegen/src/write.rs
+++ b/lib/codegen/src/write.rs
@@ -111,11 +111,12 @@ pub fn write_function(w: &mut Write, func: &Function, isa: Option<&TargetIsa>) -
     decorate_function(&mut PlainWriter, w, func, isa)
 }
 
+/// Create a reverse-alias map from a value to all aliases having that value as a direct target
 fn alias_map(func: &Function) -> EntityMap<Value, Vec<Value>> {
     let mut aliases = EntityMap::<_, Vec<_>>::new();
     for v in func.dfg.values() {
-        let k = func.dfg.resolve_aliases(v);
-        if k != v {
+        // VADFS returns the immediate target of an alias
+        if let Some(k) = func.dfg.value_alias_dest_for_serialization(v) {
             aliases[k].push(v);
         }
     }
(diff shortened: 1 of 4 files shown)


kazcw:
Added a test that aliases print exactly once and at their definition location (whether parameter or result).

lib/codegen/src/write.rs
```diff
@@ -251,27 +270,27 @@ fn type_suffix(func: &Function, inst: Inst) -> Option<Type> {
...
-        }
+// Write out any aliases to the given target
+fn write_value_aliases(
+    w: &mut Write,
+    aliases: &EntityMap<Value, Vec<Value>>,
+    target: Value,
+    indent: usize,
+) -> fmt::Result {
+    for a in &aliases[target] {
+        writeln!(w, "{1:0$}{2} -> {3}", indent, "", a, target)?;
```
sunfishcode: Could you make this look up the immediate target of the alias, rather than the fully resolved one? We want to print the IR as it is, without any simplifications.

Aliases to aliases should be rare, and in some cases could indicate a problem, so when they do occur, it's useful to see them. And, it's useful to be able to print the IR and read it back in without anything getting canonicalized.

kazcw: Pushed a squashme. Single-level alias following was defined in a `_for_parser` method, so I moved it to a normal method now that write.rs needs it too. I also had to use `make_value_alias_for_parser` in tests because it looks like only the parser can create an alias to an alias.

sunfishcode: We almost always do want to `resolve_aliases` though :-). I think the only cases where we don't are when we're serializing and deserializing the IR in some way. So, what if we left these routines as-is and renamed them from "for_parser" to "for_serialization"? Then we can call them from write.rs or other potential serialize/deserialize routines.

kazcw: OK, I did it that way

sunfishcode:
Looks good!

PR #503, opened 2018-09-07
Improve simplejit-example

joshuawarner32:
This does two things:
* Fix "cargo run --example simpljit-minimal", by not suppressing default features for the dev-dependency on the umbrella crate
* Demonstrate how to use arguments, returns, and an "add" instruction in simplejit-minimal


sunfishcode:
Cool! I didn't know about `cargo run --example` before. And great content in the example code, complete with a check for getting the correct answer :-).

PR #507, opened 2018-09-09
Added FunctionBuilder::{call_memcpy, call_memset, call_memmove}

Aaronepower:
I couldn't really figure out how to write the tests, and I don't what type pointers should be as there doesn't appear to be a type like that in `types`.


sunfishcode:
This is a good start!

See the `sample_function` test in lib/frontend/src/frontend.rs for a simple test that builds an IR function, and then you can do `func.display(None)` to get a String that you can check. Unfortunately, it's not easier to create an instance of `TargetIsa` from within a testcase right now. One option is to do something like the arm32 code in the tests in lib/codegen/src/regalloc/pressure.rs. Hard-coding a specific target for the test would also make it easier to expect a specific output string. So that's probably the best approach here.

lib/frontend/src/frontend.rs
```diff
@@ -546,6 +547,59 @@ impl<'a> FunctionBuilder<'a> {
     }
 }
 
+/// Helper functions
+impl<'a> FunctionBuilder<'a> {
+    /// Calls libc.memcpy
+    ///
+    /// Copies the `size` bytes from `src` to `dest`, assumes that `src + size`
+    /// won't overlap onto `dest`.
+    pub fn call_memcpy(&mut self, dest: Value, src: Value, size: Value) {
```
bjorn3: Could you also add a `memmove` version for overlapping memory, or add a bool for it to this function?

Aaronepower:
@sunfishcode Could you check that the test is what should and if it all looks good I'll add memset per @bjorn3 's suggestion.

lib/frontend/src/frontend.rs
```diff
@@ -546,6 +546,54 @@ impl<'a> FunctionBuilder<'a> {
...
+    ///
+    /// Copies the `size` bytes from `src` to `dest`, assumes that `src + size`
+    /// won't overlap onto `dest`.
+    pub fn call_memcpy(&mut self, isa: &TargetIsa, dest: Value, src: Value, size: Value) {
+        let pointer_type = isa.pointer_type();
+        let signature = {
+            let mut s = Signature::new(isa.flags().call_conv());
+            s.params.push(AbiParam::new(pointer_type));
+            s.params.push(AbiParam::new(pointer_type));
+            s.params.push(AbiParam::new(types::I32));
```
sunfishcode: The size parameter should be pointer_type too. 

(Technically it's size_t, but that's the same size as a pointer on all popular platforms.)

lib/frontend/src/frontend.rs
```diff
@@ -546,6 +546,54 @@ impl<'a> FunctionBuilder<'a> {
...
+
+    /// Calls libc.memset
+    ///
+    /// Stores `num` copies of `ch` to `buffer`.
+    pub fn call_memset(&mut self, isa: &TargetIsa, buffer: Value, ch: Value, num: Value) {
+        let pointer_type = isa.pointer_type();
+        let signature = {
+            let mut s = Signature::new(isa.flags().call_conv());
+            s.params.push(AbiParam::new(pointer_type));
+            s.params.push(AbiParam::new(types::I8));
```
sunfishcode: Libc memset's second argument is int, so we should declare it as i32, and emit a uextend instruction to extend `ch` to i32.

(Technically int can be other sizes, but it's ok to ignore that here for now :-)).

Aaronepower:
@sunfishcode Updated with `call_memmove` and suggested changes.

lib/frontend/src/frontend.rs
sunfishcode: Memmove's size parameter should be `pointer_type` too.

lib/frontend/src/frontend.rs
```diff
@@ -545,6 +545,79 @@ impl<'a> FunctionBuilder<'a> {
...
+            self.import_signature(s)
+        };
+
+        let libc_memset = self.import_function(ExtFuncData {
+            name: ExternalName::LibCall(LibCall::Memset),
+            signature,
+            colocated: false,
+        });
+
+        self.ins().call(libc_memset, &[buffer, ch, len]);
```
sunfishcode: Can you add a `uextend` instruction to convert `ch` from i8 to i32?

sunfishcode:
Looks good! I expect we'll be using this a lot when compiling Rust code, with values in Rust being movable and all :-).

PR #509, opened 2018-09-11
Update the reload pass to replace copies with fill/spill instructions

### LINKED ISSUE #245: Reload: Elide copy instructions
Copy instructions are inserted by two regalloc passes before the reload pass:

1. The coalescer can insert copies before branches to separate virtual registers that would conflict.
2. The spiller can insert copies before two-address instructions, and to deal with other tricky constraints.

The reload pass should handle these copy instructions better when one or both values have been spilled.

```
v10 = copy v5
```

Currently becomes:

```
v50 = fill v5
v10 = copy v50
```

or

```
v50 = copy v5
v10 = spill v50
```

These copies are not necessary. If the reload pass recognized `copy` instructions, we could do better:

```
v10 = fill v5
```

or

```
v10 = spill v5
```

In the case where both values were spilled, we need a stack-stack copy:

```
v50 = fill v5
v10 = spill v50
```

mlea13:
This fixes #245.

### DIFF
--- a/filetests/regalloc/reload.clif
+++ b/filetests/regalloc/reload.clif
@@ -17,3 +17,30 @@ ebb0:
     ; check: $(reload=$V) = fill v0
     ; check: return $reload
 }
+
+; Check that copies where the arg has been spilled are replaced with fills.
+;
+; RV32E has 6 registers for function arguments so the 7th, v6, will be placed
+; on the stack.
+function %spilled_copy_arg(i32, i32, i32, i32, i32, i32, i32) -> i32 {
+
+ebb0(v0: i32, v1: i32, v2: i32, v3: i32, v4: i32, v5: i32, v6: i32):
+    ; not: copy
+    ; check: v10 = fill v6
+    v10 = copy v6
+    return v10
+}
+
+; Check that copies where the result has been spilled are replaced with spills.
+;
+; v1 is live across a call so it will be spilled.
+function %spilled_copy_result(i32) -> i32 {
+    fn0 = %foo(i32)
+
+ebb0(v0: i32):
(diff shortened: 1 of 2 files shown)


filetests/regalloc/reload.clif
```diff
@@ -17,3 +17,30 @@ ebb0:
...
+function %spilled_copy_result(i32) -> i32 {
+    fn0 = %foo(i32)
+
+ebb0(v0: i32):
+    ; not: copy
+    ; check: v1 = spill v0
+    v1 = copy v0
+    call fn0(v1)
+    return v1
+}
```
bjorn3: Missing trailing newline

lib/codegen/src/regalloc/reload.rs
```diff
@@ -209,40 +209,14 @@ impl<'a> Context<'a> {
...
-            let reg = self.cur.ins().fill(cand.value);
-            let fill = self.cur.built_inst();
-
-            self.reloads.insert(ReloadedValue {
-                stack: cand.value,
-                reg,
-            });
-            cand.value = reg;
-
-            // Create a live range for the new reload.
```
sunfishcode: I'm curious why you removed this comment.

mlea13: Oops, meant to copy that over as well.

filetests/regalloc/reload.clif
mlea13: Fixed.

sunfishcode:
Thanks for working on this! A few tests I tried this on saw reductions of the number of instructions of 9%.

PR #512, opened 2018-09-12
clif-util: schedule pass jobs incrementally while scanning directories

fitzgen:
When scanning directories, if we were doing a non-pass test runner suite, then scanned test files would be incrementally scheduled for execution, but if we were doing a pass test runner suite, then they would not be. This seems to be because the method for scheduling jobs is different for pass runs.

This commit factors out the schedule-these-jobs-please callback from scanning directories, so that pass jobs can be incrementally scheduled and pipelined with more directory scanning just like other jobs.

r? @sunfishcode

### DIFF
--- a/lib/filetests/src/runner.rs
+++ b/lib/filetests/src/runner.rs
@@ -30,12 +30,6 @@ enum State {
     Done(TestResult),
 }
 
-#[derive(PartialEq, Eq, Debug)]
-pub enum IsPass {
-    Pass,
-    NotPass,
-}
-
 impl QueueEntry {
     pub fn path(&self) -> &Path {
         self.path.as_path()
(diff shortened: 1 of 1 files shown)


sunfishcode:
The `pass` subcommand is initially meant to only accept a single input file, because its initial purpose is to allow developers to run specific passes on specific files to see how they behave, rather than batch processing.

Unless you want batch processing now, it would be better to refactor this code to avoid calling scan_dirs, and instead just issue an error if the input is a directory.

sunfishcode:
Hi @fitzgen, are you interested in pursing this soon? Otherwise we can close it for now and reopen it later as needed.

fitzgen:
Nope -- I've dropped this for now.

PR #516, opened 2018-09-14
[WIP] Don't pass around the dfg separately from the func?

sunfishcode:
This is an experiment I did a while ago, and I thought I'd post it to see what others think. It takes all the places which pass around `&func.dfg` and makes them pass `&func` instead. This essentially shows that it would be possible to merge `DataFlowGraph`'s contents into `Function`, which is to say, move everything from `func.dfg` into `func`. So this patch demonstrates that we could easily do this, if we wanted to.

I don't have a strong feeling either way yet. On one hand, I don't like how the division between `func` and `func.dfg` is non-obvious, not to mention making common API usage more cluttered with extra `.dfg`s in various places (and "dfg" isn't as well-known an acronym). On the other, it is kind of interesting that the separation, while not immediately obvious, does produce a relatively clean separation of concerns. `DataFlowGraph` can do a lot of things without the full `Function`, and `Function` doesn't need access to several things that are private to `DataFlowGraph`.


bnjbvr:
Hi @sunfishcode ! The Python code doesn't exist anymore; can you rebase this, please? (I'm happy to review it as soon as it's rebased!) Thanks!

bnjbvr:
Closing for inactivity, feel free to reopen if you get some time to work on this. Thanks!

PR #517, opened 2018-09-19
Add a sparse Switch usable instead of JumpTable to cranelift-frontend

### LINKED ISSUE #438: Switch lowering
Cranelift's jump_table mechanism is intended to directly map to an actual jump table, and it's limited to dense ranges of cases starting at 0, and to moderate sizes. Cranelift should provide a facility to make it east to lower "switch" statements, where cases may be sparse, very large, or may not start at zero.

One way to do this within Cranelift's current design is to add code to `cranelift-frontend` which presents an API for declaring switch statements, and which then produces lowered Cranelift IR for them. It could define a `Switch` struct, containing an index type, a `Vec` of `Case`s, where `Case` is a struct containing an index value and an `Ebb` to jump to on that value, and a `Case` for a default destination. Once the user has created and populated a `Switch`, they could then pass it into the library and request code be emitted for it.

There are a lot of techniques for switch lowering, so to keep things simple to start with, we can probably focus on just two: branch trees and jump tables. A simple algorithm that would be good to start with would be to sort the cases, form groups of contiguous ranges of cases, and then do a binary search through the groups using a tree of branches. Then for each group, emit a jump table.

bjorn3:
fixes #438

I will add docs tomorrow.

### DIFF
--- a/lib/frontend/src/switch.rs
+++ b/lib/frontend/src/switch.rs
@@ -73,39 +73,34 @@ impl Switch {
             _ => val,
         };
 
-        let cases_and_jt_ebbs: Vec<Option<(EntryIndex, Ebb, Vec<Ebb>)>> = cases_tree
-            .into_iter()
-            .rev()
-            .map(|(first_index, ebbs)| {
-                if ebbs.len() == 1 {
-                    let is_good_val = bx.ins().icmp_imm(IntCC::Equal, val, first_index);
-                    bx.ins().brnz(is_good_val, ebbs[0], &[]);
-                    None
-                } else {
-                    let jt_ebb = bx.create_ebb();
-                    let is_good_val =
-                        bx.ins()
-                            .icmp_imm(IntCC::UnsignedGreaterThanOrEqual, val, first_index);
-                    bx.ins().brnz(is_good_val, jt_ebb, &[]);
-                    Some((first_index, jt_ebb, ebbs))
-                }
-            }).collect();
+        let mut cases_and_jt_ebbs: Vec<(EntryIndex, Ebb, Vec<Ebb>)> = Vec::new();
+        for (first_index, ebbs) in cases_tree.into_iter().rev() {
+            if ebbs.len() == 1 {
+                let is_good_val = bx.ins().icmp_imm(IntCC::Equal, val, first_index);
+                bx.ins().brnz(is_good_val, ebbs[0], &[]);
+            } else {
+                let jt_ebb = bx.create_ebb();
+                let is_good_val =
+                    bx.ins()
+                        .icmp_imm(IntCC::SignedGreaterThanOrEqual, val, first_index);
+                bx.ins().brnz(is_good_val, jt_ebb, &[]);
+                cases_and_jt_ebbs.push((first_index, jt_ebb, ebbs));
+            }
+        };
 
         bx.ins().jump(otherwise, &[]);
 
-        for elem in cases_and_jt_ebbs.into_iter().rev() {
-            if let Some((first_index, jt_ebb, ebbs)) = elem {
-                let mut jt_data = JumpTableData::new();
-                for ebb in ebbs {
-                    jt_data.push_entry(ebb);
-                }
-                let jump_table = bx.create_jump_table(jt_data);
-
-                bx.switch_to_block(jt_ebb);
-                let discr = bx.ins().iadd_imm(val, -first_index);
-                bx.ins().br_table(discr, jump_table);
-                bx.ins().jump(otherwise, &[]);
+        for (first_index, jt_ebb, ebbs) in cases_and_jt_ebbs.into_iter().rev() {
+            let mut jt_data = JumpTableData::new();
+            for ebb in ebbs {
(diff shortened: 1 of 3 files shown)


lib/frontend/src/switch.rs
```diff
@@ -0,0 +1,193 @@
...
+            if let Some((first_index, jt_ebb, ebbs)) = elem {
+                let mut jt_data = JumpTableData::new();
+                for ebb in ebbs {
+                    jt_data.push_entry(ebb);
+                }
+                let jump_table = bx.create_jump_table(jt_data);
+
+                bx.switch_to_block(jt_ebb);
+                let first_index = bx.ins().iconst(val_type, first_index);
+                let discr = bx.ins().isub(val, first_index);
```
sunfishcode: Can this be an `isub_imm` rather than needing a separate `iconst`?

lib/frontend/src/switch.rs
```diff
@@ -0,0 +1,193 @@
...
+
+impl Switch {
+    pub fn new() -> Self {
+        Switch {
+            cases: HashMap::new(),
+        }
+    }
+
+    pub fn set_entry(&mut self, index: EntryIndex, ebb: Ebb) {
+        assert!(self.cases.insert(index, ebb).is_none(), "Tried to set the same entry {} twice", index);
```
sunfishcode: Could you move the `insert` out of the `assert!` into a separate statement? The code works either way, but I find it easier to read when the program logic is separated from `assert!` conditions.

lib/frontend/src/switch.rs
```diff
@@ -0,0 +1,193 @@
...
+    fn build_cases_tree(self) -> Vec<(EntryIndex, Vec<Ebb>)> {
+        println!("before: {:#?}", self.cases);
+        let mut cases = self.cases.into_iter().collect::<Vec<(_, _)>>();
+        cases.sort_by_key(|&(index, _)| index);
+
+        let mut cases_tree: Vec<Vec<(EntryIndex, Ebb)>> = vec![];
+        let mut last_index = None;
+        for (index, ebb) in cases {
+            if Some(index - 1) > last_index {
+                cases_tree.push(Vec::new());
```
sunfishcode: Would it work to make `cases_tree` be a `Vec<(EntryIndex, Vec<ebb>)>`, and do `cases_tree.push((index, Vec::new())` here? If so, it seems like you could avoid the `map` step below.

lib/frontend/src/switch.rs
```diff
@@ -0,0 +1,193 @@
...
+            cases: HashMap::new(),
+        }
+    }
+
+    pub fn set_entry(&mut self, index: EntryIndex, ebb: Ebb) {
+        assert!(self.cases.insert(index, ebb).is_none(), "Tried to set the same entry {} twice", index);
+    }
+
+    fn build_cases_tree(self) -> Vec<(EntryIndex, Vec<Ebb>)> {
+        println!("before: {:#?}", self.cases);
```
sunfishcode: We are using the `log` crate now, so you could make these use `debug!` for logging this kind of information.

lib/frontend/src/switch.rs
```diff
@@ -0,0 +1,193 @@
...
+                    bx.ins().brnz(is_good_val, ebbs[0], &[]);
+                    None
+                } else {
+                    let jt_ebb = bx.create_ebb();
+                    let is_good_val = bx.ins().icmp_imm(IntCC::UnsignedGreaterThanOrEqual, val, first_index);
+                    bx.ins().brnz(is_good_val, jt_ebb, &[]);
+                    Some((first_index, jt_ebb, ebbs))
+                }
+            })
+            .collect();
```
sunfishcode: There are of course lots of refinements possible here, most of which we can do later, however there is one thing that we'll need to do before this feature is "minimally viable", which is to use a binary search tree here rather than a linear sequence of branches, so that we don't have surprisingly bad performance when there are a lot of gaps between the cases.

lib/frontend/src/switch.rs
bjorn3: `isub_imm` doesnt exist.

lib/frontend/src/switch.rs
bjorn3: Sure, I forgot that i added those.

lib/frontend/src/switch.rs
sunfishcode: Oops, yes, the reason there's no isub_imm is because you can use iadd_imm with a negative immediate.

bjorn3:
Note to self: https://github.com/CraneStation/cranelift/pull/517#discussion_r219015365 is not fixed yet.

bjorn3:
Please check that I didn't mess it up.

bjorn3:
Rebased

sunfishcode:
Looks good! I expect Rust compilation will involve a lot of switches, with all the enums and matches :-). This is a great feature, and a base to build further optimizations on!

bjorn3:
> I expect Rust compilation will involve a lot of switches, with all the enums and matches

Yes, even if's are.

PR #520, opened 2018-09-20
More introductory documentation

sunfishcode:
This adds a few more miscellaneous changes.

I recently learned about \<details\>, which seems like a nice way to keep the main README.md landing page a little tidier while still making additional information available.

And, I took the liberty of reorganzing the issue labels again. Unfortunately the issue labels themselves can't be changed via PR. But the basic idea is to follow the Rust repo's issue scheme, with `E-easy` etc indicating experience levels needed, and so on. I'm open to suggestions here!

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -27,12 +27,22 @@ We're happy to mentor people, whether you're learning Rust, learning about
 compiler backends, learning about machine code, learning about how Cranelift
 does things, or all together at once.
 
-We tag issues in the issue tracker marked [good first issue] when we can, so
-that's sometimes a good place to get started. Also, we encourage people to just
-look around and find things they're interested in. This a good time to get
-involved, as there aren't a lot of things set in stone yet.
+We categorize issues in the issue tracker using a tag scheme inspired by
+[Rust's issue tags]. For example, the [E-easy] marks good beginner issues,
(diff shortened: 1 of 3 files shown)


CONTRIBUTING.md
```diff
@@ -27,12 +27,25 @@ We're happy to mentor people, whether you're learning Rust, learning about
...
+See also the [full list of labels].
+
+
+Also, we encourage people to just look around and find things they're
+interested in. This a good time to get involved, as there aren't a lot of
+things set in stone yet.
+
+[Rust's issue tags]: https://github.com/rust-lang/rust/blob/master/CONTRIBUTING.md#issue-triage
+[E-easy]: https://github.com/CraneStation/cranelift/labels/E-easy
+[E-rust]: https://github.com/CraneStation/cranelift/labels/E-rust
```
bjorn3: Shouldnt this be an Area label, instead of an Efford label? https://github.com/rust-lang/rust-wiki-backup/blob/master/Note-tag-label-names-and-definitions.md

sunfishcode: That rust-wiki-backup page appears to be an older version; in the CONTRIBUTING.md link above, "E-" labels are for "experience" rather than "effort". The goal here is to highlight issues for people who have general Rust experience.

bjorn3: I see

PR #522, opened 2018-09-23
Replaced LibCall's fmt::Display implementation

Aaronepower:
Replacing use of the `NAME` with just calling the Debug implementation makes it less prone to accidentally printing the wrong thing or panicking in case a the enum does not match up to the array.

### DIFF
--- a/lib/codegen/src/ir/libcall.rs
+++ b/lib/codegen/src/ir/libcall.rs
@@ -40,21 +40,9 @@ pub enum LibCall {
     NearestF64,
 }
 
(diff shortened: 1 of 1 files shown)


sunfishcode:
Sounds reasonable to me.

Long term, it may also make things more consistent to avoid using CamelCase in the serialized form (since nothing else does that) however if we do that, we can easily switch to a `match` rather than go back to an explicit array.

BTW, the CI failure here is due to https://github.com/m4b/scroll/issues/40 and is unrelated to the PR here, so we don't have to worry about that here.

Issue #466, opened 2018-08-14
Tracking issue for missing i8/i16 encodings


bjorn3:
Opening this issue to prevent polluting the issue list

## Clz.i16

```
target x86_64

function %clz_i16(i16) -> i16 fast {
ebb0(v0: i16):
    v1 = clz.i16 v0
    return v1
}
```

```
function %clz_i16(i16 [%rdi]) -> i16 [%rax] fast {
                                ebb0(v0: i16):
                                    v1 = clz.i16 v0
[Op1ret#c3]                         return v1
                                    ^~~~~~ verifier inst1: v1 is a ghost value used by a real [Op1ret#c3] instruction
}
```

## ireduce i16 -> i8

```
target x86_64

function %ireduce_i8(i16) -> i8 fast {
ebb0(v0: i16):
    v1 = ireduce.i8 v0
    return v1
}
```

```
function %ireduce_i8(i16 [%rdi]) -> i8 [%rax] fast {
                                ebb0(v0: i16):
                                    v1 = ireduce.i8 v0
[Op1ret#c3]                         return v1
                                    ^~~~~~ verifier inst1: v1 is a ghost value used by a real [Op1ret#c3] instruction

}
```

## iadd_cout.i8 / isub_bout.i8

bjorn3:
## icmp_imm.i8

bjorn3:
## ushr.i8

bjorn3:
I added `isplit.i32` and `iconcat.i32`. (I need them for the bswap intrinsic)

bjorn3:
* [ ] `rot{l,r}_imm.i16`

bjorn3:
* [ ] fcvt_to_uint_sat.i8

jyn514:
`icmp.i8` is marked as completed here, but I'm getting a verifier error for `icmp ne v8, v9`, where v8 and v9 are `i8`.

```
function u0:0(i64) -> i64 system_v {
    ss0 = explicit_slot 8
    ss1 = explicit_slot 8

ebb0(v0: i64):
    v1 = stack_addr.i64 ss0
    store v0, v1
    v2 = iconst.i64 0
    v3 = stack_addr.i64 ss1
    store v2, v3
    jump ebb1

ebb1:
    v4 = stack_addr.i64 ss0
    v5 = iconst.i64 1
    v6 = load.i64 v4
    v7 = iadd v6, v5
    store v7, v4
    v8 = load.i8 v6
[...]

PR #524, opened 2018-09-23
Legalize some more i8/i16 intructions

bjorn3:
cc #466


sunfishcode:
Nice, this factoring should make it easy to add even more operators as well. Of course, the generated code is currently quite verbose, but we'll eventually have ways to optimize a lot of this away :-).

PR #531, opened 2018-09-28
Put TargetIsa's emit_inst under a "testing_hooks" feature.

### LINKED ISSUE #530: multiple copies of cranelift_codegen::isa::x86::binemit::emit_inst
Landing cranelift in firefox produced [a large regression in installer size](https://bugzilla.mozilla.org/show_bug.cgi?id=1494976#c1).  There are a couple of cranelift functions in the top 100 largest symbols, but these two stand out (readelf output):

```
 68448: 0000000004317a30 62803 FUNC    LOCAL  DEFAULT   16 _ZN17cranelift_codegen3isa3x867binemit9emit_inst17hc9a5f910e87df419E
 68447: 0000000004327120 61669 FUNC    LOCAL  DEFAULT   16 _ZN17cranelift_codegen3isa3x867binemit9emit_inst17h10353c1070eace20E
```

That is, we have two copies of this `emit_insn` function, specialized differently, each weighing in at ~60k.  Is it possible to refactor things so we avoid these separate specializations?

sunfishcode:
In practice, TargetIsa's emit_inst pulls in its own instantiation
of the target-specific `emit_inst` functions, which can be quite
large, and LTO doesn't eliminate them because they're held live
by TargetIsa's vtable.

Fortunately, this function is only used by tests, so we can put
it behind a feature flag.

Fixes #530.

### DIFF
--- a/lib/codegen/Cargo.toml
+++ b/lib/codegen/Cargo.toml
@@ -34,6 +34,9 @@ cranelift-codegen-meta = { path = "meta", version = "0.22.0" }
 default = ["std"]
 std = ["cranelift-entity/std", "cranelift-bforest/std", "target-lexicon/std"]
 core = ["hashmap_core"]
+# This enables some additional functions useful for writing tests, but which
+# can significantly increase the size of the library.
+testing_hooks = []
 
 [badges]
 maintenance = { status = "experimental" }
--- a/lib/codegen/src/isa/arm32/mod.rs
+++ b/lib/codegen/src/isa/arm32/mod.rs
@@ -7,7 +7,9 @@ mod registers;
 pub mod settings;
 
 use super::super::settings as shared_settings;
-use binemit::{emit_function, CodeSink, MemoryCodeSink};
+#[cfg(feature = "testing_hooks")]
+use binemit::CodeSink;
+use binemit::{emit_function, MemoryCodeSink};
 use ir;
 use isa::enc_tables::{self as shared_enc_tables, lookup_enclist, Encodings};
 use isa::Builder as IsaBuilder;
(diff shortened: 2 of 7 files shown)


sunfishcode:
We include the ability to run tests in the clif-util binary, which isn't built in `#[test]` mode.

So maybe one option would be to stop doing that. It's debatable whether we need the ability to do `clif-util test something.clif` when one could instead register the test and then use `cargo test --package cranelift-codegen [TESTNAME]` or similar.

sunfishcode:
Do you think this patch is good for now, or should we look at switching to #[test] now?

bnjbvr:
> Do you think this patch is good for now, or should we look at switching to #[test] now?

I think the only drawback to merge as is is that this introduces a feature flag, which will make it into a release, that we might be disabling/removing later, creating a bit of churn for embedders. In the worst case, we might end up with a feature flag that's a no-op, which would be bad. So I'll tend to say we should move to the most proper way now, which seems to be using `#[test]`, if it's not too much hassle.

That being said, we're still far from v1, so having breaking API changes seems fine, as long as they're properly documented when a release takes place.

I can't get to a strong position, so I'll leave the choice out to you; both decisions would make sense.

sunfishcode:
I've been thinking about this too, and I think what it comes down to is: is `clif-util test`, the ability to put tests in .clif files and directly test them by hand, useful? And with the way filetests currently works, I think it is.

It's possible we could move to a different style of tests in the future which wouldn't involve `clif-util test` in the same way. But if so, it seems like we should be able to remove the `testing_hooks` feature without too much trouble. I've now added a comment to `emit_inst` clarifying its purpose, so hopefully people won't unknowingly depend on it.

Issue #537, opened 2018-10-01
Logging level shouldn't be limited in release mode


bnjbvr:
The `log` library provides feature flags that disable compilation of log lines at compile-time, for log levels that are lesser than the one specified in the feature. We use `release_max_level_warn` which will only accept `warn` and `error` log messages.

Unfortunately, the code checking if these features are enabled in `log` is iterating from the most restrictive to the least restrictive: e.g. if a program embedding cranelift enables the `info` log level in release mode, it won't be taken into account because cranelift sets that the log level in release mode is `warn` (or higher).

Since Cranelift is to be used as a library, I don't think it should define any feature like this, so as to leave the most liberty to the embedder.

PR #538, opened 2018-10-01
Remove logging levels restrictions

bnjbvr:
See #537 for the reasoning.

### DIFF
--- a/README.md
+++ b/README.md
@@ -131,6 +131,22 @@ protection. Just something to think about.
 </details>
 
 <details>
+<summary>Log configuration</summary>
+
+Cranelift uses the `log` crate to log messages at various levels. It doesn't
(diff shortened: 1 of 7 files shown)


sunfishcode:
With this change, can a program embedding Cranelift still disable trace/debug/info messages at compile time?

bnjbvr:
Yes, the log features chosen by the embedder program will prevail; so if the embedder decides to disable trace/debug/info, it can.

sunfishcode:
Ok, makes sense. We should just document that users can add `features = ["release_max_level_warn"]` or so to their log dependency to reduce code size somewhere. Maybe in the top-level README.md under "Building Cranelift?"

bnjbvr:
Done!

sunfishcode:
Looks good!

PR #540, opened 2018-10-02
Fixes #504: Implement Display/Fail for SetError and LookupError

### LINKED ISSUE #504: Make LookupError and SetError implement Fail
I'd like to propose that all the errors defined in Cranelift implement the `std::error::Error` trait, so they can be used in Error boxed traits, which are quite convenient to use (e.g. to unify all the possible Result types appearing in signatures of functions called by a given function). This would be a usability improvement, in my opinion.

The trait is right there: https://doc.rust-lang.org/std/error/trait.Error.html

We'd use conditional compilation on the presence of std, to not break non-std builds.

What do people think about this?

bnjbvr:


lib/codegen/src/isa/mod.rs
```diff
@@ -129,6 +129,15 @@ pub enum LookupError {
...
 }
 
+impl fmt::Display for LookupError {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        match self {
+            LookupError::SupportDisabled => write!(f, "Support for this target is disabled"),
+            LookupError::Unsupported => write!(f, "Support for this target is not implemented yet"),
+        }
+    }
+}
```
sunfishcode: We've started to use the `#[fail(...)]` macros in a few places for this kind of code, such as [here](https://github.com/CraneStation/cranelift/blob/master/lib/codegen/src/result.rs#L14), although I don't know if being slightly more concise is worth the benefit of being somewhat more magical. Do you have a preference?

bnjbvr: Yeah, it's more concise, I like it better, thanks!

bnjbvr:
I've added a second commit that makes SettingsError more verbose, which I think is useful for debugging; let me know if you think they're a good idea!

sunfishcode:
I like it!

PR #541, opened 2018-10-03
Add support for custom passes

lachlansneff:
This commit adds support for custom, user-supplied passes during compilation. The api may not be optimal, so I'm happy to revise.

### DIFF
--- a/lib/codegen/src/context.rs
+++ b/lib/codegen/src/context.rs
@@ -33,7 +33,13 @@ use unreachable_code::eliminate_unreachable_code;
 use verifier::{verify_context, verify_locations, VerifierErrors, VerifierResult};
 
 pub trait CustomPass {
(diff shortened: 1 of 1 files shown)


sunfishcode:
I think my main concern is that while it may work for your pass to run just before register allocation, other passes might want to go after register allocation, or before legalization. Rather than hooks, what if we factored out the contents of `compile` into functions named something like `before_legalize`, `after_legalize`, and `after_regalloc`, so that users can call those directly if they want to do other things between the steps?

sunfishcode:
I like the token idea, however it doesn't quite fit with some other directions I think this code can evolve in. I envision the ability to run things like `compile_after_legalize` on IR obtained from deserialization, in which case we won't have a token to pass it.

Something we could do to help though is to move in the direction of making passes more aware of their preconditions. preopt doesn't update encodings, so it shouldn't be run on IR that has encodings already. Register allocation requires that all non-"ghost" instructions have encodings. And so on.

PR #543, opened 2018-10-04
Remove verify! macro that's used only once

bnjbvr:


sunfishcode:
@6A Do I recall correctly that an earlier version of this code used this macro in more places? It does seem like if we're only using it one place now, it'd be simpler to avoid the macro.

6A:
I think I also used it in some tests, but it's no longer the case. I see no problem in removing the macro.

bnjbvr:
Thanks for chiming in! It seems there are two ways to move forward with this:

- either take this PR as it is.
- or try to introduce more uses of the macro itself.

Since we've managed to live without more uses of the macro, I'd tend to do the former. Any thoughts about this?

6A:
We should remove the macro, rather than try to make it useful unnaturally. In any case, if we need it again we can always see it in the Git history.

sunfishcode:
Thanks! Then let's merge this, removing the macro.

PR #546, opened 2018-10-04
Remove the concept of non-dense jump tables.

sunfishcode:
WebAssembly doesn't have non-dense jump tables, and higher-level users
are better served by the facilities in lib/frontend/src/switch.rs for
working with non-dense switches.

This eliminates the concept of "absent" jump table entries, which
were represented as "0" in the text format.

Also, jump table contents are now enclosed in `[` and `]`, so that
we can unambiguously display empty jump tables. Previously, empty jump
tables were displayed as if they had a single absent entry.

### DIFF
--- a/docs/ir.rst
+++ b/docs/ir.rst
@@ -360,13 +360,12 @@ instruction in the EBB.
 .. autoinst:: brff
 .. autoinst:: br_table
 
-.. inst:: JT = jump_table EBB0, EBB1, ..., EBBn
+.. inst:: JT = jump_table [EBB0, EBB1, ..., EBBn]
 
     Declare a jump table in the :term:`function preamble`.
 
     This declares a jump table for use by the :inst:`br_table` indirect branch
-    instruction. Entries in the table are either EBB names, or ``0`` which
-    indicates an absent entry.
+    instruction. Entries in the table are EBB names.
 
     The EBBs listed must belong to the current function, and they can't have
     any arguments.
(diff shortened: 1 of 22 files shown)


lachlansneff:
Is the list syntax that the jt pseudo-instruction using new syntax for the clif text-representation?

sunfishcode:
@lachlansneff I'm a little confused. Which jt pseudo instruction?

If you mean the syntax for jump table declarations, then yes, the syntax changed: '0' is no longer permitted as an entry, and the entries are enclosed in '[' and ']'.

lachlansneff:
@sunfishcode Yes, sorry. I didn't structure my question very well.

I meant to ask if the brackets around arguments was a new syntax for clif ir. I don't remember seeing any other instructions with that method of adding arguments.

sunfishcode:
Ok, yeah. Yes, it is a new thing, syntax-wise. Cranelift's text format is one of those languages that doesn't use semicolons *and* doesn't treat newlines as significant, so it has fun ambiguities.

```
    jt0 = jump_table ebb0
```
vs
```
    jt0 = jump_table
ebb0:
```
We *could* disambiguate it by scanning ahead to the colon, but my instinct is to just change the syntax to be unambiguous rather than use extra lookahead to disambiguate.

`[` and `]` do appear elsewhere in the language, though not as lists. They specify register/stack locations for values and encodings for instructions.

PR #547, opened 2018-10-04
Move `return_at_end` out of Settings and into the wasm FuncEnvironment.

sunfishcode:
The `return_at_end` flag supports users that want to append a custom
epilogue to Cranelift-produced functions. It arranges for functions to
always return via a single return statement at the end, and users are
expected to remove this return to append their code.

This patch makes two changes:
 - First, introduce a `fallthrough_return` instruction and use that
   instead of adding a `return` at the end. That's simpler than having
   users remove the `return` themselves.

 - Second, move this setting out of the Settings and into the wasm
   FuncEnvironment. This flag isn't something the code generator uses,
   it's something that the wasm translator uses. The code generator
   needs to preserve the property, however we can give the
   `fallthrough_return` instruction properties to ensure this as needed,
   such as marking it non-cloneable.


lib/codegen/src/verifier/mod.rs
```diff
@@ -307,7 +307,7 @@ struct Verifier<'a> {
     func: &'a Function,
     expected_cfg: ControlFlowGraph,
     expected_domtree: DominatorTree,
-    flags: &'a Flags,
+    _flags: &'a Flags,
```
bnjbvr: Can it be removed, if it's unused?

sunfishcode: Done.

PR #549, opened 2018-10-04
Optimize load/store with an iadd_imm operand.

sunfishcode:
Fold the immediate into the load/store offset when possible.

I'm hoping to avoid doing too many special-case pattern matches like this, as I'm looking forward to developing a more comprehensive pattern-matching system, but it does make sense to do a few simple and common ones for now.

### DIFF
--- a/filetests/postopt/fold_offset_into_address.clif
+++ b/filetests/postopt/fold_offset_into_address.clif
@@ -1,0 +1,18 @@
+test postopt
+target x86_64
+
+; Fold the immediate of an iadd_imm into an address offset.
+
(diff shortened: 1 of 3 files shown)


tyler:
Does it make sense to do this for load/store_complex ops too?

sunfishcode:
I myself haven't seen reg+reg+imm come up in WebAssembly, and it might be because when that would happen, in theory wasm producers should be using the wasm load/store offset field, which we already do optimize.

That said, if you see this pattern happening, that optimization is something we will want to do eventually anyway, so I wouldn't object :-).

PR #551, opened 2018-10-06
Declare br_table otherwise ebb as predecessor

bjorn3:
fixes #545

PR #552, opened 2018-10-08
Fix #335: Introduce variable size recipes and remove GPR_SAFE reg classes

### LINKED ISSUE #335: panic: Ran out of registers in GPR
Cretonne crashes when trying to compile `AngryBots1.wasm` from [wasm-collection](https://github.com/sunfishcode/wasm-collection):

```
RUST_BACKTRACE=1 target/release/cton-util wasm -T --set is_64bit=1 --isa x86 ../wasm-collection/misc-valid/AngryBots1.wasm
thread 'main' panicked at 'Invalid type for vsplit', libcore/option.rs:914:5
stack backtrace:
   0: std::sys::unix::backtrace::tracing::imp::unwind_backtrace
             at libstd/sys/unix/backtrace/tracing/gcc_s.rs:49
   1: std::sys_common::backtrace::print
             at libstd/sys_common/backtrace.rs:71
             at libstd/sys_common/backtrace.rs:59
   2: std::panicking::default_hook::{{closure}}
             at libstd/panicking.rs:211
   3: std::panicking::default_hook
             at libstd/panicking.rs:227
   4: std::panicking::rust_panic_with_hook
             at libstd/panicking.rs:463
   5: std::panicking::begin_panic_fmt
             at libstd/panicking.rs:350
   6: rust_begin_unwind
             at libstd/panicking.rs:328
   7: core::panicking::panic_fmt
             at libcore/panicking.rs:71
   8: core::option::expect_failed
             at libcore/option.rs:914
   9: cretonne_codegen::legalizer::split::split_value
             at /checkout/src/libcore/option.rs:302
             at lib/codegen/src/legalizer/split.rs:216
  10: cretonne_codegen::legalizer::split::split_any
             at lib/codegen/src/legalizer/split.rs:125
  11: cretonne_codegen::legalizer::boundary::convert_to_abi
             at lib/codegen/src/legalizer/split.rs:0
  12: cretonne_codegen::legalizer::boundary::handle_call_abi
             at lib/codegen/src/legalizer/boundary.rs:476
             at lib/codegen/src/legalizer/boundary.rs:505
  13: cretonne_codegen::legalizer::legalize_function
             at lib/codegen/src/legalizer/mod.rs:47
             at lib/codegen/src/legalizer/mod.rs:101
  14: cretonne_codegen::context::Context::compile
             at lib/codegen/src/context.rs:231
             at lib/codegen/src/context.rs:131
  15: cton_util::wasm::run
             at src/wasm.rs:161
             at src/wasm.rs:52
  16: cton_util::main
             at src/cton-util.rs:117
             at src/cton-util.rs:140
  17: std::rt::lang_start::{{closure}}
             at /checkout/src/libstd/rt.rs:74
  18: std::panicking::try::do_call
             at libstd/rt.rs:59
             at libstd/panicking.rs:310
  19: __rust_maybe_catch_panic
             at libpanic_unwind/lib.rs:105
  20: std::rt::lang_start_internal
             at libstd/panicking.rs:289
             at libstd/panic.rs:374
             at libstd/rt.rs:58
  21: main
  22: __libc_start_main
  23: _start
```

bnjbvr:
This removes the GPR_SAFE_DEREF and GPR_ZERO_SAFE_DEREF register classes on x86, relaxing constraints for the solver. Each recipe (sizing) now gets an optional function pointer field `compute_size` that computes the size of the instruction given register inputs; this allows the generated code size to be variable, and thus to insert SIB bytes for r12/rsp when needed, and a zero offset for r13/rbp when needed. Then, we don't need the special register classes anymore.

It happens that a few of the register class constraints were incorrect in the first place, i.e., a recipe would say it needs GPR_SAFE_DEREF while it always emits a SIB byte. These recipes just have a fixed size in this case.

I've tried to split my work into meaningful commits, to make review easier. I'd like to test this more, to make sure the generated code actually runs. Note that with this and #551, we can compile AngryBots and many other big wasm demos with Cranelift again.

### DIFF
--- a/filetests/regalloc/gpr-deref-safe-335.clif
+++ b/filetests/regalloc/gpr-deref-safe-335.clif
@@ -1,0 +1,44 @@
+test regalloc
+target x86_64
+
+function u0:587() fast {
+ebb0:
+    v97 = iconst.i32 0
+    v169 = iconst.i32 0
+    v1729 = iconst.i32 0
+    jump ebb100(v97, v97, v97, v97, v97)
+
+ebb100(v1758: i32, v1784: i32, v1845: i32, v1856: i32, v1870: i32):
+    v1762 = iconst.i32 0
+    v1769 = iconst.i32 0
+    v1774 = iconst.i32 0
+    v1864 = iconst.i32 0
+    v1897 = iconst.i32 0
+    jump ebb102(v1774, v1784, v1845, v1856, v1870, v1758, v1762, v169, v1729, v97, v169, v169, v169, v169)
+
+ebb102(v1785: i32, v1789: i32, v1843: i32, v1854: i32, v1868: i32, v1882: i32, v1890: i32, v1901: i32, v1921: i32, v1933: i32, v2058: i32, v2124: i32, v2236: i32, v2366: i32):
+    v1929 = iconst.i32 0
+    v1943 = iconst.i32 0
+    v1949 = iconst.i32 0
+    jump ebb123(v1897, v1769)
+
+ebb123(v1950: i32, v1979: i32):
+    v1955 = iconst.i32 0
+    brz v1955, ebb125
+    jump ebb122(v1929, v1843, v1864, v2058, v1882, v1897, v1943, v1868, v2124, v1901)
+
+ebb125:
+    v1961 = iadd_imm.i32 v1949, 0
+    v1952 = iconst.i32 0
+    v1962 = iconst.i64 0
+    v1963 = load.i32 v1962
+    brz v1963, ebb123(v1952, v1961)
+    jump ebb127
+
+ebb127:
+    v1966 = iconst.i32 0
+    jump ebb122(v1963, v1966, v1966, v1966, v1966, v1966, v1966, v1966, v1966, v1966)
+
+ebb122(v1967: i32, v1971: i32, v1972: i32, v1978: i32, v2032: i32, v2041: i32, v2053: i32, v2076: i32, v2085: i32, v2096: i32):
+    trap user0
+}
--- a/lib/codegen/meta-python/cdsl/isa.py
+++ b/lib/codegen/meta-python/cdsl/isa.py
@@ -329,24 +329,25 @@ class EncRecipe(object):
 
     def __init__(
             self,
-            name,                      # type: str
-            format,                    # type: InstructionFormat
-            base_size,                 # type: int
-            ins,                       # type: ConstraintSeq
-            outs,                      # type: ConstraintSeq
-            compute_size='base_size',  # type: str
-            branch_range=None,         # type: BranchRange
-            clobbers_flags=True,       # type: bool
-            instp=None,                # type: PredNode
-            isap=None,                 # type: PredNode
-            emit=None                  # type: str
+            name,                     # type: str
+            format,                   # type: InstructionFormat
+            base_size,                # type: int
+            ins,                      # type: ConstraintSeq
+            outs,                     # type: ConstraintSeq
+            compute_size=None,        # type: str
+            branch_range=None,        # type: BranchRange
+            clobbers_flags=True,      # type: bool
+            instp=None,               # type: PredNode
+            isap=None,                # type: PredNode
+            emit=None                 # type: str
             ):
         # type: (...) -> None
         self.name = name
         self.format = format
         assert base_size >= 0
         self.base_size = base_size
-        self.compute_size = compute_size
+        self.compute_size = compute_size if compute_size is not None \
+            else 'base_size'
         self.branch_range = branch_range
         self.clobbers_flags = clobbers_flags
         self.instp = instp
--- a/lib/codegen/meta-python/cdsl/isa.py
+++ b/lib/codegen/meta-python/cdsl/isa.py
@@ -316,7 +316,8 @@ class EncRecipe(object):
     :param name: Short mnemonic name for this recipe.
     :param format: All encoded instructions must have this
             :py:class:`InstructionFormat`.
-    :param size: Number of bytes in the binary encoded instruction.
+    :param base_size: Base number of bytes in the binary encoded instruction.
+    :param compute_size: Function name to use when computing actual size.
     :param ins: Tuple of register constraints for value operands.
     :param outs: Tuple of register constraints for results.
     :param branch_range: `(origin, bits)` range for branches.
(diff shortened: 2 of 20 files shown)


lib/codegen/meta-python/cdsl/isa.py
```diff
@@ -316,7 +316,8 @@ class EncRecipe(object):
     :param name: Short mnemonic name for this recipe.
     :param format: All encoded instructions must have this
             :py:class:`InstructionFormat`.
-    :param size: Number of bytes in the binary encoded instruction.
+    :param base_size: Base number of bytes in the binary encoded instruction.
```
sunfishcode: I'm a little surprised to see this as a base plus extra bytes, rather than a max minus unneeded bytes, however the rest of this patch seems to demonstrate that we don't currently need the max size for anything. And, it is more intuitive to start with a base and add from there.

I'm trying to think of anything that we might want a max for. But before regalloc while, it might occasionally be useful to know how big things are, we won't have the fills and spills yet. And after regalloc, we can compute the actual sizes. So it may not matter for the foreseeable future.

lib/codegen/src/isa/x86/enc_tables.rs
```diff
@@ -5,15 +5,73 @@ use bitset::BitSet;
...
+use isa::encoding::base_size;
 use isa::encoding::RecipeSizing;
+use isa::RegUnit;
+use regalloc::RegDiversions;
 
 include!(concat!(env!("OUT_DIR"), "/encoding-x86.rs"));
 include!(concat!(env!("OUT_DIR"), "/legalize-x86.rs"));
 
+pub fn needs_sib_byte(reg: RegUnit) -> bool {
+    reg == RU::r12 as u16 || reg == RU::rsp as u16
```
sunfishcode: Should this use `RegUnit` rather than `u16` here?

lib/codegen/meta-python/isa/x86/recipes.py
```diff
@@ -801,37 +803,50 @@ def valid_scale(iform):
 
 # XX /r register-indirect store with no offset.
 st = TailRecipe(
-        'st', Store, size=1, ins=(GPR, GPR_ZERO_DEREF_SAFE), outs=(),
+        'st', Store, base_size=1, ins=(GPR, GPR), outs=(),
         instp=IsEqual(Store.offset, 0),
         clobbers_flags=False,
+        compute_size="size_plus_maybe_offset_1",
```
sunfishcode: When I first read this, it wasn't obvious that the `_1` referred to `in_reg1`. Could we name it something like `size_plus_maybe_offset_for_in_reg1`?

lib/codegen/src/isa/encoding.rs
```diff
@@ -78,12 +80,23 @@ impl fmt::Display for DisplayEncoding {
     }
 }
 
+type SizeCalculatorFn = fn(&RecipeSizing, Inst, &RegDiversions, &Function) -> u8;
+
+/// Returns the base size of the Recipe, assuming it's fixed. This is the default for most
+/// encodings.
```
sunfishcode: It'd be good to explicitly mention that encodings may be longer than this, depending on what registers they use.

lib/codegen/src/isa/x86/enc_tables.rs
```diff
@@ -5,15 +5,73 @@ use bitset::BitSet;
...
+}
+pub fn needs_offset(reg: RegUnit) -> bool {
+    reg == RU::r13 as u16 || reg == RU::rbp as u16
+}
+
+fn _additional_size_if(
```
sunfishcode: This function is unconditionally used below; can we remove the leading underscore?

lib/codegen/meta-python/isa/x86/recipes.py
```diff
@@ -846,66 +861,92 @@ def valid_scale(iform):
...
             sink.trap(TrapCode::HeapOutOfBounds, func.srclocs[inst]);
         }
         PUT_OP(bits, rex2(in_reg1, in_reg0), sink);
-        modrm_disp8(in_reg1, in_reg0, sink);
+        if needs_sib_byte(in_reg1) {
+            modrm_sib_disp8(in_reg0, sink);
+            sib_noindex(in_reg1, sink);
+        } else {
+            modrm_disp8(in_reg1, in_reg0, sink);
+        }
```
sunfishcode: Here and in the other load and store recipes, could we factor out this code for emitting the modrm and sib bytes into a helper function? Perhaps it could be named something like `modrm_plus_maybe_sib_1` so that there's a clear correspondence with the compute_size function.

lib/codegen/meta-python/isa/x86/registers.py
```diff
@@ -46,14 +46,7 @@
...
 
 GPR = RegClass(IntRegs)
-# Certain types of deref encodings cannot be used with all registers.
-#   R13/RBP cannot be used with zero-offset load or store instructions.
-#   R12 cannot be used with a non-SIB-byte encoding of all derefs.
-GPR_DEREF_SAFE = GPR.without(GPR.rsp, GPR.r12)
-GPR_ZERO_DEREF_SAFE = GPR_DEREF_SAFE.without(GPR.rbp, GPR.r13)
 GPR8 = GPR[0:8]
-GPR8_DEREF_SAFE = GPR8.without(GPR.rsp)
-GPR8_ZERO_DEREF_SAFE = GPR8_DEREF_SAFE.without(GPR.rbp)
```
sunfishcode: \o/

filetests/regalloc/gpr-deref-safe-335.clif
```diff
@@ -0,0 +1,44 @@
...
+    brz v1963, ebb123(v1952, v1961)
+    jump ebb127
+
+ebb127:
+    v1966 = iconst.i32 0
+    jump ebb122(v1963, v1966, v1966, v1966, v1966, v1966, v1966, v1966, v1966, v1966)
+
+ebb122(v1967: i32, v1971: i32, v1972: i32, v1978: i32, v2032: i32, v2041: i32, v2053: i32, v2076: i32, v2085: i32, v2096: i32):
+    trap user0
+}
```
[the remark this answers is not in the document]
bnjbvr: The thing is, the test wasn't crashing if i didn't use `test compile`, so there's something else going on here. I'll investigate a bit more to see what causes this.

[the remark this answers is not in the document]
bnjbvr: Also, regarding testing, I've imagined that adding a big real-world test case like one of the wasm demos could be useful to make sure we're not breaking wasm support? Then we could run it on CI, and make sure it doesn't break, or at least warn when it does.

lib/codegen/meta-python/cdsl/isa.py
bnjbvr: > And, it is more intuitive to start with a base and add from there.

This +1. Also, it's kind of an edge case when there's something to add; the most generic case is when we don't add anything.

lib/codegen/meta-python/isa/x86/recipes.py
bnjbvr: It seems there's a variety of bodies, though: some of them want to pass 2 registers, some of them 3, which type will depend (sometimes GPR, sometimes FPR for loads), sometimes we want modrm_disp8, other times we want modrm_sib_disp8... So I'm not sure it would make it more readable or factor so much code. Does it make sense to keep the code like this?

filetests/regalloc/gpr-deref-safe-335.clif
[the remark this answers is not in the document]
bnjbvr: > The thing is, the test wasn't crashing if i didn't use test compile, so there's something else going on here. I'll investigate a bit more to see what causes this.

Just looking at the structure of CompileContext::run, I'm pretty sure this difference happens because compile runs optimizations, while regalloc doesn't.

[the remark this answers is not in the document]
sunfishcode: Can you dump the IR after the optimizations? Does `test regalloc` reproduce the problem if you run it on that IR?

lib/codegen/meta-python/isa/x86/recipes.py
sunfishcode: Ok. My main observation is that we kind of have two version of the same logic, one to compute the size, and one to encode the bytes (which effectively also computes the size), so I was looking for ways to help keep them in sync. However, we do have tests for this too, so if it's not straightforward to refactor something here, it's ok to leave it as is.

filetests/regalloc/gpr-deref-safe-335.clif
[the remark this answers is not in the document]
bnjbvr: Done in a subsequent push.

sunfishcode:
Looks good!

Issue #554, opened 2018-10-09
Tracking issue for constant folding


lachlansneff:
This is the tracking issue for minimal constant-folding support. The aim is to collapse code like
```clif
v0 = iconst.i32 5
v1 = iconst.i32 37
v2 = iadd v1, v0
```
into
```clif
v2 = iconst.i32 42
```

and

```clif
ebb0:
  v0 = bconst true
  brnz v0, ebb1
  ...
ebb1
  ...
```
into
```clif
ebb0:
  jump ebb1
  ...
ebb1
  ...
```

Progress:

- [ ] Add constant folding pass

- [ ] Collapse simple operations on constants.

- [ ] Collapse branches when branch can be determined at compile time.

PR #556, opened 2018-10-10
Add early-stage optimization crate

lachlansneff:
This pr adds a simple constant folding pass. Don't merge this yet, it's a work in progress and not near completion.

The tracking issue is #554.


lib/codegen/src/timing.rs
```diff
@@ -62,6 +62,7 @@ define_passes!{
     gvn: "Global value numbering",
     licm: "Loop invariant code motion",
     unreachable_code: "Remove unreachable blocks",
+    // constant_folding: "Fold constant expressions",
```
lachlansneff: Uncommenting this to add `constant_folding` pass timing will not compile. The array generated by the macro is currently 32 items in length, and a static slice of 33 items does not implement `Default`. When const generics land, this will be fixed, but not sure what to do about it before then.

lib/codegen/src/constant_folding.rs
```diff
@@ -0,0 +1,118 @@
...
+/// It's important to note that this will not remove unused constants. It's
+/// assumed that the DCE pass will take care of them.
+pub fn fold_constants(func: &mut ir::Function) {
+    // let _tt = timing::constant_folding();
+    let mut pos = FuncCursor::new(func);
+
+    while let Some(_ebb) = pos.next_ebb() {
+        while let Some(inst) = pos.next_inst() {
+            use ir::instructions::Opcode::*;
+            match pos.func.dfg[inst].opcode() {
```
sunfishcode: An organization that I think would be advantageous would be to match on the `InstructionFormat`, which is `pos.func.dfg[inst]`, rather than pulling out the opcode right away. Then there will be one match arm for all `ir::InstructionData::Binary` opcodes, and then we can have one `fold_numerical_binary` call and pass it the opcode. This seems appealing because it seems like it could let us keep the code for manipulating the IR separate from the folding arithmetic logic. If the ultimate folding function for binary operators could be a function which takes an opcode and two i64's and returns an i64 (or maybe an Option<i64> because trapping operators), that'd (a) make it really easy to write unit tests for the folding arithmetic, and (b) make it possible to reuse the folding arithmetic for... other fun things in the future, like IR interpreters, or auto-folding InstBuilders :-).

lib/codegen/src/context.rs
```diff
@@ -123,6 +124,10 @@ impl Context {
         let _tt = timing::compile();
         self.verify_if(isa)?;
 
+        if isa.flags().enable_constant_folding() {
+            self.fold_constants(isa)?;
+        }
```
sunfishcode: A concern that I have is that this may increase the size of the compiler for users that aren't using this. I don't know how big the constant folding pass will get, but if we're going to be getting into optimizing unoptimized code, we may be adding other things besides.

Since this is running at the beginning of `compile`, what if we moved it out into a separate top-level `optimize` function on `Context`? Users that know they have unoptimized code could call `optimize` before `compile`, and for users that don't call `optimize`, it could get DCE'd.

lib/codegen/src/timing.rs
sunfishcode: Huh. Not sure what to do yet either.

lib/codegen/src/context.rs
lachlansneff: Okay, that sounds good to me. That would entail removing the `constant_folding` setting as well, right?

lib/codegen/src/timing.rs
lachlansneff: Figured it out. Just had to coerce them to an dynamically sized slice.

lib/codegen/src/constant_folding.rs
lachlansneff: Oh boy, didn't even think about an auto-folding `InstBuilder`. That would be super cool, and would probably remove most of the overhead of a constant folding pass.

lib/codegen/src/context.rs
sunfishcode: Yes, that's right.

lib/codegen/src/constant_folding.rs
```diff
@@ -0,0 +1,240 @@
...
+    match dfg[inst] {
+        UnaryImm {
+            opcode: Iconst,
+            imm,
+        } => Some(ConstImm::I64(imm.into())),
+        UnaryIeee32 {
+            opcode: F32const,
+            imm,
+        } => {
+            let imm_as_f32 = f32::from_bits(imm.bits()); // see https://doc.rust-lang.org/std/primitive.f32.html#method.from_bits for caveats
```
bjorn3: Could you put the comment before this line. This line is a bit long.

lib/codegen/src/constant_folding.rs
```diff
@@ -0,0 +1,240 @@
...
+                if imm1 == 0.0 {
+                    panic!("Cannot divide by a zero.")
+                }
+                Some(ConstImm::Ieee64(imm0 / imm1))
+            }
+            _ => unreachable!(),
+        },
+        // ir::Opcode::Fadd => Some(imm0.unwrap)
+        _ => None,
+    }
```
sunfishcode: Yes, this looks like a good overall direction. To further separate the arithmetic from the IR walking, we can pull just this match statement out into a separate function in a separate file.

lib/codegen/src/constant_folding.rs
```diff
@@ -0,0 +1,239 @@
...
+        ir::Opcode::Imul => {
+            let imm0 = Wrapping(imm0.unwrap_i64());
+            let imm1 = Wrapping(imm1.unwrap_i64());
+            Some(ConstImm::I64((imm0 * imm1).0))
+        }
+        ir::Opcode::Udiv => {
+            let imm0 = Wrapping(imm0.unwrap_i64());
+            let imm1 = Wrapping(imm1.unwrap_i64());
+            if imm1.0 == 0 {
+                panic!("Cannot divide by a zero.")
```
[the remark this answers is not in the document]
lachlansneff: Doesn't the verifier already deny division by a `0` immediate?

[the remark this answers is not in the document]
sunfishcode: That's an [open question](https://github.com/CraneStation/cranelift/pull/484#issuecomment-416721928), however this patch also has code that resolves Values defined by iconst, which is entirely valid.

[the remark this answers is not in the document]
lachlansneff: Oh, I thought that had gotten merged. I'll make it return None.

lib/codegen/src/constant_folding.rs
sunfishcode: [preopt](https://github.com/CraneStation/cranelift/blob/master/lib/codegen/src/preopt.rs#L447) does this. So it's not something we need here right away. However if this constant folding pass starts growing into a more general optimization framework, it may want to start doing that kind of thing too, at which point we should think about how to organize things.

lib/codegen/src/constant_folding.rs
lachlansneff: This got replaced by software floats.

lib/codegen/src/lib.rs
```diff
@@ -63,6 +63,8 @@ extern crate target_lexicon;
 #[macro_use]
 extern crate log;
 
+extern crate rustc_apfloat;
```
lachlansneff: `rustc_apfloat` uses the `u128` type, which is not stable on rust 1.25.0, which cranelift attempts to support.

lachlansneff:
I'm going to remove the soft float crate from this pr until we can get it to build on rust 1.25.0.

lib/preopt/src/constant_folding.rs
```diff
@@ -0,0 +1,269 @@
...
+    match dfg[inst] {
+        UnaryImm {
+            opcode: Iconst,
+            imm,
+        } => Some(ConstImm::I64(imm.into())),
+        UnaryIeee32 {
+            opcode: F32const,
+            imm,
+        } => {
+            // see https://doc.rust-lang.org/std/primitive.f32.html#method.from_bits for caveats
```
sunfishcode: Please captalize and emperiod comments which are complete sentences.

lib/preopt/src/constant_folding.rs
```diff
@@ -0,0 +1,269 @@
...
+            let ieee_f32 = f32::from_bits(imm.bits());
+            Some(ConstImm::Ieee32(ieee_f32))
+        }
+        UnaryIeee64 {
+            opcode: F64const,
+            imm,
+        } => {
+            // see https://doc.rust-lang.org/std/primitive.f32.html#method.from_bits for caveats
```
sunfishcode: ditto

lib/filetests/src/lib.rs
```diff
@@ -126,6 +128,7 @@ fn new_subtest(parsed: &TestCommand) -> subtest::SubtestResult<Box<subtest::SubT
         "shrink" => test_shrink::subtest(parsed),
         "simple-gvn" => test_simple_gvn::subtest(parsed),
         "verifier" => test_verifier::subtest(parsed),
+        "optimize" => test_optimize::subtest(parsed),
```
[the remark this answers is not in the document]
sunfishcode: Rename it to "simple_preopt", to match the new name of the thing it tests :-).

lib/preopt/src/constant_folding.rs
```diff
@@ -0,0 +1,269 @@
...
+
+    if (branch_if_zero && !truthiness) || (!branch_if_zero && truthiness) {
+        func.dfg.replace(inst).jump(ebb, &args);
+        // remove the rest of the ebb to avoid verifier errors
+        while let Some(next_inst) = func.layout.next_inst(inst) {
+            func.layout.remove_inst(next_inst);
+        }
+    } else {
+        let mut pos = FuncCursor::new(func).at_inst(inst);
+        pos.remove_inst_and_step_back();
```
nikic: To avoid skipping over one instruction, should this operator on the `FuncCursor` from `fold_constants`?

sunfishcode:
BTW, #569 has now landed, so the Rust 1.25 restriction is now lifted!

sunfishcode:
This looks good! Let's land this now, and we can continue to iterate from there.

PR #557, opened 2018-10-15
Revive the -T aka --time-passes argument to report run times on the CLI

bnjbvr:


sunfishcode:
Thanks for fixing this!

PR #561, opened 2018-10-18
Provide optimised codegen for small libc calls

### LINKED ISSUE #542: Optimize `call_memset` etc. to handle small aligned fixed-size cases
`FunctionBuilder`'s new `call_memcpy` currently always generates a call to "memcpy", which is correct, however it's very common in Rust code to need to make copies of small objects with known sizes and alignments, which can be more quickly done with a handful of loads and stores, rather than making a call.

We should introduce additional `FunctionBuilder` functions, with names along the lines of `emit_small_memcpy`, which also take source and destination pointer `Value`s, but instead of a length `Value`, have a `u64` specifying the size in bytes. It should also take two additional `u8` parameters, specifying the alignments in bytes of the source and destination pointers (and required to be powers of two).

For the first step, we can just ignore the alignments, put the size into a register using `iconst`, and then use `call_memcpy`. That will be the conservative fallback.

The next step is the main load/store optimization.
 - First, determine the access type for the loads and stores.
    - The alignments for the source and destination will often be the same, particularly in the case of struct copying, so for now we can focus on that case. So we start by computing the minimum of the source and destination values.
    - Then compute the greatest power of two that evenly divides the size and that's at most that minimum alignment value. That will be the size of the loads and stores we emit.
    - Next, divide the length by that size to determine how many loads and stores we'll need. If it's "too many", then we should use the conservative fallback rather than emitting a huge number of instructions. For now, let's say "too many" is 4 loads and 4 stores (this is a wild guess, but we canl tune it later).
    - Finally, emit loads and stores using, picking an integer type based on the access size we computed. We can use the constant offset fields of `load` and `store` so that we can always just use the base source and destination `Value`s without having to do address arithmetic. And we should interleave the loads and stores (load, store, load store, rather than load, load, store store) since that allows the register allocator to reuse a register rather than requiring all the values get their own register.

That's `memcpy`. `memmove` is similar, but the source and destination can overlap. So for memmove, we *don't* want to interleave the loads and stores. If we do all the loads first, then all the stores second, we correctly handle the overlap case. We'll need more registers that way, but that's the cost of using memmove instead of memcpy :-).

Then, `memset` is similar, except there, we should replace the byte `Value` operand with a `u8` value. We can then materialize it into an `iconst` with the value (c | (c << 8) | (c << 16) ...) out to the destination alignment width, and then do stores with that.

Aaronepower:
resolves #542


lib/frontend/src/frontend.rs
```diff
@@ -614,8 +684,74 @@ impl<'a> FunctionBuilder<'a> {
...
+    let mut greatest_power = 2u64;
+    let mut next_power = greatest_power.next_power_of_two();
+
+    while max > next_power {
+        if (multiple / next_power) % 2 == 0 {
+            greatest_power = next_power;
+        }
+
+        next_power = next_power.next_power_of_two();
     }
```
sunfishcode: Power-of-two multiples are interesting because when you multiply a number by a power of 2, it's effectively just shifting the number left, inserting zeros at the low end. The number of zeros inserted is equal to the number of trailing zeros in the power of 2.

So instead of iterating here, we can use `trailing_zeros` to find the number of trailing zeros in the size. Its greatest power-of-two divisor is then the power of two with that many trailing zeros, or a one left-shifted by that number.

lib/frontend/src/frontend.rs
```diff
@@ -571,10 +571,57 @@ impl<'a> FunctionBuilder<'a> {
...
+    /// Optimised memcpy for small copys.
+    pub fn emit_small_memcpy(&mut self,
+                             isa: &TargetIsa,
+                             dest: Value,
+                             src: Value,
+                             size: u64,
+                             dest_align: u8,
+                             src_align: u8)
+    {
+        const THRESHOLD: u64 = 4;
```
sunfishcode: Please add a comment mentioning that this size is just a guess, not based on experiments.

lib/frontend/src/frontend.rs
```diff
@@ -571,10 +571,57 @@ impl<'a> FunctionBuilder<'a> {
...
+            let access_size = greatest_divisible_power_of_two(min, size);
+            let load_and_store_amount = size / access_size;
+
+            if load_and_store_amount > THRESHOLD {
+                fallback!();
+                return
+            }
+
+            let flags = MemFlags::new();
+            flags.aligned();
```
sunfishcode: This should be `set_aligned()`.

lib/frontend/src/frontend.rs
```diff
@@ -571,10 +571,57 @@ impl<'a> FunctionBuilder<'a> {
...
+
+            for i in 0..load_and_store_amount {
+                let offset = (access_size * i) as i32;
+                let value = self.ins().load(types::I8, flags, src, offset);
```
sunfishcode: The type for the load here should be an integer type the size of `access_size`. You can use `Type::int` to get such a type.

lib/frontend/src/frontend.rs
```diff
@@ -591,14 +638,37 @@ impl<'a> FunctionBuilder<'a> {
...
+        self.ins().call(libc_memset, &[buffer, ch, size]);
+    }
+
+    /// Calls libc.memset
+    ///
+    /// Writes `size` bytes of value `ch` to memory starting at `buffer`.
+    pub fn emit_small_memset(&mut self,
+                             isa: &TargetIsa,
+                             buffer: Value,
+                             ch: Value,
```
sunfishcode: I don't think I mentioned this yet, but for the small memset function, we actually want the `ch` to be passed in as a constant too, since the main use for this function is just zeroing out memory.

Also, we want this function to take an alignment parameter similar to memcpy for its destination pointer, and we can compute an access size from it in a similar way. And in that case we'll need to convert `ch` from a single byte to an access-size-sized integer in which every byte is `ch` :).

Aaronepower:
@sunfishcode Alright, I've implemented those changes, let me know if there is anything else!

Aaronepower:
@sunfishcode I've implemented that feedback.  I also changed how the greatest divisible power of two logic based on this [stackoverflow answer](https://stackoverflow.com/questions/1551775/calculating-highest-power-of-2-that-evenly-divides-a-number-in-c), which doesn't require a starting number so I added an assert so that the alignments are always at lest smaller, let me know if that's the right thing to do or if it's completely off the mark.

sunfishcode:
Cool, this looks really good! The CI is red due to formatting diffs; would you mind running rustfmt on this PR?

Aaronepower:
@sunfishcode Done!

sunfishcode:
Thanks!

PR #562, opened 2018-10-19
Add a `readonly` flag for loads

### LINKED ISSUE #366: Add a `readonly` flag for loads

lachlansneff:
resolves #366

### DIFF
--- a/lib/codegen/src/simple_gvn.rs
+++ b/lib/codegen/src/simple_gvn.rs
@@ -36,7 +36,7 @@ fn is_load_and_not_readonly(inst_data: &InstructionData) -> bool {
             flags,
             offset: _,
         } => !flags.readonly(),
-        _ => false,
+        _ => inst_data.opcode().can_load(),
     }
 }
 
--- a/lib/codegen/src/simple_gvn.rs
+++ b/lib/codegen/src/simple_gvn.rs
@@ -18,8 +18,26 @@ fn trivially_unsafe_for_gvn(opcode: Opcode) -> bool {
         || opcode.can_trap()
         || opcode.other_side_effects()
         || opcode.can_store()
-        || opcode.can_load()
         || opcode.writes_cpu_flags()
+}
+
+/// Test that, if the specified instruction is a load, it doesn't have the `readonly` memflag.
+fn is_load_and_not_readonly(inst_data: &InstructionData) -> bool {
+    match inst_data {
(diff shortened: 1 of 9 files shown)


lib/codegen/src/simple_gvn.rs
```diff
@@ -90,18 +99,23 @@ pub fn do_simple_gvn(func: &mut Function, domtree: &mut DominatorTree) {
...
             if opcode.is_branch() && !opcode.is_terminator() {
                 scope_stack.push(func.layout.next_inst(inst).unwrap());
                 visible_values.increment_depth();
             }
+
             if trivially_unsafe_for_gvn(opcode) {
                 continue;
+            } else if is_load_and_not_readonly(&inst_data) {
+                continue;
             }
```
sunfishcode: This pattern is easier to read of we omit the else because the previous if ends with continue. And also, splitting up the two ifs means we can move the inst_data clone() down after the first if. This is a dubious micro-optimization, however I'm expecting we'll proceed to put more code here and I want to keep "check the opcode" as the clearly independent first step, because that quickly eliminates a lot of instructions from further consideration.

PR #563, opened 2018-10-19
Insert copies to support a value being used as multiple return values.

sunfishcode:
When one value is used multiple times for separate return values, we
need to copy it to produce a new value, so that each value can be
allocated a different register.

### DIFF
--- a/lib/codegen/src/regalloc/spilling.rs
+++ b/lib/codegen/src/regalloc/spilling.rs
@@ -348,6 +348,11 @@ impl<'a> Context<'a> {
         // Similarly, for return instructions, collect uses of ABI-defined
         // return values.
         if self.cur.func.dfg[inst].opcode().is_return() {
+            debug_assert_eq!(
+                self.cur.func.dfg.inst_variable_args(inst).len(),
+                self.cur.func.signature.returns.len(),
+                "The non-fixed arguments in a return should follow the function's signature."
+            );
             for (ret_idx, (ret, &arg)) in
                 self.cur.func.signature.returns.iter().zip(args).enumerate()
             {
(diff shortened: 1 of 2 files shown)


lib/codegen/src/regalloc/spilling.rs
```diff
@@ -335,6 +344,29 @@ impl<'a> Context<'a> {
...
             }
         }
+
+        // Similarly, for return instructions, collect uses of ABI-defined
+        // return values.
+        if self.cur.func.dfg[inst].opcode().is_return() {
+            for (ret_idx, (ret, &arg)) in
+                self.cur.func.signature.returns.iter().zip(args).enumerate()
+            {
+                let idx = constraints.ins.len() + ret_idx;
```
bnjbvr: This is a bit unfortunate, because if I understand correctly, this breaks encapsulation: the layout of the arguments vector of the return instruction in particular is hereby exposed. (I've seen this index is used in `process_reg_uses` as an index into the arguments vector) Maybe there's a way to encapsulate it a bit better? (could be done as a follow-up)

sunfishcode: The loop above this is iterating over the instruction args zipped with `constraints.ins.iter()`, so it already knows about the layout of the arguements vector. By starting at `constraints.ins.len()`, we're effectively just picking up the arguements where the previous loop left off.

More generally, we make these assumptions in a lot of places:
 - fixed arguments are described by the opcode constraints, variable arguments are not
 - variable arguments appear after the fixed arguments

and I think these are reasonable, though I'm open to discussing other approaches.

I have added an assert to the PR to at least check that the `return` has the number of return values the signature says it should, which should help catch any cases where things get out of sync.

bnjbvr: OK, I wasn't aware of this but this comment sheds light on this invariant, so that's good enough to me. Thanks!

bnjbvr:
Thanks!

PR #569, opened 2018-10-24
Update the minimum supported Rust version to 1.29.

sunfishcode:
We no longer need the Ubuntu LTS restriction, so now the only only
constraint I'm aware of is Firefox's policy. Fortunately, that tracks
the latest stable delayed by only two weeks. So this puts is at
Rust 1.29 now.

If anyone is aware of other important constraints that we should follow,
please bring them up!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,9 +2,9 @@
 
 language: rust
 rust:
-    - stable
-    # The version of rust in the latest Ubuntu LTS, currently Bionic.
-    - 1.25.0
+    # The oldest version we currently support. See
+    # CONTRIBUTING.md#rustc-version-support for details.
+    - 1.29.0
     - beta
     - nightly
 matrix:
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -75,17 +75,14 @@ which may be convenient when there are multiple versions installed.
 
 ### Rustc version support
 
-Our current policy is to support the version of Rustc that ships with the
-latest Ubuntu LTS release, as well as the current stable version. This means
-we don't use some of the very latest released Rust features.
+Cranelift supports stable Rust, and follows the
+[Rust Update Policy for Firefox].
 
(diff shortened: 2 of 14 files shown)


.travis.yml
```diff
@@ -2,9 +2,9 @@
 
 language: rust
 rust:
-    - stable
```
bjorn3: Why did you remove stable?

bjorn3:
```
[…]
======  Checking python source files  ======
   === flake8: 3.6.0 (mccabe: 0.6.1, pycodestyle: 2.4.0, pyflakes: 2.0.0) CPython 3.4.3 on Linux ===
./base/instructions.py:1145:-316: W605 invalid escape sequence '\p'
./cdsl/instructions.py:349:-5: W605 invalid escape sequence '\s'
./cdsl/test_ti.py:111:-405: W605 invalid escape sequence '\i'
./cdsl/ti.py:435:-296: W605 invalid escape sequence '\ '
./cdsl/ti.py:435:-29: W605 invalid escape sequence '\ '
./cdsl/typevar.py:124:-28: W605 invalid escape sequence '\i'
./cdsl/xform.py:119:-129: W605 invalid escape sequence '\i'
./semantics/__init__.py:27:-254: W605 invalid escape sequence '\i'
./semantics/__init__.py:27:-71: W605 invalid escape sequence '\i'
[…]
```

.travis.yml
sunfishcode: We don't want to immediately update to stable when new versions are released because we're following [Firefox's policy](https://wiki.mozilla.org/Rust_Update_Policy_for_Firefox).

sunfishcode:
A new version of flake8 added some new warnings. These are now fixed.

lachlansneff:
This is great! Once this lands, I'll update preopt to use rustc_apfloat.

sunfishcode:
Ok, let's land this!

PR #570, opened 2018-10-26
Introduce a `TargetFrontendConfig` type.

### LINKED ISSUE #529: Add method to Module to get TargetIsa

### LINKED ISSUE #555: call_conv shouldn't be a setting
call_conv is currently defined as a setting in lib/codegen/meta-python/base/settings.py. This specifies the default calling convention (which can be overridden as needed). However, it should be possible to infer a default calling convention from the target `Triple`, so we shouldn't need an independent setting for it.

We should remove the setting, which will remove the generated definition of `CallConv`, and instead add a plain `CallConv` enum in lib/codegen/src/ir/extfunc.rs. Then we should add a `call_conv()` function (or perhaps call it `default_call_conv()` to be more descriptive), defined in the `TargetIsa` trait, which returns a `CallConv` based on the value of `self.triple().operating_system()` -- Windows uses `WindowsFastcall`, and I believe all the other currently known operating systems use `SystemV`.

sunfishcode:
`TargetFrontendConfig` is information specific to the target which is
provided to frontends to allow them to produce Cranelift IR for the
target. Currently this includes the pointer size and the default calling
convention.

The default calling convention is now inferred from the target, rather
than being a setting. cranelift-native is now just a provider of target
information, rather than also being a provider of settings, which gives
it a clearer role.

And instead of having cranelift-frontend routines require the whole
`TargetIsa`, just require the `TargetFrontendConfig`, and add a way to
get the `TargetFrontendConfig` from a `Module`.

Fixes #529.
Fixes #555.

### DIFF
--- a/lib/frontend/src/frontend.rs
+++ b/lib/frontend/src/frontend.rs
@@ -9,7 +9,7 @@ use cranelift_codegen::ir::{
     JumpTable, JumpTableData, LibCall, MemFlags, SigRef, Signature, StackSlot, StackSlotData, Type,
     Value,
 };
-use cranelift_codegen::isa::TargetIsa;
+use cranelift_codegen::isa::{TargetFrontendConfig, TargetIsa};
 use cranelift_codegen::packed_option::PackedOption;
 use ssa::{Block, SSABuilder, SideEffects};
 use std::vec::Vec;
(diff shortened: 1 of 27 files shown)


nbp:
Doesn't that prevent the idea of having a simulator as done in SpiderMonkey / v8, if we cannot switch the `TargetIsa` and only use the native one? Or we should change the compilation of `cranelift-native` based on the targeted `ISA`?

sunfishcode:
cranelift-codegen lets you specify whatever target you want at runtime. cranelift-native is a separate crate, which provides a way to identify "the current machine", which is handy in simple use cases. We made cranelift-native a separate optional crate specifically to keep it isolated from the rest of the crates.

As it happens, we don't currently use cranelift-native [in SpiderMonkey](https://dxr.mozilla.org/mozilla-central/source/js/src/wasm/cranelift/Cargo.toml#10). However even if we did, it would be straightforward to have an option to use a different target when needed.

lib/frontend/src/frontend.rs
```diff
@@ -550,10 +550,16 @@ impl<'a> FunctionBuilder<'a> {
     /// won't overlap onto `dest`. If `dest` and `src` overlap, the behavior is
     /// undefined. Applications in which `dest` and `src` might overlap should
     /// use `call_memmove` instead.
-    pub fn call_memcpy(&mut self, isa: &TargetIsa, dest: Value, src: Value, size: Value) {
-        let pointer_type = isa.pointer_type();
+    pub fn call_memcpy(
+        &mut self,
+        config: &TargetFrontendConfig,
```
bjorn3: Why is this a reference?

PR #571, opened 2018-10-27
SimpleJit perf map

bjorn3:
Not yet tested

https://lkml.org/lkml/2009/6/8/499

### DIFF
--- a/lib/simplejit/src/backend.rs
+++ b/lib/simplejit/src/backend.rs
@@ -11,6 +11,7 @@ use libc;
 use memory::Memory;
 use std::collections::HashMap;
 use std::ffi::CString;
+use std::io::Write;
 use std::ptr;
 use target_lexicon::PointerWidth;
 #[cfg(windows)]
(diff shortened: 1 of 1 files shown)


sunfishcode:
Cool! Would it make sense to make this conditional under `#[cfg(target_os = "linux")]` and the "PERF_BUILDID_DIR" environment variable being set at runtime?

bjorn3:
Tested

bjorn3:
Build fails for rust 1.25, because process::id is not yet stabilised there. What should I do?

sunfishcode:
#569 just landed, so we no longer need to support Rust 1.25 :-).

bjorn3:
can you restart travis?

sunfishcode:
I restarted it, but I think you need to rebase on master for it to pick up the Rust-1.29 change.

bjorn3:
> but I think you need to rebase on master

Of course, done

sunfishcode:
Looks good!

PR #572, opened 2018-10-27
Add SourceLocSink

bjorn3:
This allows backends to implement line debug info.


bjorn3:
You can access the `Context` after calling `define_function` to get the same information, so closing.

Issue #550, opened 2018-10-04
Eliminate callstack recursion in `write_value_aliases`


sunfishcode:
Cranelift generally avoids using callstack recursion, to avoid stack overflow. This is useful even in IR printing code as it may be called from debugging or logging scenarios. So it would be good to convert `write_value_aliases` in lib/code/src/write.rs to use a `Vec` for a stack.

mominul:
Mentoring instructions please! :smiley:

sunfishcode:
The first step is to familiarize yourself with what `write_value_aliases` does. One way to start here would be to take a look at the testcase filetests/parser/alias.clif. You can run `clif-util cat filetests/parser/alias.clif` to see `write_value_aliases` in action. The %transitive function in that file gives an example of a case that makes use of interesting recursion, because we want to print all the aliases of v0, which includes v1, which is then aliased by v2, which is then aliased by v3, and so on.

Then can look at how it does this. Currently `write_value_aliases` achieves this by calling itself. Notice that when a function calls itself, it's essentially pushing a new frame of the function's variables on the stack, so that when it returns and this frame is popped, it goes back to the previous frame. So what we want to do is convert the local variable state of the function into a struct type that can be stored in a `Vec` so that we can push new entries (replacing the function calling itself) and pop them (analogous to the function returning).

Does that make sense?

sunfishcode:
This is now fixed in #573. I apologize, I didn't notice that that was submitted by someone else. Let me know if there are any areas you're interested in working on that I can help mentor with!

mominul:
:heart:

PR #573, opened 2018-10-28
use iterative rather than recursive method for following aliases

Giesch:
* this avoids consuming stack via plain recursive calls

addresses #550 

I feel like there's probably a more idiomatic way to do this, but factoring out the for loop seemed like it'd have to be verbose b/c of borrowing the vec.

Should there be a unit test for this fn or is the write::tests::aliases test enough?

### DIFF
--- a/lib/codegen/src/write.rs
+++ b/lib/codegen/src/write.rs
@@ -283,10 +283,21 @@ fn write_value_aliases(
     target: Value,
     indent: usize,
 ) -> fmt::Result {
+    let mut todo_stack = vec![];
+
     for &a in &aliases[target] {
         writeln!(w, "{1:0$}{2} -> {3}", indent, "", a, target)?;
-        write_value_aliases(w, aliases, a, indent)?;
+        todo_stack.push(a);
     }
+
+    while !todo_stack.is_empty() {
+        let target = todo_stack.pop().unwrap();
+        for &a in &aliases[target] {
+            writeln!(w, "{1:0$}{2} -> {3}", indent, "", a, target)?;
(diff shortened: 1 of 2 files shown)


lachlansneff:
I curious whether the original, recursive version of this has better or worse performance. I have a hunch that the cache locality characteristics of the call stack would do better than the that of a vector. I could be wrong though.

lib/codegen/src/write.rs
```diff
@@ -283,10 +283,21 @@ fn write_value_aliases(
     target: Value,
     indent: usize,
 ) -> fmt::Result {
+    let mut todo_stack = vec![];
```
bjorn3: Cleaner version:

```rust
    let mut todo_stack = vec![target];
    while let Some(target) = todo_stack.pop() {
        for &a in &aliases[target] {
            writeln!(w, "{1:0$}{2} -> {3}", indent, "", a, target)?;
            todo_stack.push(a);
        }
    }

    Ok(())
```

Giesch:
Using easybench, it looks like it's pretty close. I don't know how accurate that is for cache related things.

With this test:
https://github.com/Giesch/cranelift/commit/7697d2daf9de24b1a2a2083fcede9d3f455a52a8

running:
```cargo test --release tests::bench_recursive --manifest-path lib/codegen/Cargo.toml -- --nocapture```

I get:
```original: 991us 610ns (R²=1.000, 789 iterations in 45 samples)```
```new: 932us 313ns (R²=1.000, 789 iterations in 45 samples)```

Setting n = 10_000 with the original version also blows the stack for me.

PR #575, opened 2018-10-29
Make regalloc visit fallthrough_return instructions.

sunfishcode:
Add an explicit "is_ghost" property to selected instructions, and use
that to determine whether reload and coloring should visit instructions.
This allows them to visit fallthrough_return instructions and insert
fills and register moves as needed.

### DIFF
--- a/lib/codegen/meta-python/cdsl/instructions.py
+++ b/lib/codegen/meta-python/cdsl/instructions.py
@@ -92,6 +92,7 @@ class Instruction(object):
     :param is_indirect_branch: This is an indirect branch instruction.
     :param is_call: This is a call instruction.
     :param is_return: This is a return instruction.
+    :param is_ghost: This is a ghost instruction.
     :param can_trap: This instruction can trap.
     :param can_load: This instruction can load from memory.
     :param can_store: This instruction can store to memory.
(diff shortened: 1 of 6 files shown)


lib/codegen/meta-python/cdsl/instructions.py
```diff
@@ -108,6 +109,7 @@ class Instruction(object):
             'True for all indirect branch or jump instructions.',
             'is_call': 'Is this a call instruction?',
             'is_return': 'Is this a return instruction?',
+            'is_ghost': 'Is this a ghost instruction?',
```
bnjbvr: It'd be nice to precise what it means: my understanding is that it's an instruction with an illegal encoding that should not be visited by register allocation. (fallthrough_return has an illegal encoding but must be visited -- are there any other ones like fallthrough_return?)

bnjbvr:
Also, please update the commit message so it doesn't include "unreachable_return" anymore :). Thanks!

sunfishcode:
Yes, your understanding is exactly correct.

sunfishcode:
Ok, I've now updated the commit message, added a verifier check that ghost instructions don't have an encoding, and added better comments explaining what ghost instructions are.

bnjbvr:
Great, thanks!

PR #576, opened 2018-10-29
When lowering br_table to branches, delete the old jump tables.

sunfishcode:
This prevents codegen from emitting unused readonly data.


lib/codegen/src/legalizer/mod.rs
```diff
@@ -110,6 +110,11 @@ pub fn legalize_function(func: &mut ir::Function, cfg: &mut ControlFlowGraph, is
             }
         }
     }
+
+    // Now that we've lowered all br_tables, we don't need the jump tables anymore.
+    if !isa.flags().jump_tables_enabled() {
+        pos.func.jump_tables.clear();
```
bnjbvr: Correct me if I'm wrong, but regalloc happens after legalization, and `is_live_on_outgoing_edge` in coloring.rs uses the `func.jump_tables` information even if jump tables aren't enabled (`analyze_branch` in this code can find a `BranchTable` instruction). So this doesn't seem safe to do?

Instead, could we just not emit the readonly table data if jump tables are disabled?

bnjbvr: It seems the branch relaxation pass uses this information too

sunfishcode: When jump tables are disabled, we lower all br_tables into sequences of conditional branches, such that no `BranchTable` instructions remain after legalization.

bnjbvr: Ah right, I didn't see the `remove_inst()` call in this function (and/or thought it was applying to some other instructions that was inserted in the meanwhile, but no, the position remains the same after the creation of the cursor). Interestingly, even the other function (that expands to a jump table) removes the `BranchTable` one after usage (replacing it with BranchTableBase/BranchTableEntry/IndirectJump, as one could expect).

bnjbvr: So this makes me think the code I mentioned in the first comment is unreachable. If I cause the code to panic in this branch, all the tests keep on passing, but I'm not sure our test coverage makes it so that we can be confident that it's actually unreachable; will wait for the next SM update to confirm. Anyway, we can do this later.

bnjbvr:
LGTM, thanks!

PR #577, opened 2018-10-29
Make `FuncEnvironment::return_mode()` default to `NormalReturns`.

sunfishcode:
`NormalReturns` is what we expect most users will normally want.


bnjbvr:
Makes sense, thanks!

PR #579, opened 2018-10-29
Split the default edge of a br_table.

sunfishcode:
When splitting critical edges for a br_table to handle arguments being
passed, split the default edge along with the normal table edges.


bnjbvr:
Makes sense, thanks!

PR #581, opened 2018-10-29
Use `Display` rather than having an explicit `name()` function for types.

sunfishcode:
This is more idiomatic Rust.


bnjbvr:
Sweet, thanks.

PR #585, opened 2018-11-01
Prefix verifier errors with ;

JoeHowarth:
extract arrow drawing code into helper
print ; at beginning of line followed by arrow and error body

### DIFF
--- a/lib/codegen/src/print_errors.rs
+++ b/lib/codegen/src/print_errors.rs
@@ -29,7 +29,8 @@ pub fn pretty_verifier_error<'a>(
         &mut w,
         func,
         isa,
-    ).unwrap();
+    )
+    .unwrap();
     w
 }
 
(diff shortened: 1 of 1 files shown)


lib/codegen/src/print_errors.rs
```diff
@@ -152,6 +144,16 @@ fn pretty_preamble_error(
     Ok(())
 }
 
+/// prints ^~~~~~ verifier [ERROR BODY]
+fn print_error(w: &mut Write, indent: usize, s: String, err: VerifierError) -> fmt::Result {
+    write!(w, ";{1:0$}^", indent - 1, "")?;
```
sunfishcode: I don't think the indent is ever zero in the current verifier, so it's fine to not support that, but it'd be good to have an assert checking that.

lib/codegen/src/print_errors.rs
```diff
@@ -152,6 +144,16 @@ fn pretty_preamble_error(
     Ok(())
 }
 
+/// prints ^~~~~~ verifier [ERROR BODY]
```
sunfishcode: Minor style nit: Please capitalize sentence-like comments.

sunfishcode:
Cool, thanks!

PR #590, opened 2018-11-08
Renamings

bnjbvr:
This removes one warning seen when compiling on Linux, and renames `fixed_results / fixed_value_arguments` to the same names prefixed with `num_` to make clear they return a number (and not a slice/array of what they describe).


sunfishcode:
Makes sense!

PR #592, opened 2018-11-08
Build: Introduce a TargetIsaBuilder to better encapsulate the TargetIsa

bnjbvr:
It was the caller's responsibility to call TargetIsa::check() before;
now one can't manipulate a TargetIsa without calling the
TargetIsaBuilder::finish() method, which is less error-prone and more in
line with what's coming for other things we're going to generate in the
meta crate.

(SettingGroup will require a similar finish() method that computes the layout of the byte vector, for instance)


bnjbvr:
Sorry for the churn, found a way to make this even better by splitting the construction of the RegClass in two parts. No more awkward passing of the `TargetIsa` to the `RegClass` ctors anymore! That's how it should have been in the first place.

bnjbvr:
Sneaking an extra commit that removes dead code in the Python code generator and translates Python assertions that were there to Rust.

PR #593, opened 2018-11-09
Fix bit count bug in emit_small_{memcpy, memset}.

Aaronepower:
Also moved creating the `int_type` in `emit_small_memcpy` from outside the loop.

### DIFF
--- a/lib/frontend/src/frontend.rs
+++ b/lib/frontend/src/frontend.rs
@@ -605,12 +605,12 @@ impl<'a> FunctionBuilder<'a> {
             return;
         }
 
+        let int_type = Type::int((access_size * 8) as u16).unwrap();
         let mut flags = MemFlags::new();
         flags.set_aligned();
 
         for i in 0..load_and_store_amount {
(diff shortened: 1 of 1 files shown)


lib/frontend/src/frontend.rs
```diff
@@ -605,12 +605,12 @@ impl<'a> FunctionBuilder<'a> {
             return;
         }
 
+        let int_type = Type::int((access_size * 8) as u16).unwrap();
```
bnjbvr: Light suggestion: should we rename `size` and `access_size` to `byte_size` and `access_byte_size` to make this clearer? (maybe for callers of `emit_small_memcpy` too)

sunfishcode: I'm sympathetic, but my sense is it's not super important in this context, because the memcpy/memset/etc. interfaces are pretty well known.

Aaronepower:
@sunfishcode Okay I've added the test.

sunfishcode:
Cool, thanks!

PR #594, opened 2018-11-09
Make spilling visit fallthrough_return instructions too.

sunfishcode:
This is a followup to af2a952aabd82cf401cc664d0262b139ff92d86b. It
teaches the spilling pass to use the is_ghost() property to test whether
to visit instructions. This fixes a bug handling multiple return values
with fallthrough_return.


bnjbvr:
Makes sense, thanks!

PR #598, opened 2018-11-12
Licm fixes

sunfishcode:
This fixes 2 bugs in the LICM pass, fixing the problem reported [here](https://github.com/bjorn3/rustc_codegen_cranelift/issues/133#issuecomment-437687747).

LICM wasn't creating dedicated preheader blocks in the case where the existing preheader has other successors, and when LICM creates a preheader, it wasn't assigning an encoding.


lib/codegen/src/licm.rs
```diff
@@ -90,11 +90,18 @@ fn create_pre_header(
...
     }
     {
-        let mut pos = EncCursor::new(func, isa).at_top(header);
+        let mut pos = FuncCursor::new(func).at_top(header);
         // Inserts the pre-header at the right place in the layout.
         pos.insert_ebb(pre_header);
         pos.next_inst();
-        pos.ins().jump(header, pre_header_args_value.as_slice(pool));
+        let jump = pos.ins().jump(header, pre_header_args_value.as_slice(pool));
+        if !pos.func.encodings.is_empty() {
```
bnjbvr: Why wasn't the EncCursor enough here? (In any case, this hunk could be moved to the previous commit)

lib/codegen/src/licm.rs
```diff
@@ -111,21 +118,28 @@ fn has_pre_header(
...
         if !domtree.dominates(header, last_inst, layout) {
-            if found {
+            if result.is_some() {
                 // We have already found one, there are more than one
                 return None;
-            } else {
-                result = Some((pred_ebb, last_inst));
-                found = true;
             }
+            if last_inst != layout.last_inst(pred_ebb).unwrap() {
```
bnjbvr: Pre-existing: I think `last_inst` is a bad name (I was wondering why the last instruction wouldn't be equal to the last instruction!); maybe we could rename it `branch_inst`, or `pred_branch_inst` to make it clear that it's the actual branch instruction that jumps into the header?

Also, isn't this test redundant with the one checking the number of successors below?

PR #602, opened 2018-11-13
Port typevar and InstructionList from meta-python

lachlansneff:
This ports more of meta-python over to meta. Right now, the added code isn't used, but it will be soon, once the instructions start getting moved over.

### DIFF
--- a/lib/codegen/meta/src/cdsl/instructions.rs
+++ b/lib/codegen/meta/src/cdsl/instructions.rs
@@ -6,9 +6,7 @@ pub struct InstructionList {
 
 impl InstructionList {
     pub fn new() -> Self {
-        Self {
-            groups: Vec::new(),
-        }
+        Self { groups: Vec::new() }
     }
 
     pub fn group(&mut self, name: &'static str, f: impl FnOnce(&mut InstructionGroup)) {
@@ -96,12 +94,8 @@ impl InstructionBuilder {
 
 #[derive(Debug, PartialEq, Eq)]
 pub enum Constraint {
-    WiderOrEq {
-        ty: [ValueType; 2],
-    },
-    TypesEqual {
-        ty: [ValueType; 2],
-    },
+    WiderOrEq { ty: [ValueType; 2] },
+    TypesEqual { ty: [ValueType; 2] },
 }
 
 impl Constraint {
--- a/lib/codegen/meta/src/cdsl/instructions.rs
+++ b/lib/codegen/meta/src/cdsl/instructions.rs
@@ -1,0 +1,118 @@
+use super::types::ValueType;
+
+pub struct InstructionList {
+    groups: Vec<InstructionGroup>,
+}
+
+impl InstructionList {
+    pub fn new() -> Self {
+        Self {
+            groups: Vec::new(),
+        }
+    }
+
+    pub fn group(&mut self, name: &'static str, f: impl FnOnce(&mut InstructionGroup)) {
+        let mut group = InstructionGroup::new(name);
+        f(&mut group);
+        self.groups.push(group);
+    }
+}
+
+pub struct InstructionGroup {
+    name: &'static str,
+    insts: Vec<Instruction>,
+}
+
+impl InstructionGroup {
+    fn new(name: &'static str) -> Self {
+        Self {
+            name,
+            insts: Vec::new(),
+        }
+    }
+
+    pub fn add(&mut self, inst: Instruction) {
+        self.insts.push(inst);
+    }
+}
+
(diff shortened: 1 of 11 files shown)


lib/codegen/meta/src/cdsl/instructions.rs
```diff
@@ -0,0 +1,112 @@
...
+pub struct InstructionList {
+    groups: Vec<InstructionGroup>,
+}
+
+impl InstructionList {
+    pub fn new() -> Self {
+        Self { groups: Vec::new() }
+    }
+
+    pub fn group(&mut self, name: &'static str, f: impl FnOnce(&mut InstructionGroup)) {
```
sunfishcode: Can you add a comment explaining how this would be used? I'm a little unclear, because I don't see an obviously corresponding construct in meta-python, and it's not immediately obvious on its own.

lib/codegen/meta/src/cdsl/instructions.rs
```diff
@@ -0,0 +1,112 @@
...
+        self.insts.push(inst);
+    }
+}
+
+pub struct Instruction {
+    name: &'static str,
+    doc: &'static str,
+    ins: &'static [()],
+    outs: &'static [()],
+    constraints: &'static [Constraint],
```
sunfishcode: We can make these owned `String`s and `Vec`s, because this is meta code so we don't need to micro-manage allocations, and so that meta code can dynamically build things like operand lists when needed.

lib/codegen/meta/src/cdsl/typevar.rs
```diff
@@ -0,0 +1,345 @@
...
+    base: Option<Box<TypeVar>>,
+    derived_func: Option<&'static str>,
+}
+
+impl TypeVar {
+    pub fn build(name: impl Into<String>) -> TypeVarBuilder {
+        TypeVarBuilder {
+            name: name.into(),
+            doc: "",
+            type_set_builder: TypeSet::build(),
```
sunfishcode: We should decide on `TypeSet::build` vs `TypeSetBuilder::new`, or `Instruction`/`InstructionBuilder`, because they're currently different from each other and it'd be good to pick one way or the other :-).

I know it's more verbose, but I slightly prefer `TypeSetBuilder::new`. It's my vague impression that this style is more common in Rust code, though I'm open to learning otherwise.

lib/codegen/meta/src/srcgen.rs
```diff
@@ -45,6 +45,18 @@ impl Formatter {
         ret
     }
 
+    pub fn indent_with<T>(
```
bnjbvr: I consider having something like this too, but I never implemented it, because I thought it made it actually slightly harder to follow the generated code. At least with explicit `line()` calls before and after, we don't have to jump around to see what's after the block we just indented. If that makes sense, I'd just advise to use plain `line()` calls before and after instead.

lib/codegen/meta/src/cdsl/instructions.rs
```diff
@@ -0,0 +1,112 @@
...
+}
+
+#[derive(Debug, PartialEq, Eq)]
+pub enum Constraint {
+    WiderOrEq { ty: [ValueType; 2] },
+    TypesEqual { ty: [ValueType; 2] },
+}
+
+impl Constraint {
+    pub fn eval(&self) {
```
bnjbvr: Maybe rename `check`? Also, would it make sense to instead make the enum non public, and providing a public ctor that would create the instances and do the checks in the ctor itself? (so one can't forget to call `eval` or whatever it'll be called and miss the assertions)

lib/codegen/meta/src/cdsl/instructions.rs
```diff
@@ -0,0 +1,112 @@
...
+        self.outs = outs;
+        self
+    }
+
+    pub fn constraints(mut self, constraints: &'static [Constraint]) -> Self {
+        self.constraints = constraints;
+        self
+    }
+
+    pub fn build(self) -> Instruction {
```
bnjbvr: This name isn't consistent with the rest, where it's usually called `finish()`. An alternative could be to pass the Builder itself to `InstructionGroup::add`, which would consume the Builder argument.

bnjbvr:
Hi @lachlansneff (and happy new year y'all reading)! Was there any reason why you closed this PR early, or was this an error maybe? I hope our review comments didn't come as aggressive or we didn't hurt in any way. Your PR was off to a great start and this work is quite valuable, even though it's tedious (I know since I've had to do similar stuff!). Please let us know if we can help in any way!

PR #604, opened 2018-11-14
Remove the build time and directory from cranelift-codegen-meta's output

sunfishcode:
This makes its output non-deterministic, making it friendlier for compilation caching tools such as [sccache](https://github.com/mozilla/sccache).

Printing out the build directory is nice to do, as Cargo tends to keep a few versions around under the target directory and it's useful to know which one is for the current build, but I wasn't able to find a better way to do that.

This is my current theory as to what's causing the significant build time regressions in mozilla-central reported here: https://bugzilla.mozilla.org/show_bug.cgi?id=1506511

### DIFF
--- a/lib/codegen/build.rs
+++ b/lib/codegen/build.rs
@@ -24,11 +24,7 @@ use meta::isa::Isa;
 use std::env;
 use std::process;
 
-use std::time::Instant;
-
 fn main() {
-    let start_time = Instant::now();
-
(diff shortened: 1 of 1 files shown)


bnjbvr:
Sorry about that; I found them quite useful to have, in particular the output directory for comparing outputs of the build scripts. LGTM.

sunfishcode:
Yeah, I agree. I wish I knew of a better way to do this.

PR #605, opened 2018-11-14
Better incompatible declaration error

bjorn3:
When the signature is not compatible, the previous and current signatures are now included. This is very useful when debugging this kind of error.

### DIFF
--- a/lib/module/src/module.rs
+++ b/lib/module/src/module.rs
@@ -129,10 +129,10 @@ pub enum ModuleError {
     /// Indicates a function identifier was declared with a
     /// different signature than declared previously
     #[fail(
-        display = "Incompatible declaration of function identifier: {} ({:?} vs {:?})",
+        display = "Function {} signature {:?} is incompatible with previous declaration {:?}",
         _0,
-        _1,
-        _2
+        _2,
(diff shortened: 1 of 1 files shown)


lib/module/src/module.rs
```diff
@@ -123,9 +123,18 @@ pub enum ModuleError {
...
     #[fail(display = "Undeclared identifier: {}", _0)]
     Undeclared(String),
-    /// Indicates an identifier was used contrary to the way it was declared
+    /// Indicates an identifier was used as data/function first, but then used as the other
     #[fail(display = "Incompatible declaration of identifier: {}", _0)]
     IncompatibleDeclaration(String),
+    /// Indicates a function identifier was declared with a
+    /// different signature than declared previously
+    #[fail(
+        display = "Incompatible declaration of function identifier: {} ({:?} vs {:?})",
```
sunfishcode: Can we word the error message to make it clear which `{:?}` is the existing declaration and which is the new one?

Perhaps something like "Function {} signature {:?} is incompatible with previous declaration {:?}"?

sunfishcode:
Thanks!

PR #613, opened 2018-11-19
Make TargetIsa thread-safe

syrusakbary:
In order to parallelize function compilation (`compile_and_emit` from `Context`), the trait `TargetIsa` needs to be `Sync`.


sunfishcode:
Sounds right! The `TargetIsa` should be effectively immutable once constructed.

PR #615, opened 2018-11-23
Move settings generation from Python to Rust

bnjbvr:
This is almost done; it took me slightly more time than I expected because I wanted to have a nice interface for predicates and presets.

Unfortunately, because of anonymous predicates used in x86 encodings or legalization (can't remember), we can't just use this yet, because we need to migrate other build steps before we can. Maybe we could just change the code there to use named predicates instead...


sunfishcode:
Looks good! The `predicate!` macro is a nice simplification!

PR #617, opened 2018-11-26
Fix incorrect wasm subcommand 'about' text.

zachreizner:


src/clif-util.rs
```diff
@@ -128,7 +128,7 @@ fn get_vec(argument_vec: Option<clap::Values>) -> Vec<String> {
 
 fn add_wasm_or_compile<'a>(cmd: &str) -> clap::App<'a, 'a> {
     let about_str = match cmd {
-        "wasm" => "Compiles Cranelift IR into target language",
+        "wasm" => "Compiles Wasm binary/text into Cranelift IR",
```
bnjbvr: This is not quite correct: it actually translates the wasm IR to Cranelift IR before compiling it into the target language. Can you revise this comment, please?

zachreizner: I'm not quite sure what you would like this to say, but I tried.

bnjbvr:
Nice job, this is now correct. Great documentation matters, thank you!

PR #619, opened 2018-11-28
Clarify Cranelift's design with respect to mid-level optimization.

sunfishcode:
Cranelift doesn't currently do much mid-level optimization, however it
is something we're thinking about, so remove text describing it as out of
scope, and add more text explaining the vision for how it would fit into
the overall system.

### DIFF
--- a/docs/compare-llvm.rst
+++ b/docs/compare-llvm.rst
@@ -21,18 +21,7 @@ highlighting some of the differences and similarities. Both projects:
 - Can target multiple ISAs.
 - Can cross-compile by default without rebuilding the code generator.
 
-Cranelift's scope is much smaller than that of LLVM. The classical three main
-parts of a compiler are:
-
-1. The language-dependent front end parses and type-checks the input program.
-2. Common optimizations that are independent of both the input language and the
-   target ISA.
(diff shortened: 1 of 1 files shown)


docs/compare-llvm.rst
```diff
@@ -112,10 +118,15 @@ In LLVM IR, the largest representable unit is the *module* which corresponds
...
-references to shared data structures. Cranelift does not have any
-inter-procedural optimizations like inlining.
+In `Cranelift's IR` <https://cranelift.readthedocs.io/en/latest/ir.html>`_,
+used by the `cranelift-codegen <https://docs.rs/cranelift-codegen/>`_ crate,
+functions are self-contained, allowing them to be compiled independently. At
+this level, there is no explicit module that contains the functions.
+
+Module functionality in Cranelift is provided as an optional library layer, in
+the `cranelift-module <https://docs.rs/cranelift-module/>`_ crate. It provides
+facilities for working with with modules, which can contain multiple functions
```
bnjbvr: nit: `with with` => `with`

sunfishcode: fixed

PR #621, opened 2018-11-29
Support heaps with no offset-guard pages.

sunfishcode:
Also, say "guard-offset pages" rather than just "guard pages" to describe the
region of a heap which is never accessible and which exists to support
optimizations for heap accesses with offsets.

And, introduce a `Uimm64` immediate type, and make all heap fields use
`Uimm64` instead of `Imm64` since they really are unsigned.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -34,7 +34,7 @@ term = "0.5.1"
 capstone = { version = "0.5.0", optional = true }
 wabt = { version = "0.7.0", optional = true }
 target-lexicon = "0.2.0"
-pretty_env_logger = "0.2.4"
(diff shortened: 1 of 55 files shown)


frehberg:
My two cents:

`InstanceWorld::new(..)` is reading wasm-data and creating an instance. It would be closer to real-world example (and as template for other developers), splitting into two steps: 1) creating a module from wasm-data, 2) creating an instance from module. 

What do you think?

sunfishcode:
@frehberg I replied over in https://github.com/CraneStation/wasmtime/pull/34. The PR here is just adding some things on the Cranelift side needed by that PR :-).

PR #623, opened 2018-12-05
Fixes #614: Implement a debug trap

### LINKED ISSUE #614: debugger breakpoint trap
It would be useful for Cranelift to have a debugger-breakpoint instruction. It should be similar to the exising `trap` instruction, except that it doesn't need an immediate operand. On x86, this would be encoded as an `int 3` (0xcc).

bnjbvr:
It's quite simple, and I've checked that it works in Spidermonkey (which had me add the `other_side_effects` flag so it wouldn't get removed by DCE). There weren't any other nullary recipe I could use for this, so I implemented it the simplest way; if we want more nullary instructions in the future, we can probably factor the recipe better.

### DIFF
--- a/filetests/isa/x86/binary64.clif
+++ b/filetests/isa/x86/binary64.clif
@@ -784,6 +784,9 @@ ebb1:
     ; asm: jnbe .+4; ud2
     trapif ule v11, user0                          ; bin: 77 02 user0 0f 0b
 
+    ; Debug trap.
+    debugtrap ; bin: cc
+
     ; Stack check.
     ; asm: cmpq %rsp, %rcx
     [-,%rflags]         v40 = ifcmp_sp v1       ; bin: 48 39 e1
--- a/lib/codegen/meta-python/base/instructions.py
+++ b/lib/codegen/meta-python/base/instructions.py
@@ -181,6 +181,10 @@ indirect_jump_table_br = Instruction(
     ins=(addr, JT),
     is_branch=True, is_indirect_branch=True, is_terminator=True)
 
+debugtrap = Instruction('debugtrap', r"""
+    Encodes an assembly debug trap.
+    """, can_load=True, can_store=True, other_side_effects=True)
+
 code = Operand('code', trapcode)
 trap = Instruction(
         'trap', r"""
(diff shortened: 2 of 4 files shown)


lib/codegen/meta-python/isa/x86/encodings.py
```diff
@@ -521,12 +521,17 @@ def enc_i32_i64_ld_st(inst, w_bit, recipe, *args, **kwargs):
...
 enc_x86_64(base.indirect_jump_table_br.i64, r.indirect_jmp, 0xff, rrr=4)
 X86_32.enc(base.indirect_jump_table_br.i32, *r.indirect_jmp(0xff, rrr=4))
+
 #
 # Trap as ud2
 #
 X86_32.enc(base.trap, *r.trap(0x0f, 0x0b))
 X86_64.enc(base.trap, *r.trap(0x0f, 0x0b))
 
+# Debug traps
```
pepyakin: We could document it as being emitted as `int3`.

filetests/isa/x86/binary64.clif
```diff
@@ -784,6 +784,9 @@ ebb1:
     ; asm: jnbe .+4; ud2
     trapif ule v11, user0                          ; bin: 77 02 user0 0f 0b
 
+    ; Debug trap.
+    debugtrap; bin: cc
```
pepyakin:
```suggestion
    debugtrap ; bin: cc
```

lib/codegen/meta-python/base/instructions.py
```diff
@@ -181,6 +181,10 @@
     ins=(addr, JT),
     is_branch=True, is_indirect_branch=True, is_terminator=True)
 
+debugtrap = Instruction('debugtrap', r"""
+    Encodes an assembly debug trap.
+    """, other_side_effects=True)
```
sunfishcode: We should set `can_load` and `can_store` on this too, to represent the fact that a user stopping the program at a `debugtrap` can use a debugger to read and write memory.

bnjbvr:
Done!

sunfishcode:
Cool!

PR #624, opened 2018-12-07
WIP: Add support for an HeapBase argument

bnjbvr:
That's mostly for Spidermonkey, which uses a pinned register to contain the HeapBase on x64. It is supposed to be callee preserved, so we need a way to keep it alive (by passing it as an argument and returning it).

I'm not too happy with the changes that leak in here. Avoiding to push special purpose return values onto the wasm value stack seems fine, but the change in the frontend change feels wrong. I've had to do this, because otherwise the exit block will have the HeapBase as an EBB input; in which case every single jump to the exit block (natural jumps or jumps threaded to the return in FallthroughReturn mode) will need to have the HeapBase argument, and thus rematerialize it sometimes. I don't know much what to do here, so we'll need to think it through.

### DIFF
--- a/lib/codegen/src/ir/extfunc.rs
+++ b/lib/codegen/src/ir/extfunc.rs
@@ -265,6 +265,12 @@ pub enum ArgumentPurpose {
     /// caller in an indirect call. The callee can verify that the expected signature ID matches.
     SignatureId,
 
+    /// A pointer to the base of a single heap.
+    ///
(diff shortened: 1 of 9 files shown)


bnjbvr:
See Spidermonkey counterpart in [bug 1507820](https://bugzilla.mozilla.org/show_bug.cgi?id=1507820).

bnjbvr:
Closing, since I'm not working on this anymore at this time. I'll open an issue with the state of our discussions and what were the remaining work items.

PR #629, opened 2018-12-21
Branch comparison folding and order optimizations

### LINKED ISSUE #606: Optimize branches
Given input WebAssembly like this:

```wasm
  (block $label$1
   (if
    (i32.eqz
     (get_global $global$6)
    )
    (return
     (get_local $11)
    )
...
```

The Cranelift IR looks like this:

```clif
...
@000b   v10 = load.i32 notrap aligned v3+136
        v58 = ifcmp_imm v10, 0
@000e   brif ne v58, ebb3
@0012   jump ebb1(v8)

ebb3:
...
```

We should optimize this by swapping the conditional and unconditional branches, so that we have

```
         brif eq v58, ebb1(v8)
         jump ebb3
```
where the `jump` can then turn into a fallthrough. I think we can do this in `simple_preopt`, since it's unlikely the legalization will expose more opportunities to do this.

tyler:
Fixes #606.

This PR add two optimizations to `simple_preopt.rs`. The first is to fold code where we compare against zero, then use an explicit conditional branch. For instance:
```
v1 = icmp_imm v0, 0
brz v1, ebb1
```
becomes
```
brnz v0, ebb1
```
For x86-64 this doesn't actually make much difference, as it compiles to a `test` rather than a `cmp`, which should be very near equivalent. But it does save a byte on a very common operation, so hey.

The second is reordering branches at the end of ebbs to encourage fallthroughs. As pointed out in #606, it's common for wasm code to be compiled like this:
```
    brz v0, ebb1
    jump ebb2
ebb1:
    ...
ebb2:
    ...
```
Which ends up being compiled as a conditional jump followed by a jump. However, we can switch those instructions by flipping the destinations and inverting the conditional jump's condition, like so:
```
    brnz v0, ebb2
    fallthrough ebb1
ebb1:
    ...
ebb2:
    ...
```
And when we do that we eliminate the latter jump entirely.

### DIFF
--- a/cranelift-codegen/src/simple_preopt.rs
+++ b/cranelift-codegen/src/simple_preopt.rs
@@ -545,14 +545,8 @@ fn simplify(pos: &mut FuncCursor, inst: Inst) {
 struct BranchOptInfo {
     br_inst: Inst,
     cmp_arg: Value,
-    destination: Ebb,
     args: ValueList,
-    kind: BranchOptKind,
-}
-
-enum BranchOptKind {
-    EqualZero,
-    NotEqualZero,
+    new_opcode: Opcode,
 }
 
 /// Fold comparisons into branch operations when possible.
@@ -561,80 +555,86 @@ enum BranchOptKind {
 /// result in a `brz` or `brnz` branch. It folds those two operations into a
 /// single `brz` or `brnz`.
 fn branch_opt(pos: &mut FuncCursor, inst: Inst) {
-    let info = match pos.func.dfg[inst] {
-        InstructionData::Branch {
-            opcode,
-            destination,
-            ref args,
-        } => {
-            let first_arg = {
-                let args = pos.func.dfg.inst_args(inst);
-                args[0]
-            };
+    let mut info = if let InstructionData::Branch {
+        opcode, ref args, ..
+    } = pos.func.dfg[inst]
+    {
+        let first_arg = {
+            let args = pos.func.dfg.inst_args(inst);
+            args[0]
+        };
 
-            let iconst_inst =
-                if let ValueDef::Result(iconst_inst, _) = pos.func.dfg.value_def(first_arg) {
-                    iconst_inst
-                } else {
-                    return;
-                };
-
-            if let InstructionData::IntCompareImm {
-                opcode: Opcode::IcmpImm,
-                arg: cmp_arg,
-                cond: cmp_cond,
-                imm: cmp_imm,
-            } = pos.func.dfg[iconst_inst]
-            {
-                let cmp_imm: i64 = cmp_imm.into();
-                if cmp_imm != 0 {
-                    return;
-                }
-
-                // icmp_imm returns non-zero when the comparison is true. So, if
-                // we're branching on zero, we need to invert the condition.
-                let cond = match opcode {
-                    Opcode::Brz => cmp_cond.inverse(),
-                    Opcode::Brnz => cmp_cond,
-                    _ => return,
-                };
-
-                let kind = match cond {
-                    IntCC::Equal => BranchOptKind::EqualZero,
-                    IntCC::NotEqual => BranchOptKind::NotEqualZero,
-                    _ => return,
-                };
-
-                BranchOptInfo {
-                    br_inst: inst,
-                    cmp_arg: cmp_arg,
-                    destination: destination,
-                    args: args.clone(),
-                    kind: kind,
-                }
+        let iconst_inst =
+            if let ValueDef::Result(iconst_inst, _) = pos.func.dfg.value_def(first_arg) {
+                iconst_inst
             } else {
                 return;
+            };
+
+        if let InstructionData::IntCompareImm {
+            opcode: Opcode::IcmpImm,
+            arg: cmp_arg,
+            cond: cmp_cond,
+            imm: cmp_imm,
+        } = pos.func.dfg[iconst_inst]
+        {
+            let cmp_imm: i64 = cmp_imm.into();
+            if cmp_imm != 0 {
+                return;
             }
+
+            // icmp_imm returns non-zero when the comparison is true. So, if
+            // we're branching on zero, we need to invert the condition.
+            let cond = match opcode {
+                Opcode::Brz => cmp_cond.inverse(),
+                Opcode::Brnz => cmp_cond,
+                _ => return,
+            };
+
+            let new_opcode = match cond {
+                IntCC::Equal => Opcode::Brz,
+                IntCC::NotEqual => Opcode::Brnz,
+                _ => return,
+            };
+
+            BranchOptInfo {
+                br_inst: inst,
+                cmp_arg: cmp_arg,
+                args: args.clone(),
+                new_opcode: new_opcode,
+            }
+        } else {
+            return;
         }
-        _ => return,
+    } else {
+        return;
     };
 
-    match info.kind {
-        BranchOptKind::EqualZero => {
-            let args = info.args.as_slice(&pos.func.dfg.value_lists)[1..].to_vec();
-            pos.func
-                .dfg
-                .replace(info.br_inst)
-                .brz(info.cmp_arg, info.destination, &args);
-        }
-        BranchOptKind::NotEqualZero => {
-            let args = info.args.as_slice(&pos.func.dfg.value_lists)[1..].to_vec();
-            pos.func
-                .dfg
-                .replace(info.br_inst)
-                .brnz(info.cmp_arg, info.destination, &args);
-        }
+    info.args.as_mut_slice(&mut pos.func.dfg.value_lists)[0] = info.cmp_arg;
+    if let InstructionData::Branch { ref mut opcode, .. } = pos.func.dfg[info.br_inst] {
+        *opcode = info.new_opcode;
+    } else {
+        panic!();
     }
+
+    /*
+        match info.kind {
+            BranchOptKind::EqualZero => {
+                let args = info.args.as_slice(&pos.func.dfg.value_lists)[1..].to_vec();
+                pos.func
+                    .dfg
+                    .replace(info.br_inst)
+                    .brz(info.cmp_arg, info.destination, &args);
+            }
+            BranchOptKind::NotEqualZero => {
+                let args = info.args.as_slice(&pos.func.dfg.value_lists)[1..].to_vec();
+                pos.func
+                    .dfg
+                    .replace(info.br_inst)
+                    .brnz(info.cmp_arg, info.destination, &args);
+            }
+        }
+    */
 }
 
 struct BranchOrderInfo {
@@ -696,7 +696,6 @@ fn branch_order(pos: &mut FuncCursor, cfg: &mut ControlFlowGraph, ebb: Ebb, inst
                     opcode,
                     args: ref prev_args,
                     destination: cond_dest,
-                    ..
                 } => {
                     let cond_arg = {
                         let args = pos.func.dfg.inst_args(prev_inst);
@@ -762,7 +761,7 @@ fn branch_order(pos: &mut FuncCursor, cfg: &mut ControlFlowGraph, ebb: Ebb, inst
             pos.func
                 .dfg
                 .replace(info.term_inst)
-                .fallthrough(info.cond_dest, &cond_args[1..]);
+                .jump(info.cond_dest, &cond_args[1..]);
             pos.func
                 .dfg
                 .replace(info.cond_inst)
@@ -772,7 +771,7 @@ fn branch_order(pos: &mut FuncCursor, cfg: &mut ControlFlowGraph, ebb: Ebb, inst
             pos.func
                 .dfg
                 .replace(info.term_inst)
-                .fallthrough(info.cond_dest, &cond_args[1..]);
+                .jump(info.cond_dest, &cond_args[1..]);
             pos.func
                 .dfg
                 .replace(info.cond_inst)
@@ -782,7 +781,7 @@ fn branch_order(pos: &mut FuncCursor, cfg: &mut ControlFlowGraph, ebb: Ebb, inst
             pos.func
                 .dfg
                 .replace(info.term_inst)
-                .fallthrough(info.cond_dest, &cond_args[2..]);
+                .jump(info.cond_dest, &cond_args[2..]);
             pos.func.dfg.replace(info.cond_inst).br_icmp(
                 cond.inverse(),
                 x_arg,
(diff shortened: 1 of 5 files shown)


lib/codegen/src/simple_preopt.rs
```diff
@@ -536,11 +542,279 @@ fn simplify(pos: &mut FuncCursor, inst: Inst) {
...
+}
+
+fn branch_order(pos: &mut FuncCursor, cfg: &mut ControlFlowGraph, ebb: Ebb, inst: Inst) {
+    let info = match pos.func.dfg[inst] {
+        InstructionData::Jump {
+            opcode: Opcode::Jump,
+            destination,
+            ref args,
+        } => {
+            if let Some(next_ebb) = pos.func.layout.next_ebb(ebb) {
```
bjorn3: Could you please do it like:

```rust
let next_ebb = if let Some(next_ebb) = pos.func.layout.next_ebb(ebb) {
    next_ebb
} else {
    return;
};

if destination == next_ebb {
    return;
}

// ...
```

Here and below to reduce the indentation significantly.

lib/codegen/src/simple_preopt.rs
```diff
@@ -536,11 +542,279 @@ fn simplify(pos: &mut FuncCursor, inst: Inst) {
...
+                if destination == next_ebb {
+                    return;
+                }
+
+                if let Some(prev_inst) = pos.func.layout.prev_inst(inst) {
+                    let prev_inst_data = &pos.func.dfg[prev_inst];
+                    if !prev_inst_data.opcode().is_branch() {
```
sunfishcode: This test is redundant, as `branch_destination()` below will just return None if prev_inst_data is not a branch.

lib/codegen/src/simple_preopt.rs
```diff
@@ -536,11 +542,279 @@ fn simplify(pos: &mut FuncCursor, inst: Inst) {
...
+                            args: args.clone(),
+                            kind: BranchOptKind::NotEqualZero,
+                        },
+                        IntCC::Equal => BranchOptInfo {
+                            br_inst: inst,
+                            cmp_arg: cmp_arg,
+                            destination: destination,
+                            args: args.clone(),
+                            kind: BranchOptKind::EqualZero,
+                        },
```
sunfishcode: The duplication between the NotEqual and Equal arms here can be factored out by doing `let kind = match br_cond { ... };` to compute just the kind, and then you can do a single `BranchOptInfo { ... }`.

lib/codegen/src/simple_preopt.rs
```diff
@@ -536,11 +542,279 @@ fn simplify(pos: &mut FuncCursor, inst: Inst) {
...
+                                    Opcode::Brnz => BranchOrderInfo {
+                                        term_inst: inst,
+                                        term_inst_args: args.clone(),
+                                        term_dest: destination,
+                                        cond_inst: prev_inst,
+                                        cond_arg: cond_arg,
+                                        cond_inst_args: prev_args.clone(),
+                                        cond_dest: *cond_dest,
+                                        kind: BranchOrderKind::BrnzToBrz,
+                                    },
```
sunfishcode: Ditto here. The kind is the only thing that's different between the two cases, so you can factor that out.

lib/codegen/src/simple_preopt.rs
```diff
@@ -536,11 +542,279 @@ fn simplify(pos: &mut FuncCursor, inst: Inst) {
...
+    args: ValueList,
+    kind: BranchOptKind,
+}
+
+enum BranchOptKind {
+    EqualZero,
+    NotEqualZero,
+}
+
+fn branch_opt(pos: &mut FuncCursor, inst: Inst) {
```
sunfishcode: Please add a brief comment here describing what this function does.

lib/codegen/src/simple_preopt.rs
```diff
@@ -536,11 +542,279 @@ fn simplify(pos: &mut FuncCursor, inst: Inst) {
...
+}
+
+enum BranchOrderKind {
+    BrzToBrnz,
+    BrnzToBrz,
+    InvertIntCond(IntCC),
+    InvertFloatCond(FloatCC),
+}
+
+fn branch_order(pos: &mut FuncCursor, cfg: &mut ControlFlowGraph, ebb: Ebb, inst: Inst) {
```
sunfishcode: This too :-).

lib/codegen/src/simple_preopt.rs
```diff
@@ -536,11 +542,279 @@ fn simplify(pos: &mut FuncCursor, inst: Inst) {
...
+            destination,
+            ref args,
+        } => {
+            let first_arg = {
+                let args = pos.func.dfg.inst_args(inst);
+                args[0]
+            };
+            if let ValueDef::Result(iconst_inst, _) = pos.func.dfg.value_def(first_arg) {
+                if let InstructionData::BinaryImm {
+                    opcode: Opcode::IfcmpImm,
```
sunfishcode: This is admittedly not document as clearly as it could be, but we currently don't expect frontends to produce flags-using instructions like `ifcmp` or `ifcmp_imm` themselves. Regular `icmp` is lowered to `ifcmp` by the legalizer on platforms that use a flags register on x86 and in the future other architectures. Since the code here is in simple_preopt, the legalizer hasn't run yet, so in theory we don't expect to see `ifcmp_imm` occur at this point. So one question here is: are you seeing `ifmp_imm` at this point?

If so, I'm not strongly attached to the "frontends shouldn't use flags" rule; we could just as easily have frontends produce flags-using code and have the legalizer lower it into other things on platforms that don't use flags, and it's not immediately obvious which way is better.

lib/codegen/src/simple_preopt.rs
```diff
@@ -536,11 +542,279 @@ fn simplify(pos: &mut FuncCursor, inst: Inst) {
...
+            opcode: Opcode::Brif,
+            cond: br_cond,
+            destination,
+            ref args,
+        } => {
+            let first_arg = {
+                let args = pos.func.dfg.inst_args(inst);
+                args[0]
+            };
+            if let ValueDef::Result(iconst_inst, _) = pos.func.dfg.value_def(first_arg) {
```
bjorn3: Please do this like the following to avoid that immense amount of indentation:

```rust
let iconst_inst = if let ValueDef::Result(iconst_inst, _) = pos.func.dfg.value_def(first_arg) {
    iconst_inst
} else {
    return;
};
```

lib/codegen/src/simple_preopt.rs
tyler: Error on my end. I've fixed this to match the branch ops that frontends are actually supposed to use. I was blindly going by the code example in #606 and hadn't tested it against real code until today. :)

Cargo.toml
```diff
@@ -14,7 +14,7 @@ name = "clif-util"
 path = "src/clif-util.rs"
 
 [dependencies]
-cfg-if = "0.1"
+cfg-if = "0.1.6"
```
tyler: This appears to have been underspecified. With v0.1.4 locally, I was getting errors about missing macros. Presumably due to lack of this commit: https://github.com/alexcrichton/cfg-if/commit/67e01e5f0ea0e27d0bba38d3da7d73cda52d8be4

tyler:
I think this is ready for another round of reviews.

bnjbvr:
Hi @tyler ! Any idea when/if you'd have time to address the review comments and rebase this PR here? Or can we help you moving this forward by taking over the remaining changes?

tyler:
hey @bnjbvr, yeah I can get back to this soon. I'll try to get to that this afternoon.

tyler:
(If I don't, feel free to take it over. :))

tyler:
@sunfishcode or @bnjbvr, do you want to take another look before merge? I think I addressed all the things.

cranelift-codegen/src/simple_preopt.rs
```diff
@@ -536,11 +542,264 @@ fn simplify(pos: &mut FuncCursor, inst: Inst) {
...
+            }
+            BranchOptKind::NotEqualZero => {
+                let args = info.args.as_slice(&pos.func.dfg.value_lists)[1..].to_vec();
+                pos.func
+                    .dfg
+                    .replace(info.br_inst)
+                    .brnz(info.cmp_arg, info.destination, &args);
+            }
+        }
+    */
```
bjorn3: Remove this?

bnjbvr:
Thanks @tyler! I'll try to take a look at this early next week.

cranelift-codegen/src/simple_preopt.rs
tyler: I'm an idiot. Thanks. :)

cranelift-codegen/src/simple_preopt.rs
```diff
@@ -536,11 +542,245 @@ fn simplify(pos: &mut FuncCursor, inst: Inst) {
...
+            }
+        }
+
+        _ => return,
+    };
+
+    let cond_args = {
+        info.cond_inst_args
+            .as_slice(&pos.func.dfg.value_lists)
+            .to_vec()
```
bnjbvr: It seems all the uses of `cond_args` can use a slice; could we remove the vec heap allocation here?

cranelift-codegen/src/simple_preopt.rs
```diff
@@ -536,11 +542,245 @@ fn simplify(pos: &mut FuncCursor, inst: Inst) {
...
+    };
+    let term_args = {
+        info.term_inst_args
+            .as_slice(&pos.func.dfg.value_lists)
+            .to_vec()
```
bnjbvr: ditto

cranelift-codegen/src/simple_preopt.rs
```diff
@@ -536,11 +542,245 @@ fn simplify(pos: &mut FuncCursor, inst: Inst) {
...
+
+    info.args.as_mut_slice(&mut pos.func.dfg.value_lists)[0] = info.cmp_arg;
+    if let InstructionData::Branch { ref mut opcode, .. } = pos.func.dfg[info.br_inst] {
+        *opcode = info.new_opcode;
+    } else {
+        panic!();
+    }
+}
+
+struct BranchOrderInfo {
```
bnjbvr: Since this struct is only used internally within a function, could we use a tuple in the function instead, and that would remove the need for this structure?

cranelift-codegen/src/simple_preopt.rs
```diff
@@ -536,11 +542,245 @@ fn simplify(pos: &mut FuncCursor, inst: Inst) {
...
+            if let Some(prev_dest) = prev_inst_data.branch_destination() {
+                if prev_dest != next_ebb {
+                    return;
+                }
+            } else {
+                return;
+            }
+
+            match prev_inst_data {
+                InstructionData::Branch {
```
bnjbvr: Do we want to handle BranchInt (for `brif)` and BranchFloat (for `brff`) here too? (we could do this in a separate PR)

bjorn3:
I believe you meant `other` instead of `order` in the pr title :)

cranelift-codegen/src/simple_preopt.rs
```diff
@@ -536,11 +542,245 @@ fn simplify(pos: &mut FuncCursor, inst: Inst) {
...
+}
+
+/// Fold comparisons into branch operations when possible.
+///
+/// This matches against operations which compare against zero, then use the
+/// result in a `brz` or `brnz` branch. It folds those two operations into a
+/// single `brz` or `brnz`.
+fn branch_opt(pos: &mut FuncCursor, inst: Inst) {
+    let mut info = if let InstructionData::Branch {
+        opcode, ref args, ..
```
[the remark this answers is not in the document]
tyler: Good point. Done.

cranelift-codegen/src/simple_preopt.rs
```diff
@@ -536,11 +542,245 @@ fn simplify(pos: &mut FuncCursor, inst: Inst) {
...
+
+        if let InstructionData::IntCompareImm {
+            opcode: Opcode::IcmpImm,
+            arg: cmp_arg,
+            cond: cmp_cond,
+            imm: cmp_imm,
+        } = pos.func.dfg[iconst_inst]
+        {
+            let cmp_imm: i64 = cmp_imm.into();
+            if cmp_imm != 0 {
```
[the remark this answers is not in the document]
tyler: Doing this would require changing the definition of `Imm64(i64)` to `Imm64(pub i64)`, as far as I can tell. I could certainly do that, but figured there might be a reason why it's private.

I'd love to do that as `Imm64::from(0)` but https://github.com/rust-lang/rust/issues/57240 is not yet a thing.

cranelift-codegen/src/simple_preopt.rs
tyler: The trouble is that just below this I borrow `pos.func.dfg` as mutable. I haven't been able to figure out how to appease the borrow checker in this case. I'm all ears if you have suggestions though.

cranelift-codegen/src/simple_preopt.rs
tyler: I went ahead and did this. But I'm really not sure that it improved readability of the function. Take a look and let me know what you think?

cranelift-codegen/src/simple_preopt.rs
tyler: I don't *think* those should be necessary here, because they're not supposed to be constructed by frontends and this is a preopt pass. Please correct me if I'm thinking about that incorrectly though.

tyler:
> I believe you meant `other` instead of `order` in the pr title :)

Surprisingly, I didn't. I was referring to re-ordering branches. :)

cranelift-codegen/src/simple_preopt.rs
bnjbvr: Sounds fair, I asked just in case.

cranelift-codegen/src/simple_preopt.rs
bnjbvr: Good point! I asked this since I didn't know these opcodes were private.

cranelift-codegen/src/simple_preopt.rs
bnjbvr: Definitely an improvement *after* the let statement. It's maybe less clear within the body, but it seems there are few enough lines in between that it's easy to check the tuple definition, so I'd be good with keeping it the new way. Thanks!

bnjbvr:
Thanks, looks great!

PR #632, opened 2018-12-22
Update to Rust 2018 edition

### LINKED ISSUE #628: Update to Rust 2018 edition
It's now two weeks after Rust 1.31 is released, as we're following the [Firefox Rust Update policy](https://wiki.mozilla.org/Rust_Update_Policy_for_Firefox), so we can now update to the Rust 2018 edition.

We should update the Rust version in the top-level README.md and .travis.yml.

See [the documentation](https://rust-lang-nursery.github.io/edition-guide/editions/transitioning-an-existing-project-to-a-new-edition.html) for information on how to transition an existing project to a new edition.

mominul:
It's been one hell of a ride! :rocket: 
Please review the `codegen` crate carefully as I dont have much knowlede about the code generation scripts.
I have ran `format-all.sh` and `test-all.sh`.

Closes #628 

Thanks!


mominul:
@sunfishcode I think we can remove the dependency on `cranelift-entity` crate when we already depend on the `cranelift-codegen` crate as it re-exports `cranelift-entity` crate. I guess we had to depend on it explicitly because of the `entity_impl!` macro. But because of the new module system changes, we can import it with `cranelift_codegen::entity::entity_impl`. Any thoughts?

bjorn3:
I am getting some warnings locally:

```
warning: use of deprecated item 'core::str::<impl str>::trim_left': superseded by `trim_start`
   --> /Users/bjorn/Documents/cranelift/lib/codegen/meta/src/srcgen.rs:154:19
    |
154 |         let t = s.trim_left();
    |                   ^^^^^^^^^
    |
    = note: #[warn(deprecated)] on by default

warning: use of deprecated item 'core::str::<impl str>::trim_left': superseded by `trim_start`
   --> /Users/bjorn/Documents/cranelift/lib/codegen/meta/src/srcgen.rs:172:30
    |
172 |         .map(|l| l.len() - l.trim_left().len())
    |                              ^^^^^^^^^

warning: use of deprecated item 'core::str::<impl str>::trim_right': superseded by `trim_end`
   --> /Users/bjorn/Documents/cranelift/lib/codegen/meta/src/srcgen.rs:189:24
    |
189 |             .map(|l| l.trim_right())
    |                        ^^^^^^^^^^

warning: use of deprecated item 'core::str::<impl str>::trim_right': superseded by `trim_end`
   --> /Users/bjorn/Documents/cranelift/lib/codegen/meta/src/srcgen.rs:194:24
    |
194 |             .map(|l| l.trim_right())
    |                        ^^^^^^^^^^
```

sunfishcode:
@mominul Thanks for working on this! I hope the ride has been fun :smile:! Concerning cranelift_entity vs cranelift_codegen::entity, I don't have a strong preference either way. Does rexporting crates like this help in the case where the end user is using a different version of cranelift-entity for unrelated purposes?

.travis.yml
```diff
@@ -7,7 +7,7 @@ language: rust
 rust:
     # The oldest version we currently support. See
     # CONTRIBUTING.md#rustc-version-support for details.
-    - 1.30.1
+    - 1.31.0
```
sunfishcode: Rust 1.31.1 was just released; we should update to that here.

mominul:
@sunfishcode Yes, indeed it was a fun ride :grinning: I have resolved the tidy issues and conflicts so the PR is green now. I think you can start the reviewing process now. About cranelift_entity vs cranelift_codegen::entity, if we use re-exported crate this would remove an unneeded dependency from the dependency graph as the codegen crate already re-exports the entity crate. I am a rust beginner so I can't help much with the decision. If you agree, then I can do the experiment in another PR.

@bjorn3 I think you are using the nightly channel because those two functions are going to be deprecated in the future 1.33.0 version. But thanks for the hint, I have replaced those two with the new ones.

Thanks!

bjorn3:
Yes I am using the nightly channel. Looking at the stable docs I thought they were already deprecated, but now that I read it again it says "deprecat**ing** in 1.33.0". :)

sunfishcode:
Cool, thanks for working on this!

mominul:
Feeling great to help!

sunfishcode:
If you're interested in a followup, another thing we can do here is grep for the various places in the code that are commented 'non-lexical lifetimes' and `Non-lexical lifetimes`, which hopefully can be simplified now :-).

mominul:
Ok, I'll try to do a follow-up! I am thinking about making the `entity_impl! ` into a derivable proc-macro, what's your opinion?

PR #634, opened 2018-12-24
Fix emit_small_mem{cpy,set} when size == 0 or access_size > 8

bjorn3:


lib/frontend/src/frontend.rs
```diff
@@ -599,6 +603,13 @@ impl<'a> FunctionBuilder<'a> {
             access_size >= ::std::cmp::min(src_align, dest_align) as u64,
             "`size` is smaller than `dest` and `src`'s alignment value."
         );
+
+        let (access_size, int_type) = if let Some(int_type) = Type::int((access_size * 8) as u16) {
```
sunfishcode: To be safe, we shouldn't use `as u16` here, as that will silently wrap if access_size is large. Since this code knows that the largest supported access size is 8 below anyway, I think we can just do `if access_size > 8` here to test whether it needs to be adjusted.

lib/frontend/src/frontend.rs
```diff
@@ -670,6 +684,13 @@ impl<'a> FunctionBuilder<'a> {
             access_size >= buffer_align as u64,
             "`size` is smaller than `dest` and `src`'s alignment value."
         );
+
+        let (access_size, int_type) = if let Some(int_type) = Type::int((access_size * 8) as u16) {
+            (access_size, int_type)
+        } else {
+            (8, types::I64)
+        };
```
sunfishcode: Ditto.

sunfishcode:
Looks good, thanks!

PR #645, opened 2019-01-03
Use snake case for magic numbers computation

bnjbvr:
Just a small cleanup that was sitting silently on my machine. There's no good reason to not use snake case here, so let's do that.


sunfishcode:
Let's do it.

PR #651, opened 2019-01-12
Peekable define function

bjorn3:
This is the least intrusive way I could find to get the necessary things for bjorn3/rustc_codegen_cranelift#166.

### DIFF
--- a/lib/module/src/module.rs
+++ b/lib/module/src/module.rs
@@ -506,7 +506,7 @@ where
 
     /// Define a function, producing the function body from the given `Context`.
     pub fn define_function(&mut self, func: FuncId, ctx: &mut Context) -> ModuleResult<()> {
-        self.define_function_peek_compiled(func, ctx, |_, _| ())
+        self.define_function_peek_compiled(func, ctx, |_, _, _| ())
     }
 
     /// Define a function, allowing to peek at the compiled function and producing the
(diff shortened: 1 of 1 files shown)


sunfishcode:
Is it necessary to call the peek_compiled function before doing the `self.backend.define_function`?

bjorn3:
No, calling it later simplified it a lot.

sunfishcode:
Now that the call is at the end, would it be possible for the caller of `define_function_peek_compiled` to instead call the function itself after `define_function_peek_compiled` returns, rather than passing in a function to be called at the end?

bjorn3:
Hadn't thought about that yet. It will require exposing `code_size` and `isa` though.

sunfishcode:
We could have `define_function` return the `code_size` instead of `()`. And `Module` can provide an accessor to its `isa` if needed, although do you need the whole `isa` or just the `TargetFrontendConfig`? The latter is already available as `module.target_config()`.

bjorn3:
I need the whole isa to get the `encoding_info`.

sunfishcode:
Ok. This PR feels somewhat specialized, and I'd ideally like Cranelift's API to make this kind of thing easier in a more general way, though I don't have any specific ideas at the moment. So let's merge the PR, but keep an eye out for ideas to refactor this :-).

PR #656, opened 2019-01-22
avoid directories for cargo build dependencies

froydnj:
We are facing peculiar Windows-only regressions in build times in
https://bugzilla.mozilla.org/show_bug.cgi?id=1506511 and while the build
times might just be Windows being slow, putting directories in
`rerun-if-changed` might also be causing problems.  The build only
depends on the files, anyway, so let's just say that.

### DIFF
--- a/lib/codegen/meta-python/gen_build_deps.py
+++ b/lib/codegen/meta-python/gen_build_deps.py
@@ -22,22 +22,11 @@ except ImportError:
     pass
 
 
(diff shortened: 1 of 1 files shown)


sunfishcode:
Sounds good.

froydnj:
The build is failing in files I didn't modify due to `rustfmt` differences?  That seems...peculiar.

PR #657, opened 2019-01-22
Preserve global wasm module offset in SourceLoc.

yurydelendik:
Making SourceLoc(u32) unique across the module.

### DIFF
--- a/cranelift-wasm/src/func_translator.rs
+++ b/cranelift-wasm/src/func_translator.rs
@@ -54,10 +54,15 @@ impl FuncTranslator {
     pub fn translate<FE: FuncEnvironment + ?Sized>(
         &mut self,
         code: &[u8],
+        code_offset: usize,
         func: &mut ir::Function,
         environ: &mut FE,
     ) -> WasmResult<()> {
-        self.translate_from_reader(BinaryReader::new(code), func, environ)
+        self.translate_from_reader(
+            BinaryReader::new_with_offset(code, code_offset),
+            func,
+            environ,
+        )
     }
 
     /// Translate a binary WebAssembly function from a `BinaryReader`.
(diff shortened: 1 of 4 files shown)


lib/wasm/src/func_translator.rs
```diff
@@ -226,7 +231,7 @@ fn cur_srcloc(reader: &BinaryReader) -> ir::SourceLoc {
     // This will wrap around of a single function's byte code is larger than 4 GB, but a) the
     // WebAssembly format doesn't allow for that, and b) that would hit other Cranelift
     // implementation limits anyway.
-    ir::SourceLoc::new(reader.current_position() as u32)
+    ir::SourceLoc::new(reader.original_position() as u32)
```
sunfishcode: Should this be `reader.original_position() + reader.current_position()` to get the full offset to the current opcode?

Also, we should coordinate with SpiderMonkey to see whether changing from function-relative offsets to module offsets will affect things there. @bnjbvr Do you know if baldrdash has the function start offsets around when it calls `emit_metadata`? If so, it could use them to translate back to function-relative offsets if needed.

And we should update the comment here in `cur_srcloc` :-).

yurydelendik: `original_position()` already [adds](https://github.com/yurydelendik/wasmparser.rs/blob/master/src/binary_reader.rs#L123) offset (that is provided during `new_with_offset`)

bnjbvr: In `emit_metadata`, Baldrdash already saves only the function relative offset; the function offset base is added in the C++ code. So if we change this to use module-relative offsets instead, my understanding is we can just update the C++ code.

sunfishcode: Sounds good, let's merge it.

cranelift-wasm/src/func_translator.rs
```diff
@@ -222,11 +227,9 @@ fn parse_function_body<FE: FuncEnvironment + ?Sized>(
 
 /// Get the current source location from a reader.
 fn cur_srcloc(reader: &BinaryReader) -> ir::SourceLoc {
-    // We record source locations as byte code offsets relative to the beginning of the function.
-    // This will wrap around of a single function's byte code is larger than 4 GB, but a) the
-    // WebAssembly format doesn't allow for that, and b) that would hit other Cranelift
-    // implementation limits anyway.
-    ir::SourceLoc::new(reader.current_position() as u32)
+    // We record source locations as byte code offsets relative to the beginning of the file.
+    // This will wrap around of byte code is larger than 4 GB.
```
bnjbvr: preexisting nit: `of` => `if`

bnjbvr:
Thanks!

PR #659, opened 2019-01-23
[WIP] Emit source location map: wasm bytecode to native byte offsets.

yurydelendik:
Adds the `SourceLocSink` to the `compile` process to record the map of all bitemit produced offsets to the wasm read `SourceLoc`'s.

### DIFF
--- a/lib/codegen/src/binemit/memorysink.rs
+++ b/lib/codegen/src/binemit/memorysink.rs
@@ -36,6 +36,7 @@ pub struct MemoryCodeSink<'a> {
     pub code_size: isize,
     relocs: &'a mut RelocSink,
     traps: &'a mut TrapSink,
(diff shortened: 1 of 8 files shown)


philipc:
FYI, an alternative way to do this is to get the srclocs in a separate pass, such as in https://github.com/bjorn3/rustc_codegen_cranelift/pull/291/files#diff-16055c9225ae7e0bb50fc787ee2da587R453. Might be simpler, depending on what you are doing with them.

yurydelendik:
Agree. For experimentation and pilot work, it [possible](https://github.com/yurydelendik/wasmtime/commit/85e1e666429c028a57308a4803bd6a7cabe162e5#diff-b7292fcfd4672502321175915088c25d) to do it without this PR. Closing as now.

Issue #647, opened 2019-01-06
Directory structure for crates


sunfishcode:
Currently Cranelift uses a cargo workspace structure in which there's a top-level crate, and then subcrates under a `lib` directory with shortened names:

   lib/wasm/ - the cranelift-wasm crate
   lib/codegen/ - the cranelift-codegen crate
   etc.

I've not seen any other Rust projects organized in this way. What I have seen is that projects create directories under the top level with the full name of their crates. For Cranelift, that would look something like this:

  cranelift-wasm/
  cranelift-codegen/
  etc.

Has anyone found the `/lib/foo` structure to be confusing? Would the full crate name approach be easier to understand?

lazypassion:
I probably wouldn't have found those crates if I didn't read this issue. I feel like the way rustc is organized is at least easier to discover all the parts of the overall crate.

sunfishcode:
Ok, it sounds like it'd be good to do this. My guess is we should include the `cranelift-` prefixes, to separate the from the non-crate directories. Is anyone interested in working on this? :smiley:

lazypassion:
I'm willing. What would I have to do?

sunfishcode:
You'd have to rename the directories, update the paths in the top-level Cargo.toml, as well as changing paths like "../codegen" to "../cranelift-codegen" in the subdirectory Cargo.toml files. And then search for "lib/" in the rest of the tree and update accordingly. And then make sure everything still builds :-).

lazypassion:
Okay sounds good

bnjbvr:
Thanks @lazypassion, closing.

PR #660, opened 2019-01-24
moved crates in lib/ to src/, renamed crates, modified some files' text

lazypassion:
#647 
Hopefully I didn't miss anything or had a typo. The test-all script had no problems.

Is there some way to format the python code? The build failed because some lines are longer than 80.


sunfishcode:
Overall this looks really good!

My impression from repos that I've seen is that the convention is to put the crate directories at the top level of the repository, rather than under the src directory. For some examples:

https://github.com/rajcspsg/multi_module_cargo_project
https://github.com/Smithay/wayland-rs/

However I haven't done a formal survey or anything, but I'm ok doing it either way, as long as it's conventional.

Actually, moving them out of src just might shrink those python source lines within 80 columns ;-). If not, let me know and I can help wrap those lines.

lazypassion:
Ahhhhhh, gotcha. I overthought what you meant there. You simply wanted to move them all out from the lib folder into the top level directory, correct? And yeah that will probably fix the line length issue.

sunfishcode:
Looks great, thanks a lot!

PR #662, opened 2019-01-28
Fix link to cargo workspace documentation

Voultapher:
Small thing I came across while reading the README. Old link leads to `The second edition of the book is no longer distributed with Rust's documentation.`.


sunfishcode:
Yes, thanks for spotting that, and for the fix!

PR #667, opened 2019-02-05
Fix #666: Change the way we consider a block has been visited in relaxation

### LINKED ISSUE #666: Branch relaxation ignores non-entry blocks with destination offset 0
The following test case triggers an invalid forward jump instead of a backward jump. What's happening is the following:

- The compiled function entry block is not generating code, so the second block `Ebb3` has offset 0.
- Cranelift optimistically selects the 1-byte encoded jump for a jump that goes from the Nth block to `Ebb3`.
- Since the offset of the Nth block is 206, and the target block is at offset 0, the branch target is out of range and needs to be relaxed.
- The branch relaxation code does see that the offset is out of range, but the second condition checks that the `dest_offset == 0 || dest block == entry block`. So the second condition fails (`dest_offset` is 0, but the dest block isn't the empty entry block), and the branch isn't relaxed.
- Generated code is invalid: the jump encoding is `eb 30                   jmp     0x30`, that is, a forward jump (and in my case, it jumps into the middle of two instructions, causing a crash).

Test case:

```
test compile
set opt_level=best
set avoid_div_traps
set baldrdash_prologue_words=3
set allones_funcaddrs
set probestack_enabled=false
target x86_64 haswell

function u0:2691(i32, i32, i64 vmctx) -> i32 uext baldrdash {
    gv0 = vmctx
    gv1 = iadd_imm.i64 gv0, 48
    gv2 = load.i64 notrap aligned readonly gv0
    heap0 = static gv2, min 0xd839_6000, bound 0x0001_0000_0000, offset_guard 0x8000_0000, index_type i32

                                ebb0(v0: i32, v1: i32, v2: i64):
@0001                               v4 = iconst.i32 0
@0005                               jump ebb3(v0, v1)

                                ebb3(v8: i32, v19: i32):
@0005                               v6 = global_value.i64 gv1
@0005                               v7 = load.i64 v6
@0005                               trapnz v7, interrupt
@0009                               v9 = iconst.i32 1
@000b                               v10 = iadd v8, v9
                                    v80 -> v10
@0010                               v11 = heap_addr.i64 heap0, v8, 1
@0010                               v12 = sload8.i32 v11
@0015                               v13 = iconst.i32 24
@0017                               v14 = ishl v12, v13
@0018                               v15 = iconst.i32 24
@001a                               v16 = sshr v14, v15
@001d                               v17 = iconst.i32 32
@001f                               v18 = iadd v16, v17
@0024                               v20 = iconst.i32 1
@0026                               v21 = iadd v19, v20
                                    v82 -> v21
@002b                               v22 = heap_addr.i64 heap0, v19, 1
@002b                               v23 = sload8.i32 v22
                                    v55 -> v23
@0030                               v24 = iconst.i32 24
@0032                               v25 = ishl v23, v24
@0033                               v26 = iconst.i32 24
@0035                               v27 = sshr v25, v26
                                    v69 -> v27
@0038                               v28 = iconst.i32 32
@003a                               v29 = iadd v27, v28
                                    v68 -> v29
@003f                               v30 = iconst.i32 -65
@0042                               v31 = iadd v12, v30
@0043                               v32 = iconst.i32 24
@0045                               v33 = ishl v31, v32
@0046                               v34 = iconst.i32 24
@0048                               v35 = sshr v33, v34
@0049                               v36 = iconst.i32 255
@004c                               v37 = band v35, v36
@004d                               v38 = iconst.i32 26
@004f                               v39 = icmp slt v37, v38
@004f                               v40 = bint.i32 v39
@0050                               brz v40, ebb6
@0054                               jump ebb5(v18, v18)

                                ebb6:
@0059                               jump ebb5(v16, v16)

                                ebb5(v41: i32, v84: i32):
                                    v83 -> v84
@005a                               v42 = iconst.i32 255
@005d                               v43 = band v41, v42
@0060                               v44 = iconst.i32 24
@0062                               v45 = ishl v43, v44
@0063                               v46 = iconst.i32 24
@0065                               v47 = sshr v45, v46
@0066                               v48 = iconst.i32 0
@0068                               v49 = icmp ne v47, v48
@0068                               v50 = bint.i32 v49
@006b                               v51 = iconst.i32 24
@006d                               v52 = ishl v43, v51
@006e                               v53 = iconst.i32 24
@0070                               v54 = sshr v52, v53
@0073                               v56 = iconst.i32 -65
@0076                               v57 = iadd.i32 v23, v56
@0077                               v58 = iconst.i32 24
@0079                               v59 = ishl v57, v58
@007a                               v60 = iconst.i32 24
@007c                               v61 = sshr v59, v60
@007d                               v62 = iconst.i32 255
@0080                               v63 = band v61, v62
@0081                               v64 = iconst.i32 26
@0083                               v65 = icmp slt v63, v64
@0083                               v66 = bint.i32 v65
@0084                               brz v66, ebb8
@0088                               jump ebb7(v29, v10, v21, v29)

                                ebb8:
@008d                               jump ebb7(v27, v80, v82, v27)

                                ebb7(v67: i32, v79: i32, v81: i32, v87: i32):
@008e                               v70 = iconst.i32 255
@0091                               v71 = band v67, v70
@0092                               v72 = iconst.i32 24
@0094                               v73 = ishl v71, v72
@0095                               v74 = iconst.i32 24
@0097                               v75 = sshr v73, v74
@0098                               v76 = icmp.i32 eq v54, v75
@0098                               v77 = bint.i32 v76
@0099                               v78 = band.i32 v50, v77
@009a                               brz v78, ebb9
@00a4                               jump ebb3(v79, v81)

                                ebb9:
@00a7                               jump ebb4

                                ebb4:
@00aa                               v85 = iconst.i32 255
@00ad                               v86 = band.i32 v84, v85
@00b0                               v88 = iconst.i32 255
@00b3                               v89 = band.i32 v87, v88
@00b4                               v90 = isub v86, v89
@00b5                               jump ebb2(v90)

                                ebb2(v5: i32):
@00b6                               jump ebb1(v5)

                                ebb1(v3: i32):
@00b6                               fallthrough_return v3
}
```

bnjbvr:
This was previously using the following condition to decide that a block
hadn't been visited yet: either dest_offset is non-0 or the block isn't
the entry block.  Unfortunately, this didn't work when the first block
would be non-empty but wouldn't generate code at all. This uses
immutable indexing with an Option<> to check if the block's offset is
defined, instead.

### DIFF
--- a/cranelift-codegen/src/binemit/relaxation.rs
+++ b/cranelift-codegen/src/binemit/relaxation.rs
@@ -55,13 +55,26 @@ pub fn relax_branches(func: &mut Function, isa: &TargetIsa) -> CodegenResult<Cod
     let mut offset = 0;
     let mut divert = RegDiversions::new();
 
-    // The relaxation algorithm iterates to convergence.
+    // First, compute initial offsets for every EBB.
+    {
+        let mut cur = FuncCursor::new(func);
+        while let Some(ebb) = cur.next_ebb() {
+            divert.clear();
+            cur.func.offsets[ebb] = offset;
+            while let Some(inst) = cur.next_inst() {
+                let enc = cur.func.encodings[inst];
+                offset += encinfo.byte_size(enc, inst, &divert, &cur.func);
+            }
+        }
+    }
+
+    // Then, run the relaxation algorithm until it converges.
     let mut go_again = true;
     while go_again {
         go_again = false;
         offset = 0;
 
-        // Visit all instructions in layout order
+        // Visit all instructions in layout order.
         let mut cur = FuncCursor::new(func);
         while let Some(ebb) = cur.next_ebb() {
             divert.clear();
@@ -81,15 +94,12 @@ pub fn relax_branches(func: &mut Function, isa: &TargetIsa) -> CodegenResult<Cod
 
                 let enc = cur.func.encodings[inst];
 
-                // See if this might be a branch that is out of range.
+                // See if this is a branch has a range and a destination, and if the target is in
+                // range.
                 if let Some(range) = encinfo.branch_range(enc) {
                     if let Some(dest) = cur.func.dfg[inst].branch_destination() {
                         let dest_offset = cur.func.offsets[dest];
-                        // This could be an out-of-range branch.
-                        // Relax it unless the destination offset has not been computed yet.
-                        if !range.contains(offset, dest_offset)
-                            && (dest_offset != 0 || Some(dest) == cur.func.layout.entry_block())
-                        {
+                        if !range.contains(offset, dest_offset) {
                             offset +=
                                 relax_branch(&mut cur, &divert, offset, dest_offset, &encinfo, isa);
                             continue;
--- a/filetests/isa/x86/relax_branch.clif
+++ b/filetests/isa/x86/relax_branch.clif
@@ -1,0 +1,123 @@
+test binemit
+set opt_level=best
+set avoid_div_traps
+set baldrdash_prologue_words=3
+set allones_funcaddrs
+set probestack_enabled=false
+target x86_64 haswell
+
+; This checks that a branch that is too far away is getting relaxed. In
+; particular, the first block has to be non-empty but its encoding size must be
+; zero (i.e. not generate any code). See also issue #666 for more details.
+
+function u0:2691(i32 [%rdi], i32 [%rsi], i64 vmctx [%r14]) -> i64 uext [%rax] baldrdash {
+    ss0 = incoming_arg 24, offset -24
+    gv0 = vmctx
+    gv1 = iadd_imm.i64 gv0, 48
+    gv2 = load.i64 notrap aligned readonly gv0
+    heap0 = static gv2, min 0xd839_6000, bound 0x0001_0000_0000, offset_guard 0x8000_0000, index_type i32
+
+                                ebb0(v0: i32 [%rdi], v1: i32 [%rsi], v2: i64 [%r14]):
+@0005 [-]                           fallthrough ebb3(v0, v1)
+
+                                ebb3(v8: i32 [%rdi], v19: i32 [%rsi]):
+@0005 [RexOp1ldDisp8#808b,%rax]     v7 = load.i64 v2+48
+@0005 [RexOp1rcmp_ib#f083,%rflags]  v91 = ifcmp_imm v7, 0
+@0005 [trapif#00]                   trapif ne v91, interrupt
+[Op1umr#89,%rax]                    v105 = copy v8
+@000b [Op1r_ib#83,%rax]             v10 = iadd_imm v105, 1
+                                    v80 -> v10
+@0010 [Op1umr#89,%rcx]              v92 = uextend.i64 v8
+@0010 [RexOp1ld#808b,%rdx]          v93 = load.i64 notrap aligned readonly v2
+                                    v95 -> v93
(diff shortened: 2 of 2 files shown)


bnjbvr:
I'll add a test.

bnjbvr:
This patch is not correct; I misunderstood the condition under which the SecondaryMap would return None (all entries have the default value).

bnjbvr:
I've got a fix that seems to be right, now. Unfortunately, I couldn't add my orginal test case, because for some reason, most instructions in it don't get a valid encoding when run under the `binemit` directive. I've spent hours on trying to add a test case; in particular I tried to copy/paste the CLIF IR that's received just before regalloc is done, so it's definitely valid and legal, but this didn't work either. I can see that compiling and printing assembly for the test yields the right type of jump, so this is "manually" verified, which is unsatisfying.

bnjbvr:
So I think I understand why my test case doesn't get legal encodings: it seems I need to run both the legalization and regalloc phases to get the jump I want, in the binemit test. Now I don't think that the binemit tests want to run regalloc. Instead, should we allow the regalloc filetests to contain `bin:` directives that are checked?

sunfishcode:
The patch looks good. For a test, it's surprising that we can't write a plain binemit test for this. Is there something that regalloc can produce that you can't get from clif directly?

bnjbvr:
I think regalloc gives register assignments to instructions, which in turn gives instructions real sizes, which in turn makes it that the jump and destination are far away enough from each other.

sunfishcode:
clif files can represent register assignments. For example, in filetests/isa/x86/binary64.clif, a line like
```
[-,%rax]            v521 = iconst.i64 1
```
means an instruction with no encoding assigned yet, but which uses %rax for its output value.

bnjbvr:
That's right! I could actually just copy the last CLIF output by `cargo run compile -d -p` and use this in the binemit test, which I've added here.

filetests/isa/x86/relax_branch.clif
```diff
@@ -0,0 +1,119 @@
+test binemit
+set opt_level=best
+set avoid_div_traps
+set baldrdash_prologue_words=3
+set allones_funcaddrs
+set probestack_enabled=false
+target x86_64 haswell
+
+function u0:2691(i32 [%rdi], i32 [%rsi], i64 vmctx [%r14]) -> i64 uext [%rax] baldrdash {
```
sunfishcode: Can you add a comment mentioning what this test is testing? Perhaps link to the issue, and point out that the key feature of the code here is an entry block that gets zero bytes of encoding.

bnjbvr:
Done, merging. Thanks!

cranelift-codegen/src/binemit/relaxation.rs
```diff
@@ -55,13 +55,26 @@ pub fn relax_branches(func: &mut Function, isa: &TargetIsa) -> CodegenResult<Cod
...
     let mut divert = RegDiversions::new();
 
-    // The relaxation algorithm iterates to convergence.
+    // First, compute initial offsets for every EBB.
+    {
+        let mut cur = FuncCursor::new(func);
+        while let Some(ebb) = cur.next_ebb() {
+            divert.clear();
+            cur.func.offsets[ebb] = offset;
+            while let Some(inst) = cur.next_inst() {
```
philipc: The main loop does: `divert.apply(&cur.func.dfg[inst]);`. Is this needed here too?

bnjbvr: Yes indeed, good catch.

PR #671, opened 2019-02-09
Removed FunctionBuilderContext from public API.

### LINKED ISSUE #567: Eliminate the FunctionBuilder vs FunctionBuilderContext divide
Cranelift uses `FunctionBuilderContext` to hold onto memory allocations that can be reused between instances of `FunctionBuilder`s. However, because `FunctionBuilder` has a reference to the `FunctionBuilderContext`, it can't live in the same struct as a `FunctionBuilderContext`, and this can be inconvenient for some frontends, which want to store all Cranelift-related state in a single struct.

I think it'd be good to just merge `FunctionBuilder` and `FunctionBuilderContext`, and allow the result to be cleared while retaining the underlying allocations.

XAMPPRocky:
closes #567

### DIFF
--- a/cranelift-frontend/src/lib.rs
+++ b/cranelift-frontend/src/lib.rs
@@ -3,7 +3,7 @@
 //! Provides a straightforward way to create a Cranelift IR function and fill it with instructions
 //! corresponding to your source program written in another language.
 //!
-//! To get started, create an [`FunctionBuilderContext`](struct.FunctionBuilderContext.html) and
+//! To get started, create an [`Function`](../cranelift_codegen/ir/function/struct.Function.html) and
 //! pass it as an argument to a [`FunctionBuilder`](struct.FunctionBuilder.html).
 //!
 //! # Mutable variables and Cranelift IR values
@@ -59,7 +59,7 @@
 //! }
 //! ```
 //!
-//! Here is how you build the corresponding Cranelift IR function using `FunctionBuilderContext`:
+//! Here is how you build the corresponding Cranelift IR function using `Function`:
 //!
 //! ```rust
 //! extern crate cranelift_codegen;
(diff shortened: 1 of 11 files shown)


cranelift-frontend/src/lib.rs
```diff
@@ -58,104 +57,6 @@
...
-//!     }
-//!
-//!     let flags = settings::Flags::new(settings::builder());
-//!     let res = verify_function(&func, &flags);
-//!     println!("{}", func.display(None));
-//!     if let Err(errors) = res {
-//!         panic!("{}", errors);
-//!     }
-//! }
-//! ```
```
sunfishcode: Instead of deleting this large comment, could you instead update the comment to work with the new API? This comment serves as an intro to the whole cranelift-frontend crate, and it's useful to have a basic getting-started example like this.

cranelift-wasm/src/func_translator.rs
```diff
@@ -20,15 +20,13 @@ use wasmparser::{self, BinaryReader};
 /// by a `FuncEnvironment` object. A single translator instance can be reused to translate multiple
 /// functions which will reduce heap allocation traffic.
 pub struct FuncTranslator {
-    func_ctx: FunctionBuilderContext,
```
sunfishcode: This should hold a `FuncBuilder` now, which is cleared between functions, so that it achieves the same effect of holding on to the memory allocations.

Aaronepower: Hello, sorry I'm only getting back to this now. I was wondering how do I initialise a regular `FunctionBuilder`, as `FuncTranslator` is not given a `Function` until `Function::translate_from_reader` is called?

sunfishcode: Ah, that's a good question. One option would be to change FunctionBuilder from holding the Function by reference to holding a Function value. Then it could have a method that does `std::replace(&mut self.func, Function::new())` and returns the old Function, so allow consumers to take the Function out when it's done.

Aaronepower:
@sunfishcode Okay I've completly rewritten the PR, let me know what you think of the new API. I made `finalize` return the `Function` as I found that's when you want the function back in the code in cranelift, rather than a seperate method.

XAMPPRocky:
Hey, I went back to this PR and updated it against the latest changes in the codebase. @sunfishcode Would you or another reviewer be able to review this PR?

sunfishcode:
Yes, and thanks!

PR #673, opened 2019-02-11
Conditional compilation per ISA target

bnjbvr:
This series of commits does the following:

- disentangle the selection of ISAs from the env variable CRANELIFT_TARGETS
- then rename "base" to "shared" in the Rust meta crate; the "base" things are actually entities that are shared among all platforms (i hesitated with "common" which seemed clearer too). We could even put this directory into the `isa` directory, since it's really something that's common to all ISAs (but also contain stuff that's not specific to ISAs, so I didn't do it).
- it bothered me that Registers and Settings were not constructed the same way. It was a bit weird that the TargetIsaBuilder had the responsibility of holding state during creation of the registers, so i put this out of it into an `IsaRegsBuilder` struct that builds into an `IsaRegs`. After this, the `TargetIsa` can be built immutably, which is nice. (And a few names are simplified along the way)
- last commt implements conditional compilation of ISAs with features, taking some inspiration from #137 (Thanks @zummenix for the initial work!). I've tried it by making all the Cranelift subcrates define `default-features=false` when they depended on Codegen, and could then use the features to select one or the other ISAs.

(For what it's worth, I tried to get one step further by having the Python script get a new runtime flag `--isas` to select which ISAs to target, but I ran into mysterious issues where the shared legalizer functions were not being generated, and stopped looking into it, since it is code that is meant to disappear)


cranelift-codegen/build.rs
```diff
@@ -30,29 +25,36 @@ fn main() {
...
-            }
-        }
-        Err(err) => {
-            eprintln!("Error: {}", err);
-            process::exit(1);
+    let isa_targets = meta::isa::Isa::all()
+        .into_iter()
+        .map(|isa| isa.to_string())
+        .filter(|isa| {
+            let env_key = format!("CARGO_FEATURE_{}", isa.to_uppercase());
```
sunfishcode: Would it work to move the `to_string()` here into the `filter` body, rather than using a `map` first?

cranelift-codegen/meta/src/lib.rs
```diff
@@ -2,12 +2,69 @@
...
+
 mod constant_hash;
+mod shared;
 mod srcgen;
 mod unique_table;
+
+pub fn isa_from_arch(arch: &str) -> Result<isa::Isa, String> {
+    isa::Isa::from_arch(arch)
+        .ok_or_else(|| format!("no supported isa found for arch `{}`", arch))
+        .and_then(|isa| Ok(isa))
```
sunfishcode: Is this `and_then` line redundant?

cranelift-codegen/build.rs
```diff
@@ -30,29 +25,36 @@ fn main() {
...
+        .collect::<Vec<_>>();
+
+    let isas = if isa_targets.len() == 0 {
+        // Try to match native target.
+        let target_name = target_triple.split('-').next().unwrap();
+        let found_isa = meta::isa_from_arch(&target_name);
+        if let Ok(isa) = found_isa {
+            println!("cargo:rustc-cfg=feature=\"{}\"", isa);
         }
+        found_isa.and_then(|x| Ok(vec![x]))
```
sunfishcode: Instead of `and_then(|x| Ok(...))`, this can use `map(|x| ...)`.

cranelift-codegen/build.rs
```diff
@@ -30,29 +25,36 @@ fn main() {
...
+    let isa_targets = meta::isa::Isa::all()
+        .into_iter()
+        .map(|isa| isa.to_string())
+        .filter(|isa| {
+            let env_key = format!("CARGO_FEATURE_{}", isa.to_uppercase());
+            env::var(env_key).is_ok()
+        })
+        .collect::<Vec<_>>();
+
+    let isas = if isa_targets.len() == 0 {
```
sunfishcode: This can use `is_empty()` rather than `len() == 0`.

cranelift-codegen/build.rs
bnjbvr: Yes, it's nice: the `isa_targets` just is the vector of ISA targets we want in the end, so that removes an entire function `isa_from_targets` in the `meta` crate.

bnjbvr:
Updated!

sunfishcode:
Looks good!

PR #674, opened 2019-02-13
Meta: make the Formatter more pleasant to use

bnjbvr:
This does the following:

- accept both `&str` and `String` as arguments of the `Formatter::line` function, so one doesn't need to add a reference for String arguments.
- introduce the `fmtln!` macro to pass formatted strings to `Formatter::line` in an easier way. It was discussed on IRC to reuse the `writeln!` macro, but unfortunately the latter is fallible while the former isn't, so it would pollute the code by adding a lot of `?` or `unwrap/expect` which don't bring additional value (since our `Write` impl for the Formatter would always return `Ok(())`).
- the last commit I'm less sure of: I wondered if adding a `Formatter::block(text_before, |fmt| {})` function would make it easier to use the formatter (and not have to worry about block scoping). Unfortunately, in many cases we don't want just a pair of enclosing braces, but sometimes we want an additional semicolon, sometimes we want an array, etc. There was a function in the Python code to do this, allowing one to specify the opening and closing content, but I found it less pleasant to use, because of the mental back-and-forth due to the position of the closing content (maybe that's something to retry with a different arg position, but I can't foresee it being much nicer to use one way or the other).

Any opinions about this?

### DIFF
--- a/cranelift-codegen/meta/src/srcgen.rs
+++ b/cranelift-codegen/meta/src/srcgen.rs
@@ -3,6 +3,8 @@
 //! The `srcgen` module contains generic helper routines and classes for
 //! generating source code.
 
+#![macro_use]
+
 use std::cmp;
 use std::collections::{BTreeMap, BTreeSet};
 use std::fs;
@@ -12,6 +14,18 @@ use std::path;
 use crate::error;
 
 static SHIFTWIDTH: usize = 4;
+
+/// A macro that simplifies the usage of the Formatter by allowing format
+/// strings.
+macro_rules! fmtln {
+    ($fmt:ident, $fmtstring:expr, $($fmtargs:expr),*) => {
+        $fmt.line(format!($fmtstring, $($fmtargs),*));
+    };
+
+    ($fmt:ident, $arg:expr) => {
+        $fmt.line($arg);
+    };
+}
 
 pub struct Formatter {
     indent: usize,
@@ -104,7 +118,7 @@ impl Formatter {
 
     /// Add a comment line.
     pub fn comment(&mut self, s: impl AsRef<str>) {
-        self.line(format!("// {}", s.as_ref()));
+        fmtln!(self, "// {}", s.as_ref());
     }
 
     /// Add a (multi-line) documentation comment.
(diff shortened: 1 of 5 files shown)


bnjbvr:
Since I wasn't convinced the last commit (which introduced the `Formatter::block()` function) was an enhancement, I removed it from this PR.

sunfishcode:
I haven't looked at the patch in detail yet, but I like the idea. It's similar to how Rust lets us say `write!(w, ...)` rather than `write_str(w, format!(...))`.

cranelift-codegen/meta/src/srcgen.rs
```diff
@@ -120,7 +137,7 @@ impl Formatter {
 
     /// Add a match expression.
     pub fn add_match(&mut self, m: Match) {
-        self.line(&format!("match {} {{", m.expr));
+        self.line(format!("match {} {{", m.expr));
```
sunfishcode: Can these use `fmtln!(self, "match {} {{", m.expr)`?

cranelift-codegen/meta/src/srcgen.rs
```diff
@@ -135,7 +152,7 @@ impl Formatter {
                     })
                     .collect();
                 let lhs = conditions.join(" | ");
-                fmt.line(&format!("{} => {{", lhs));
+                fmt.line(format!("{} => {{", lhs));
```
sunfishcode: Another `fmtln!` opportunity?

cranelift-codegen/meta/src/srcgen.rs
```diff
@@ -99,14 +117,13 @@ impl Formatter {
     }
 
     /// Add a comment line.
-    pub fn comment(&mut self, s: &str) {
-        let commented_line = format!("// {}", s);
-        self.line(&commented_line);
+    pub fn comment(&mut self, s: impl AsRef<str>) {
+        self.line(format!("// {}", s.as_ref()));
```
sunfishcode: Can this be `fmtln!(self, "// {}", s.as_ref())`?

bnjbvr:
Thanks for the review! Added all these new uses.

PR #679, opened 2019-02-19
Track frame layout changes.

yurydelendik:
Adds basic frame layout (or unwinding) capabilities to the x86 abi.

### DIFF
--- a/cranelift-codegen/src/ir/framelayout.rs
+++ b/cranelift-codegen/src/ir/framelayout.rs
@@ -51,7 +51,7 @@ pub struct FrameLayout {
     /// Initial frame layout.
     pub initial: FrameLayoutChanges,
 
-    /// Instruction frame layout (changes). The map will not be dense,
+    /// Instruction frame layout (changes). Because the map will not be dense,
     /// a HashMap is used instead of a SecondaryMap.
     pub instructions: HashMap<Inst, FrameLayoutChanges>,
 }
--- a/cranelift-codegen/src/ir/framelayout.rs
+++ b/cranelift-codegen/src/ir/framelayout.rs
@@ -5,7 +5,7 @@ use crate::isa::RegUnit;
 use std::boxed::Box;
 
 #[cfg(not(feature = "std"))]
-use hashmap_core::HashMap;
+use crate::HashMap;
 #[cfg(feature = "std")]
 use std::collections::HashMap;
 
--- a/cranelift-codegen/src/ir/framelayout.rs
+++ b/cranelift-codegen/src/ir/framelayout.rs
@@ -31,6 +31,15 @@ pub enum FrameLayoutChange {
         /// Offset in the frame (offset from CFA).
         cfa_offset: isize,
     },
+    /// The entire frame layout must be preserved somewhere to be restored at a corresponding
+    /// `Restore` change.
+    ///
+    /// This likely maps to the DWARF call frame instruction `.cfa_remember_state`
+    Preserve,
+    /// Restore the entire frame layout from a corresponding prior `Preserve` frame change.
+    ///
+    /// This likely maps to the DWARF call frame instruction `.cfa_restore_state`
+    Restore,
 }
 
 /// Set of frame layout changes.
--- a/cranelift-codegen/src/ir/framelayout.rs
+++ b/cranelift-codegen/src/ir/framelayout.rs
@@ -1,0 +1,64 @@
+//! Frame layout item changes.
+
+use crate::ir::entities::Inst;
+use crate::isa::RegUnit;
+use std::boxed::Box;
+
+#[cfg(not(feature = "std"))]
+use hashmap_core::HashMap;
+#[cfg(feature = "std")]
+use std::collections::HashMap;
+
+/// Change in the frame layout information.
+#[derive(Clone, Copy, Debug, Eq, PartialEq, Hash)]
+pub enum FrameLayoutChange {
+    /// Base CFA pointer moved to different register/offset.
+    CallFrameAddressAt {
+        /// CFA register.
+        reg: RegUnit,
+        /// CFA offset.
+        offset: isize,
+    },
+    /// Register saved at.
+    RegAt {
+        /// Saved register.
+        reg: RegUnit,
+        /// Offset in the frame (offset from CFA).
+        cfa_offset: isize,
+    },
+    /// Return address saved at.
+    ReturnAddressAt {
+        /// Offset in the frame (offset from CFA).
+        cfa_offset: isize,
+    },
+}
+
+/// Set of frame layout changes.
+pub type FrameLayoutChanges = Box<[FrameLayoutChange]>;
+
+/// Frame items layout for (prologue/epilogue) instructions.
+#[derive(Debug, Clone)]
+pub struct FrameLayout {
+    /// Initial frame layout.
+    pub initial: FrameLayoutChanges,
+
+    /// Instruction frame layout (changes). The map will not be dense,
+    /// a HashMap is used instead of a SecondaryMap.
+    pub instructions: HashMap<Inst, FrameLayoutChanges>,
+}
+
+impl FrameLayout {
+    /// Creates instance of FrameLayout.
+    pub fn new() -> Self {
+        FrameLayout {
+            initial: vec![].into_boxed_slice(),
+            instructions: HashMap::new(),
+        }
+    }
+
+    /// Clear the structure.
+    pub fn clear(&mut self) {
+        self.initial = vec![].into_boxed_slice();
+        self.instructions.clear();
+    }
+}
--- a/cranelift-codegen/src/ir/function.rs
+++ b/cranelift-codegen/src/ir/function.rs
@@ -11,7 +11,7 @@ use crate::ir::{
     Ebb, ExtFuncData, FuncRef, GlobalValue, GlobalValueData, Heap, HeapData, Inst, JumpTable,
     JumpTableData, SigRef, StackSlot, StackSlotData, Table, TableData,
 };
-use crate::ir::{EbbOffsets, InstEncodings, SourceLocs, StackSlots, ValueLocations};
+use crate::ir::{EbbOffsets, FrameLayout, InstEncodings, SourceLocs, StackSlots, ValueLocations};
 use crate::ir::{JumpTableOffsets, JumpTables};
 use crate::isa::{CallConv, EncInfo, Encoding, Legalize, TargetIsa};
 use crate::regalloc::{EntryRegDiversions, RegDiversions};
@@ -83,6 +83,13 @@ pub struct Function {
     /// Track the original source location for each instruction. The source locations are not
     /// interpreted by Cranelift, only preserved.
     pub srclocs: SourceLocs,
+
+    /// Frame layout for the instructions.
+    ///
+    /// The stack unwinding requires to have information about which registers and where they
+    /// are saved in the frame. This information is created during the prologue and epilogue
+    /// passes.
+    pub frame_layout: Option<FrameLayout>,
 }
 
 impl Function {
@@ -104,6 +111,7 @@ impl Function {
             offsets: SecondaryMap::new(),
             jt_offsets: SecondaryMap::new(),
             srclocs: SecondaryMap::new(),
+            frame_layout: None,
         }
     }
 
@@ -123,6 +131,7 @@ impl Function {
         self.offsets.clear();
         self.jt_offsets.clear();
         self.srclocs.clear();
+        self.frame_layout = None;
     }
 
     /// Create a new empty, anonymous function with a Fast calling convention.
@@ -232,6 +241,7 @@ impl Function {
     /// Starts collection of debug information.
     pub fn collect_debug_info(&mut self) {
         self.dfg.collect_debug_info();
+        self.frame_layout = Some(FrameLayout::new());
     }
 
     /// Changes the destination of a jump or branch instruction.
--- a/cranelift-codegen/src/isa/x86/abi.rs
+++ b/cranelift-codegen/src/isa/x86/abi.rs
@@ -17,6 +17,7 @@ use crate::regalloc::RegisterSet;
 use crate::result::CodegenResult;
 use crate::stack_layout::layout_stack;
 use core::i32;
+use std::boxed::Box;
 use target_lexicon::{PointerWidth, Triple};
 
 /// Argument registers for x86-64
@@ -722,13 +723,11 @@ fn insert_common_epilogues(
         if let Some(inst) = pos.current_inst() {
             if pos.func.dfg[inst].opcode().is_return() {
                 // figure out if we need to insert end-of-function-aware frame layout information
-                let following_inst = pos.next_ebb().and_then(|next_ebb| {
-                    pos.goto_first_inst(next_ebb);
-                    pos.current_inst()
-                });
-
-                // and rewind back to where we want to insert prologue instructions
-                pos.goto_last_inst(ebb);
+                let following_inst = pos
+                    .func
+                    .layout
+                    .next_ebb(ebb)
+                    .and_then(|next_ebb| pos.func.layout.first_inst(next_ebb));
 
                 if following_inst.is_some() {
                     if let Some(ref mut frame_layout) = pos.func.frame_layout {
@@ -819,10 +818,12 @@ fn insert_common_epilogue(
             .instructions
             .entry(inst)
             .and_modify(|insts| {
-                let mut new_instructions = insts.to_vec();
-                new_instructions.push(new_cfa);
-                *insts = new_instructions.into_boxed_slice();
+                *insts = insts
+                    .into_iter()
+                    .map(|x| *x)
+                    .chain(std::iter::once(new_cfa))
+                    .collect::<Box<[_]>>();
             })
-            .or_insert_with(|| vec![new_cfa].into_boxed_slice());
+            .or_insert_with(|| Box::new([new_cfa]));
     }
 }
--- a/cranelift-codegen/src/isa/x86/abi.rs
+++ b/cranelift-codegen/src/isa/x86/abi.rs
@@ -372,10 +372,26 @@ fn baldrdash_prologue_epilogue(func: &mut ir::Function, isa: &dyn TargetIsa) -> 
     Ok(())
 }
 
+/// CFAState is `cranelift`'s model of the call frame layout at any particular point in a function.
+/// Changes in this layout are used to derive appropriate `ir::FrameLayoutChange` to record for
+/// relevant instructions.
 #[derive(Clone)]
 struct CFAState {
+    /// The register from which we can derive the call frame address. On x86_64, this is typically
+    /// `rbp`, but at function entry and exit may be `rsp` while the call frame is being
+    /// established.
     cf_ptr_reg: RegUnit,
+    /// Given that `cf_ptr_reg` is a register containing a pointer to some memory, `cf_ptr_offset`
+    /// is the offset from that pointer to the address of the start of this function's call frame.
+    ///
+    /// For a concrete x86_64 example, we will start this at 8 - the call frame begins immediately
+    /// before the return address. This will typically then be set to 16, after pushing `rbp` to
+    /// preserve the parent call frame. It is very unlikely the offset should be anything other
+    /// than one or two `usize`.
     cf_ptr_offset: isize,
+    /// The offset between the start of the call frame and the current stack pointer. This is
+    /// primarily useful to point to where on the stack preserved registers are, but is maintained
+    /// through the whole function for consistency.
     current_depth: isize,
 }
 
--- a/cranelift-codegen/src/isa/x86/abi.rs
+++ b/cranelift-codegen/src/isa/x86/abi.rs
@@ -371,6 +371,7 @@ fn baldrdash_prologue_epilogue(func: &mut ir::Function, isa: &dyn TargetIsa) -> 
     Ok(())
 }
 
+#[derive(Clone)]
 struct CFAState {
     cf_ptr_reg: RegUnit,
     cf_ptr_offset: isize,
@@ -458,7 +459,7 @@ fn fastcall_prologue_epilogue(func: &mut ir::Function, isa: &dyn TargetIsa) -> C
 
     // Reset the cursor and insert the epilogue
     let mut pos = pos.at_position(CursorPosition::Nowhere);
-    insert_common_epilogues(&mut pos, local_stack_size, reg_type, &csrs, &cfa_state);
+    insert_common_epilogues(&mut pos, local_stack_size, reg_type, &csrs, isa, &cfa_state);
 
     Ok(())
 }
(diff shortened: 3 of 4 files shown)


bjorn3:
Using `.debug_frame` will hopefully fix gdb backtraces in debug mode. (gdb gets confused by `rex push rbp` https://github.com/bjorn3/rustc_codegen_cranelift/issues/146#issuecomment-449474527)

lachlansneff:
It'd be useful to generate the `.eh_frame` and equivalent sections so that we can use the system unwinder.

cranelift-codegen/src/ir/function.rs
```diff
@@ -202,6 +221,24 @@ impl Function {
     pub fn encode(&self, inst: ir::Inst, isa: &TargetIsa) -> Result<Encoding, Legalize> {
         isa.encode(&self, &self.dfg[inst], self.dfg.ctrl_typevar(inst))
     }
+
+    /// Gets list of all instructions with frame layout changes.
+    pub fn frame_layout_instrs(&self) -> Box<[&ir::Inst]> {
```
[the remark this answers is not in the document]
yurydelendik: As I understand, `SecondaryMap` allocates memory for all Inst in the function. Though we need a few at the start (and end). `HashMap` will be a better choice IMHO.

I changed that to return a `HashSet` instead of `Box` and now it is used in wasmtime.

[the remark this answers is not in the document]
philipc: > I changed that to return a HashSet instead of Box and now it is used in wasmtime.

I still don't see the point of this function. So your usage in wasmtime is these snippets:
```rust
let frame_layout_instrs = func.frame_layout_instrs();
...
if !frame_layout_instrs.contains(&inst) {
    continue;
}
let frame_layout_cmds = func.inst_frame_layout_changes(inst).expect("frame_layout_cmds");
```
A `HashSet` is just a `HashMap` with `()` as the value. So you're taking the keys of a `HashMap` to create a `HashSet`, then looking up the keys in the `HashSet` in order to see if the `HashMap` contains them, and then immediately looking up the key in the `HashMap`. I don't see what this is accomplishing compared to just looking it up in the `HashMap` directly.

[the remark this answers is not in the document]
yurydelendik: okay let's do just that

cranelift-codegen/src/isa/x86/abi.rs
```diff
@@ -361,14 +367,27 @@ fn fastcall_prologue_epilogue(func: &mut ir::Function, isa: &TargetIsa) -> Codeg
         func.signature.returns.push(csr_arg);
     }
 
+    let mut cfa_state = CFAState {
+        reg: RU::rsp as RegUnit,
+        offset: word_size as isize,
+        position: -(word_size as isize),
+    };
```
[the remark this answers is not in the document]
yurydelendik: Added the `initial` field to the `FrameLayout` and x86 ISA data -- I'll add logic to the wasmtime later.

philipc:
> It'd be useful to generate the .eh_frame

I've implemented write support for .eh_frame in gimli (https://github.com/gimli-rs/gimli/pull/419). I expect that it should require no changes to this PR, and only a minor change to the wasmtime PR, but I haven't tested.

cranelift-codegen/src/isa/x86/abi.rs
```diff
@@ -445,21 +479,58 @@ fn insert_common_prologue(
...
     let ebb = pos.current_ebb().expect("missing ebb under cursor");
     let fp = pos.func.dfg.append_ebb_param(ebb, reg_type);
     pos.func.locations[fp] = ir::ValueLoc::Reg(RU::rbp as RegUnit);
 
-    pos.ins().x86_push(fp);
-    pos.ins()
+    let push_fp_inst = pos.ins().x86_push(fp);
+    let word_size = word_size as isize;
+    cfa_state.position -= word_size;
+    cfa_state.offset += word_size;
```
bnjbvr: Are we sure no compiler pass won't add more instructions in between? (in particular thinking about register allocation which could insert copies/fills, or constant pools being generated, whenever we have them?)

cranelift-codegen/src/isa/x86/abi.rs
```diff
@@ -361,14 +367,27 @@ fn fastcall_prologue_epilogue(func: &mut ir::Function, isa: &dyn TargetIsa) -> C
         func.signature.returns.push(csr_arg);
     }
 
+    let mut cfa_state = CFAState {
+        reg: RU::rsp as RegUnit,
+        offset: word_size as isize,
+        position: -(word_size as isize),
```
bnjbvr: These members names are a bit unclear: my understanding is that `position` points to the return address position, while `reg` / `offset` are information telling what's the current register + offset pointing to the call frame. Is that correct? If so, we should probably rename these.

cranelift-codegen/src/ir/framelayout.rs
```diff
@@ -0,0 +1,63 @@
...
+pub type FrameLayoutChanges = Box<[FrameLayoutChange]>;
+
+/// Frame items layout for (prologue/epilog) instructions.
+#[derive(Debug, Clone)]
+pub struct FrameLayout {
+    /// Initial frame layout.
+    pub initial: FrameLayoutChanges,
+
+    /// Instruction frame layout (changes).
+    pub instructions: HashMap<Inst, FrameLayoutChanges>,
```
bnjbvr: `Inst` implements `cranelift_entity::Entity`, which means we could and should use a `SecondaryMap` here.

That being said: this means that whenever a user wants to track the call frame position at any point, they have to apply all the frame layout changes from the beginning up to the current instruction they're looking at. Is there enough information in the frame layout that it's justified to not just store the whole state for each instruction?

cranelift-codegen/src/ir/framelayout.rs
```diff
@@ -0,0 +1,63 @@
...
+    },
+    /// Register saved at.
+    RegAt {
+        /// Saved register.
+        reg: RegUnit,
+        /// Offset in the frame (offset from CFA).
+        cfa_offset: isize,
+    },
+    /// Return address saved at.
+    RaAt {
```
bnjbvr: nit: I'd expand this name to `RetAddrAt`, or even `ReturnAddressAt`.

cranelift-codegen/src/isa/x86/abi.rs
```diff
@@ -469,7 +540,16 @@ fn insert_common_prologue(
         pos.func.locations[csr_arg] = ir::ValueLoc::Reg(reg);
 
         // Remember it so we can push it momentarily
-        pos.ins().x86_push(csr_arg);
+        let reg_push_inst = pos.ins().x86_push(csr_arg);
+        cfa_state.position -= word_size;
+        pos.func.frame_layout.instructions.insert(
+            reg_push_inst,
+            vec![FrameLayoutChange::RegAt {
+                reg: reg,
```
bnjbvr: Are we interested in the position of every register? (I guess so, if I understand correctly this is for tracking where local variables are.) If so, should every instruction in the function's body also have its own frame layout changes (in case regalloc moves a variable from one register to another)?

cranelift-codegen/src/ir/function.rs
```diff
@@ -74,6 +74,13 @@ pub struct Function {
     /// Track the original source location for each instruction. The source locations are not
     /// interpreted by Cranelift, only preserved.
     pub srclocs: SourceLocs,
+
+    /// Frame layout for the instructions.
+    ///
+    /// The stack unwinding requires to have information about which registers and where they
+    /// are saved in the frame. This information is created during the prologue and epilogue
+    /// passes.
+    pub frame_layout: FrameLayout,
```
bnjbvr: Could this be an Option type, so it is generated and filled only when a flag is set?

cranelift-codegen/src/ir/framelayout.rs
```diff
@@ -0,0 +1,63 @@
...
+        /// Offset in the frame (offset from CFA).
+        cfa_offset: isize,
+    },
+}
+
+/// Set of frame layout changes.
+pub type FrameLayoutChanges = Box<[FrameLayoutChange]>;
+
+/// Frame items layout for (prologue/epilog) instructions.
```
bnjbvr: nit: epilogue

cranelift-codegen/src/isa/x86/abi.rs
yurydelendik: the `prologue_epilogue` runs after `regalloc` but before `shrink_instructions`, `relax_branches` -- I hope the latter does not insert anything to change stack layout.

cranelift-codegen/src/isa/x86/abi.rs
yurydelendik: With this PR (and .debug_frame) we are interested in basics: proper frame unwinding (i.e. BP, SP, IP) and maybe registers used for args. (The tracking of original variables values is happening somewhere else).

cranelift-codegen/src/ir/framelayout.rs
yurydelendik: `SecondaryMap` is not sparse enough -- we need to do it only for prologue/epilogue instructions at this moment. I added the comment to the field description.

cranelift-codegen/src/isa/x86/abi.rs
bnjbvr: Right. There should be @julian-seward1 's late stage optimizations happening after that (and they can remove instructions), so we should be careful to put this somewhere the data flow graph can't change (i.e. at the very end).

bnjbvr:
Deferring to @sunfishcode, since my understanding is that this is mostly for wasmtime.

bjorn3:
This is useful for cg_clif too for debuginfo and unwinding.

cranelift-codegen/src/isa/x86/abi.rs
```diff
@@ -547,25 +635,28 @@ fn insert_common_epilogues(
...
+                insert_common_epilogue(inst, stack_size, pos, reg_type, csrs, &cfa_state);
             }
         }
     }
 }
 
 /// Insert an epilogue given a specific `return` instruction.
 /// This is used by common calling conventions such as System V.
+/// TODO implement and handle _cfa_state more than one epilogue.
 fn insert_common_epilogue(
```
iximeow: Unlikely but important edge case, there is a `FrameLayoutChange` missing for the `pop rbp` added here. If `FrameLayoutChange` are used for unwinding, a signal delivered _at_ the following `ret`, it would unwind with the wrong frame pointer and probably cause the unwinder to abort or segfault. I think a `FrameLayoutChange::CallFrameAddressAt { reg: RU::rsp, as RegUnit, offset: 0 }` would be appropriate to refer to the current frame (eg the span starting at the return address, `[rsp]`), like the one for the function's initial layout

sunfishcode:
I think this patch looks good overall, though [@iximeow's comment](https://github.com/CraneStation/cranelift/pull/679/files#r314517618) looks worth addressing.

iximeow:
Just a note: the above merged PR adds frame changes to track the last few instructions as we exit functions on x86. I'd forgotten when commenting but we _can_ have multiple returns, so it's a little more complex than I'd initially thought. I'm not sure how to get multi-return sys-v functions out of Cranelift to test, but to my understanding of the spec (and a comparison against gcc) it looks correct.

So much as my word matters here: these changes look :heavy_check_mark:

cranelift-codegen/src/isa/x86/abi.rs
```diff
@@ -300,6 +300,13 @@ fn baldrdash_prologue_epilogue(func: &mut ir::Function, isa: &dyn TargetIsa) ->
     Ok(())
 }
 
+#[derive(Clone)]
+struct CFAState {
+    cf_ptr_reg: RegUnit,
+    cf_ptr_offset: isize,
+    current_depth: isize,
```
bjorn3: Please add doc comments. I don't know what `current_depth` is used for.

cranelift-codegen/src/isa/x86/abi.rs
```diff
@@ -547,43 +643,115 @@ fn insert_common_epilogues(
...
 ) {
     while let Some(ebb) = pos.next_ebb() {
         pos.goto_last_inst(ebb);
         if let Some(inst) = pos.current_inst() {
             if pos.func.dfg[inst].opcode().is_return() {
-                insert_common_epilogue(inst, stack_size, pos, reg_type, csrs);
+                // figure out if we need to insert end-of-function-aware frame layout information
+                let following_inst = pos.next_ebb().and_then(|next_ebb| {
+                    pos.goto_first_inst(next_ebb);
+                    pos.current_inst()
```
bjorn3: You can use [`pos.func.layout.first_inst`](https://docs.rs/cranelift-codegen/0.43.1/cranelift_codegen/ir/layout/struct.Layout.html#method.first_inst) instead of `goto_first_inst`, `current_inst` and `goto_last_inst`.

cranelift-codegen/src/isa/x86/abi.rs
```diff
@@ -547,43 +643,115 @@ fn insert_common_epilogues(
...
+            reg: cfa_state.cf_ptr_reg,
+            offset: cfa_state.cf_ptr_offset,
+        };
+
+        frame_layout
+            .instructions
+            .entry(inst)
+            .and_modify(|insts| {
+                let mut new_instructions = insts.to_vec();
+                new_instructions.push(new_cfa);
```
bjorn3:
```rust
let new_instructions = insts.iter().chain(std::iter::once(new_cfa)).collect::<Box<[_]>>();
*insts = new_instructions
```

Will save one or two allocation. Before it would have to reallocate the vec when pushing if `new_instructions.len() == new_intructions.capacity()` and in case of excess capacity when turning it back into a boxed slice, after it would just allocate a `Box<[_]>` with the correct size.

cranelift-codegen/src/isa/x86/abi.rs
```diff
@@ -547,43 +643,115 @@ fn insert_common_epilogues(
...
+                *insts = new_instructions.into_boxed_slice();
+            })
+            .or_insert_with(|| vec![new_cfa].into_boxed_slice());
```
bjorn3:
```suggestion
            .or_insert_with(|| Box::new([new_cfa]) as Box<[_]>);
```

Before the `vec!` macro created a boxed slice, turned it into a vec and then you turned it into a boxed slice again. This change should not matter for perf in release mode, but could save a bit of time during compilation for optimizing it.

cranelift-codegen/src/isa/x86/abi.rs
iximeow: ah so `following_inst` would be
```
let following_inst = pos.func.layout().next_ebb(ebb)
  .and_then(|next_ebb| pos.func.layout().first_inst(next_ebb))
```
? that cleans this up nicely.

cranelift-codegen/src/ir/framelayout.rs
```diff
@@ -0,0 +1,73 @@
+//! Frame layout item changes.
+
+use crate::ir::entities::Inst;
+use crate::isa::RegUnit;
+use std::boxed::Box;
+
+#[cfg(not(feature = "std"))]
+use hashmap_core::HashMap;
```
jyn514:
```suggestion
use crate::HashMap;
```

This should work both before and after my PR.

iximeow:
I just realized that I tried to re-request a review on this a few weeks ago, but it didn't go through. Maybe I can't re-request reviews for other folks' PRs? Anyway, if you have a moment @bjorn3 I'd like to get this one sorted out and resume getting #902 ship-shape soon.

cranelift-codegen/src/isa/x86/abi.rs
```diff
@@ -547,43 +660,115 @@ fn insert_common_epilogues(
...
+            offset: cfa_state.cf_ptr_offset,
+        };
+
+        frame_layout
+            .instructions
+            .entry(inst)
+            .and_modify(|insts| {
+                *insts = insts
+                    .into_iter()
+                    .map(|x| *x)
```
bjorn3: .cloned()

bjorn3:
@iximeow I am not a member of `@CraneStation`.

iximeow:
oh whoops, I'd thought you were. Sorry about that!

bnjbvr:
Hi! Just as a preamble to the review, can you make sure this commit doesn't introduce merge commits please? (Rebasing should work just fine.)

bnjbvr:
@yurydelendik Hey, can you help with this rebase, please? (Once that's done, I'm happy to review this.)

yurydelendik:
>  can you help with this rebase

rebased all commits

cranelift-codegen/src/isa/x86/abi.rs
```diff
@@ -618,43 +731,115 @@ fn insert_common_epilogues(
...
 ) {
+    let word_size = isa.pointer_bytes() as isize;
     if stack_size > 0 {
         pos.ins().adjust_sp_up_imm(Imm64::new(stack_size));
     }
 
     // Pop all the callee-saved registers, stepping backward each time to
     // preserve the correct order.
     let fp_ret = pos.ins().x86_pop(reg_type);
+    // account for CFA state in the reverse of `insert_common_prologue`.
```
bnjbvr: nit: for all the comments here, please capitalize the first letter.

cranelift-codegen/src/isa/x86/abi.rs
```diff
@@ -618,43 +731,115 @@ fn insert_common_epilogues(
...
+
+        frame_layout
+            .instructions
+            .entry(inst)
+            .and_modify(|insts| {
+                *insts = insts
+                    .into_iter()
+                    .map(|x| *x)
+                    .chain(std::iter::once(new_cfa))
+                    .collect::<Box<[_]>>();
```
bnjbvr: Any reason we don't use a vec! for the changes? We could then use `push` here.

cranelift-codegen/src/isa/x86/abi.rs
```diff
@@ -618,43 +731,115 @@ fn insert_common_epilogues(
...
     }
 
     // Pop all the callee-saved registers, stepping backward each time to
     // preserve the correct order.
     let fp_ret = pos.ins().x86_pop(reg_type);
+    // account for CFA state in the reverse of `insert_common_prologue`.
+    cfa_state.current_depth += word_size;
+    cfa_state.cf_ptr_offset -= word_size;
+    // and now that we're going to overwrite `rbp`, `rsp` is the only way to get to the call frame.
+    cfa_state.cf_ptr_reg = RU::rsp as RegUnit;
```
bnjbvr: Don't we need to insert the layout changes after this instruction?

cranelift-codegen/src/isa/x86/abi.rs
```diff
@@ -371,6 +372,29 @@ fn baldrdash_prologue_epilogue(func: &mut ir::Function, isa: &dyn TargetIsa) ->
...
+    /// is the offset from that pointer to the address of the start of this function's call frame.
+    ///
+    /// For a concrete x86_64 example, we will start this at 8 - the call frame begins immediately
+    /// before the return address. This will typically then be set to 16, after pushing `rbp` to
+    /// preserve the parent call frame. It is very unlikely the offset should be anything other
+    /// than one or two `usize`.
+    cf_ptr_offset: isize,
+    /// The offset between the start of the call frame and the current stack pointer. This is
+    /// primarily useful to point to where on the stack preserved registers are, but is maintained
+    /// through the whole function for consistency.
```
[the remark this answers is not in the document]
bjorn3: We dont have alloca support, so both esp and ebp arent modified between prologue and epilogue, right?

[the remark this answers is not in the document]
bnjbvr: I was thinking about regalloc's spills, but maybe regalloc just mandates a specific amount of spill slots, and something else allocates them in advance once and for all?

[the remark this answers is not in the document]
bjorn3: Spill slots are a kind of stack slots. I believe they get allocated together with the rest of the stack slots in prologue.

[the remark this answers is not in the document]
bnjbvr: And i imagine the emergency spill slot belongs to that group? In this case, current comment is fine.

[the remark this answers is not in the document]
bjorn3: Yes, it is a kind of stack slot: https://docs.rs/cranelift-codegen/0.46.1/cranelift_codegen/ir/stackslot/enum.StackSlotKind.html#variant.EmergencySlot

cranelift-codegen/src/isa/x86/abi.rs
```diff
@@ -618,43 +731,115 @@ fn insert_common_epilogues(
...
+    cfa_state.cf_ptr_reg = RU::rsp as RegUnit;
+
     pos.prev_inst();
 
     pos.func.locations[fp_ret] = ir::ValueLoc::Reg(RU::rbp as RegUnit);
     pos.func.dfg.append_inst_arg(inst, fp_ret);
 
     for reg in csrs.iter(GPR) {
         let csr_ret = pos.ins().x86_pop(reg_type);
+        cfa_state.current_depth += word_size;
```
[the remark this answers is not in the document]
iximeow: We *could*, but it's not strictly necessary - this is in the epilogue, so the preserved register won't be changed until we return, and the value on the stack won't be overwritten by subsequent instructions. Not that "no one else does!" is a great argument, but I don't think there's anything to gain from encoding this while it comes at the cost of an additional ~3 bytes per callee-save per function, unless you forsee Cranelift doing some really funky stuff in prologues one day?

cranelift-codegen/src/isa/x86/abi.rs
iximeow: https://github.com/CraneStation/cranelift/pull/679#discussion_r327762353 :grimacing:

bjorn3: That was based on the fact that `FrameLayoutChanges` is a `Box<[_]>` and not a `Vec<_>`. Using `Vec<_>` for `FrameLayoutChanges` and then using `pop` would be faster. I didn't actually look why there was a `Box<[_]>` in the first place when writing that comment.

cranelift-codegen/src/isa/x86/abi.rs
iximeow: Great eye, https://github.com/CraneStation/cranelift/pull/679/files/f6582f0eafb1c9202b589944c59f4d8ffbb1c311#diff-08ef6b0c3d44ab300b5158dab40ea830R835 should be `fp_ret`.

cranelift-codegen/src/isa/x86/abi.rs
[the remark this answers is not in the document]
bnjbvr: Fair enough, thanks for explaining.

cranelift-codegen/src/isa/x86/abi.rs
bnjbvr: Ok, we could use a Vec here, or keep the Box. Not sure it's very performance-sensitive since it's tied to debug collection mode.

bjorn3: > Not sure it's very performance-sensitive since it's tied to debug collection mode.

It would be perf sensitive for cg_clif once I add unwinding support based on this.

iximeow:
@yurydelendik will you have time to address the comments here or should I do so and PR into your fork again?

yurydelendik:
> @yurydelendik will you have time to address the comments here or should I do so and PR into your fork again?

@iximeow at this moment, I lost track on what is going on with this PR. It will take me some time to address the review comments. If you can address the comments, which will be awesome, I recommend to just submit your fork's branch as a new/different PR -- and I'll close this one.

iximeow:
I'll open a new PR and address the remaining comments there. Thanks!

bnjbvr:
Thanks both! I'll close this one since work has been continued in #1204.

PR #683, opened 2019-02-22
Fix bitrev intrinsic

### LINKED ISSUE #652: Bitreverse intrinisic seems to work wrong
```rust
assert_eq!(
    core::intrinsics::bitreverse(0b10101000u8),
                                 0b00010101u8,
);
```

panics

cc https://github.com/bjorn3/rustc_codegen_cranelift/issues/168

robojumper:
The incorrect version references a [stackoverflow answer](https://stackoverflow.com/questions/746171/most-efficient-algorithm-for-bit-reversal-from-msb-lsb-to-lsb-msb-in-c). The algorithm described there appears to be correct, but the transcription in cranelift contained one error where it would use a right-shift instead of a left-shift. I have fixed this error and the intrinsic now seems to work as intended.

Fixes #652.


sunfishcode:
Great spot!

PR #687, opened 2019-02-24
Implement sshr_imm.i8

bjorn3:


cranelift-codegen/meta-python/base/legalize.py
```diff
@@ -375,39 +375,14 @@ def widen_imm(signed, op):
...
-        Rtl(
-            x << uextend.i32(b),
-            z << ushr.i32(x, c),
-            a << ireduce.bind(int_ty)(z)
-        ))
-
-    widen.legalize(
-        a << sshr.bind(int_ty)(b, c),
-        Rtl(
-            x << sextend.i32(b),
```
bnjbvr: Note this uses `sextend` here, not `uextend`, so it can't be refactored with the rest. I assume `sshr_imm` will need to use `sextend` too.

bjorn3: Oops, fixed.

bnjbvr:
Thanks, lgtm!

PR #692, opened 2019-03-02
Fix emit_small_memset

bjorn3:
It used to try to extend an `i32` to `i32`.


jgouly:
Is there a test you could add?

bjorn3:
`cranelift-frontend` unit tests don't seem to be run.

bnjbvr:
> cranelift-frontend unit tests don't seem to be run.

Great catch! Tests stopped being run when we switched from a namespace to... no more namespace.

sunfishcode:
Great! Thanks for finding and fixing this!

Issue #691, opened 2019-02-27
The x64 calling convention (windows) is implemented incorrectly.


lachlansneff:
As per the microsoft [documentation](https://docs.microsoft.com/en-us/cpp/build/x64-calling-convention?view=vs-2017#example-of-argument-passing-3---mixed-ints-and-floats), floating point registers are picked by their general parameter index, not their fp parameter index.

bnjbvr:
Should be fixed by #693, please re-open if not. Thanks for reporting this issue!

CryZe:
Cool, I'll give it a try later today. Thanks for fixing this :)

CryZe:
https://github.com/CraneStation/cranelift/pull/693/files#diff-08ef6b0c3d44ab300b5158dab40ea830R122

The debug_assert_eq! here is firing for me now. 
```
thread '<unnamed>' panicked at 'assertion failed: `(left == right)`
  left: `2`,
 right: `1`', cranelift\cranelift-codegen\src\isa\x86\abi.rs:122:13
```

So this doesn't seem to work yet, @bnjbvr

CryZe:
I roughly figured out the problem
```rust
/// Return value registers for x86-64, when using windows fastcall
static RET_GPRS_WIN_FASTCALL_X64: [RU; 1] = [RU::rax];
```
this has a single register, but is being passed to Arg::new here
```rust
    let regs = if sig.call_conv == CallConv::WindowsFastcall {
        &RET_GPRS_WIN_FASTCALL_X64[..]
    } else {
        &RET_GPRS[..]
    };

    let mut rets = Args::new(bits, regs, 2, sig.call_conv);
```
which causes the assert to eventually trigger here
```rust
debug_assert_eq!(self.fpr_limit, self.gpr.len());
```
as the fpr_limit is the 2 from the Args::new and the gpr is the static regs array that has a single register.

steffengy:
Likely a function with a float return value, that isn't covered by tests and trips up here since the same code is used for register assignment, I'll look into that case as well.

PR #693, opened 2019-03-02
Use single index for param register allocation for windows callconv (…

steffengy:
…#691)

The used registers depend entirely on the parameter index (1st, 2nd, 3rd, 4th, ... param)
and we cannot shift unused registers to other indexes, if they are not designated for
the use for that parameter index.

### DIFF
--- a/cranelift-codegen/src/isa/x86/abi.rs
+++ b/cranelift-codegen/src/isa/x86/abi.rs
@@ -116,10 +116,10 @@ impl ArgAssigner for Args {
 
         // Try to use an FPR.
         let fpr_offset = if self.call_conv == CallConv::WindowsFastcall {
-            // For registers and Win64, the registers always depend on the parameter index.
-            // The first argument either uses RCX (int)/XMM0 (float) and the second RDX/XMM1.
-            // XMM0 is never used for the second parameter, even if the first parameter uses RCX
-            // and XMM0 theoretically would still be "available".
+            // Float and general registers on windows share the same parameter index.
+            // The used register depends entirely on the parameter index: Even if XMM0
+            // is not used for the first parameter, it cannot be used for the second parameter.
+            debug_assert_eq!(self.fpr_limit, self.gpr.len());
             &mut self.gpr_used
         } else {
             &mut self.fpr_used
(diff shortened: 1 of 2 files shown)


jgouly:
LGTM. You could change the last parameter to `f32` if you wanted to match the example in the Microsoft docs.

steffengy:
@jgouly Right, done.

cranelift-codegen/src/isa/x86/abi.rs
```diff
@@ -115,9 +115,18 @@ impl ArgAssigner for Args {
...
+        let fpr_offset = if self.call_conv == CallConv::WindowsFastcall {
+            // For registers and Win64, the registers always depend on the parameter index.
+            // The first argument either uses RCX (int)/XMM0 (float) and the second RDX/XMM1.
+            // XMM0 is never used for the second parameter, even if the first parameter uses RCX
+            // and XMM0 theoretically would still be "available".
+            &mut self.gpr_used
+        } else {
+            &mut self.fpr_used
+        };
+        if ty.is_float() && *fpr_offset < self.fpr_limit {
```
bnjbvr: There's an implicit invariant for the Win64 fastcall convention, assumed here, that `fpr_limit` is equal to `gpr.len()` (which is true). What do you think about adding a `debug_assert!` in the then branch of the `if self.call_conv == CallConv::WindowsFastcall` to ensure it's true and make it more explicit?

cranelift-codegen/src/isa/x86/abi.rs
```diff
@@ -115,9 +115,18 @@ impl ArgAssigner for Args {
...
 
         // Try to use an FPR.
-        if ty.is_float() && self.fpr_used < self.fpr_limit {
-            let reg = FPR.unit(self.fpr_used);
-            self.fpr_used += 1;
+        let fpr_offset = if self.call_conv == CallConv::WindowsFastcall {
+            // For registers and Win64, the registers always depend on the parameter index.
+            // The first argument either uses RCX (int)/XMM0 (float) and the second RDX/XMM1.
+            // XMM0 is never used for the second parameter, even if the first parameter uses RCX
+            // and XMM0 theoretically would still be "available".
```
bnjbvr: This is nicely exhaustive. As a light suggestion, we could make it shorter by telling "Float and general registers on Win64 share the same parameter index, independently from the fact they're actually used or not" (of course, since I'm no native speaker, please fix grammar issues / reformulate to make it sound better etc. if you address this :))

bnjbvr:
That's great, thank you for addressing those!

PR #694, opened 2019-03-04
Migrate Instructions to the Rust meta crate;

bnjbvr:
Not finished yet, but since other people might be interested to work on the Python stuff, let's try to limit stepping on each other's toes as much as we can :)

After this PR, the remaining elements to be ported will be binemit and legalization, at least.


bnjbvr:
I can split this PR into smaller ones (one for TypeVar, one for Operand based off TypeVar's, one for InstructionFormat based off Operand's, Instruction based on all of that), and/or split into smaller commits. Do people have a preference? @sunfishcode maybe?  (also cc @jgouly who might be working on this too)

jgouly:
This change is big, would be nice to see it broken into smaller commits for sure! (and possibly committed under separate PRs too)

sunfishcode:
I agree, it would be nice to split this up, so that we can think about the layering going on here. The TypeVar, Operand, InstructionFormat, Instruction sequence you describe sounds like a good sequence to do that with. Then, if that turns out to be awkward or a lot of work, it'll be interesting to learn why that happens :-).

bnjbvr:
Small status report: haven't properly split the commits yet, but this can generate all the Cranelift instructions, including x86 specific ones. It doesn't generate them correctly, though (I've been tricked by the Python code using `is` and I've mistranslated this to structural equality in Rust), which is unfortunate, but this should be pretty easy to debug.

I'll fix this and will split the PR correctly early next week.

cranelift-codegen/meta/src/lib.rs
```diff
@@ -40,8 +40,8 @@ pub fn generate(isas: &Vec<isa::Isa>, out_dir: &str) -> Result<(), error::Error>
     gen_inst::generate(
         all_inst_groups,
         &shared_defs.format_registry,
-        "new_opcodes.rs",
-        "new_inst_builder.rs",
+        "opcodes.rs",
+        "inst_builder.rs",
```
bjorn3: Could you please remove generation of these files from meta-python at https://github.com/CraneStation/cranelift/blob/e9a81f6cce11c8115da3116bbfb205fb4e04f962/cranelift-codegen/meta-python/gen_instr.py#L794-L812

bnjbvr: I can remove most of gen_instr, but some functions it contains are still in use by the legalizer; I'll move them there.

bnjbvr:
This is ready for review. I've tried to split commits as much as I can (and made sure that each commit could be compiled independently, but even though I've been careful, some commits might be just adding code that's not compiled).

It appears that Operand and InstructionFormat are tightly coupled, because there's a global hash map called the FormatRegistry which contains all the existing InstructionFormat instances, which keys are derived from Operand fields. The commit containing those also contains all the InstructionFormat definitions, copied from Python code.

The commit that introduces Instruction/InstructionGroup in the series also shows how x86 specific instructions are implemented. The next commit porting all the remaining instructions has been automatically generated (and then manually fixed for the few errors it contained), using the Python meta-program itself.

The last commit enables usage of the new generator, which passes all Cranelift tests, and all Spidermonkey tests too (when Cranelift is used as the code generator for wasm).

bnjbvr:
This is now based off https://github.com/CraneStation/cranelift/pull/707.

bnjbvr:
To tell just a bit more about a possible order for review: #705, #706, #707 then this one. (Or if you want to have the full look, directly this one)

There's one thing I haven't reproduced from the Python code, which is the numbering of instructions: they're numbered as they're implemented. This is an implicit contract between the Rust crate and the Python code at the moment; it is going to be explicit when other meta code is ported later to Rust, and this meta code makes use of the instruction order. In the meanwhile, the instruction has been manually preserved (by inserting shared instructions first, then x86 specific ones).

bnjbvr:
All the other dependencies have been checked in; the last two commits did the following:

1. generate new equivalent of `opcodes.rs` and `inst_builder.rs` (prefixed with `new_`, as for the settings) based on the Rust code.
2. remove the Python generation and use the Rust code to generate these two files.

Now that I'm deep down in the legalization part, I've removed the second commit, because some of the Rust data structures (TypeVar, Inst, at least) will need to be updated again, and that could regress the code generation. So I propose to keep the Python generation until the final switch to the Rust crate, in the meanwhile we can compare that the Rust and Python code still generate the same things. (I've kept the second commit around, should we decide otherwise)

@pchickey @jgouly (or anybody interested in this) Would you be interested in reviewing this last part? It's the Rust port of the Python code `gen_instr.py`. Thanks!

cranelift-codegen/meta/src/gen_inst.rs
```diff
@@ -0,0 +1,1092 @@
...
+
+        All instructions from all supported ISAs are present.
+    "#,
+    );
+    fmt.line("#[derive(Copy, Clone, PartialEq, Eq, Debug, Hash)]");
+
+    // We explicitly set the discriminant of the first variant to 1, which
+    // allows us to take advantage of the NonZero optimization, meaning that
+    // wrapping enums can use the 0 discriminant instead of increasing the size
+    // of the whole type, and so SIZEOF(Option<Opcode>>) == SIZEOF(Opcode).
```
jgouly: std::mem::size_of::<Option<Opcode>() :P

jgouly:
LGTM, this one is pretty close to the original python.

The only question I have is this difference:
```
+ const OPERAND_CONSTRAINTS: [OperandConstraint; 78] = [
- const OPERAND_CONSTRAINTS: [OperandConstraint; 89] = [
```

Do you know if this incorrect?

cranelift-codegen/meta/src/gen_inst.rs
bjorn3: ^ missing second `>`

cranelift-codegen/meta/src/gen_inst.rs
```diff
@@ -0,0 +1,1092 @@
...
+fn gen_arguments_method(registry: &FormatRegistry, fmt: &mut Formatter, is_mut: bool) {
+    let (method, mut_, rslice, as_slice) = if is_mut {
+        ("arguments_mut", "mut ", "ref_slice_mut", "as_mut_slice")
+    } else {
+        ("arguments", "", "ref_slice", "as_slice")
+    };
+
+    fmtln!(
+        fmt,
+        "pub fn {}<'a>(&'a {}self, pool: &'a {} ir::ValueListPool) -> &{}[Value] {{",
```
bnjbvr: nit: remove space between `{}` and `ir::ValueListPool`. (Thanks @jgouly !)

bnjbvr:
> Do you know if this incorrect?

The Python implementation of UniqueSeqTable.add is imperfect; I can see for instance that to find the constraint offset for InsertLane, the Python impl will re-insert the sequence, while it's already there. The Rust implementation, while naive and slow, figures it out properly, so it reuses the sequence. This results in the Rust impl inserting fewer elements in the table.

It's because the Python impl is only aware of all the subsequences of a subsequence, and it cannot tie together two arbitrary subsequences, while the Rust impl does. (Say you've inserted (A, B) and (C, D) and then want to insert (A, B, C, D); the Python impl will reinsert it because it only knows indexes for the sequences ((A), (B), (C), (D), (A,B), (B,C)), while the Rust one will enumerate all the values and find that the sequence already exists starting at A.)

I've also spotted an opportunity for improvement: when the sequence table ends with (A, B), and we want to insert (B, C) in the table, we could reuse the existing B and just insert C. The current implementation will insert (B, C) again just after (A, B).

bnjbvr:
> I've also spotted an opportunity for improvement: when the sequence table ends with (A, B), and we want to insert (B, C) in the table, we could reuse the existing B and just insert C. The current implementation will insert (B, C) again just after (A, B).

I've implemented this locally, and this reduces the size from 78 to 71. It's not very complex, but I'll file a separate PR for this.

bnjbvr:
We had a discussion with Dan and agreed to merge this, enabling it in place of the Python code, so that the Rust code doesn't stay dormant (and rot!). I'll do a manual visual diffing first to make sure that both files coincide, and then will merge this today.

PR #696, opened 2019-03-05
Fixes #695: Reintroduce a workspace declaration to run all tests

### LINKED ISSUE #695: Re-enable testing for all crates
`cargo test --all` runs all tests for all crates in a given workspace, but we've stopped using workspace recently when we moved to directories. So we can just use a small script to find Cargo.toml files and run tests for the crates we've found along the way.

bnjbvr:


sunfishcode:
r+. Thanks for adding a comment!

PR #701, opened 2019-03-09
windows/x64 call convention: only use XMM0 for float return values

steffengy:
Previously we triggered the debug_assert for functions  returning floats.
This adds a testcase and fixes the behavior.

#691


bnjbvr:
Good catch, thanks for adding a test case too! Another possible fix would have been to tweak the assert so that `num float regs <= num GPRs`, but being precise is even better. Cheers!

PR #702, opened 2019-03-11
Beginnings of arm32.

carolinecullen:
Initial PR of ARM32 so that the unimplemented() call no longer happens and development can begin.

### DIFF
--- a/cranelift-codegen/src/isa/arm32/abi.rs
+++ b/cranelift-codegen/src/isa/arm32/abi.rs
@@ -1,18 +1,103 @@
 //! ARM ABI implementation.
 
 use super::registers::{D, GPR, Q, S};
-use crate::ir;
+use crate::abi::{legalize_args, ArgAction, ArgAssigner, ValueConversion};
+use crate::ir::{self, AbiParam, ArgumentExtension, ArgumentLoc, ArgumentPurpose, Type};
 use crate::isa::RegClass;
 use crate::regalloc::RegisterSet;
-use crate::settings as shared_settings;
+use core::i32;
+use target_lexicon::Triple;
+
+struct Args {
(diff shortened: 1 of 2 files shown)


jgouly:
This seems to be copied from RISC-V, I think it's worth removing the bits that are obviously wrong for ARM.

jgouly:
Just to clarify, I didn't mean to close the PR, just to modify the contents (in case my message was misunderstood)

sunfishcode:
I think you're both right :-). At this early phase in development, it's ok to fill in things like the ABI-lowering code just enough to support testcases for instruction encodings, and it's fine to start with RISC-V code for those parts here, even if we'll eventually need to revisit it. So I think if you just add comments saying "This is from the RISC-V target and will need to be updated for ARM32", that should help us avoid getting confused about it later on.

cranelift-codegen/src/isa/arm32/abi.rs
```diff
@@ -1,18 +1,103 @@
...
+pub fn legalize_signature(sig: &mut ir::Signature, triple: &Triple, current: bool) {
+    let bits = triple.pointer_width().unwrap().bits();
+
+    let mut args = Args::new(bits);
+    legalize_args(&mut sig.params, &mut args);
+
+    let mut rets = Args::new(bits);
+    legalize_args(&mut sig.returns, &mut rets);
+
+    if current {
```
jgouly: What about removing this bit as well? I think it may need similar handling eventually, but can probably be left off to start with (and it's mentioning jalr and x1 which are riscv)

sunfishcode:
Cool, thanks for cleaning up the RISC-V parts, and adding comments!

PR #705, opened 2019-03-12
Port Instructions to the meta crate : part 1, prerequisites

bnjbvr:


cranelift-codegen/meta/src/srcgen.rs
```diff
@@ -151,8 +164,27 @@ impl Formatter {
                         }
                     })
                     .collect();
-                let lhs = conditions.join(" | ");
-                fmtln!(fmt, "{} => {{", lhs);
+
+                for (i, condition) in conditions.iter().enumerate() {
```
jgouly: Is there any point in caring about the output? Can't we just run rustfmt on the file if we want to read it?

cranelift-codegen/meta/src/unique_table.rs
```diff
@@ -1,15 +1,54 @@
+use std::collections::HashMap;
+use std::hash::Hash;
 use std::slice;
 
+/// Collect items into the `table` list, removing duplicates.
+pub struct UniqueTable<'entries, T: Eq + Hash> {
```
jgouly: Maybe at some point this can be replaced with something from crates.io? However I know sunfish wants to keep the dependencies down and it's pretty simple as-is..

bnjbvr: Yeah, it is quite simple. Now that I think about it, I even wonder why a simple BTreeMap couldn't be used (and an OrderedDict in the Python code).

cranelift-codegen/meta/src/srcgen.rs
bnjbvr: Running rustfmt on the output (from the build program itself) is a fun idea! In the meanwhile, we've tried to format code ourselves properly; using `join` here is probably clearer (I don't what why I reimplemented it from scratch myself).

cranelift-codegen/meta/src/unique_table.rs
jgouly: Probably IndexSet or IndexMap, I think it's quite a popular crate!

jgouly:
Also what about sprinkling some `#[allow(dead_code)]` in a  few places until things are used, so we don't get all the warnings? Either that or wait until everything is reviewed before committing.

bnjbvr:
> Also what about sprinkling some #[allow(dead_code)] in a few places until things are used, so we don't get all the warnings? Either that or wait until everything is reviewed before committing.

I appreciate your care of having code compiling without warnings, but unless this is **really really** wanted, I would like to avoid spending time massaging these PRs, considering that now it's pretty hard to do so. Each change I'd apply here needs to be dispatched to the right commit, and will inevitably break the next commits waiting in line, which will need to undo what was added to massage the PR for landing. For instance, I've noticed an improvement I could do on top of the entire Instruction PR (the biggest one), and the code changes had to be split into 4 commits, which had to be fused with each existing commit: it was all very time consuming while not adding a lot of value. So I'd like to avoid rebase hell here again ;)

cranelift-codegen/meta/src/srcgen.rs
bnjbvr: The reason I was reimplementing join by hand was because it resulted in a multi-line string, but we have the `multi_line` method designed exactly for this case, so I could use that.

cranelift-codegen/meta/src/unique_table.rs
bnjbvr: BTreeMap required Ord on TypeSet, which they don't obviously implement; UniqueTable is a bit ugly but it's quite small, it works fine at this time and should be fast enough, so I would think it's fine to keep it (and avoid adding a new dependency).

jgouly:
I think if you put an #[allow(dead_code)] in the meta crate in part 1 and removed it in part 4 that would cover probably everything. My suggestion was to try allow you to get this committed sooner and just reduce the compilation noise, so you didn't rebase too often etc (already said this on irc, but writing here for redundancy :))

Regardless, LGTM!

bnjbvr:
Uhh, I would have expected Github to catch that this was merged with #706. Closing.

PR #706, opened 2019-03-12
Port Instructions to meta crate : part 2, typevar and operands

bnjbvr:
Based off https://github.com/CraneStation/cranelift/pull/705

### DIFF
--- a/cranelift-codegen/meta/src/cdsl/typevar.rs
+++ b/cranelift-codegen/meta/src/cdsl/typevar.rs
@@ -1,0 +1,738 @@
+use std::collections::BTreeSet;
+use std::iter::FromIterator;
+use std::ops;
+use std::rc::Rc;
+
+use crate::cdsl::types::{BVType, LaneType, SpecialType, ValueType};
+
+const MAX_LANES: u16 = 256;
+const MAX_BITS: u16 = 64;
+const MAX_BITVEC: u16 = MAX_BITS * MAX_LANES;
+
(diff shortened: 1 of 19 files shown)


cranelift-codegen/meta/src/cdsl/typevar.rs
```diff
@@ -0,0 +1,743 @@
...
+        if self.lanes.contains(&1) {
+            copy.bools.insert(1);
+        }
+        copy
+    }
+
+    /// Return a TypeSet describing the image of self across halfwidth.
+    fn half_width(&self) -> TypeSet {
+        let mut copy = self.clone();
+        copy.ints = NumSet::from_iter(self.ints.iter().map(|&x| if x > 8 { x / 2 } else { x }));
```
jgouly: This seems slightly different from the Python, but maybe it doesn't make a difference in practise.

    [x//2 for x in self.ints if x > 8]
 
This filters out elements less than 8, but the rust code keeps them as-is?

jgouly:
Looked just at the Typevar stuff, it's pretty dense. Looks good, pretty much a straight port from Python.

cranelift-codegen/meta/src/cdsl/typevar.rs
bnjbvr: Good catch, I probably need to filter before the map (the cases where this was applied probably had both 8 and 16 in the int set, hiding this issue).

bnjbvr:
Thanks for the reviews!

PR #707, opened 2019-03-12
Port Instructions to meta ; part 3, Instruction/InstructionGroup + definitions

bnjbvr:
Based off https://github.com/CraneStation/cranelift/pull/706

### DIFF
--- a/cranelift-codegen/meta/src/cdsl/inst.rs
+++ b/cranelift-codegen/meta/src/cdsl/inst.rs
@@ -1,0 +1,458 @@
+use crate::cdsl::camel_case;
+use crate::cdsl::formats::{FormatRegistry, InstructionFormat, InstructionFormatIndex};
+use crate::cdsl::operands::Operand;
+use crate::cdsl::type_inference::Constraint;
+use crate::cdsl::typevar::TypeVar;
+
+use std::fmt;
+use std::slice;
+
+/// Every instruction must belong to exactly one instruction group. A given
+/// target architecture can support instructions from multiple groups, and it
+/// does not necessarily support all instructions in a group.
+pub struct InstructionGroup {
+    _name: &'static str,
+    _doc: &'static str,
+    instructions: Vec<Instruction>,
+}
+
+impl InstructionGroup {
+    pub fn new(name: &'static str, doc: &'static str) -> Self {
+        Self {
+            _name: name,
+            _doc: doc,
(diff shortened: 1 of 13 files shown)


cranelift-codegen/meta/src/cdsl/inst.rs
```diff
@@ -0,0 +1,459 @@
...
+impl Instruction {
+    pub fn snake_name(&self) -> &'static str {
+        if self.name == "return" {
+            "return_"
+        } else {
+            self.name
+        }
+    }
+
+    // XXX was named blurb
```
jgouly: I guess this can be removed?

cranelift-codegen/meta/src/cdsl/inst.rs
```diff
@@ -0,0 +1,459 @@
...
+    pub fn doc_comment_first_line(&self) -> &'static str {
+        for line in self.doc.split("\n") {
+            let stripped = line.trim();
+            if stripped.len() > 0 {
+                return stripped;
+            }
+        }
+        return "";
```
jgouly: Just replace with ""

cranelift-codegen/meta/src/cdsl/inst.rs
```diff
@@ -0,0 +1,459 @@
...
+            can_load: false,
+            can_store: false,
+            can_trap: false,
+            other_side_effects: false,
+        }
+    }
+
+    pub fn operands_in(mut self, operands: Vec<&Operand>) -> Self {
+        assert!(self.operands_in.is_none());
+        self.operands_in = Some(operands.iter().map(|x| (*x).clone()).collect());
```
jgouly: Can you use .cloned() here? (and below)

jgouly:
LGTM too, just a few very minor comments. I also didn't do a deep review of `cranelift-codegen/meta/src/shared/instructions.rs`, just randomly looked at some of the instructions. Hopefully tests will catch issues if there are any!

cranelift-codegen/meta/src/cdsl/inst.rs
bnjbvr: No, because the parameter `operands` is a `Vec<&Operand>` while the `operands_in` is a `Vec<Operand>`.

bnjbvr:
Thanks for the reviews!

PR #708, opened 2019-03-12
Fix #686: Allow code shrink in relaxation if the shrinking pass hasn't run

### LINKED ISSUE #686: thread 'rustc' panicked at 'Code shrinking during relaxation'

bnjbvr:
Also:
- make sure to apply diversions when determining offsets for code
relaxation.
- select the smallest encoding when selecting a relaxed branch
instruction.

I didn't know where to put the test case, so I added a "regress" directory for regression tests (found by fuzzers), but I could move it to another directory. (`isa/` didn't seem quite appropriate, but other relaxation tests are there, so maybe it is)

@sunfishcode Can you have a look, please? Thanks!

### DIFF
--- a/cranelift-codegen/src/binemit/relaxation.rs
+++ b/cranelift-codegen/src/binemit/relaxation.rs
@@ -62,6 +62,7 @@ pub fn relax_branches(func: &mut Function, isa: &TargetIsa) -> CodegenResult<Cod
             divert.clear();
             cur.func.offsets[ebb] = offset;
             while let Some(inst) = cur.next_inst() {
+                divert.apply(&cur.func.dfg[inst]);
                 let enc = cur.func.encodings[inst];
                 offset += encinfo.byte_size(enc, inst, &divert, &cur.func);
             }
@@ -81,10 +82,6 @@ pub fn relax_branches(func: &mut Function, isa: &TargetIsa) -> CodegenResult<Cod
 
             // Record the offset for `ebb` and make sure we iterate until offsets are stable.
             if cur.func.offsets[ebb] != offset {
-                debug_assert!(
-                    cur.func.offsets[ebb] < offset,
-                    "Code shrinking during relaxation"
-                );
                 cur.func.offsets[ebb] = offset;
                 go_again = true;
             }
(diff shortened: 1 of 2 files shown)


filetests/regress/allow-relaxation-shrink.clif
```diff
@@ -0,0 +1,53 @@
+test compile
+target x86_64-apple-darwin
```
bjorn3: maybe

```suggestion
target x86_64
```

cranelift-codegen/src/binemit/relaxation.rs
```diff
@@ -178,7 +183,7 @@ fn relax_branch(
     let ctrl_type = dfg.ctrl_typevar(inst);
```
jgouly: Change the comment to 'smallest', not first? (I can't seem to put a review note on the right line for some reason)

bnjbvr:
Updated commit. Props to @jgouly for suggesting to return a flag indicating if code could have shrunk; it's cleaner than the weird flag indicating whether code shrink ran or not; thank you!

cranelift-codegen/src/binemit/relaxation.rs
```diff
@@ -82,7 +85,7 @@ pub fn relax_branches(func: &mut Function, isa: &TargetIsa) -> CodegenResult<Cod
             // Record the offset for `ebb` and make sure we iterate until offsets are stable.
             if cur.func.offsets[ebb] != offset {
                 debug_assert!(
-                    cur.func.offsets[ebb] < offset,
+                    cur.func.offsets[ebb] < offset || could_have_shrunk,
                     "Code shrinking during relaxation"
                 );
```
philipc: This assert seems pointless to me now. How could the `offset` ever be less if `could_have_shrunk` is false, since `offset` is calculated using the same values that are used to determine `could_have_shrunk`?

bnjbvr: Yeah, it's a fair point. @sunfishcode Should we just remove the assertion?

cranelift-codegen/src/binemit/relaxation.rs
```diff
@@ -100,8 +103,10 @@ pub fn relax_branches(func: &mut Function, isa: &TargetIsa) -> CodegenResult<Cod
                     if let Some(dest) = cur.func.dfg[inst].branch_destination() {
                         let dest_offset = cur.func.offsets[dest];
                         if !range.contains(offset, dest_offset) {
-                            offset +=
+                            let (new_enc_size, did_shrunk) =
```
sunfishcode: Grammar nit: This should be spelled `did_shrink`.

cranelift-codegen/src/binemit/relaxation.rs
sunfishcode: Sure, removing it works for me.

bnjbvr: Cool, that even removes the need for the `could_have_shrunk` bool and in the signature of the relax function.

bjorn3:
:tada:

PR #709, opened 2019-03-15
[meta] Port recipes

bjorn3:
Based on #694

While `gen_binemit.py` has been ported, I still need to port `Encoding` and parts of `TargetIsa` to make it usable. Those are a bit harder as they seem to rely on the fact that python objects are not copied on assignment, but shared. Eg `EncRecipe` contains a `number` field which gets manipulated afterwards.

### DIFF
--- a/cranelift-codegen/meta/src/cdsl/formats.rs
+++ b/cranelift-codegen/meta/src/cdsl/formats.rs
@@ -200,6 +200,12 @@ impl FormatRegistry {
             .expect("unknown InstructionFormat; please define it in shared/formats.rs first")
     }
 
+    pub fn lookup_by_name(&self, name: &str) -> InstructionFormatIndex {
+        self.map.iter().find(|(_key, value)| {
+            value.name == name
+        }).unwrap().0
+    }
+
     pub fn get(&self, index: InstructionFormatIndex) -> &InstructionFormat {
         self.map.get(index).unwrap()
     }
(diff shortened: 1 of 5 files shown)


bnjbvr:
Hi @bjorn3 ! This is becoming increasingly important for us at Mozilla to finish the Rust port (so as to be able to add new backends), so that's the next item on my list of priorities. I don't want to impose our priorities onto you, so I propose the following:

- if you'd like to, you can rebase your work on top of the current master branch. (That means rebasing and other git foo to get rid of potential conflicts between commits marked as mine in this PR and commits that actually got merged.)
- then, if it compiles, we can land this, even if the generated output is unused and/or incomplete and/or broken.
- I'm happy to take over the rest of the work and finish it as soon as possible, so we can get rid of the Python code entirely.

Alternatively, if you don't have time to rebase and make sure it compiles, it is also fine; in this case, I'll start the work from scratch again. How does that sound?

bjorn3:
I rebases the `gen_binemit` part. I removed my experiment to actually port some encodings to rust meta. (You can find the old state of this PR in the `meta_port_encodings` branch of mine)

bjorn3:
Sorry for neglecting this PR for 2 months.

bnjbvr:
No worries, thanks for working on it in the first place, I appreciate it. I'll fix up this PR (code formatting!) and will merge it.

PR #713, opened 2019-03-25
Add a --disasm option to clif-util `wasm` and `compile`

lars-t-hansen:
Both the `wasm` and `compile` commands get this new subcommand, and it defaults to false.  This means that test runs with `wasm` can request disassembly (the main reason I am doing this) while test runs with `compile` now must request it, this changes current behavior.  Discuss.

### DIFF
--- a/src/clif-util.rs
+++ b/src/clif-util.rs
@@ -30,6 +30,7 @@ use std::process;
 
 mod cat;
 mod compile;
+mod disasm;
 mod print_cfg;
 mod utils;
 
@@ -67,6 +68,19 @@ fn add_time_flag<'a>() -> clap::Arg<'a, 'a> {
     Arg::with_name("time-passes")
         .short("T")
         .help("Print pass timing report for test")
+}
+
+fn add_size_flag<'a>() -> clap::Arg<'a, 'a> {
+    Arg::with_name("print-size")
+        .short("X")
+        .help("Print bytecode size")
+}
+
+fn add_disasm_flag<'a>() -> clap::Arg<'a, 'a> {
+    Arg::with_name("disasm")
+        .long("disasm")
+        .short("D")
+        .help("Disassemble machine code")
 }
 
 fn add_set_flag<'a>() -> clap::Arg<'a, 'a> {
@@ -120,6 +134,8 @@ fn add_wasm_or_compile<'a>(cmd: &str) -> clap::App<'a, 'a> {
         .arg(add_verbose_flag())
         .arg(add_print_flag())
         .arg(add_time_flag())
+        .arg(add_size_flag())
+        .arg(add_disasm_flag())
         .arg(add_set_flag())
         .arg(add_target_flag())
         .arg(add_input_file_arg())
@@ -226,6 +242,7 @@ fn main() {
             compile::run(
                 get_vec(rest_cmd.values_of("file")),
                 rest_cmd.is_present("print"),
+                rest_cmd.is_present("disasm"),
                 rest_cmd.is_present("time-passes"),
                 &get_vec(rest_cmd.values_of("set")),
                 target_val,
@@ -247,6 +264,7 @@ fn main() {
                     rest_cmd.is_present("just-decode"),
                     rest_cmd.is_present("check-translation"),
                     rest_cmd.is_present("print"),
+                    rest_cmd.is_present("disasm"),
                     &get_vec(rest_cmd.values_of("set")),
                     target_val,
                     rest_cmd.is_present("print-size"),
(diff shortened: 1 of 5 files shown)


jgouly:
All the travis tests passed, so it doesn't seem like any tests were relying on that behaviour?

Would be good to squash the commits, but otherwise seems fine to me!

src/wasm.rs
```diff
@@ -196,6 +206,21 @@ fn handle_module(
...
 
+        if let Some(compiled_size) = saved_compiled_size {
+            let mut mem = vec![0; compiled_size as usize];
+            let mut relocs = PrintRelocs { flag_print };
+            let mut traps = PrintTraps { flag_print };
+            let mut code_sink: binemit::MemoryCodeSink;
+            unsafe {
+                code_sink = binemit::MemoryCodeSink::new(mem.as_mut_ptr(), &mut relocs, &mut traps);
+            }
+            isa.emit_function_to_memory(&context.func, &mut code_sink);
```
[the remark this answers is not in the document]
lars-t-hansen: It's actually not completely fine that the order changes.  I thought it was going to be fine too, but from a reader's point of view it's desirable that the IR (which is the output of the "just an additional print") comes before the machine code.

Once you start looking at it that way, I believe there's little benefit in factoring this code, which is why I chose not to do it, and I'd prefer not to do it now either.

src/clif-util.rs
```diff
@@ -69,6 +70,19 @@ fn add_time_flag<'a>() -> clap::Arg<'a, 'a> {
...
+    Arg::with_name("print-size")
+        .short("X")
+        .help("Print bytecode size")
+}
+
+fn add_disasm_flag<'a>() -> clap::Arg<'a, 'a> {
+    Arg::with_name("disasm")
+        .long("disasm")
+        .short("D")
+        .help("Disassemble machine code")
```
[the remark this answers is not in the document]
lars-t-hansen: Sure, I can do that.

src/wasm.rs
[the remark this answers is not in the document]
lars-t-hansen: All that said, I see there's a compile_and_emit() function in context.rs that we could use instead (I think), and that would reduce the amount of duplicated code here by quite a bit.

lars-t-hansen:
I agree about refactoring this further into cranelift-codegen, I'd really like to see this in the SpiderMonkey shell.  But my preference would be to save that for another patch.

Take a look at the new refactoring, let me know what you think - it's cleaner now, plus I fixed some bugs with how flags were handled.

Still not happy about how the traps and relocations print, the offsets look wrong to me (eg, try to run `compile -D` on filetests/wasm/control.clif), but I don't think I'll go after that bug right now, because it's preexisting.

src/disasm.rs
```diff
@@ -0,0 +1,215 @@
...
+                    .arm64()
+                    .mode(arch::arm64::ArchMode::Arm)
+                    .build(),
+                _ => return Err(String::from("Unknown ISA")),
+            };
+
+            cs.map_err(|err| err.to_string())
+        }
+
+        pub fn print_disassembly(isa: &TargetIsa, mem: &[u8]) -> Result<(), String> {
```
bnjbvr: nit: doesn't need to be public

src/disasm.rs
```diff
@@ -0,0 +1,215 @@
...
+                if let Some(s) = i.op_str() {
+                    write!(&mut line, "{}", s).unwrap();
+                }
+
+                println!("{}", line);
+            }
+            Ok(())
+        }
+    } else {
+        pub fn print_disassembly(_: &TargetIsa, _: &[u8]) -> Result<(), String> {
```
bnjbvr: nit: doesn't need to be public

src/disasm.rs
```diff
@@ -0,0 +1,215 @@
...
+    traps: &PrintTraps,
+) -> Result<(), String> {
+    print_bytes(&mem);
+    print_disassembly(isa, &mem[0..code_size as usize])?;
+    print_readonly_data(&mem[code_size as usize..(code_size + rodata_size) as usize]);
+    println!("\n{}\n{}", &relocs.text, &traps.text);
+    Ok(())
+}
+
+pub fn print_bytes(mem: &[u8]) {
```
bnjbvr: nit: doesn't need to be public

src/disasm.rs
```diff
@@ -0,0 +1,215 @@
...
+            first = false;
+        } else {
+            print!(", ");
+        }
+        print!("{}", byte);
+    }
+    println!();
+}
+
+pub fn print_readonly_data(mem: &[u8]) {
```
bnjbvr: nit: doesn't need to be public

lars-t-hansen:
FWIW, I left those functions public on purpose, so that it's possible to use the disasm module more flexibly in contexts we've not yet encountered.  Eg, though print_all prints the bytes, that is mostly completely annoying, ditto traps and relocs and even rodata, so I could see people wanting to generate ad-hoc code that calls the printing functions individually and as-needed.

bnjbvr:
Sounds good to me, thanks.

PR #714, opened 2019-03-26
Move wasmtime-debug to the crantelift-wasm

yurydelendik:
* Move cranelift-debug from wasmtime-debug: direct copy
* Add data structures declarations from wasmtime-environ
* Update faerie dependency


cranelift-debug/LICENSE
```diff
@@ -0,0 +1,220 @@
...
+      To apply the Apache License to your work, attach the following
+      boilerplate notice, with the fields enclosed by brackets "[]"
+      replaced with your own identifying information. (Don't include
+      the brackets!)  The text should be enclosed in the appropriate
+      comment syntax for the file format. We also recommend that a
+      file or class name and description of purpose be included on the
+      same "printed page" as the copyright notice for easier
+      identification within third-party archives.
+
+   Copyright [yyyy] [name of copyright owner]
```
pepyakin: Is it ok to have a placeholder here?

sunfishcode: Yes; this is in the APPENDIX section, which is not part of the TERMS AND CONDITIONS. It's not an actual copyright declaration, it's an example of a copyright declaration, so we don't need to modify it.

cranelift-debug/src/write_debuginfo.rs
```diff
@@ -0,0 +1,209 @@
...
+
+pub trait SymbolResolver {
+    fn resolve_symbol(&self, symbol: usize, addend: i64) -> ResolvedSymbol;
+}
+
+pub fn emit_dwarf(
+    artifact: &mut Artifact,
+    mut dwarf: TransformedDwarf,
+    symbol_resolver: &SymbolResolver,
+) {
```
bjorn3: With latest gimli this can be simplified to:

```rust
let mut sections = Sections::new(WriterRelocate::new(endian, symbol_resolver));
dwarf.write(&mut sections).unwrap();

let _: Result<()> = sections.for_each_mut(|id, section| {
    if !section.writer.slice().is_empty() {
        artifact
   .declare_with(id.name(), Decl::debug_section(), section.writer.take())
   .unwrap();
    }
    Ok(())
});

let _: Result<()> = sections.for_each(|id, section| {
    for reloc in &section.relocs {
        artifact
            .link_with(
                faerie::Link {
                    from: id.name(),
                    to: reloc.name(self),
                    at: u64::from(reloc.offset),
                },
                faerie::Reloc::Debug {
                    size: reloc.size,
                    addend: reloc.addend as i32,
                },
            )
            .expect("faerie relocation error");
    }
    Ok(())
});
```

(adapted from https://github.com/bjorn3/rustc_codegen_cranelift/blob/47b1ef24e48bd3c250d14c756eb5e9de6c1ece20/src/debuginfo.rs#L188-L217)

philipc:
This is currently all very specific to transforming wasm DWARF, but that isn't reflected in the name of the crate (cranelift-debug). What are your plans in this area? e.g. do you intend to refactor this to move the wasm parts elsewhere? I'm just thinking about what the plan is for sharing code with the rustc backend. I think there is potential for reuse for at least line number info and call frame info.

yurydelendik:
> This is currently all very specific to transforming wasm DWARF, but that isn't reflected in the name of the crate (cranelift-debug). What are your plans in this area?  e.g. do you intend to refactor this to move the wasm parts elsewhere? 

Good point. There was no plans yet, but I was hoping to introduce abstractions later to use different source than wasm DWARF.
 
> I'm just thinking about what the plan is for sharing code with the rustc backend. I think there is potential for reuse for at least line number info and call frame info.

Agree. I did not look into the rustc backend yet, but can refactor this code to extract any common parts.

yurydelendik:
Based on the feedback above, I think placement under cranelift-wasm will be a better approach for now. Though we can move portion to the cranelift-debug and cranelift-faerie.

cranelift-wasm/Cargo.toml
```diff
@@ -20,6 +20,9 @@ failure = { version = "0.1.1", default-features = false, features = ["derive"] }
 failure_derive = { version = "0.1.1", default-features = false }
 log = { version = "0.4.6", default-features = false }
 cast = { version = "0.2.2", default-features = false }
+target-lexicon = { version = "0.3.0", default-features = false }
+gimli = "0.17.0"
+faerie = "0.9.1"
```
sunfishcode: gimli and especially faerie are fairly heavyweight dependencies that not all cranelift-wasm users will want. Would you mind either putting these under control of a feature, or (at the risk of suggesting yet another crate change) moving this code into a seperate crate, maybe cranelift-wasmdebug?

yurydelendik: added a feature

yurydelendik:
It is not clear where is the best place for the produce-DWARF logic, closing this PR in favor of https://github.com/CraneStation/wasmtime/pull/63 for now.

PR #715, opened 2019-03-27
Update target-lexicon to 0.4.0

bnjbvr:
This depends on https://github.com/m4b/faerie/pull/73 and will require a version bump there.

Having this would be nice for Spidermonkey/Gecko, which currently uses @glandium's clone of `target-lexicon` as the Gecko dependency (so as to have support for some ARM devices I guess).

(Build will be broken until faerie is updated)


bnjbvr:
Thanks @sunfishcode for the new faerie release! This unlocks this PR, which will unlock updating Cranelift in Spidermonkey.

sunfishcode:
Sounds good!

PR #717, opened 2019-03-28
Fixes #716: Stub out parse_x86_cpuid in the non-x86 case

### LINKED ISSUE #716: build fails on arm64
Building on a Packet c1.large.arm (arm64, ThunderX) system I get this error:

```
error[E0425]: cannot find function `parse_x86_cpuid` in this scope
  --> cranelift-native/src/lib.rs:45:9
   |
45 |         parse_x86_cpuid(&mut isa_builder)?;
   |         ^^^^^^^^^^^^^^^ not found in this scope
error: Could not compile `cranelift-native`.
```

is there arm64 support in this package? I did not expect to get this codepath.

bnjbvr:
See #716 for discussion; an alternative solution is to add the dependency `cfg-if` and use `cfg_if!` here.

### DIFF
--- a/cranelift-native/src/lib.rs
+++ b/cranelift-native/src/lib.rs
@@ -88,6 +88,11 @@ fn parse_x86_cpuid(isa_builder: &mut isa::Builder) -> Result<(), &'static str> {
     Ok(())
 }
 
+#[cfg(not(any(target_arch = "x86", target_arch = "x86_64")))]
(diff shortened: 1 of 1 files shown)


sunfishcode:
If cranelift-native starts having more places where `cfg-if` would be helpful, I think it'll make sense to use it, but this PR looks good for now.

PR #718, opened 2019-03-28
Fix a disasm formatting error

lars-t-hansen:
When I factored the disassembler I tried to be smart about how it prints spaces, in order to avoid a formatting error in an edge condition.  But I wasn't smart enough: the space should be printed conditionally before the byte value, not after.  This fixes that.


bnjbvr:
Indeed. Thanks.

PR #720, opened 2019-03-29
[meta] TypeVar/TypeSet tests

bnjbvr:
This does a few things:

- add a `num_set!` macro to avoid repeating `NumSet::from_iter(vec![...])`.
- Move the TypeSet building code from the TypeVarBuilder into its own TypeSetBuilder structure. This doesn't change the all behavior, while it moves some code around.
- next commits add more tests, some of them were passing, but one of them helped identify two issues (that I saw when porting the legalization actions!):
  - the name of derived type variables wasn't computed correctly
  - nor was the type set, resulting in weird assertions triggering (e.g. when doing `TypeSet(ints=32..64).half_width().double_width()`)

I'm also pretty sure I found a bug in `gen_instr.py` which directly reads `type_set` from some typevars (it might need to use `get_typeset()` instead; this doesn't change behavior in the Python case, because none of the type variables which we read the type sets are derived). Since I'm not sure other Python generators rely on this behavior, I haven't changed/fixed it yet. That might happen later.

### DIFF
--- a/cranelift-codegen/meta/src/cdsl/types.rs
+++ b/cranelift-codegen/meta/src/cdsl/types.rs
@@ -21,13 +21,13 @@ static LANE_BASE: u8 = 0x70;
 // Rust name prefix used for the `rust_name` method.
 static _RUST_NAME_PREFIX: &'static str = "ir::types::";
 
-// ValueType variants (i8, i32, ...) are provided in `base::types.rs`.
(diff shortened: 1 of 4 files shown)


jgouly:
Thanks, it's good to have more tests for this code! And easier to review since it was broken done into smaller commits.

LGTM.

bnjbvr:
Thanks!

PR #721, opened 2019-03-30
Make it possible to define data alignment

bjorn3:
This is necessary to fix (some text to prevent auto close) bjorn3/rustc_codegen_cranelift#348

### DIFF
--- a/cranelift-module/src/module.rs
+++ b/cranelift-module/src/module.rs
@@ -185,6 +185,7 @@ pub struct DataDeclaration {
     pub name: String,
     pub linkage: Linkage,
     pub writable: bool,
+    pub align: Option<u8>,
 }
 
 /// A data object belonging to a `Module`.
@@ -433,6 +434,7 @@ where
         name: &str,
         linkage: Linkage,
         writable: bool,
+        align: Option<u8>, // An alignment bigger than 128 is unlikely
     ) -> ModuleResult<DataId> {
         // TODO: Can we avoid allocating names so often?
         use super::hash_map::Entry::*;
@@ -442,7 +444,7 @@ where
                     let existing = &mut self.contents.data_objects[id];
                     existing.merge(linkage, writable);
                     self.backend
-                        .declare_data(name, existing.decl.linkage, existing.decl.writable);
+                        .declare_data(name, existing.decl.linkage, existing.decl.writable, existing.decl.align);
                     Ok(id)
                 }
 
@@ -456,11 +458,12 @@ where
                         name: name.to_owned(),
                         linkage,
                         writable,
+                        align,
                     },
                     compiled: None,
                 });
                 entry.insert(FuncOrDataId::Data(id));
-                self.backend.declare_data(name, linkage, writable);
+                self.backend.declare_data(name, linkage, writable, align);
                 Ok(id)
             }
         }
@@ -558,6 +561,7 @@ where
             Some(self.backend.define_data(
                 &info.decl.name,
                 info.decl.writable,
+                info.decl.align,
                 data_ctx,
                 &ModuleNamespace::<B> {
                     contents: &self.contents,
--- a/cranelift-module/src/module.rs
+++ b/cranelift-module/src/module.rs
@@ -203,9 +203,10 @@ impl<B> ModuleData<B>
 where
     B: Backend,
 {
-    fn merge(&mut self, linkage: Linkage, writable: bool) {
+    fn merge(&mut self, linkage: Linkage, writable: bool, align: Option<u8>) {
         self.decl.linkage = Linkage::merge(self.decl.linkage, linkage);
         self.decl.writable = self.decl.writable || writable;
+        self.decl.align = self.decl.align.max(align);
     }
 }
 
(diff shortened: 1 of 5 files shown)


cranelift-simplejit/src/backend.rs
```diff
@@ -177,7 +183,7 @@ impl<'simple_jit_backend> Backend for SimpleJITBackend {
         let size = code_size as usize;
         let ptr = self
             .code_memory
-            .allocate(size)
+            .allocate(size, 0x10)
```
bjorn3: This is the same default as the one faerie uses.

cranelift-simplejit/src/memory.rs
```diff
@@ -97,7 +97,12 @@ impl Memory {
     }
 
     /// TODO: Use a proper error type.
-    pub fn allocate(&mut self, size: usize) -> Result<*mut u8, String> {
+    pub fn allocate(&mut self, size: usize, align: u8) -> Result<*mut u8, String> {
+        if self.position % align as usize != 0 {
+            self.position += align as usize - self.position % align as usize;
```
bjorn3: Is this correct?

cranelift-module/src/module.rs
```diff
@@ -441,8 +443,12 @@ where
                 FuncOrDataId::Data(id) => {
                     let existing = &mut self.contents.data_objects[id];
                     existing.merge(linkage, writable);
```
bnjbvr: I think you'd need to merge alignment constraints too.

cranelift-simplejit/src/memory.rs
bnjbvr: I think so, at least it matches what Spidermonkey [does](https://searchfox.org/mozilla-central/source/js/src/jsutil.h#156) (although in the Spidermonkey code there's an additional modulo in case `self.position` is 0 or a multiple of alignment, but the explicit check makes it obvious here).

cranelift-simplejit/src/backend.rs
```diff
@@ -221,11 +228,11 @@ impl<'simple_jit_backend> Backend for SimpleJITBackend {
         let size = init.size();
         let storage = if writable {
             self.writable_memory
-                .allocate(size)
+                .allocate(size, align.unwrap_or(0x8))
                 .expect("TODO: handle OOM etc.")
         } else {
             self.readonly_memory
-                .allocate(size)
+                .allocate(size, align.unwrap_or(1))
```
bnjbvr: nit: for consistency, can you use `0x1` here?

cranelift-simplejit/src/backend.rs
```diff
@@ -221,11 +228,11 @@ impl<'simple_jit_backend> Backend for SimpleJITBackend {
         let size = init.size();
         let storage = if writable {
             self.writable_memory
-                .allocate(size)
+                .allocate(size, align.unwrap_or(0x8))
```
bnjbvr: It'd be nice to make this a constant, with a comment explaining where it comes from (here it's because it's the alignment of the biggest data type we can write at the moment, `f64`, right?).

cranelift-simplejit/src/backend.rs
bnjbvr: It's indeed the right constant for x64; for ARM it can be only 0x8, it seems. Being conservative and having 0x10 all the time sounds fine to me. Can you make this a global constant, please?

cranelift-simplejit/src/backend.rs
bjorn3: It took all alignment constants from faerie. https://github.com/m4b/faerie/blob/49aac0ca5b49c6ef95a57a95d8a006ab73e998e5/src/elf.rs#L268-L276

cranelift-simplejit/src/backend.rs
bjorn3: Yes

bjorn3:
I resolved all comments.

cranelift-simplejit/src/backend.rs
```diff
@@ -17,6 +17,10 @@ use target_lexicon::PointerWidth;
 #[cfg(windows)]
 use winapi;
 
+const EXECUTABLE_DATA_ALIGNMENT: u8 = 0x10;
+const WRITABLE_DATA_ALIGNMENT: u8 = 0x8;
```
sunfishcode: These two numbers appear somewhat magical. Why 0x10 and 0x8? If these are meant to be target-specific, we should have `TargetMachine` provide them.

bjorn3: Quoting myself:

> It took all alignment constants from faerie. https://github.com/m4b/faerie/blob/49aac0ca5b49c6ef95a57a95d8a006ab73e998e5/src/elf.rs#L268-L276

bnjbvr: Indeed, but Faerie didn't invent these constants either. I think these should be good for x64 at least. Agreed it would be cleaner to have this in the TargetMarchine, but we can do this as a follow-up; @bjorn3, do you prefer to defer this, or do it now?

bjorn3: I would prefer to defer this.

bnjbvr: Works fine for me, opened #758, thanks for answering.

philipc: I would expect data alignment to always be decided by the caller, instead of passing in an option. At the least, using different values for writable data vs readable data seems strange.

bnjbvr:
Thanks! Merging and somebody else can take care of the follow-up.

PR #725, opened 2019-04-05
[meta] Optimize seq table

bnjbvr:
This allows the unique sequence table to reuse the last elements of the whole table as a prefix when possible, that is:

- if we have `[A, B, C, D]` in the whole table
- and we want to insert `[C, D, E, F]` as a new sequence,

Then the `add()` function will just insert `[E, F]` and point to C's offset.


sunfishcode:
Neat! Do you know how often this comes up?

bnjbvr:
There are three uses of UniqueSeqTable in the Python code: for settings' ENUMERATORS, for instructions' OPERAND_CONSTRAINTS and encodings' ENCLISTS. The one I've checked was `OPERAND_CONSTRAINTS`, and this optimization reduces the size by 7 entries, from 79 to 71 entries (9% overall size reduction).

Issue #726, opened 2019-04-05
LICM does not hoist safe loads


lars-t-hansen:
On a [dot-product microbenchmark](https://github.com/lars-t-hansen/embenchen/blob/master/micro/wasmdot.js), the entire difference in speed between Ion-generated and Cranelift-generated code is due to the Cranelift code reloading the heap pointer from the tls (vmctx) on every loop iteration, thus increasing memory traffic and dependencies for the subsequent value loads.  Yet this heap pointer load would be safe to hoist - the value loaded never changes, and the load will never trap.

One problem is that licm, unlike gvn, does not consider read-only loads specially - any load is flagged unsafe for licm.  Changing this experimentally, I find that the heap pointer load is hoisted and the performance improves by 15%.  But presumably this hoisting is only safe if the load is also guaranteed never to trap.  So it's possible that that bit of information needs to be encoded, or if it is already encoded it must be used, or the heap pointer load must be encoded in some other way entirely.

lars-t-hansen:
Oooh, we have the notrap() flag for this.

bnjbvr:
Fixed by #727.

PR #727, opened 2019-04-05
Allow readonly nontrapping loads to be hoisted by licm

lars-t-hansen:
This addresses #726.

### DIFF
--- a/cranelift-codegen/src/licm.rs
+++ b/cranelift-codegen/src/licm.rs
@@ -5,7 +5,7 @@ use crate::dominator_tree::DominatorTree;
 use crate::entity::{EntityList, ListPool};
 use crate::flowgraph::{BasicBlock, ControlFlowGraph};
 use crate::fx::FxHashSet;
-use crate::ir::{DataFlowGraph, Ebb, Function, Inst, InstBuilder, Layout, Opcode, Type, Value};
+use crate::ir::{DataFlowGraph, Ebb, Function, Inst, InstBuilder, InstructionData, Layout, Opcode, Type, Value};
 use crate::isa::TargetIsa;
 use crate::loop_analysis::{Loop, LoopAnalysis};
 use crate::timing;
@@ -145,8 +145,7 @@ fn change_branch_jump_destination(inst: Inst, new_ebb: Ebb, func: &mut Function)
 
 /// Test whether the given opcode is unsafe to even consider for LICM.
 fn trivially_unsafe_for_licm(opcode: Opcode) -> bool {
-    opcode.can_load()
-        || opcode.can_store()
+    opcode.can_store()
         || opcode.is_call()
         || opcode.is_branch()
         || opcode.is_terminator()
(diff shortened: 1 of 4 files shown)


cranelift-codegen/src/licm.rs
```diff
@@ -156,12 +155,25 @@ fn trivially_unsafe_for_licm(opcode: Opcode) -> bool {
...
 
+fn is_unsafe_load(inst_data: &InstructionData) -> bool {
+    match *inst_data {
+        InstructionData::Load { flags, .. } | InstructionData::LoadComplex { flags, .. } => {
+            !flags.readonly() || !flags.notrap()
+        }
+        _ => inst_data.opcode().can_load(),
+    }
+}
+
```
sunfishcode: This is now the same code as simple_gvn's `is_load_and_not_readonly`. I'm not particular on which name we use, but at this point it'd be good to factor this function out into a utility function in cranelift-codegen/src/ir/instructions.rs. It looks like we could move `trivially_unsafe_for_licm`/`trivially_unsafe_for_gvn` there too.

cranelift-codegen/src/licm.rs
```diff
@@ -156,12 +155,25 @@ fn trivially_unsafe_for_licm(opcode: Opcode) -> bool {
...
 }
 
+fn is_unsafe_load(inst_data: &InstructionData) -> bool {
+    match *inst_data {
+        InstructionData::Load { flags, .. } | InstructionData::LoadComplex { flags, .. } => {
+            !flags.readonly() || !flags.notrap()
+        }
+        _ => inst_data.opcode().can_load(),
+    }
+}
```
sunfishcode: This is now the same code as simple_gvn.rs' `is_load_and_not_readonly`. I'm not particular about the name here, but this would be a good point to factor out this logic into a shared helper function in cranelift/cranelift-codegen/src/ir/instructions.rs. It looks like we can factor out `trivially_safe_for_licm`/`trivially_safe_for_gvn` now too.

cranelift-codegen/src/licm.rs
sunfishcode: (Oops, I originally had a duplicate of this comment here.)

jgouly:
I think you could also add a test in `filetests/licm/` for this too.

cranelift-codegen/src/licm.rs
lars-t-hansen: Actually, it's not the same code.  Here I also check !notrap(), which gvn does not do (and probably should not need to do).

lars-t-hansen:
> I think you could also add a test in `filetests/licm/` for this too.

Will do.

cranelift-codegen/src/licm.rs
bnjbvr: Agreed.

bnjbvr:
Looks good, thanks for the positive and negative tests too!

PR #731, opened 2019-04-08
Fix the publish script to include the remote when pushing the tag

bnjbvr:


bnjbvr:
(Merging; this is just a mistake I made when adding tagging to the original script)

PR #733, opened 2019-04-11
[meta] Port legalization to the Rust crate

bnjbvr:
This now works and passes all the tests, including the Spidermonkey tests (when Cranelift is used as the main compiler).

When selecting specific targets at compile time, it does even more than the current Python module: the latter was creating all the legalizations for all the platforms independently of the selected ISA targets. The Rust code does only create legalizations that are used for the current target.

cc @bjorn3 because more code than I expected might be shared (instruction predicates in particular), so we might need to sync.

TODO:

- [x] Re-read and remove some XXX and TODOs.
- [x] Rename XForm to Transform, to make it explicit.
- [x] Slightly reformat the output (add empty lines between functions, wrap conditions on several lines to make them more readable).
- [x] see if the `def!` macro can do slightly more and remain simple enough (for example, `DummyVar` could just get generated by the macro itself).
- [x] Split the PR into smaller ones.

### DIFF
--- a/cranelift-codegen/meta/src/cdsl/ast.rs
+++ b/cranelift-codegen/meta/src/cdsl/ast.rs
@@ -1,0 +1,653 @@
+use crate::cdsl::formats::FormatRegistry;
+use crate::cdsl::inst::{BoundInstruction, Instruction, InstructionPredicate};
+use crate::cdsl::operands::{OperandKind, OperandKindFields};
+use crate::cdsl::types::{LaneType, ValueType};
+use crate::cdsl::typevar::{TypeSetBuilder, TypeVar};
+
+use cranelift_entity::{entity_impl, PrimaryMap};
+
+use std::fmt;
+
+pub enum Expr {
+    Var(VarIndex),
+    Literal(Literal),
+    Apply(Apply),
+}
+
+impl Expr {
+    pub fn maybe_literal(&self) -> Option<&Literal> {
+        match &self {
+            Expr::Literal(lit) => Some(lit),
+            _ => None,
+        }
+    }
+
+    pub fn maybe_var(&self) -> Option<VarIndex> {
+        if let Expr::Var(var) = &self {
+            Some(*var)
+        } else {
+            None
+        }
+    }
+
+    pub fn unwrap_var(&self) -> VarIndex {
+        self.maybe_var()
+            .expect("tried to unwrap a non-Var content in Expr::unwrap_var")
+    }
+
+    pub fn to_rust_code(&self, var_pool: &VarPool) -> String {
+        match self {
+            Expr::Var(var_index) => var_pool.get(*var_index).to_rust_code(),
+            Expr::Literal(literal) => literal.to_rust_code(),
+            Expr::Apply(a) => a.to_rust_code(var_pool),
+        }
+    }
+}
+
+/// An AST definition associates a set of variables with the values produced by an expression.
+pub struct Def {
+    pub apply: Apply,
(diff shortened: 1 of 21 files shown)


bnjbvr:
>  see if the def! macro can do slightly more and remain simple enough (for example, DummyVar could just get generated by the macro itself).

It was not the case that variables could be generated by the macro itself, because in some cases we want to use an identifier that's referring to a constant value defined somewhere else, so the variable can't just be a literal input to the macro. However, by having the definition arguments be expressions, I could get rid of a few constant values defined only once.

bnjbvr:
This is now ready for review! I've split it into a lot of commits, but some of them are still big, notably the one that introduces XForm / type inference / etc. But it's quite mechanical and mostly a port of the Python code! (I renamed variables and added comments and made things more explicit.)

I've made sure that each commits compiles on its own. Each of them introduces dead code until the last one, which makes use of everything and compiles cleanly.

If people prefer, I could split this into multiple PRs; there's already #735 for the first part, and then I'll rebase and open PRs once this one has been approved.

This passes all the Cranelift tests and Spidermonkey tests when it's used in place of the Python generating code.

bnjbvr:
cc @jgouly @pchickey @sunfishcode if you're interested in reviewing this, partially or entirely, that'd be great, thanks!

cranelift-codegen/meta/src/cdsl/ast.rs
```diff
@@ -0,0 +1,658 @@
...
+            .assign_to(vec![$($dest.clone()),*])
+    };
+
+    // x = inst(a, b, c)
+    ($dest:ident = $inst:ident($($src:expr),*)) => {
+        ExprBuilder::apply($inst.into(), vec![$($src.clone().into()),*])
+            .assign_to(vec![$dest.clone()])
+    };
+
+    // x = inst.i32(a, b, c)
```
bjorn3:
```suggestion
    // x = inst.type(a, b, c)
```

pchickey:
Sorry, I meant to get around to reviewing this sooner, I will try to tonight or tomorrow

bnjbvr:
No worries, it's a big one! I just rebased over the part 1 (and I can split it in smaller PRs if you'd like to).

cranelift-codegen/meta/src/cdsl/cpu_modes.rs
```diff
@@ -0,0 +1,84 @@
+use crate::cdsl::types::LaneType;
+use crate::cdsl::xform::{TransformGroup, TransformGroupIndex, TransformGroups};
+
+use std::collections::{HashMap, HashSet};
+use std::iter::FromIterator;
+
```
pchickey: While I believe I understand this code, and I understand why CpuMode makes sense at the use site of this code over in the ISA definitions, the name "CpuMode" throws me off because none of the contents seem to have to do with CpuMode (and there is no corresponding name in the python dsl). I can't think of a name for this module that is better - LegalizationScheme is the best thing I could come up with.

bnjbvr: Good question; it's actually something I've slightly adapted from the Python code, where there are many indirections to iterate through all the XForms (go through the TargetIsa, then each CpuMode to register all the XForms in a local TargetIsa attribute, that is then iterated on). I propose to keep it like this until encodings/binemit are ported too, to try to keep it closer to the Python code and reduce porting confusion, and then I'm definitely happy to rename this / separate it into a different data structure.

bnjbvr:
Thanks! Rebased and merging.

PR #739, opened 2019-04-16
[meta] Finish migrating settings

bnjbvr:
This finishes the migration of settings to the Rust crate, by:

- using named predicates instead of anonymous predicates in the encodings, in the Python code. Also ensure a deterministic and easy to reproduce order for predicates.
- do the same thing in Rust
- fix a few small issues in the Rust generation
- use the Rust generation in lieu of the Python code to generate the settings files


bnjbvr:
@pchickey Hi! Would you like to review this PR, please? I think it's straightforward and earns us one more set of files generated by the Rust crate. Thanks in advance!

cranelift-codegen/meta-python/isa/x86/settings.py
```diff
@@ -35,6 +35,13 @@
 use_bmi1 = And(has_bmi1)
 use_lzcnt = And(has_lzcnt)
 
+is_pic = And(shared.is_pic)
+not_is_pic = Not(shared.is_pic)
+all_ones_funcaddrs_and_not_is_pic = And(shared.allones_funcaddrs,
+                                        Not(shared.is_pic))
+not_all_ones_funcaddrs_and_not_is_pic = And(Not(shared.allones_funcaddrs),
+                                            Not(shared.is_pic))
```
sunfishcode: Can you comment on why `is_pic` and other target-independent settings are duplicated here in the x86-specific settings.py, and similarly below in x86/mod.rs?

pchickey:
Sorry, I forgot to come back to this one :( Please bug me via discord or email if I forget in the future...

cranelift-codegen/meta-python/isa/x86/settings.py
bnjbvr: Yes, the Python code doesn't make it quite obvious: `shared.is_pic` is a BooleanSetting, while `is_pic` is a predicate. The python code uses the `And` function with one argument to transform a setting into a predicate; the Rust code is more verbose and explicit, e.g. `settings.add_predicate("is_pic", predicate!(shared.get_bool("is_pic")))` does the same thing in Rust code.

Then, it is locally duplicated in the target specific settings definition, because predicates belong to a `SettingGroup`, which are tied to the target ISA. Before this PR, the same predicates would end up in the setting group, but were anonymous (that is, instruction predicates would use the settings and the predicate would be created on the fly). This change makes them explicit (by naming them) so we just remove the concept of anonymous predicates (and don't have to port it, when porting encodings).

Do you want me to add a comment around here to explain this, or in the Rust crate?

sunfishcode: A comment in the Rust crate would be good.

bnjbvr: Done!

bnjbvr:
@sunfishcode or @pchickey, gentle review ping

bnjbvr:
Thanks and sorry for the hassle! (This area should get quieter over time.)

PR #741, opened 2019-04-17
[WIP] Delay slot filler framework

### LINKED ISSUE #728: Support architectural delay slots
Some architectures make use of branch delay slots, such as MIPS and SPARC; some even more exotic ones have load delay slots as well. Often enough there are also instructions with results only usable after certain number of cycles.

I'm trying to port Cranelift to MIPS64. I glanced over the codebase and found no support for filling the delay slots; I had to emit NOPs after every branch in the meantime.

So what's the current plan regarding this? A roadmap or implementation guide would be great!

xen0n:
Ripped out from the ongoing MIPS port for earlier review.

Some outstanding questions:

* [ ] How to tag instructions moved into delay slots as such, for verifier use?
* [ ] Should the `has_delay_slot` flag be moved into encinfo realm? Seems a little more work to do, but I'm willing to because delay slots are inherently ISA-specific.

Also some comments are not adjusted after copy-pasting... will fix these afterwards.

Closes #728.

### DIFF
--- a/cranelift-codegen/meta-python/base/instructions.py
+++ b/cranelift-codegen/meta-python/base/instructions.py
@@ -59,7 +59,8 @@ jump = Instruction(
         EBB arguments. The number and types of arguments must match the
         destination EBB.
         """,
-        ins=(EBB, args), is_branch=True, is_terminator=True)
+        ins=(EBB, args), is_branch=True, is_terminator=True,
+        has_delay_slot=True)
 
 fallthrough = Instruction(
         'fallthrough', r"""
@@ -72,7 +73,8 @@ fallthrough = Instruction(
         relaxation pass. There is no reason to use this instruction outside
         that pass.
         """,
-        ins=(EBB, args), is_branch=True, is_terminator=True)
+        ins=(EBB, args), is_branch=True, is_terminator=True,
+        has_delay_slot=True)
 
 brz = Instruction(
         'brz', r"""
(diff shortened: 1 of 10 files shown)


bnjbvr:
Hi! Sorry for not getting back to you earlier. We've discussed this yesterday in an internal meeting and came to the conclusion that each target would get its own target-specific peephole optimizer pass, very late in the pipeline, and that delay slot filling would happen there for MIPS.

This code was operating on the Python meta code, which has been migrated to Rust in the meanwhile, so it requires to be rebased. I wonder if putting the `has_delay_slot` property on the instruction makes sense (target-independent), or if it could be a property of the encoding (which is target specific).

bnjbvr:
@xen0n ping? (see previous comment)

xen0n:
@bnjbvr 

Hi, sorry for the delay. My research focus has since shifted mid-2019 and presently I don't have enough time to rework this on top of the current codebase, unfortunately.

As for whether the `has_delay_slot` property is target-independent or not, definitely it is encoding-specific; for example, MIPS R6 branches don't have delay slots while their pre-R6 counterparts do. So even treating all branch insns of a target as having the same delay slot property won't work.

eqrion:
> @bnjbvr
> 
> Hi, sorry for the delay. My research focus has since shifted mid-2019 and presently I don't have enough time to rework this on top of the current codebase, unfortunately.
> 

That's not a problem, feel free to open a new PR if you ever take another look at this.

PR #742, opened 2019-04-18
Legalize fcvt_from_u/sint.f32/f64.i8/i16

bjorn3:
cc #466 (seems like I didn't list the instructions there though)

### DIFF
--- a/cranelift-codegen/meta-python/base/legalize.py
+++ b/cranelift-codegen/meta-python/base/legalize.py
@@ -261,6 +261,23 @@ for binop in [iadd_imm, imul_imm, udiv_imm, urem_imm]:
 for binop in [sdiv_imm, srem_imm]:
     widen_imm(True, binop)
 
+# Use expand instead of widen, because widen only gets applied for i8 and i16
+# base type vars, but these take f32 and f64 as base type var.
+for ty in [types.f32, types.f64]:
+    for int_ty in [types.i8, types.i16]:
+        expand.legalize(
+            a << insts.fcvt_from_uint.bind(ty).bind(int_ty)(b),
+            Rtl(
+                x << uextend.i32(b),
+                a << insts.fcvt_from_uint.bind(ty).i32(x),
+            ))
+        expand.legalize(
+            a << insts.fcvt_from_sint.bind(ty).bind(int_ty)(b),
+            Rtl(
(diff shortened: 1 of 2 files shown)


filetests/isa/x86/legalize-fcvt_from_usint-i16.clif
```diff
@@ -0,0 +1,14 @@
+test compile
+target x86_64
+
+function u0:0(i16) -> f64 fast {
+ebb0(v0: i16):
+    v1 = fcvt_from_uint.f64 v0
```
bnjbvr: Can you add check statements for what's expected in place of this instruction?

filetests/isa/x86/legalize-fcvt_from_usint-i16.clif
```diff
@@ -0,0 +1,14 @@
...
+    return v1
+}
+
+function u0:1(i16) -> f64 fast {
+ebb0(v0: i16):
+    v1 = fcvt_from_sint.f64 v0
```
bnjbvr: Ditto here.

cranelift-codegen/meta-python/base/legalize.py
```diff
@@ -261,6 +261,23 @@ def widen_imm(signed, op):
 for binop in [sdiv_imm, srem_imm]:
     widen_imm(True, binop)
 
+# Use expand instead of widen, because widen only gets applied for i8 and i16
+# base type vars, but these take f32 and f64 as base type var.
+for ty in [types.f32, types.f64]:
```
bnjbvr: Can you move this chunk of code next to the other expand legalizations, please?

bnjbvr:
Hi @bjorn3 ! Now that the legalization port to Rust is over, you should be able to define it in Rust (and still need to define it in Python for the encoding table, I'm afraid). Would you like to take care of this, or do you want me to take over?

bjorn3:
I am quite busy this and next week, so please take over.

bnjbvr:
Closing in favor of #768; the filetest checks couldn't be added because there's actually a lot of code that's added to expand the fcvt_from_sint/uint. To wit: these instructions have the wasm semantics if I remember correctly, so you want to make sure that they work as you expect for edge case values.

PR #743, opened 2019-04-19
Expose function definitions, populate FaerieCompiledFunction

iximeow:
Hi! I'm working on some [changes](https://github.com/fastly/lucet/pull/136) for Lucet that need to get at the whole list of compiled functions after they've been translated to the target architecture.

This is a pair of changes that does that, and also populates `FaerieCompiledFunction` with the same kinds of data as `SimpleJITCompiledFunction`. I really just need function sizes, but I figure making them consistent keeping the code bytes in `FaerieCompiledFunction` too couldn't hurt.

### DIFF
--- a/cranelift-faerie/src/backend.rs
+++ b/cranelift-faerie/src/backend.rs
@@ -96,7 +96,15 @@ pub struct FaerieBackend {
     libcall_names: Box<Fn(ir::LibCall) -> String>,
 }
 
-pub struct FaerieCompiledFunction {}
+pub struct FaerieCompiledFunction {
+    code_length: u32,
+}
+
+impl FaerieCompiledFunction {
+    pub fn code_length(&self) -> u32 {
+        self.code_length
+    }
+}
 
 pub struct FaerieCompiledData {}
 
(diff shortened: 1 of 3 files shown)


sunfishcode:
If the function sizes are all that's needed, I'd prefer to just store those, rather than cloning the generated code. The reason `SimpleJITCompiledFunction` has a pointer to the code is that its users may want the pointer so that they can call it. With the faerie backend, once we've written out the .o file, typical users shouldn't have any further use for the actual code.

I'm also wondering if the `name` field of `FaerieCompiledFunction` is needed. If you have the `ModuleFunction`, you can get the name as `decl.name`. If you just have the `ir::ExternalName`, you can get the decl via `ModuleNamespace::get_function_decl`. That said, if having the name in `FaerieCompiledFunction` is more convenient, that's fine, I'm less worried about function name strings than I am about function bodies :-).

pchickey:
Lets just keep the length of the code for now, and refactor Faerie so we can borrow a ref to the code at some later point if we end up needing it.

iximeow:
Sounds good! I'd missed that `decl.name` even exists initially, whoops

PR #747, opened 2019-04-22
cranelift-simplejit: add a translation mechanism for LibCalls

### LINKED ISSUE #688: thread 'rustc' panicked at 'unexpected ExternalName kind %Memset'
This happens when using `FunctionBuilder::call_memset` and then calling `Module::finalize_definitions`.

antoyo:
Fix #688 by using a similar approach to #321.

### DIFF
--- a/cranelift-simplejit/src/backend.rs
+++ b/cranelift-simplejit/src/backend.rs
@@ -30,7 +30,7 @@ impl SimpleJITBuilder {
     /// The `libcall_names` function provides a way to translate `cranelift_codegen`'s `ir::LibCall`
     /// enum to symbols. LibCalls are inserted in the IR as part of the legalization for certain
     /// floating point instructions, and for stack probes. If you don't know what to use for this
-    /// argument, use `FaerieBuilder::default_libcall_names()`.
+    /// argument, use `SimpleJITBuilder::default_libcall_names()`.
     pub fn new(libcall_names: Box<Fn(ir::LibCall) -> String>) -> Self {
         let flag_builder = settings::builder();
         let isa_builder = cranelift_native::builder().unwrap_or_else(|msg| {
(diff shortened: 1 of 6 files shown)


cranelift-simplejit/src/backend.rs
```diff
@@ -373,18 +418,27 @@ impl<'simple_jit_backend> Backend for SimpleJITBackend {
             let ptr = data.storage;
             debug_assert!((offset as usize) < data.size);
             let at = unsafe { ptr.offset(offset as isize) };
-            let base = if namespace.is_function(name) {
-                let (def, name_str, _signature) = namespace.get_function_definition(&name);
-                match def {
-                    Some(compiled) => compiled.code,
-                    None => self.lookup_symbol(name_str),
+            let base = match *name {
```
bnjbvr: Any chance this code can be factored out, since this is duplicated twice?

cranelift-simplejit/src/backend.rs
```diff
@@ -21,17 +21,23 @@ use winapi;
...
 }
 
 impl SimpleJITBuilder {
     /// Create a new `SimpleJITBuilder`.
-    pub fn new() -> Self {
+    ///
+    /// The `libcall_names` function provides a way to translate `cranelift_codegen`'s `ir::LibCall`
+    /// enum to symbols. LibCalls are inserted in the IR as part of the legalization for certain
+    /// floating point instructions, and for stack probes. If you don't know what to use for this
+    /// argument, use `FaerieBuilder::default_libcall_names()`.
```
bnjbvr: In this comment and the one below, do you mean `SimpleJITBuilder::default_libcall_names()` instead?

antoyo:
@bnjbvr: Thanks for your comments. I fixed this.

cranelift-simplejit/tests/basic.rs
```diff
@@ -144,3 +150,53 @@ fn switch_error() {
...
+
+        bcx.call_memset(module.target_config(), buffer, zero, size);
+        module.finalize_definitions();
+
+        bcx.ins().return_(&[]);
+    }
+
+    module.define_function(func_id, &mut ctx).unwrap();
+
+    module.finalize_definitions();
```
bnjbvr: It's a bit strange that `finalize_definitions()` can be called twice, but this is outside the scope of this bug. Can you remove the first call, that's a few lines above, please?

bjorn3:
> @bjorn3 do you use the SimpleJit backend, or do you implement your own?

Yes, that's why I discovered #688. Currently `rustc_codegen_cranelift` can only run executables with absolutely no dependencies in jit mode though, but I want to fix that. (https://github.com/bjorn3/rustc_codegen_cranelift/issues/167)

bnjbvr:
> Yes, that's why I discovered #688. Currently rustc_codegen_cranelift can only run executables with absolutely no dependencies in jit mode though, but I want to fix that. (bjorn3/rustc_codegen_cranelift#167)

Ok, in this case maybe we don't need to go full generic, and we can just hoist the default function to cranelift-module.

antoyo:
@bnjbvr: I moved the default function to cranelift-module. Is there anything else to do?

bnjbvr:
Looks good to me. (As an aside: TIL one can use `Box` et al. in a no-std build!)
Thanks!

PR #748, opened 2019-04-24
Avoid inserting checks during div/rem legalization

bnjbvr:
... when the input is a constant immediate.


cranelift-codegen/src/isa/x86/enc_tables.rs
```diff
@@ -91,6 +91,23 @@ fn size_plus_maybe_sib_or_offset_for_in_reg_1(
     sizing.base_size + additional_size_if(1, inst, divert, func, needs_sib_byte_or_offset)
 }
 
+fn maybe_iconst_imm(pos: &FuncCursor, value: ir::Value) -> Option<i64> {
```
sunfishcode: Would you mind adding a brief one-line doc comment on this function?

bnjbvr:
Thanks for the review!

PR #749, opened 2019-04-24
[cranelift-simplejit] Remove unused symbols table

bnjbvr:
I was looking at SimpleJit / Faerie to understand more how they work (in the context of reviewing #747), and I figured that this symbols table in SimpleJit is actually unused. @sunfishcode is this expected? If so, can we remove it?


bjorn3:
The symbols table is actually used to make it possible for users of `SimpleJIT` to provide symbols to the compiled program without making them global. It was added in #445.

bnjbvr:
Oops, I didn't know about this, thanks for the pointer.

PR #750, opened 2019-04-25
[meta] Switch to the Rust crate to generate legalizations

bnjbvr:
This just toggles the switch from Python to Rust for legalization. I can confirm it works perfectly fine with testing and Spidermonkey, and a manual diff inspection showed no significant differences.


Ericson2314:
What's the ETA on this? I know switching to replacements is a big step, so just wondering roughly.

bnjbvr:
Patrick, since you've reviewed the previous parts, would you be interested in checking this too, please? This switches the legalization generation from the Python code to the Rust one that you have reviewed in the past. Please let me know if I should ask somebody else :)

bnjbvr:
Thanks!

PR #754, opened 2019-04-29
[wasm] Make FuncEnvironment functions fallible

bnjbvr:
@pchickey Great idea to use `failure::Error`! It requires an `std` build because it's only defined there; I've made the `User` variant in `WasmError` defined only in this case, which is a bit weird. Any better idea?

The rest of the patch is pretty mechanical:

- I've grouped infallible getters in the `FuncEnvironment` trait (and dummy impl).
- I've removed `PartialEq` and `Eq` from `WasmError`, because 1. `failure::Error` doesn't support those, and 2. I argue that one will explicitly need to match on the possible enum values anyways most of the time. (Maybe tests might suffer from this, though?)
- Transitively propagate `WasmResult` to their callers.
- There are changes around hash maps, where instead of using `hashmap.entry(key).or_insert_with(F)` where hashmap has type `T` and F returns a `T`, we now have to explicitly match on the entry, because the func F now returns `Result<T>` and we want to properly handle errors.

### DIFF
--- a/cranelift-wasm/Cargo.toml
+++ b/cranelift-wasm/Cargo.toml
@@ -27,7 +27,7 @@ target-lexicon = "0.4.0"
 
 [features]
 default = ["std"]
-std = ["cranelift-codegen/std", "cranelift-frontend/std", "wasmparser/std"]
+std = ["cranelift-codegen/std", "cranelift-frontend/std", "wasmparser/std", "failure/std"]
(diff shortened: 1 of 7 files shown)


bnjbvr:
Bonus commit: also raise `UnsupportedError` instead of panicking when a type is not supported in the function local declaration section. (Global context of all of this: I'm trying to reduce panickability of Cranelift in Spidermonkey so we can fuzz it more)

bnjbvr:
Thanks for the review! Implemented your suggestion as well, it's a nice idea.

PR #756, opened 2019-04-29
Clarify that FixedTied constraints are not Tied

lars-t-hansen:
What I mean is, `ConstraintKind::Tied` forces the `tied_ops` flags to be true, but `ConstraintKind::FixedTied` does not - it only forces `fixed_ins` or `fixed_outs`, as the case may be.  (These appear only to be used in coloring, and `tied_ops` and `fixed_outs` are only used once each, while `fixed_ins` is not used at all.)

### DIFF
--- a/cranelift-codegen/src/isa/constraints.rs
+++ b/cranelift-codegen/src/isa/constraints.rs
@@ -105,13 +105,13 @@ pub struct RecipeConstraints {
     /// constraints must be derived from the calling convention ABI.
     pub outs: &'static [OperandConstraint],
 
-    /// Are any of the input constraints `FixedReg`?
+    /// Are any of the input constraints `FixedReg` or `FixedTied`?
     pub fixed_ins: bool,
 
-    /// Are any of the output constraints `FixedReg`?
+    /// Are any of the output constraints `FixedReg` or `FixedTied`?
     pub fixed_outs: bool,
 
-    /// Are there any tied operands?
+    /// Are any of the input/output constraints `Tied`?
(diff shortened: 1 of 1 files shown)


bnjbvr:
I see what you mean; tied_ops means "tied but not fixed tied", this is a bit confusing. We could just remove `fixed_ins` from the meta codegen since it's unused. It seems the two other flags that are used once during coloring help preventing loops from being entered, so they're probably useful (it'd be interesting to remove them and see if a plain constraints iteration isn't as fast).

cranelift-codegen/src/isa/constraints.rs
```diff
@@ -105,13 +105,13 @@ pub struct RecipeConstraints {
...
-    /// Are any of the input constraints `FixedReg`?
+    /// Are any of the input constraints `FixedReg` or `FixedTied`?
     pub fixed_ins: bool,
 
-    /// Are any of the output constraints `FixedReg`?
+    /// Are any of the output constraints `FixedReg` or `FixedTied`?
     pub fixed_outs: bool,
 
-    /// Are there any tied operands?
+    /// Are any of the input/output constraints `Tied`?
```
bnjbvr: For extreme clarity, how about even precising that they're `Tied` but not `FixedTied`?

lars-t-hansen:
I can for sure fix the comment as suggested (had been thinking along those lines myself) but I don't think I want to take on removing fixed_ins in this patch, that requires more discussion + work.  All I wanted to do was reduce confusion for readers of the code.

PR #759, opened 2019-05-02
Add extensive test cases for integer division-by-constant magic numbe…

julian-seward1:
…r generation.

This adds test cases to ensure, to a reasonably high degree of certainty, that
the magic-number generators `magic_u32`, `magic_s32`, `magic_u64` and
`magic_s64` work correctly.  This is done by iterating through a large number
of `(n, d)` pairs, generating the magic numbers for `d`, interpreting the
magic numbers so as to perform the division, and comparing against the result
produced directly by the hardware.  The distribution of numbers is arranged so
that particular emphasis is given to corner cases -- the range ends and
midpoints -- but also so that there is at least some cover for values away
from those areas.  In total 50,148,000 tests are performed.

### DIFF
--- a/cranelift-codegen/src/divconst_magic_numbers.rs
+++ b/cranelift-codegen/src/divconst_magic_numbers.rs
@@ -647,4 +647,439 @@ mod tests {
         }
         assert_eq!(total, 563301797155560970);
     }
+
+    #[test]
+    fn test_magic_generators_give_correct_numbers() {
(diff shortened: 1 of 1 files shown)


bjorn3:
How much time does it take to run he tests?

julian-seward1:
2.50 to 2.55 seconds, for a debug build, running on an Intel Xeon at 2.6 GHz.

PR #766, opened 2019-05-17
Use 'xor r, r' to set registers to 0 instead of mov

iximeow:
This change is pretty closely modeled after the similar change in #344, and extends xor-to-clear to integer registers as well

Couple comments for questions I have, but this looks like it does the right things to me, so if y'all say it's good, then by all means :)

### DIFF
--- a/cranelift-codegen/meta/src/isa/x86/encodings.rs
+++ b/cranelift-codegen/meta/src/isa/x86/encodings.rs
@@ -598,6 +598,7 @@ pub(crate) fn define(
     let rec_trapif = r.recipe("trapif");
     let rec_trapff = r.recipe("trapff");
     let rec_u_id = r.template("u_id");
+    let rec_u_id_z = r.template("u_id_z");
     let rec_umr = r.template("umr");
     let rec_umr_reg_to_ssa = r.template("umr_reg_to_ssa");
     let rec_ur = r.template("ur");
@@ -733,6 +734,18 @@ pub(crate) fn define(
         e.enc_both(bconst.bind(ty), rec_pu_id_bool.opcodes(vec![0xb8]));
     }
     e.enc64(bconst.bind(B64), rec_pu_id_bool.opcodes(vec![0xb8]).rex());
+
+    let is_zero_int = InstructionPredicate::new_is_zero_int(f_unary_imm, "imm");
+    e.enc_both_instp(
+        iconst.bind(I32),
+        rec_u_id_z.opcodes(vec![0x31]),
+        is_zero_int.clone(),
+    );
+    e.enc_x86_64_instp(
+        iconst.bind(I64),
+        rec_u_id_z.opcodes(vec![0x31]),
+        is_zero_int,
+    );
 
     // Shifts and rotates.
     // Note that the dynamic shift amount is only masked by 5 or 6 bits; the 8-bit
--- a/cranelift-codegen/meta/src/isa/x86/encodings.rs
+++ b/cranelift-codegen/meta/src/isa/x86/encodings.rs
@@ -737,6 +737,23 @@ pub(crate) fn define(
 
     let is_zero_int = InstructionPredicate::new_is_zero_int(f_unary_imm, "imm");
     e.enc_both_instp(
+        iconst.bind(I8),
+        rec_u_id_z.opcodes(vec![0x30]),
+        is_zero_int.clone(),
+    );
+    // You may expect that i16 encodings would have an 0x66 prefix on the opcode to indicate that
+    // encodings should be on 16-bit operands (f.ex, "xor %ax, %ax"). Cranelift currently does not
+    // know that it can drop the 0x66 prefix and clear the upper half of a 32-bit register in these
+    // scenarios, so we explicitly select a wider but permissible opcode.
+    //
+    // This effectively formalizes the i16->i32 widening that Cranelift performs when there isn't
+    // an appropriate i16 encoding available.
+    e.enc_both_instp(
+        iconst.bind(I16),
+        rec_u_id_z.opcodes(vec![0x31]),
+        is_zero_int.clone(),
+    );
+    e.enc_both_instp(
         iconst.bind(I32),
         rec_u_id_z.opcodes(vec![0x31]),
         is_zero_int.clone(),
--- a/cranelift-codegen/meta/src/isa/x86/recipes.rs
+++ b/cranelift-codegen/meta/src/isa/x86/recipes.rs
@@ -1002,6 +1002,22 @@ pub(crate) fn define<'shared>(
             ),
     );
 
+    // XX+rd id unary with 32-bit immediate.
+    {
+        let format = formats.get(f_unary_imm);
+        recipes.add_template_recipe(
+            EncodingRecipeBuilder::new("u_id_z", f_unary_imm, 1)
+                .operands_out(vec![gpr])
+                .inst_predicate(InstructionPredicate::new_is_zero_int(format, "imm"))
+                .emit(
+                    r#"
+                        {{PUT_OP}}(bits, rex2(out_reg0, out_reg0), sink);
+                        modrm_rr(out_reg0, out_reg0, sink);
+                    "#,
+                ),
+        );
+    }
+
     // XX /n Unary with floating point 32-bit immediate equal to zero.
     {
         let format = formats.get(f_unary_ieee32);
--- a/cranelift-codegen/meta/src/isa/x86/recipes.rs
+++ b/cranelift-codegen/meta/src/isa/x86/recipes.rs
@@ -1002,21 +1002,17 @@ pub(crate) fn define<'shared>(
             ),
     );
 
-    // XX+rd id unary with 32-bit immediate.
-    {
-        let format = formats.get(f_unary_imm);
-        recipes.add_template_recipe(
-            EncodingRecipeBuilder::new("u_id_z", f_unary_imm, 1)
-                .operands_out(vec![gpr])
-                .inst_predicate(InstructionPredicate::new_is_zero_int(format, "imm"))
-                .emit(
-                    r#"
-                        {{PUT_OP}}(bits, rex2(out_reg0, out_reg0), sink);
-                        modrm_rr(out_reg0, out_reg0, sink);
-                    "#,
-                ),
-        );
-    }
+    // XX+rd id unary with zero immediate.
+    recipes.add_template_recipe(
+        EncodingRecipeBuilder::new("u_id_z", f_unary_imm, 1)
+            .operands_out(vec![gpr])
+            .emit(
+                r#"
+                    {{PUT_OP}}(bits, rex2(out_reg0, out_reg0), sink);
+                    modrm_rr(out_reg0, out_reg0, sink);
+                "#,
+            ),
+    );
 
     // XX /n Unary with floating point 32-bit immediate equal to zero.
     {
--- a/filetests/isa/x86/optimized-zero-constants-32bit.clif
+++ b/filetests/isa/x86/optimized-zero-constants-32bit.clif
@@ -1,5 +1,6 @@
-; Check that floating-point constants equal to zero are optimized correctly.
+; Check that floating-point and integer constants equal to zero are optimized correctly.
 test binemit
+set opt_level=best
 target i686
 
 function %foo() -> f32 fast {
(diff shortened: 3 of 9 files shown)


cranelift-codegen/meta-python/isa/x86/encodings.py
```diff
@@ -213,6 +222,9 @@ def enc_i32_i64_ld_st(inst, w_bit, recipe, *args, **kwargs):
 # Finally, the 0xb8 opcode takes an 8-byte immediate with a REX.W prefix.
 X86_64.enc(base.iconst.i64, *r.pu_iq.rex(0xb8, w=1))
 
+# bool 0 can be written as an xor rather than mov.
+enc_both(base.bconst.b1, r.u_id_bool_z, 0x30)
```
iximeow: I'm not sure if this is actually very useful, but saw the opportunity to use the same change as a shorthand for `false`.

Should this clear the entire register for a `b1`? I just realized the other encoding rule uses `b8` for full registers rather than `b0` for byte registers.

cranelift-codegen/meta-python/isa/x86/encodings.py
```diff
@@ -198,9 +198,18 @@ def enc_i32_i64_ld_st(inst, w_bit, recipe, *args, **kwargs):
...
 # band_imm.i32. Can even use the single-byte immediate for 0xffff_ffXX masks.
 
+# 32-bit rules for iconst variants with immediate 0.
+X86_32.enc(base.iconst.i32, *r.u_id_z(0x31))
+
 # Immediate constants.
 X86_32.enc(base.iconst.i32, *r.pu_id(0xb8))
 
+# 64-bit rules for iconst variants with immediate 0.
+X86_64.enc(base.iconst.i32, *r.u_id_z.rex(0x31))
```
iximeow: I have these added before the more broad recipes so they get picked first, because (if I understood correctly), the squeezing step of binemit isn't run for filetests. Since this seems like an extra step if optimizations are enabled, should this be moved later so a more permissive catchall could apply first?

filetests/isa/x86/optimized-zero-constants-32bit.clif
```diff
@@ -16,3 +16,18 @@ ebb0:
...
+  [-,%rax]     v0 = iconst.i32 0       ; bin: 31 c0
+  ; asm: xor %edi, %edi
+  [-,%rdi]     v1 = iconst.i32 0       ; bin: 31 ff
+  return v0
+}
+
+function %zero_byte() -> b1 fast {
+ebb0:
+  ; asm: xor %eax, %eax
+  [-,%rax]     v1 = bconst.b1 false       ; bin: 30 c0
```
iximeow: if the `b1` false recipe is something to keep, I'll copy this into the 64bit tests as well. If not, I'll delete it

iximeow:
A broader comment, which I have about the floating point change as well but applies here: how are flag states tracked? `xor` will clobber some flags, but I didn't have to change any tests around flag preservation. If it's well-defined that flags for conditional operations might be produced from instructions more than the immediate predecessor, I'll add tests around that as well. And of course, that complicates this change :)

lars-t-hansen:
> A broader comment, which I have about the floating point change as well but applies here: how are flag states tracked? `xor` will clobber some flags, .... If it's well-defined that flags for conditional operations might be produced from instructions more than the immediate predecessor, ...

This very question came up in a meeting yesterday here at Mozilla in the context of this exact proposed change :)  I don't think we're convinced yet that Cranelift tracks and considers flags in a consistent / correct manner.  This'll need to be investigated for sure.

bnjbvr:
> > 
> 
> This very question came up in a meeting yesterday here at Mozilla in the context of this exact proposed change :) I don't think we're convinced yet that Cranelift tracks and considers flags in a consistent / correct manner. This'll need to be investigated for sure.

Cranelift tracks flags in two ways: one at the IR level with the instruction attributes `writes_cpu_flags`, that is set when an instruction *returns* flags (the ifcmp family and ffcmp, as far as I can see). Then encoding constraints can signal they clobber flags; this is the default on x86. There's a verifier pass to check that flags are not being clobbered before being used, so it's deferred to the Cranelift users to make sure that this can't happen in practice. The verifier gets run a lot of times during compilation, so it should get caught, whenever this happens; at least the error won't be silently swallowed. But if this can transform an initially valid IR graph into an invalid one, it's of course a compiler error.

Maybe we want to have a way to tell that flags are live when performing legalization, but this might require a second pass over the EBBs in reverse order so as to be able to observe uses. (DCE happens after legalization, so we can't assume that flags observed when iterating in normal order will actually be used.)

lars-t-hansen:
I thought this had been merged, actually...  Anyway, I ran into a related problem here when experimenting with rematerializing constants naively.  In this case, an instruction that generates a constant is ignored, and instead a new instruction that generates the same constant is inserted before every use of the original constant and the use then uses the new name.  This will tend to shrink live ranges and might be good for register allocation (though in reality, good rematerialization must be regalloc-aware).  But constants whose generation set the flags are a problem for this and similar schemes, as the effect of the flags must be tracked when inserting code, see eg https://github.com/lars-t-hansen/cranelift/blob/call_splitting/cranelift-codegen/src/rematerialize.rs#L61. 

I think that on x86, ARM, and ARM64 the only case when generating a constant will set the flags is when we use `xor` to generate a zero: on ARM and ARM64 we control the setting of the flags completely, and on x86 we can load any 32-bit constant directly with `mov` and a 64-bit constant from memory, also with `mov`, without setting flags.

This is all sort of inconsequential; unless we make it an invariant that constant generation can't set the flag then even this type of simple code insertion will need to worry about the flags, and i'm not sure i'm ready to propose that.  Just wanted to record the experience of encountering this problem.

cranelift-codegen/meta-python/isa/x86/encodings.py
iximeow: decided to not deal with booleans here

cranelift-codegen/meta-python/isa/x86/encodings.py
iximeow: ended up putting these rules later so first-pass encoding can happen faster, then opting a few tests into `opt_level=best` to test that squeezing picks these smaller encodings. "instruction length" isn't a great selection metric for these, because they're better alternatives for other reasons, but it'll do

iximeow:
I've fixed up conflicts (as the prior diff here was against python codegen. just a bit dusty...) and dropped the bool `false` encodings here, because I'd like to cover all of `B1`, `B8`, `B16`, `B32`, and `B64` at once, but want to read a bit about `b1` before changing that stuff.

@lars-t-hansen because the only IR instructions Cranelift tracks as writing flags are the `ifcmp`s and `ffcmp`s, and x86 instructions are considered flag-clobber by default (maybe wider than it ought to be?) we shouldn't be in a situation where a hoisted `xor` would change the flags as visible to the instruction selected to consume flags afterward, right? That's not a strong argument for correctness in the face of future changes, but at least lends some confidence that right now this would not be an incorrect change.

Anyway, I can't merge this until someone's approved it, and @bnjbvr I choose you!

lars-t-hansen:
> @lars-t-hansen because the only IR instructions Cranelift tracks as writing flags are the `ifcmp`s and `ffcmp`s, and x86 instructions are considered flag-clobber by default (maybe wider than it ought to be?) we shouldn't be in a situation where a hoisted `xor` would change the flags as visible to the instruction selected to consume flags afterward, right? That's not a strong argument for correctness in the face of future changes, but at least lends some confidence that right now this would not be an incorrect change.

Not quite sure what you're saying here but ...

The problem is not with using the `xor` to generate the constant, that should be fine.  The problem is the hoops one has to jump through to insert a flag-clobbering constant at an arbitrary point in the instruction stream.  For example, it cannot be inserted between the icmp and a subsequent branch, even though (for example) that branch passes a constant via an ebb parameter to its target and it would be natural to generate that constant into the required register just before the branch, after the compare.

I'm not really suggesting that we change anything here; x86 is far too clobber-happy for it to matter very much at the moment.  I'm just saying it's a shame :-)

iximeow:
Ahh I had only thought about the sites we would remove an `xor` to produce the constant earlier, not about the constraints in inserting the moved `xor`. I understand now, thanks :)

cranelift-codegen/meta/src/isa/x86/recipes.rs
```diff
@@ -1002,6 +1002,22 @@ pub(crate) fn define<'shared>(
             ),
     );
 
+    // XX+rd id unary with 32-bit immediate.
+    {
+        let format = formats.get(f_unary_imm);
+        recipes.add_template_recipe(
+            EncodingRecipeBuilder::new("u_id_z", f_unary_imm, 1)
+                .operands_out(vec![gpr])
+                .inst_predicate(InstructionPredicate::new_is_zero_int(format, "imm"))
```
bnjbvr: Ok, I know this is confusing (this confuses me at least), but one of the instruction predicates here or in the encodings instruction predicate is redundant, since they have the same effect. It's better to only keep one of them, until we figure out how to handle this better at the meta crate level (probably just keeping encodings predicates could be sufficient). I'd suggest to remove this one, because the encodings code generator is able to group encodings with the same predicates (removing checks in the legalizer interpreter).

filetests/isa/x86/optimized-zero-constants-32bit.clif
```diff
@@ -1,5 +1,6 @@
-; Check that floating-point constants equal to zero are optimized correctly.
+; Check that floating-point and integer constants equal to zero are optimized correctly.
 test binemit
+set opt_level=best
```
bnjbvr: nit: Instead of changing existing tests, can you add new files instead, please? It's cheap and doesn't require modifying the behavior of current tests, which might lower coverage. (Ditto for the other file.)

cranelift-codegen/meta/src/isa/x86/encodings.rs
```diff
@@ -734,6 +735,18 @@ pub(crate) fn define(
     }
     e.enc64(bconst.bind(B64), rec_pu_id_bool.opcodes(vec![0xb8]).rex());
 
+    let is_zero_int = InstructionPredicate::new_is_zero_int(f_unary_imm, "imm");
+    e.enc_both_instp(
+        iconst.bind(I32),
```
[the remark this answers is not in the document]
bnjbvr: For what it's worth, if there aren't any encodings for these types, legalization will properly create new i32 constants and reduce them.

In the i8 test that's failing, I suspect this changes the number of live values, so adapting to the new numbers should work. Ideally, we'd use the regexp directive and would catch the value number with the regexp, as is done in a few other tests now.

For i16, we could use `rec_u_id_z.opcodes.(vec![0x66, 0x31])` instead, but as @bjorn3 said, upper halves are undefined in this case, and since the instruction operating on larger types is smaller than the one operating on i16, this seems better to use the xor $eax encoding. How about updating the test, and adding a comment (here, and maybe in the test) that the i32 encoding is smaller?

filetests/isa/x86/optimized-zero-constants-32bit.clif
abrown: Also, if you want to ensure the code actually runs as a part of the CI you can add `test run` at the top of those tests and `; run` after any functions you want executed on the current machine. The one limitation now is that the function must return a boolean indicating the test's success or failure. Example here: https://github.com/CraneStation/cranelift/blob/master/filetests/isa/x86/binary64-run.clif.

cranelift-codegen/meta/src/isa/x86/recipes.rs
iximeow: Ah, the redundant predicate bit was a holdover from updating the initial attempt from the old python meta code. I was also confused about that :) I've removed this predicate in favor of the ones in encodings.

filetests/isa/x86/optimized-zero-constants-32bit.clif
iximeow: I'd argue that the existing tests not having `opt_level=best` is incorrect - the test file name _is_ "optimized". I think there's a read that the `pxor r, r` rule for floating point 0's has too high a precedence, since elsewhere we pick a naive but more flexible encoding first for quick unoptimized codegen (like iconst's zero-extending i64 `movl` rule currently).

How's it sound to move the existing pxor tests to another file ("floating-point-zero-constants.clif"?) and keep these under `opt_level=best`?

iximeow: (ended up moving existing pxor tests to a "floating-point-zero-constants" which is an exact copy of the "optimized-zero-constants" test on master. same for the 32-bit versions.)

cranelift-codegen/meta/src/isa/x86/encodings.rs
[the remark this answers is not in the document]
iximeow: Okay, done and done!

filetests/isa/x86/optimized-zero-constants-32bit.clif
bnjbvr: Nice, thank you!

PR #769, opened 2019-05-17
Add x86 encoding for rotr_imm and rotl_imm.

nbp:
While looking at #526, I noticed that x86 encoding does not exists for rotation with immediate values.
This pull request is adding encoding for left and right rotation with immediate value, as well as test cases to check that we legalize & generate the expected bytes.

However, I have no idea if the current python rules for encoding have any checks for the validity of the immediate value range.

### DIFF
--- a/filetests/isa/x86/legalize-rotate.clif
+++ b/filetests/isa/x86/legalize-rotate.clif
@@ -1,0 +1,35 @@
+test compile
+target x86_64
+
+; regex: V=v\d+
+; regex: R=%[a-z0-9]+
+
+function %i32_rotr(i32, i32) -> i32 fast {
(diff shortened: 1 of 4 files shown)


filetests/isa/x86/legalize-rotate.clif
```diff
@@ -0,0 +1,35 @@
...
+    v2 = rotl v0, v1
+    return v2
+}
+
+function %i32_rotl_imm_1(i32) -> i32 fast {
+ebb0(v0: i32):
+    ; check: $V = rotl_imm v0, 1
+    v2 = rotl_imm v0, 1
+    return v2
+}
```
bjorn3: Missing trailing newline

Issue #734, opened 2019-04-12
Some improvements for legalization in Rust


bnjbvr:
I'll use this issue as a dump for things that I've noticed could be enhanced for the legalization module in the Rust meta crate, probably better to do later once we've ported all of the meta code to Rust.

- [ ] Remove the ordering hack in the `x86` legalizer file, which was needed because we want to ensure the same ordering of transform groups as the Python module does (for encodings).
- [ ] When there are several legalizations available for a given Cranelift opcode (e.g. when the opcode is type-dependent), the generated code does unwrap the arguments from the instruction several times, and generates a predicate too. We could unwrap the arguments at the top once and for all, and then generate predicates for each type the legalization affects.
- [ ] Remove the condition when the value of `predicate` is a plain `true`.
- [ ] Try to gather the predicate conditions together, instead of refining the predicate (e.g. `let predicate = value_type == I64; let predicate = predicate && type_sets[42].contains(something);` etc).
- [ ] (In general for the meta crate) Consider adding a `ToRustCode` and `ToRustComment` trait that structs can implement.

bnjbvr:
Adding two items as proposed by @nbp in #799 

- [ ] rename `BranchRange` field `origin` to something else (suggestion was `rel_pc_base`).
- [ ] make `gen_encodings` encodings list generation simpler: see also [comment](https://github.com/CraneStation/cranelift/pull/799#discussion_r300679510).

bnjbvr:
A few items in this list were fixed by #1103 \o/.

PR #771, opened 2019-05-23
[meta] bind() should be a {Bound,}Instruction method and not a static func

bnjbvr:
It's neater and avoids the extra `bind` import when it's needed. @nbp I think you wanted something like this when we looked at a previous PR, so asking you for review please :)

### DIFF
--- a/cranelift-codegen/meta/src/cdsl/ast.rs
+++ b/cranelift-codegen/meta/src/cdsl/ast.rs
@@ -1,5 +1,5 @@
 use crate::cdsl::formats::FormatRegistry;
-use crate::cdsl::inst::{ApplyTarget, Instruction, InstructionPredicate};
+use crate::cdsl::inst::{InstSpec, Instruction, InstructionPredicate};
 use crate::cdsl::operands::{OperandKind, OperandKindFields};
 use crate::cdsl::types::ValueType;
 use crate::cdsl::typevar::{TypeSetBuilder, TypeVar};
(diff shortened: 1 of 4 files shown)


cranelift-codegen/meta/src/cdsl/ast.rs
```diff
@@ -630,7 +574,7 @@ macro_rules! def_rhs {
 
     // inst.type(a, b, c)
```
nbp: follow-up nit: you should also add support for `inst.t1.t2(a, b, c)`, to support producing instructions such as `fcvt_from_uint`

bnjbvr:
Thanks for the quick review! Re-reading my own code makes me wonder if we couldn't remove the `InstSpec` (née `ApplyTarget`) struct entirely, by including the `bound_types` vector in the `Instruction` type. Having a distinct type for bound instructions would make it less likely that we forget about bound types, but reducing the number of different concepts to the minimum sounds appealing.

bnjbvr:
> T[...] me wonder if we couldn't remove the `InstSpec` (née `ApplyTarget`) struct entirely, by including the `bound_types` vector in the `Instruction` type.

Nevermind me: Instruction is a thin `Rc` wrapper on its content, and it shouldn't be modified once it's created, so it makes sense to keep separate concepts here, since `BoundInstruction` can be bound several times and should be mutable because of this.

cranelift-codegen/meta/src/cdsl/ast.rs
bnjbvr: added this to https://github.com/CraneStation/cranelift/issues/734

PR #773, opened 2019-05-23
Cranelift: Redundant stack-slot-to-stack-slot copy removal.

julian-seward1:
This is also https://bugzilla.mozilla.org/show_bug.cgi?id=1552737.

Cranelift currently has a tendency to create redundant copies (self-copies) of
values from a stack slot back to the same stack slot.  This generates a
pointless load and store and an unnecessary register use.  The copies are
created by `visit_inst` in regalloc/reload.rs.  They appear to occur mostly,
but not exclusively, at loop heads.  It's unclear why this happens.

This patch adds a special case to `visit_inst` to find such copies.  They are
converted into a new instruction, `copy_nop`, which takes and produces the
same SSA names, so as not to break any of the SSA invariants, but which has a
zero-length encoding, hence removing the copy at emission time.

`copy_nop`s source and destination operands must be stack slots and of course
the *same* stack slot.  The verifier has been enhanced to check this, since
misuse of `copy_nop` will likely lead to hard-to-find incorrect-code bugs.

Attempts were made to write a standalone .clif test case.  But these failed
because it appears the .clif parser accepts but ignores location hints that
are stack slots.  So it's impossible to write, in clif, the exact form of
`copy` instruction that triggers the transformation.

### DIFF
--- a/cranelift-codegen/meta-python/base/instructions.py
+++ b/cranelift-codegen/meta-python/base/instructions.py
@@ -764,6 +764,16 @@ copy_special = Instruction(
         ins=(src, dst),
         other_side_effects=True)
 
+copy_nop = Instruction(
+        'copy_nop', r"""
+        Stack-slot-to-the-same-stack-slot copy, which is guaranteed to turn
+        into a no-op.  This instruction is for use only within Cranelift
+        itself.
+
+        This instruction copies its input, preserving the value type.
+        """,
+        ins=x, outs=a)
+
 delta = Operand('delta', Int)
 adjust_sp_down = Instruction(
     'adjust_sp_down', r"""
--- a/cranelift-codegen/meta-python/isa/x86/encodings.py
+++ b/cranelift-codegen/meta-python/isa/x86/encodings.py
@@ -341,6 +341,13 @@ enc_x86_64(x86.pop.i64, r.popq, 0x58)
 X86_64.enc(base.copy_special, *r.copysp.rex(0x89, w=1))
 X86_32.enc(base.copy_special, *r.copysp(0x89))
 
+# Stack-slot-to-the-same-stack-slot copy, which is guaranteed to turn
+# into a no-op.
+X86_64.enc(base.copy_nop.i64, r.stacknull, 0)
+X86_64.enc(base.copy_nop.i32, r.stacknull, 0)
+X86_64.enc(base.copy_nop.f64, r.stacknull, 0)
+X86_64.enc(base.copy_nop.f32, r.stacknull, 0)
+
 # Adjust SP down by a dynamic value (or up, with a negative operand).
 X86_32.enc(base.adjust_sp_down.i32, *r.adjustsp(0x29))
 X86_64.enc(base.adjust_sp_down.i64, *r.adjustsp.rex(0x29, w=1))
--- a/cranelift-codegen/src/regalloc/reload.rs
+++ b/cranelift-codegen/src/regalloc/reload.rs
@@ -13,7 +13,7 @@ use crate::cursor::{Cursor, EncCursor};
 use crate::dominator_tree::DominatorTree;
 use crate::entity::{SparseMap, SparseMapValue};
 use crate::ir::{AbiParam, ArgumentLoc, InstBuilder};
-use crate::ir::{Ebb, Function, Inst, InstructionData, Opcode, Value};
+use crate::ir::{Ebb, Function, Inst, InstructionData, Opcode, Value, ValueLoc};
 use crate::isa::RegClass;
 use crate::isa::{ConstraintKind, EncInfo, Encoding, RecipeConstraints, TargetIsa};
 use crate::regalloc::affinity::Affinity;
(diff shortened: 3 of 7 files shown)


cranelift-codegen/meta-python/base/instructions.py
```diff
@@ -764,6 +764,15 @@
         ins=(src, dst),
         other_side_effects=True)
 
+copy_nop = Instruction(
+        'copy_nop', r"""
+        Stack-slot-to-the-same-stack-slot copy, which is guaranteed to turn
```
bnjbvr: Can you add that it's intended to be internally-used only? (and mirror this change in the Rust code)

cranelift-codegen/meta-python/isa/x86/encodings.py
```diff
@@ -341,6 +341,13 @@ def enc_i32_i64_ld_st(inst, w_bit, recipe, *args, **kwargs):
 X86_64.enc(base.copy_special, *r.copysp.rex(0x89, w=1))
 X86_32.enc(base.copy_special, *r.copysp(0x89))
 
+# Stack-slot-to-the-same-stack-slot copy, which is guaranteed to turn
+# into a no-op.
+X86_64.enc(base.copy_nop.i64, r.stacknull, 0)
+X86_64.enc(base.copy_nop.i32, r.stacknull, 0)
+X86_64.enc(base.copy_nop.f64, r.stacknull, 0)
+X86_64.enc(base.copy_nop.f32, r.stacknull, 0)
```
bnjbvr: There's something I've seen which might be equivalent to these 4 lines (worth a try at least!):

```
 X86_64.enc(base.copy_nop.any, r.stacknull, 0) 
```

cranelift-codegen/src/regalloc/reload.rs
```diff
@@ -211,6 +211,46 @@ impl<'a> Context<'a> {
...
+        } = self.cur.func.dfg[inst]
+        {
+            let dst_vals = self.cur.func.dfg.inst_results(inst);
+            if dst_vals.len() == 1 {
+                let dst_val = dst_vals[0];
+                let src_loc = self.cur.func.locations[arg];
+                let dst_loc = self.cur.func.locations[dst_val];
+                let src_aff = self.liveness[arg].affinity;
+                let dst_aff = self.liveness[dst_val].affinity;
+                let can_transform = match (src_loc, dst_loc) {
```
bnjbvr: nit: since `src_loc`, `dst_loc`, `src_aff`, `dst_aff` are used only once, can you replace their uses with their definitions, please? (or does the borrow check not like that)

cranelift-codegen/src/regalloc/reload.rs
```diff
@@ -211,6 +211,46 @@ impl<'a> Context<'a> {
...
+                    (ValueLoc::Stack(src_slot), ValueLoc::Stack(dst_slot)) => {
+                        src_slot == dst_slot && src_aff.is_stack() && dst_aff.is_stack()
```
bnjbvr: Can you explain why we need to at the values' affinities? Even if a value is supposed to live in a register and is not (because of a diversion, I imagine?), I can't imagine a reason why we couldn't avoid the spill.

cranelift-codegen/src/verifier/mod.rs
```diff
@@ -1469,6 +1472,45 @@ impl<'a> Verifier<'a> {
...
+            arg,
+        } = self.func.dfg[inst]
+        {
+            let dst_vals = self.func.dfg.inst_results(inst);
+            if dst_vals.len() != 1 {
+                return fatal!(errors, inst, "copy_nop must produce exactly one result");
+            }
+            let dst_val = dst_vals[0];
+            let src_type = self.func.dfg.value_type(arg);
+            let dst_type = self.func.dfg.value_type(dst_val);
```
bnjbvr: nit: since these two variables are used only once, can you replace their uses by their definitions?

cranelift-codegen/src/verifier/mod.rs
```diff
@@ -1469,6 +1472,45 @@ impl<'a> Verifier<'a> {
...
+            if dst_type != src_type {
+                return fatal!(errors, inst, "copy_nop src and dst types must be the same");
```
bnjbvr: This check isn't done in the code added to perform the optimization: is it that regalloc ensures the types were the same, so this is a sanity check?

cranelift-codegen/src/verifier/mod.rs
```diff
@@ -1469,6 +1472,45 @@ impl<'a> Verifier<'a> {
...
+            }
+            let src_loc = self.func.locations[arg];
+            let dst_loc = self.func.locations[dst_val];
```
bnjbvr: same nit about single-use of definition that above

cranelift-codegen/meta-python/isa/x86/encodings.py
julian-seward1: That produces
```
  File "/home/sewardj/MOZ/CL/may27/cranelift-codegen/meta-python/cdsl/instructions.py", line 341, in __getattr__
    assert name != 'any', 'Wildcard not allowed for ctrl_typevar'
```
so I backed it out.  I do notice other uses of `.any` in encodings.py, but they all have a type-descriptor fragment before the `.any`, for example `base.istore32.i64.any` and `base.jump_table_entry.i64.any.any`.

cranelift-codegen/src/regalloc/reload.rs
julian-seward1: Hmm, excess paranoia on my part.  I removed it.  On reading the comments for `Affinity` I see it's not relevant here.

cranelift-codegen/src/verifier/mod.rs
julian-seward1: These are used twice, once in the `match` and once in the error message.

cranelift-codegen/src/verifier/mod.rs
julian-seward1: Yes, this is just a sanity check.

PR #774, opened 2019-05-24
Generalize information about compiled code

lars-t-hansen:
This patch generalizes the information that is available about the compiled-code vector in the following ways:

- it records the offsets to and sizes of the different sections of the compiled-code vector
  (machine code, jump tables, raw bits)
- it records (by means of the existing reloc system) the locations in the code portion that need to be
  fixed up if the jump tables or raw bits are moved relative to the code.
- it propagates a new structure, CodeOffsets, around where needed as a generalization of the
  existing ad-hoc system for code_size / total_size.

(At the moment, there are no "raw bits" but we keep talking about needing constant pools, so I put in the trivial bits to support this, to suggest the way forward.  There are two remaining TODO comments that would need to be dealt with for full support.  Discuss.)

The context here is that SpiderMonkey needs to move the jump tables out of the way to make space for its epilogue (https://bugzilla.mozilla.org/show_bug.cgi?id=1547752).  One alternative solution requires SpiderMonkey to create IR nodes for its prologue/epilogue so that their sizes can be accounted for properly, but this is hard because the prologue/epilogue can't be emitted until the rest of the code has been generated.  Simply splitting the code vector is easier.

### DIFF
--- a/cranelift-codegen/src/binemit/memorysink.rs
+++ b/cranelift-codegen/src/binemit/memorysink.rs
@@ -14,7 +14,7 @@
 //! relocations to a `RelocSink` trait object. Relocations are less frequent than the
 //! `CodeSink::put*` methods, so the performance impact of the virtual callbacks is less severe.
 
-use super::{Addend, CodeOffset, CodeSink, Reloc};
+use super::{Addend, CodeInfo, CodeOffset, CodeSink, Reloc};
 use crate::ir::{ExternalName, JumpTable, SourceLoc, TrapCode};
 use core::ptr::write_unaligned;
 
@@ -30,12 +30,14 @@ use core::ptr::write_unaligned;
 /// Note that `MemoryCodeSink` writes multi-byte values in the native byte order of the host. This
 /// is not the right thing to do for cross compilation.
 pub struct MemoryCodeSink<'a> {
+    /// Pointer to start of sink's preallocated memory.
     data: *mut u8,
+    /// Offset is isize because its major consumer needs it in that form.
     offset: isize,
-    /// Size of the machine code portion of output
-    pub code_size: isize,
     relocs: &'a mut RelocSink,
     traps: &'a mut TrapSink,
+    /// Information about the generated code and read-only data.
+    pub info: CodeInfo,
 }
 
 impl<'a> MemoryCodeSink<'a> {
@@ -47,7 +49,12 @@ impl<'a> MemoryCodeSink<'a> {
         Self {
             data,
             offset: 0,
-            code_size: 0,
+            info: CodeInfo {
+                code_size: 0,
+                jumptables_size: 0,
+                rodata_size: 0,
+                total_size: 0,
+            },
             relocs,
             traps,
         }
@@ -75,40 +82,35 @@ pub trait TrapSink {
     fn trap(&mut self, _: CodeOffset, _: SourceLoc, _: TrapCode);
 }
 
+impl<'a> MemoryCodeSink<'a> {
+    fn write<T>(&mut self, x: T) {
+        unsafe {
+            #[cfg_attr(feature = "cargo-clippy", allow(clippy::cast_ptr_alignment))]
+            write_unaligned(self.data.offset(self.offset) as *mut T, x);
+            self.offset += std::mem::size_of::<T>() as isize;
+        }
+    }
+}
+
 impl<'a> CodeSink for MemoryCodeSink<'a> {
     fn offset(&self) -> CodeOffset {
         self.offset as CodeOffset
     }
 
     fn put1(&mut self, x: u8) {
-        unsafe {
-            write_unaligned(self.data.offset(self.offset), x);
-        }
-        self.offset += 1;
+        self.write(x);
     }
 
     fn put2(&mut self, x: u16) {
-        unsafe {
-            #[cfg_attr(feature = "cargo-clippy", allow(clippy::cast_ptr_alignment))]
-            write_unaligned(self.data.offset(self.offset) as *mut u16, x);
-        }
-        self.offset += 2;
+        self.write(x);
     }
 
     fn put4(&mut self, x: u32) {
-        unsafe {
-            #[cfg_attr(feature = "cargo-clippy", allow(clippy::cast_ptr_alignment))]
-            write_unaligned(self.data.offset(self.offset) as *mut u32, x);
-        }
-        self.offset += 4;
+        self.write(x);
     }
 
     fn put8(&mut self, x: u64) {
-        unsafe {
-            #[cfg_attr(feature = "cargo-clippy", allow(clippy::cast_ptr_alignment))]
-            write_unaligned(self.data.offset(self.offset) as *mut u64, x);
-        }
-        self.offset += 8;
+        self.write(x);
     }
 
     fn reloc_ebb(&mut self, rel: Reloc, ebb_offset: CodeOffset) {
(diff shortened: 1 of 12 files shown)


cranelift-codegen/src/binemit/memorysink.rs
```diff
@@ -48,10 +54,25 @@ impl<'a> MemoryCodeSink<'a> {
...
             code_size: 0,
+            jumptables_size: 0,
+            rodata_size: 0,
+            total_size: 0,
             relocs,
             traps,
         }
     }
+
+    /// Return information about generarated code as a CodeOffsets structure.
```
bnjbvr: nit: `generated`

cranelift-codegen/src/binemit/mod.rs
```diff
@@ -63,6 +66,35 @@ impl fmt::Display for Reloc {
...
+#[derive(PartialEq)]
+pub struct CodeOffsets {
+    /// Number of bytes of machine code (the code starts at offset 0).
+    pub code_size: CodeOffset,
+
+    /// Offset of any relocatable jump tables, or equal to rodata if there are no jump tabes.
+    pub jumptables: CodeOffset,
+
+    /// Number of bytes of jumptables.
+    pub jumptables_size: CodeOffset,
```
bnjbvr: Here and below (for `rodata_size`): it seems this information duplicates what's provided by `jumptables`, since `jumptables == code_size + jumptables_size`. Maybe we could get rid of one of these two?

If we don't want to get rid of them, it seems weird that a CodeOffset may represent a size; could we just use an `usize` there?

bnjbvr:
> One alternative solution requires SpiderMonkey to create IR nodes for its prologue/epilogue so that their sizes can be accounted for properly, but this is hard because the prologue/epilogue can't be emitted until the rest of the code has been generated.

Interesting idea, I can see it working, if we have the ability to predict fixed sizes for the prologue/epilogues: the Cranelift instruction could have an encoding with a variable size, set by the Cranelift user; then it would emit only 0s for the defined number of bytes, and Spidermonkey could override them. Or these could just be nops.

cranelift-codegen/src/binemit/mod.rs
lars-t-hansen: I could get rid of eg jumptables and rodata since their offsets can be computed from the sizes, but I'd want to introduce methods to compute them because it's unreasonable for clients to have to perform those computations themselves.  I can investigate that.

Re usize vs CodeOffset, I thought about this but was not sure what to do here.  CodeOffset is perhaps the "wrong" type because it's an offset, but it's the "right" type because no size can be greater than the CodeOffset's range, while usize does allow for a size greater than that.

I think the problem here is the word "offset" in that type name.  This is not actually a type that represents an offset abstractly; it is for sure an unsigned integer type that represents positions in a code segment and as such can (and should) be used to represent the length of the segment as well (the position after the last byte).

With that background, I think I will continue to use the CodeOffset to represent the length.

lars-t-hansen:
> if we have the ability to predict fixed sizes for the prologue/epilogues: the Cranelift instruction could have an encoding with a variable size, set by the Cranelift user; then it would emit only 0s for the defined number of bytes, and Spidermonkey could override them. Or these could just be nops.

I did think about that but decided that trying to predict the prologue/epilogue sizes properly without leaving a lot of extra nops in (and using extra code space) was too much work for too little gain.

cranelift-codegen/src/binemit/mod.rs
lars-t-hansen: Created getters for those two fields.  Turns out nobody was accessing them yet anyway.

cranelift-codegen/src/binemit/mod.rs
```diff
@@ -63,6 +66,41 @@ impl fmt::Display for Reloc {
...
+
+    /// Number of bytes of rodata.
+    pub rodata_size: CodeOffset,
+
+    /// Number of bytes in total.
+    pub total_size: CodeOffset,
+}
+
+impl CodeOffsets {
+    /// Offset of any relocatable jump tables, or equal to rodata if there are no jump tabes.
```
bnjbvr: nit: jump tables

lars-t-hansen: There's always something :)

cranelift-codegen/src/binemit/memorysink.rs
```diff
@@ -34,6 +34,12 @@ pub struct MemoryCodeSink<'a> {
     offset: isize,
     /// Size of the machine code portion of output
     pub code_size: isize,
+    /// Size of jumptable portion of output (following code)
+    pub jumptables_size: isize,
+    /// Size of read-only data portion of output (following jump tables)
+    pub rodata_size: isize,
+    /// Size of everything
+    pub total_size: isize,
```
bjorn3: Maybe add a `CodeOffsets` field instead.

lars-t-hansen: Yeah, I was pondering that but then I saw "isize" which is signed and I started worrying about semantics of overflow/underflow.  Memorysink is small but these are public fields, so I'll have to worry about all the consumers here.  I'll investigate.

Which reminds me, how much do we know about consumers of Cranelift outside of Firefox?  Is everyone listening in on the traffic in this repo?

lars-t-hansen:
Another sad reality is that after the fix I did earlier today, there are no "offset" fields left in CodeOffsets, all it has are sizes of the different sections of the generated code+rodata.  So CodeOffsets probably should not really be called that :-/

cranelift-codegen/src/binemit/memorysink.rs
bjorn3: A size can never be negative right? So it should probably use an unsigned integer.

> how much do we know about consumers of Cranelift outside of Firefox? 

That I am one :)

lars-t-hansen: Well, no, a size can never be negative but the difference of two sizes can be.

lars-t-hansen: OK, updated this and cleaned up MemoryCodeSink a little in the process.  The offset field remains isize because its dominant use requires that form but I've commented to clarify this.  The structure is now called CodeInfo which is generic but more appropriate than CodeOffsets.

lars-t-hansen: Turns out MemoryCodeSink can be simplified further, more updates coming.

cranelift-codegen/src/binemit/memorysink.rs
```diff
@@ -30,12 +30,14 @@ use core::ptr::write_unaligned;
 /// Note that `MemoryCodeSink` writes multi-byte values in the native byte order of the host. This
 /// is not the right thing to do for cross compilation.
 pub struct MemoryCodeSink<'a> {
+    // Pointer to start of sink's preallocated memory.
```
bnjbvr: nit: use `///` instead of `//` so this shows up in IDE help

cranelift-codegen/src/binemit/memorysink.rs
```diff
@@ -30,12 +30,14 @@ use core::ptr::write_unaligned;
...
     data: *mut u8,
+    // Offset is isize because its major consumer needs it in that form.
```
bnjbvr: ditto

src/wasm.rs
```diff
@@ -176,16 +176,16 @@ fn handle_module(
                 return Err(pretty_verifier_error(&context.func, fisa.isa, None, errors));
             }
         } else {
-            let (code_size, rodata_size) = context
+            let offsets = context
```
bnjbvr: nit: maybe rename `offsets` to `code_info`?

src/compile.rs
```diff
@@ -65,7 +65,7 @@ fn handle_module(
         let mut mem = vec![];
 
         // Compile and encode the result to machine code.
-        let (code_size, rodata_size) = context
+        let offsets = context
```
bnjbvr: nit: maybe rename `offsets` to `code_info`?

cranelift-codegen/src/binemit/memorysink.rs
lars-t-hansen: Done.

cranelift-codegen/src/binemit/memorysink.rs
lars-t-hansen: Done.

src/wasm.rs
lars-t-hansen: Done.

src/compile.rs
lars-t-hansen: Done.

PR #775, opened 2019-05-28
Fix url in comments

11Takanori:
`https://webassembly.github.io/spec/binary/modules.html#code-section` is not found.


bnjbvr:
Thank you!

PR #777, opened 2019-05-28
Rename SimpleGVN to GVN everywhere

bnjbvr:


bnjbvr:
For what it's worth, @sunfishcode didn't like this idea: the "simple" meant here that Cranelift will not do anything complicated, since we expect the CLIR producer to have done so. My objection to "simple" is that "simple" is a very relative wording (that is, what you think is simple may not be simple to somebody else) thus calling something "simple" may be discouraging.

Just the fact there's been such a confusion over the word "simple" makes me think that this is a problem, and that we should use an alternative wording that 1. isn't confusing, 2. clearly spells out that we're not doing any more subtle / complete GVN there. "BasicGVN" and "MinimalGVN" come to mind here, although the former potentially carries the same confusion, and the latter could mean a GVN that results in something minimal, which is unfortunate.

An alternative could be to rename the files to GVN as this PR does, and tweak the comments to clearly state that the GVN applied here does the least amount of work.

@sunfishcode Any preference? We can also just close the PR and let things as is anyway, since this is just wording and low priority.

bnjbvr:
Closing, there's no clear mutual agreement over naming here. In the future there might be a more complicated, more powerful GVN, and maybe the topic of naming can be brought back. (For the record, we thought about calling it "Fast GVN", since it's supposed to run quickly, as opposed to the future GVN algorithm, but there was no strong consensus/will to do so.)

PR #778, opened 2019-05-28
Add bugpoint command

### LINKED ISSUE #532: Bugpoint for clif
https://llvm.org/docs/Bugpoint.html

I currently have to reduce error reproductions by hand, which is annoying.

bjorn3:
You should probably run this in release mode, as it is even more slow otherwise.

Fixes #532 

## Passes
* remove instruction
* replace instruction with single ret val with iconst
* replace instruction with trap
* remove ebb

Edit: added additional shrinking passes

### DIFF
--- a/src/bugpoint.rs
+++ b/src/bugpoint.rs
@@ -199,10 +199,8 @@ fn reduce(isa: &TargetIsa, mut func: Function, verbose: bool) {
                 PhaseStepResult::NextPhase(msg, count) => {
                     progress.set_message("done");
                     progress.finish();
-                    progress = ProgressBar::with_draw_target(
-                        count as u64,
-                        ProgressDrawTarget::stdout_nohz(),
-                    );
+                    progress =
+                        ProgressBar::with_draw_target(count as u64, ProgressDrawTarget::stdout());
                     progress.set_style(
                         ProgressStyle::default_bar()
                             .template("{bar:80} {prefix:40} {pos:>4}/{len:>4} {msg}"),
--- a/src/bugpoint.rs
+++ b/src/bugpoint.rs
@@ -73,7 +73,7 @@ fn next_inst_ret_prev(func: &Function, ebb: &mut Ebb, inst: &mut Inst) -> Option
 }
 
 fn reduce(isa: &TargetIsa, mut func: Function) {
-    'outer_loop: for _ in 0..100 {
+    'outer_loop: for pass_idx in 0..100 {
         let mut was_reduced = false;
         let first_ebb = func.layout.entry_block().unwrap();
         let mut phase = Phase::RemoveInst(first_ebb, func.layout.first_inst(first_ebb).unwrap());
@@ -81,16 +81,17 @@ fn reduce(isa: &TargetIsa, mut func: Function) {
         'inner_loop: for _ in 0..1000 {
             let mut func2 = func.clone();
 
-            match phase {
+            let msg = match phase {
                 Phase::RemoveInst(ref mut ebb, ref mut inst) => {
                     if let Some(prev_inst) = next_inst_ret_prev(&func2, ebb, inst) {
-                        print!("Remove inst {}: ", prev_inst);
                         func2.layout.remove_inst(prev_inst);
+                        format!("Remove inst {}", prev_inst)
                     } else {
                         phase = Phase::ReplaceInstWithIconst(
                             first_ebb,
                             func2.layout.first_inst(first_ebb).unwrap(),
                         );
+                        continue 'inner_loop;
                     }
                 }
                 Phase::ReplaceInstWithIconst(ref mut ebb, ref mut inst) => {
@@ -98,8 +99,8 @@ fn reduce(isa: &TargetIsa, mut func: Function) {
                         let results = func2.dfg.inst_results(prev_inst);
                         if results.len() == 1 {
                             let ty = func2.dfg.value_type(results[0]);
-                            print!("Replace inst {} with iconst.{}: ", prev_inst, ty);
                             func2.dfg.replace(prev_inst).iconst(ty, 0);
+                            format!("Replace inst {} with iconst.{}", prev_inst, ty)
                         } else {
                             continue 'inner_loop; // No change, continue with next instruction
                         }
@@ -108,44 +109,31 @@ fn reduce(isa: &TargetIsa, mut func: Function) {
                             first_ebb,
                             func2.layout.first_inst(first_ebb).unwrap(),
                         );
+                        continue 'inner_loop;
                     }
                 }
                 Phase::ReplaceInstWithTrap(ref mut ebb, ref mut inst) => {
                     if let Some(prev_inst) = next_inst_ret_prev(&func2, ebb, inst) {
-                        print!("Replace inst {} with trap: ", prev_inst);
                         func2.dfg.replace(prev_inst).trap(TrapCode::User(0));
+                        format!("Replace inst {} with trap", prev_inst)
                     } else {
                         phase = Phase::RemoveEbb(first_ebb);
+                        continue 'inner_loop;
                     }
                 }
                 Phase::RemoveEbb(ref mut ebb) => {
+                    let prev_ebb = *ebb;
                     if let Some(next_ebb) = func2.layout.next_ebb(*ebb) {
-                        println!("Remove ebb {}: ", ebb);
-                        func2.layout.remove_ebb(*ebb);
                         *ebb = next_ebb;
+                        func2.layout.remove_ebb(*ebb);
+                        format!("Remove ebb {}", prev_ebb)
                     } else {
                         break 'inner_loop;
                     }
                 }
-            }
+            };
 
-            match std::panic::catch_unwind(std::panic::AssertUnwindSafe(|| {
-                if let Err(err) = cranelift_codegen::verifier::verify_function(&func2, isa) {
-                    // Shrinking produced invalid clif, discard changes.
-                    println!("verifier error {:?}", err);
-                    false
-                } else {
-                    true
-                }
-            })) {
-                Ok(false) => continue,
-                Ok(true) => {}
-                Err(_err) => {
-                    // FIXME prevent verifier panic on removing ebb1
-                    println!("verifier panicked");
-                    continue;
-                }
-            }
+            print!("{}: ", msg);
 
             match check_for_crash(isa, &func2) {
                 Res::Succeed => {
@@ -162,9 +150,12 @@ fn reduce(isa: &TargetIsa, mut func: Function) {
                     // Panic remained while shrinking, make changes definitive.
                     was_reduced = true;
                     func = func2;
+                    println!("{}: shrink", msg);
                 }
             }
         }
+
+        println!("Pass {} finished", pass_idx);
 
         if !was_reduced {
             // No new shrinking opportunities have been found this pass. This means none will ever
@@ -180,7 +171,7 @@ fn reduce(isa: &TargetIsa, mut func: Function) {
 
 enum Res {
     Succeed,
-    Verifier(cranelift_codegen::CodegenError),
+    Verifier(String),
     Panic,
 }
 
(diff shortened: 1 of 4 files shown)


bjorn3:
This is ready for review.

Usage:

```sh
cargo run --release bugpoint crash.clif 2>/dev/null
████████████████████████████████████████████████████████████ pass 0 phase remove inst                 1124/1124 done
████████████████████████████████████████████████████████████ pass 0 phase replace inst with iconst     744/ 744 done
████████████████████████████████████████████████████████████ pass 0 phase replace inst with trap       744/ 744 done
████████████████████████████████████████████████████████████ pass 0 phase remove ebb                   121/ 121 done
████████████████████████████████████████████████████████████ pass 1 phase remove inst                  743/ 743 done
████████████████████████████████████████████████████████████ pass 1 phase replace inst with iconst     266/ 266 done
████████████████████████████████████████████████████████████ pass 1 phase replace inst with trap       266/ 266 done
████████████████████████████████████████████████████████████ pass 1 phase remove ebb                    87/  87 done
████████████████████████████████████████████████████████████ pass 2 phase remove inst                  263/ 263 done
████████████████████████████████████████████████████████████ pass 2 phase replace inst with iconst     220/ 220 done
████████████████████████████████████████████████████████████ pass 2 phase replace inst with trap       220/ 220 done
████████████████████████████████████████████████████████████ pass 2 phase remove ebb                    86/  86 done
████████████████████████████████████████████████████████████ pass 3 phase remove inst                  220/ 220 done
████████████████████████████████████████████████████████████ pass 3 phase replace inst with iconst     220/ 220 done
████████████████████████████████████████████████████████████ pass 3 phase replace inst with trap       220/ 220 done
████████████████████████████████████████████████████████████ pass 3 phase remove ebb                    86/  86 done

[... reduced function]

121 ebbs 1124 insts -> 86 ebbs 220 insts
```

src/bugpoint.rs
```diff
@@ -0,0 +1,312 @@
...
+
+    let mut relocs = PrintRelocs::new(false);
+    let mut traps = PrintTraps::new(false);
+    let mut mem = vec![];
+
+    use std::io::Write;
+    std::io::stdout().flush().unwrap(); // Flush stdout to sync with panic messages on stderr
+
+    match std::panic::catch_unwind(std::panic::AssertUnwindSafe(|| {
+        if let Err(err) = cranelift_codegen::verifier::verify_function(&func, isa) {
```
bnjbvr: I think there's a method to convert a Result into an Option directly.

src/bugpoint.rs
```diff
@@ -0,0 +1,312 @@
...
+    println!(
+        "{} ebbs {} insts -> {} ebbs {} insts",
+        orig_ebb_count,
+        orig_inst_count,
+        ebb_count(&func),
+        inst_count(&func)
+    );
+}
+
+enum Res {
```
bnjbvr: This name is a bit terse, how about CheckResult?

src/bugpoint.rs
```diff
@@ -0,0 +1,312 @@
...
+        if let Err(err) = cranelift_codegen::verifier::verify_function(&func, isa) {
+            Some(err)
+        } else {
+            None
+        }
+    })) {
+        Ok(Some(err)) => return Res::Verifier(err.to_string()),
+        Ok(None) => {}
+        Err(err) => {
+            // FIXME prevent verifier panic on removing ebb1
```
bnjbvr: FIXME here.

src/bugpoint.rs
```diff
@@ -0,0 +1,312 @@
...
+        return Err(String::from("compilation requires a target isa"));
+    };
+
+    std::env::set_var("RUST_BACKTRACE", "0"); // Disable backtraces to reduce verbosity
+
+    for (func, _) in test_file.functions {
+        reduce(isa, func, verbose);
+    }
+
+    //print!("{}", timing::take_current());
```
bnjbvr: nit: comment

src/bugpoint.rs
```diff
@@ -0,0 +1,312 @@
...
+                    *self = Phase::ReplaceInstWithTrap(
+                        first_ebb,
+                        func.layout.first_inst(first_ebb).unwrap(),
+                    );
+                    PhaseStepResult::NextPhase("replace inst with trap", inst_count(func))
+                }
+            }
+            Phase::ReplaceInstWithTrap(ref mut ebb, ref mut inst) => {
+                if let Some((_prev_ebb, prev_inst)) = next_inst_ret_prev(func, ebb, inst) {
+                    func.dfg.replace(prev_inst).trap(TrapCode::User(0));
```
bnjbvr: Replacing an instruction by a trap (which is a terminal instruction) in the middle of a block means that this block has to be split, which is likely to reduce the number of verifier errors encountered while reducing. Maybe a good idea for a follow-up!

src/bugpoint.rs
```diff
@@ -0,0 +1,312 @@
...
+                    break 'inner_loop;
+                }
+            };
+
+            progress.set_message(&msg);
+
+            match check_for_crash(isa, &func2) {
+                Res::Succeed => {
+                    // Shrinking didn't hit the problem anymore, discard changes.
+                    //progress.println("succeeded");
```
bnjbvr: nit: here and below, commented calls to progress

src/bugpoint.rs
```diff
@@ -0,0 +1,312 @@
+//! CLI tool to read Cranelift IR files and compile them into native code.
```
bnjbvr: nit: this comment needs an update

src/bugpoint.rs
```diff
@@ -0,0 +1,312 @@
...
+    filename: &str,
+    flag_set: &[String],
+    flag_isa: &str,
+    verbose: bool,
+) -> Result<(), String> {
+    let parsed = parse_sets_and_triple(flag_set, flag_isa)?;
+    let fisa = parsed.as_fisa();
+
+    let path = Path::new(&filename).to_path_buf();
+    let name = String::from(path.as_os_str().to_string_lossy());
```
bnjbvr: Why is the round trip from filename to name needed, here?

src/bugpoint.rs
```diff
@@ -0,0 +1,312 @@
...
+fn resolve_aliases(func: &mut Function) {
+    for ebb in func.layout.ebbs() {
+        for inst in func.layout.ebb_insts(ebb) {
+            func.dfg.resolve_aliases_in_arguments(inst);
+        }
+    }
+}
+
+fn reduce(isa: &TargetIsa, mut func: Function, verbose: bool) {
+    let (orig_ebb_count, orig_inst_count) = (ebb_count(&func), inst_count(&func));
```
bnjbvr: We probably don't want to reduce a function which compilation doesn't panic in the first place, right?

src/bugpoint.rs
```diff
@@ -0,0 +1,312 @@
...
+    );
+}
+
+enum Res {
+    Succeed,
+    Verifier(String),
+    Panic,
+}
+
+fn check_for_crash(isa: &TargetIsa, func: &Function) -> Res {
```
bnjbvr: There can be several kinds of panics in the pipeline, and I'm afraid that checking for a crash as is could hide the initial one we're looking for, and instead cause a panic in an unrelated location. It would be nice to handle this case, e.g. by checking that the panic reason is the same as the one we observed during the first run I'm talking about above. Bonus points if, in this case, we can still have the clif causing the new panic.

src/bugpoint.rs
bjorn3: It isnt I think. I copy pasted and inlined some functions from elsewhere.

src/bugpoint.rs
bjorn3: Yes

bjorn3:
I will make the requested changes in a few days.

bjorn3:
@bnjbvr I have some more reduction methods in mind (eg replacing ebb params with `iconst` (`jump ebb1(v1) ebb1(v1: i8): ...` to `jump ebb1 ebb1: v1 = iconst.i8 0 ...`) Do you want me to open a new issue to track them or do you want to keep #532 open?

src/bugpoint.rs
bjorn3: The panic is: "EBB node must have one reachable predecessor".

If I run `cargo run compile` on the ir shown when inserting `println!("{}", func);` here, it gives a normal verifier error:

```
[...]
    jump ebb1
;   ^~~~~~~~~
; error: inst102: invalid ebb reference ebb1


ebb2:
    trap user0
[...]
```

bnjbvr: I think there's a method to recompute successors and predecessors after an EBB has been introduced or removed, so using it might avoid this verifier issue.

bnjbvr:
> @bnjbvr I have some more reduction methods in mind (eg replacing ebb params with iconst (jump ebb1(v1) ebb1(v1: i8): ... to jump ebb1 ebb1: v1 = iconst.i8 0 ...)) Do you want me to open a new issue to track them or do you want to keep #532 open?

New issue sounds fine, thanks.

src/bugpoint.rs
bjorn3: I believe you are refering to [`Context::compute_cfg`](https://docs.rs/cranelift-codegen/0.30.0/cranelift_codegen/struct.Context.html#method.compute_cfg). Calling it wont solve it, as the `Context` is not reused.

bjorn3:
Ping

bnjbvr:
I'll try to review this this week.

bnjbvr:
@nbp said he could help with this review. Thanks!

nbp:
I will look at this issue tomorrow.

src/bugpoint.rs
```diff
@@ -0,0 +1,320 @@
...
+enum PhaseStepResult {
+    Shrinked(String),
+    Replaced(String),
+    NoChange,
+    NextPhase(&'static str, usize),
+    Finished,
+}
+
+impl Phase {
+    fn step(&mut self, func: &mut Function) -> PhaseStepResult {
```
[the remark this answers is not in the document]
bjorn3: Is it better now?

src/bugpoint.rs
```diff
@@ -0,0 +1,423 @@
...
+}
+
+/// Try to remove instructions.
+struct RemoveInst {
+    ebb: Ebb,
+    inst: Inst,
+}
+
+impl RemoveInst {
+    fn new(func: &Function) -> Box<dyn Mutator> {
```
nbp: nit: This is an unexpected signature for `new` function in Rust. Maybe move the boxing outside of this function.

src/bugpoint.rs
```diff
@@ -0,0 +1,423 @@
...
+            ProgressStyle::default_bar()
+                .template("{bar:60} {prefix:40} {pos:>4}/{len:>4} {msg}"),
+        );
+
+        progress.set_prefix(&(progress_bar_prefix + &format!(" phase {}", self.name())));
+
+        for _ in 0..10000 {
+            progress.inc(1);
+
+            let (mutated_func, msg, mutation_kind) = match self.mutate(func.clone()) {
```
nbp: Note: There is one thing which conceptually slightly troubles me a bit about this code. Here we do a `func.clone()`, but a mutator such as `RemoveInst` keep both an instruction and block index.  Technically these indexes should only be valid within a given function, and using them in a clone is abusing the entity system of Cranelift for not keeping track of its primary and secondary vectors with which they can safely be used.

While I admit that this works in practice, this could be a pain point if we were to make a more secure entity system.

src/bugpoint.rs
```diff
@@ -0,0 +1,423 @@
...
+                return;
+            }
+            CheckResult::Panic => {}
+        }
+    }
+    let (orig_ebb_count, orig_inst_count) = (ebb_count(&func), inst_count(&func));
+
+    resolve_aliases(&mut func);
+
+    'outer_loop: for pass_idx in 0..100 {
```
nbp: nit: remove the `outer_loop` label.

src/bugpoint.rs
nbp: nit: One way would be to only reduce on panics which are keeping a similar error message. This way, even if the panic message changes, it would not be accepted as a reducing case.

src/bugpoint.rs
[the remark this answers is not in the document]
nbp: Yes!

src/bugpoint.rs
bjorn3: It has to clone anyway, to be able to revert mutations that are not resulting in the wanted error.

src/bugpoint.rs
bjorn3: I prefer to just fix the verifier. Having other panics being allowed would make it easier to find panics which you didnt even know existed. Maybe even make bugpoint reduce all different panics it found.

bjorn3:
I added a pass to remove unused entities (stack slots, ext funcs, global values). I also added a test to check that bugpoint does actually succeed in reducing a test case.

PR #782, opened 2019-05-29
Add register-specific affinities and use them for ABI register parameters

iximeow:
The motivation for this change is to avoid the unnecessary `mov`s in code generated for preparing call arguments - what was producing
```
mov rax, qword [rsp]
mov eax, dword [rax + rdx]
mov rdx, qword [rsp]
mov rdi, rdx
mov esi, ecx
mov edx, eax
call <fn>
```

produces, after these changes,
```
mov rax, qword [rsp]
mov edx, dword [rax + rdx]
mov rdi, qword [rsp]
call <fn>
```
yay! :sparkles: 

I'm putting up a PR at this point because while I think these changes are still very messy, I'm at a point where the logical changes seem sound and I'd like to get eyes on that. And if someone has specific style thoughts, I'm more than happy to hear them!

Fundamentally the interesting changes are adding `Affinity::RegUnit`, switching `Affinity::abi` to use that, instead of `Affinity::Reg`, and adjusting coloring to take `Affinity::RegUnit` as a hint on how to color values. Then, when building live ranges, constrain affinity to the ABI-appropriate register. Additionally, because function arguments in my test case were the result of a spill/load, I adjusted `ReloadCandidate` to optionally hint at a specific register to reload into.

Most of the rest of the changes fell out of chasing what used `Affinity::Reg` and needed to learn about `Affinity::RegUnit`, though there's a few cases that don't make a ton of sense to me but seem to result in correct codegen. Comments on those inline.

~~A lot of the messiness here is from adjusting `if let Affinity::Reg(rci) = ... { A }` to `if let Affinity::Reg(rci) = ... { A } else if let Affinity::RegUnit(unit) { B; A }` throughout regalloc. I didn't want to get too tied up in cleaning that up until being sure these changes are the right way to go, but I'm still not sure quite what a nice way to write them would be.~~ (ended up cleaning up most of that here)

### DIFF
--- a/cranelift-codegen/src/regalloc/affinity.rs
+++ b/cranelift-codegen/src/regalloc/affinity.rs
@@ -9,7 +9,9 @@
 //! larger register class instead.
 
 use crate::ir::{AbiParam, ArgumentLoc};
-use crate::isa::{ConstraintKind, OperandConstraint, RegClassIndex, RegInfo, TargetIsa};
+use crate::isa::{
+    ConstraintKind, OperandConstraint, RegClass, RegClassIndex, RegInfo, RegUnit, TargetIsa,
+};
 use core::fmt;
 
 /// Preferred register allocation for an SSA value.
@@ -25,7 +27,10 @@ pub enum Affinity {
     Stack,
 
     /// This value prefers a register from the given register class.
-    Reg(RegClassIndex),
+    RegClass(RegClassIndex),
+
+    /// This value prefers a specific register unit
+    RegUnit(RegUnit),
 }
 
 impl Default for Affinity {
@@ -43,7 +48,7 @@ impl Affinity {
         if constraint.kind == ConstraintKind::Stack {
             Self::Stack
         } else {
-            Self::Reg(constraint.regclass.into())
+            Self::RegClass(constraint.regclass.into())
         }
     }
 
@@ -51,7 +56,7 @@ impl Affinity {
     pub fn abi(arg: &AbiParam, isa: &dyn TargetIsa) -> Self {
         match arg.location {
             ArgumentLoc::Unassigned => Self::Unassigned,
-            ArgumentLoc::Reg(_) => Self::Reg(isa.regclass_for_abi_type(arg.value_type).into()),
+            ArgumentLoc::Reg(_) => Self::RegClass(isa.regclass_for_abi_type(arg.value_type).into()),
             ArgumentLoc::Stack(_) => Self::Stack,
         }
     }
@@ -67,7 +72,7 @@ impl Affinity {
     /// Is this the `Reg` affinity?
     pub fn is_reg(self) -> bool {
         match self {
-            Self::Reg(_) => true,
+            Self::RegUnit(_) | Self::RegClass(_) => true,
             _ => false,
         }
     }
@@ -80,6 +85,22 @@ impl Affinity {
         }
     }
 
+    pub fn rc_in(&self, reginfo: &RegInfo) -> Option<RegClass> {
+        match self {
+            Affinity::RegUnit(reg) => Some(reginfo.toprc_containing_regunit(*reg)),
+            Affinity::RegClass(rci) => Some(reginfo.rc(*rci)),
+            _ => None,
+        }
+    }
+
+    pub fn toprc_in(&self, reginfo: &RegInfo) -> Option<RegClass> {
+        match self {
+            Affinity::RegUnit(reg) => Some(reginfo.toprc_containing_regunit(*reg)),
+            Affinity::RegClass(rci) => Some(reginfo.toprc(*rci)),
+            _ => None,
+        }
+    }
+
     /// Merge an operand constraint into this affinity.
     ///
     /// Note that this does not guarantee that the register allocator will pick a register that
@@ -87,18 +108,63 @@ impl Affinity {
     pub fn merge(&mut self, constraint: &OperandConstraint, reginfo: &RegInfo) {
         match *self {
             Self::Unassigned => *self = Self::new(constraint),
-            Self::Reg(rc) => {
+            Self::RegClass(rc) => {
                 // If the preferred register class is a subclass of the constraint, there's no need
                 // to change anything.
                 if constraint.kind != ConstraintKind::Stack && !constraint.regclass.has_subclass(rc)
                 {
                     // If the register classes overlap, try to shrink our preferred register class.
                     if let Some(subclass) = constraint.regclass.intersect_index(reginfo.rc(rc)) {
-                        *self = Self::Reg(subclass);
+                        *self = Self::RegClass(subclass);
                     }
                 }
             }
+            // Either the constraint is a stack constraint, and we would just keep this affinity,
+            // or it's a register constraint. If it's a register constraint, it's either the same,
+            // and we wouldn't change the affinity, or it's a different constraint, which has no
+            // intersection with this register, and wouldn't change the affinity.
+            //
+            // In all cases, we wouldn't change the affinity.
+            Self::RegUnit(_) => {}
             Self::Stack => {}
+        }
+    }
+
+    /// Compute the maximal intersection between this affinity and some other affinity.
+    ///
+    /// In cases where the two affinities conflict, this affinity wins out.
+    pub fn intersect(&mut self, other: &Affinity, reginfo: &RegInfo) {
+        match *self {
+            Affinity::Unassigned => {
+                // Unassigned is the least picky affinity, we can accept anything more precise.
+                *self = other.clone();
+            }
+            Affinity::RegClass(curr_rci) => {
+                // Register class affinities can be narrowed, if a narrower class exists.
+                match other {
+                    Affinity::RegUnit(new_unit) => {
+                        // if the other affinity is for a compatible register, we can use it
+                        if reginfo.rc(curr_rci).contains(*new_unit) {
+                            *self = Affinity::RegUnit(*new_unit);
+                        }
+                    }
+                    Affinity::RegClass(new_rci) => {
+                        // or if the affinity is for a register class more precise than the
+                        // current, we can also use it
+                        if let Some(subclass) =
+                            reginfo.rc(curr_rci).intersect_index(reginfo.rc(*new_rci))
+                        {
+                            *self = Affinity::RegClass(subclass);
+                        }
+                    }
+                    // other cases are either wider or equivalent
+                    _ => {}
+                }
+            }
+            // Either this is already a RegUnit affinity, and either the same or not useful, or
+            // this is a Stack affinity, and the new affinity is conflicting or a no-more-precise
+            // Stack affinity
+            _ => {}
         }
     }
 
@@ -117,9 +183,13 @@ impl<'a> fmt::Display for DisplayAffinity<'a> {
         match self.0 {
             Affinity::Unassigned => write!(f, "unassigned"),
             Affinity::Stack => write!(f, "stack"),
-            Affinity::Reg(rci) => match self.1 {
+            Affinity::RegClass(rci) => match self.1 {
                 Some(regs) => write!(f, "{}", regs.rc(rci)),
                 None => write!(f, "{}", rci),
+            },
+            Affinity::RegUnit(unit) => match self.1 {
+                Some(regs) => write!(f, "{}", regs.display_regunit(unit)),
+                None => write!(f, "{}", unit),
             },
         }
     }
--- a/cranelift-codegen/src/regalloc/affinity.rs
+++ b/cranelift-codegen/src/regalloc/affinity.rs
@@ -9,9 +9,7 @@
 //! larger register class instead.
 
 use crate::ir::{AbiParam, ArgumentLoc};
-use crate::isa::{
-    ConstraintKind, OperandConstraint, RegClass, RegClassIndex, RegInfo, RegUnit, TargetIsa,
-};
+use crate::isa::{ConstraintKind, OperandConstraint, RegClass, RegClassIndex, RegInfo, RegUnit};
 use core::fmt;
 
 /// Preferred register allocation for an SSA value.
@@ -45,18 +43,20 @@ impl Affinity {
     /// This will never create an `Affinity::Unassigned`.
     /// Use the `Default` implementation for that.
     pub fn new(constraint: &OperandConstraint) -> Self {
-        if constraint.kind == ConstraintKind::Stack {
-            Self::Stack
-        } else {
-            Self::RegClass(constraint.regclass.into())
+        match constraint.kind {
+            ConstraintKind::Stack => Affinity::Stack,
+            ConstraintKind::FixedReg(unit) | ConstraintKind::FixedTied(unit) => {
+                Affinity::RegUnit(unit)
+            }
+            _ => Affinity::RegClass(constraint.regclass.into()),
         }
     }
 
     /// Create an affinity that matches an ABI argument for `isa`.
-    pub fn abi(arg: &AbiParam, isa: &dyn TargetIsa) -> Self {
+    pub fn abi(arg: &AbiParam) -> Self {
         match arg.location {
             ArgumentLoc::Unassigned => Self::Unassigned,
-            ArgumentLoc::Reg(_) => Self::RegClass(isa.regclass_for_abi_type(arg.value_type).into()),
+            ArgumentLoc::Reg(unit) => Self::RegUnit(unit),
             ArgumentLoc::Stack(_) => Self::Stack,
         }
     }
@@ -109,14 +109,28 @@ impl Affinity {
         match *self {
             Self::Unassigned => *self = Self::new(constraint),
             Self::RegClass(rc) => {
-                // If the preferred register class is a subclass of the constraint, there's no need
-                // to change anything.
-                if constraint.kind != ConstraintKind::Stack && !constraint.regclass.has_subclass(rc)
-                {
-                    // If the register classes overlap, try to shrink our preferred register class.
-                    if let Some(subclass) = constraint.regclass.intersect_index(reginfo.rc(rc)) {
-                        *self = Self::RegClass(subclass);
+                match constraint.kind {
+                    ConstraintKind::RegClass => {
+                        // If the preferred register class is a subclass of the constraint, there's no need
+                        // to change anything.
+                        if !constraint.regclass.has_subclass(rc) {
+                            // If the register classes don't overlap, `intersect` returns `Unassigned`, and
+                            // we just keep our previous affinity.
+                            if let Some(subclass) =
+                                constraint.regclass.intersect_index(reginfo.rc(rc))
+                            {
+                                // This constraint shrinks our preferred register class.
+                                *self = Self::RegClass(subclass);
+                            }
+                        }
                     }
+                    ConstraintKind::FixedReg(reg) | ConstraintKind::FixedTied(reg) => {
+                        // If the constraint register indicates a more precise affinity, update it
+                        if reginfo.rc(rc).contains(reg) {
+                            *self = Affinity::RegUnit(reg);
+                        }
+                    }
+                    _ => {}
                 }
             }
             // Either the constraint is a stack constraint, and we would just keep this affinity,
--- a/cranelift-codegen/src/regalloc/coloring.rs
+++ b/cranelift-codegen/src/regalloc/coloring.rs
@@ -304,8 +304,8 @@ impl<'a> Context<'a> {
                     .get(lv.value, &self.cur.func.locations)
                     .display(&self.reginfo)
             );
-            if let Affinity::Reg(rci) = lv.affinity {
-                let rc = self.reginfo.rc(rci);
+
+            if let Some(rc) = lv.affinity.rc_in(&self.reginfo) {
                 let loc = self.cur.func.locations[lv.value];
                 let reg = match loc {
                     ValueLoc::Reg(reg) => reg,
@@ -349,29 +349,45 @@ impl<'a> Context<'a> {
         let mut regs = AvailableRegs::new(&self.usable_regs);
 
         for (lv, abi) in args.iter().zip(&sig.params) {
-            match lv.affinity {
-                Affinity::Reg(rci) => {
-                    let rc = self.reginfo.rc(rci);
+            // the RegClass for the affinity of the value we need to spill
+            let rc = match lv.affinity {
+                Affinity::RegUnit(unit) => {
+                    // if unit != reg, we've introduced inconsistent data in that the value's
+                    // Affinity disagrees with where the value actually exists. At this point,
+                    // those two should be consistent.
                     if let ArgumentLoc::Reg(reg) = abi.location {
-                        if !lv.is_dead {
-                            regs.take(rc, reg, lv.is_local);
-                        }
-                        self.cur.func.locations[lv.value] = ValueLoc::Reg(reg);
-                    } else {
-                        // This should have been fixed by the reload pass.
-                        panic!(
-                            "Entry arg {} has {} affinity, but ABI {}",
-                            lv.value,
-                            lv.affinity.display(&self.reginfo),
-                            abi.display(&self.reginfo)
-                        );
+                        debug_assert!(unit == reg);
                     }
+                    Some(self.reginfo.toprc_containing_regunit(unit))
                 }
+                Affinity::RegClass(rci) => Some(self.reginfo.rc(rci)),
                 // The spiller will have assigned an incoming stack slot already.
-                Affinity::Stack => debug_assert!(abi.location.is_stack()),
+                Affinity::Stack => {
+                    debug_assert!(abi.location.is_stack());
+                    None
+                }
                 // This is a ghost value, unused in the function. Don't assign it to a location
                 // either.
-                Affinity::Unassigned => {}
+                Affinity::Unassigned => None,
+            };
+
+            if let Some(rc) = rc {
+                // alright, it's not on the stack or a ghost value, so we need to actually spill
+
+                if let ArgumentLoc::Reg(reg) = abi.location {
+                    if !lv.is_dead {
+                        regs.take(rc, reg, lv.is_local);
+                    }
+                    self.cur.func.locations[lv.value] = ValueLoc::Reg(reg);
+                } else {
+                    // This should have been fixed by the reload pass.
+                    panic!(
+                        "Entry arg {} has {} affinity, but ABI {}",
+                        lv.value,
+                        lv.affinity.display(&self.reginfo),
+                        abi.display(&self.reginfo)
+                    );
+                }
             }
         }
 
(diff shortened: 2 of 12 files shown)


cranelift-codegen/src/regalloc/coloring.rs
```diff
@@ -1081,6 +1266,11 @@ impl<'a> Context<'a> {
                     regs.global
                         .free(rc, self.cur.func.locations[lv.value].unwrap_reg());
                 }
+            } else {
+                // This is sufficient as-is to pass tests - didn't understand what causes
+                // `process_ghost_kills` and it seems tests might be missing if `RegUnit` being
+                // omitted isn't causing issues here
+                panic!("bring this up in code review");
```
iximeow: I think for completeness this should mimic what's done for `Affinity::Reg`, but I'm not sure how to produce a ghost instruction that would trigger this code path, nor if that not being covered by tests is noteworthy

cranelift-codegen/src/regalloc/coloring.rs
```diff
@@ -226,7 +226,22 @@ impl<'a> Context<'a> {
                 lv.affinity.display(&self.reginfo),
                 self.cur.func.locations[lv.value].display(&self.reginfo)
             );
-            if let Affinity::Reg(rci) = lv.affinity {
+            if let Affinity::RegUnit(reg) = lv.affinity {
+                let bank = self.reginfo.bank_containing_regunit(reg).unwrap();
```
iximeow: This is just the body of `spilling::toprc_containing_regunit` - it seems like that function could/should be moved over to `RegInfo`, but also not sure if I'm using it terribly wrong :)

cranelift-codegen/src/regalloc/coloring.rs
```diff
@@ -256,6 +271,46 @@ impl<'a> Context<'a> {
...
+                                let rc = self.reginfo.classes
+                                    [bank.first_toprc..(bank.first_toprc + bank.num_toprcs)]
+                                    .iter()
+                                    .find(|&rc| rc.contains(unit))
+                                    .expect("reg unit should be in a toprc");
+                                regs.take(rc, reg, lv.is_local);
+                            }
+                            self.cur.func.locations[lv.value] = ValueLoc::Reg(reg);
+                        } else {
+                            // TODO: what does this mean? A value is assigned somewhere by the ABI
```
iximeow: So this is a really weird one to me. If `unit != reg`, my understanding is that the ABI-assigned location disagrees with where we think the value should be (so if a value comes in as a function argument, passed to an internal call in some other location?). In that case, I initially deferred to the declared Affinity, but that resulted in invalid register constraints. As-is, this uses whatever the ABI-assigned location is, and that seems to do things right, but that's still exactly opposite of what I expected.

cranelift-codegen/src/regalloc/liveness.rs
```diff
@@ -447,6 +447,19 @@ impl Liveness {
                     if let Some(constraint) = operand_constraints.next() {
                         lr.affinity.merge(constraint, &reginfo);
                     }
+
+                    let call_sig = func.dfg.call_signature(inst);
+                    if let Some(sig) = call_sig {
+                        if let crate::ir::InstructionData::CallIndirect { .. } = func.dfg[inst] {
```
iximeow: my great shame of this PR: the `CallIndirect` instruction takes one more argument than the number of arguments for the function to be called, which throws the affinity rewriting here all out of whack. Is there some nice way I couldn't find to generically talk about instruction arguments that line up with the call signature? I suppose worst case would be collecting signature params and zipping with instruction arguments in reverse, walking together, and assuming non-call-argument arguments will be at the start of the list.

cranelift-codegen/src/regalloc/spilling.rs
```diff
@@ -395,6 +433,11 @@ impl<'a> Context<'a> {
             if abi.location.is_reg() {
                 let (rci, spilled) = match self.liveness[arg].affinity {
                     Affinity::Reg(rci) => (rci, false),
+                    Affinity::RegUnit(_unit) => (
```
iximeow: RegUse tracks at register class granularity, so I'm following that and assuming that the value's affinity will not conflict with where the ABI says its value type should be. That seems like a safe assumption, but I wanted to call it out here in case it's not

cranelift-codegen/src/regalloc/spilling.rs
```diff
@@ -498,6 +541,13 @@ impl<'a> Context<'a> {
...
                 // Viable candidates are registers in one of the `mask` classes, and not already in
                 // the spill set.
+                if let Affinity::RegUnit(unit) = lv.affinity {
+                    let rc = toprc_containing_regunit(unit, &self.reginfo);
+                    if (mask & (1 << rc.toprc)) != 0 && !self.spills.contains(&lv.value) {
+                        // Here, `lv` is a viable spill candidate.
+                        return Some(lv.value);
+                    }
+                }
                 if let Affinity::Reg(rci) = lv.affinity {
```
iximeow: .. this should be an `else if let` but doesn't _break_ anything here. I'll fix that anyway.

iximeow:
Last high level comment, none of the content of the tests had to change for this (lightly surprising) - it seems like I should be able to write a test that a call does not generate the aforementioned unnecessary movs, so I'll try to do that for a test that can live in filetests. I've been working with a wasm test case, so as it is maybe not the best way to write a cranelift test.

cranelift-codegen/src/regalloc/liveness.rs
iximeow: Figured out I should have been using `inst_variable_args` here and this is much nicer now :)

cranelift-codegen/src/regalloc/coloring.rs
iximeow: went ahead and did this because looking at this again it feels like an obvious change

cranelift-codegen/src/regalloc/spilling.rs
sunfishcode: It's not immediately obvious that it's safe. See `Affinity::merge` for example where it makes no change if there are conflicting register class constraints.

cranelift-codegen/src/regalloc/coloring.rs
sunfishcode: This is a good question, and I don't have an answer right now. Does this happen as a result of the code not being able to move ABI arguements into specified `RegUnit` affinities?

cranelift-codegen/src/regalloc/coloring.rs
```diff
@@ -226,7 +226,17 @@ impl<'a> Context<'a> {
...
+            if let Affinity::RegUnit(reg) = lv.affinity {
+                let rc = self.reginfo.toprc_containing_regunit(reg);
+                let loc = self.cur.func.locations[lv.value];
+                match loc {
+                    ValueLoc::Reg(reg) => regs.take(rc, reg, lv.is_local),
+                    ValueLoc::Unassigned => panic!("Live-in {} wasn't assigned", lv.value),
+                    ValueLoc::Stack(ss) => {
+                        panic!("Live-in {} is in {}, should be register", lv.value, ss)
+                    }
+                }
```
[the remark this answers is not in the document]
iximeow: I think I'd just pushed https://github.com/CraneStation/cranelift/pull/782/commits/ebdf736d9b602d311c6998502334e01d1f9c042d when you made this comment - done!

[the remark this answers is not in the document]
iximeow: Unless you mean getting the register class corresponding to an Affinity, which is a bit tricker. In some places it's correct to use the top `RegClass`, where in others `RegClassIndex` is used to get just the matching `RegClass`. I realized in the above change that it might actually be desirable to get the most precise matching class for a `RegUnit`, but either way a helper for `Affinity -> Option<RegClass>` seems good

cranelift-codegen/src/regalloc/spilling.rs
iximeow: Wouldn't a conflict between the constraints here be a logical error? A value's affinity shouldn't place it in a register class that can't hold those values. (I realize since then that the `Reg` case is similar -- uses the declared register affinity regardless of `abi.value_type`. The equivalent here would be `self.reginfo.toprc_for_reguint(unit)`)

cranelift-codegen/src/regalloc/coloring.rs
iximeow: That's my understanding, though the cause is interesting. After making [`parameter affinity more granular`](https://github.com/CraneStation/cranelift/pull/782/commits/4511b31a230998ec05002970128c09b86cf7ac8d) I brought back this [`debug_assert!`](https://github.com/CraneStation/cranelift/pull/782/commits/59603adc5d401f078c22ba36af5649d09dabfc92#diff-ac2d83406d2fa759e3c3ffa71ae7d551R264). It now passes all tests, including the `spill.clif` filetest, which was the failure before.

My understanding is that a value going into a call in `spill` was in `r10` that ends up on the stack by ABI, while I was overwriting another value's affinity to also be `r10`. I think two values having the same affinity is what resulted in the invalid constraints.

cranelift-codegen/src/regalloc/affinity.rs
```diff
@@ -100,6 +105,13 @@ impl Affinity {
                     }
```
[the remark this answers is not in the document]
iximeow: Yes! Thank you for pointing that out. See: [`d783a64`](https://github.com/CraneStation/cranelift/pull/782/commits/d783a64f7b73c552fc5aed9b7d7fab7611c37b66).

filetests/regalloc/constraints.clif
```diff
@@ -61,12 +60,10 @@ ebb0:
 ; Tied use of a diverted register.
 function %fixed_op_twice() -> i32 {
 ebb0:
-    ; check: ,%rax]
```
iximeow: I'm pretty sure this change violates the intent of this test: that `v0` is in some register and must be diverted for multiple uses. Passing along the `FixedReg(rcx)` affinity ends up placing `v0` directly in rcx, so this no longer tests diversions as it originally did.

I think it would preserve the spirit of this test to have some value, `vN` used as a shift parameter, so it's placed in `rcx`, and subsequently used somewhere as _not_ the shift parameter twice. If that all sounds right, I'll adjust the test.

iximeow:
There's a bug here somewhere that I've seen through fuzzing, but I'm currently struggling to get creduce to do its thing and make the test case a manageable amount of code. Will update when I know more. The observed failure is that `rsi` is double freed (causing a panic) :(

bjorn3:
@iximeow can you upload the test case anyway?

> but I'm currently struggling to get creduce to do its thing

Have you tried #778 on the clif file?

regehr:
@iximeow I would also like to see the full repro info (test case + interestingness test script + cranelift version + platform) for this one. if C-Reduce is working badly here, perhaps there are some easy things we can do to make it work better.

iximeow:
Various updates:
@bjorn3: bugpoint is a really good idea. I'll try that in a bit.
@regehr: the `creduce` problems were entirely user error - I was initially missing `creduce` entirely (figured out Friday), accidentally added a bad path in our harness shortly afterward, and then an unhelpful interestingness script (the way we typically use `csmith`/`creduce` is to fuzz what we assume is a working toolchain, so Cranelift panicking does no favors).

For posterity, the test case is from csmith with seed `9512311676052123889`, and the preprocessed resulting file: [gen_preprocessed.tar.gz](https://github.com/CraneStation/cranelift/files/3299298/gen_preprocessed.tar.gz). Cranelift version is `395737b` (latest commit on this branch) but being built through clang+lucetc (master with that Cranelift commit as the submodule version). which leads me to the interestingness script,
```
#! /bin/bash
/opt/wasi-sdk/bin/clang --target=wasm32-unknown-wasi gen_preprocessed.c -o /home/awortman/lucet/lucet-wasi-fuzz/out2.wasm -Wl,--no-threads
if [[ $? -ne 0 ]]; then exit 1; fi

/home/awortman/lucet/target/debug/lucetc /home/awortman/lucet/lucet-wasi-fuzz/out2.wasm --bindings /home/awortman/lucet/lucet-wasi/bindings.json
if [[ $? -eq 0 ]]; then exit 1; else exit 0; fi
```

iximeow:
for the spectators at home, creduce has correctly concluded that
```
a;
struct {
  char a;
  char g
} fn1() {
}
main() {
  char b = 1;
  char c[1] = {};
  int d = 5, e = 4, f = 2;
  fn1();
}
```
is a minimal reproduction of the register double-free behavior. hum.

iximeow:
I've turned the above bug-displaying code into a filetest, and shrunk it down a bit more (with some help from bugpoint, @bjorn3!)

```
test regalloc
target x86_64

function u0:11(i64 vmctx) -> i32 system_v {
    gv0 = vmctx
    gv1 = load.i64 notrap aligned gv0-8
    heap0 = static gv0, min 0x0002_0000, bound 0x0040_0000, offset_guard 0x0040_0000, index_type i32
    sig0 = (i64 vmctx, i32) system_v
    fn0 = colocated u0:10 sig0

    ebb0(v0: i64):
        v6 = iconst.i32 0
        v10 = iadd v6, v6
        v11 = iconst.i32 0
        v12 = iconst.i32 0
        v15 = iconst.i32 0
        v18 = iconst.i32 0
        v19 = heap_addr.i64 heap0, v6, 1
        istore8 v18, v6+31
        v23 = iconst.i64 0
        store v15, v23+20
        store v12, v23+16
        store v11, v23+16
        call fn0(v0, v10)
        return v6
}
```

it _looks_ like the bug is an issue where setting up ebb parameters at a branch when under some amount of register pressure, but I don't really understand how `add_live_var` actually adds a variable to the solver, which is where this goes sideways.

iximeow:
For anyone reading up from PR comments, I've fixed the above issue, which was due to [`me neglecting to check diversions`](https://github.com/CraneStation/cranelift/pull/782/commits/bfa2aec129ce3d46b18dd59ecaa8c36a099cbd89#diff-ac2d83406d2fa759e3c3ffa71ae7d551R909) in `try_add_var`. This didn't come up in existing tests, but the easiest way I can find to trigger the resulting bug was a situation like the above reduction:
* ~8 live variables over the end of an EBB
* one variable constrained to a register X
* _the same_ variable constrained to a register Y later

By not checking diversions, I added a value to the solver in a register it had been diverted out from. That pretty quickly gets the solver into a state that's totally invalid, resulting in the panic.

Fixed now! And I added the file I was working with as a testcase.

iximeow:
bump - at this point I'm just waiting for review. @sunfishcode do you think you'll be able to take another look?

bjorn3:
@iximeow I believe the policy is to rebase instead of merge master.

filetests/regalloc/constraints.clif
iximeow: Realized while cleaning up the branch today it's much easier to preserve the intent of this test by forcing `v0` into `rax` by having it be the return value of a function. So it's that now, instead of expecting `v0` to be in the first available register.

iximeow:
I've rebased and fixed conflicts here - took the opportunity to reshape this branch into two commits while I was at it. The first introduces `RegUnit` and some corresponding refactoring, but doesn't change actual affinities anywhere. The second commit actually wires it up in a few places and has matching test changes.

cranelift-codegen/src/regalloc/liveness.rs
```diff
@@ -448,6 +448,43 @@ impl Liveness {
...
+                                if let Some(subclass) =
+                                    reginfo.rc(lr_rci).intersect_index(reginfo.rc(abi_rci))
+                                {
+                                    Affinity::RegClass(subclass)
+                                } else {
+                                    Affinity::RegClass(lr_rci)
+                                }
+                            }
+                            (lr_affinity, _) => lr_affinity,
+                        };
```
sunfishcode: This `match` expression looks like a good candidate to be refactored out into a helper function.

cranelift-codegen/src/regalloc/liveness.rs
```diff
@@ -448,6 +448,43 @@ impl Liveness {
...
+                if let Some(sig) = func.dfg.call_signature(inst) {
+                    let sig = &func.dfg.signatures[sig];
+                    let args = func.dfg.inst_variable_args(inst);
+                    if sig.params.len() != args.len() {
+                        panic!(
+                            "Assumption violated: args count == sig.params count ({} != {})",
+                            args.len(),
+                            sig.params.len()
+                        );
+                    }
```
sunfishcode: This `if` with `panic!` could be replaced by `assert_eq!`, which has the nice property of handling the `({} != {})` part automatically.

cranelift-codegen/src/regalloc/reload.rs
```diff
@@ -402,11 +404,30 @@ impl<'a> Context<'a> {
...
+                            value: arg,
+                            reg: (op.regclass, Some(reg)),
+                        })
+                    } else {
+                        self.candidates.push(ReloadCandidate {
+                            argidx,
+                            value: arg,
+                            reg: (op.regclass, None),
+                        })
+                    }
```
sunfishcode: The only thing that differs between the then and else branches of this if is the second element of `reg`'s initializer. Would it make sense to factor out?
```rust
let unit = if let ConstraintKind::FixedReg(reg) = op.kind {
    Some(reg)
} else {
    None
};
self.candidates.push(ReloadCandidate {
    argidx,
    value: arg,
    reg: (op.regclass, unit),
};

File: cranelift-codegen/src/regalloc/spilling.rs (hunk 5 above)
Comment (sunfishcode) [reply cranelift-codegen/src/regalloc/spilling.rs]: The ABI might request the value in one sub-register class, and the affinity might be for a different sub-register class. They'd both be part of the same parent register class, which is valid for the type, but they wouldn't intersect. Would that cause a problem here?

Comment (iximeow) [reply cranelift-codegen/src/regalloc/spilling.rs]: I see, that's a good question. Here's my understanding, which leads me to think this can be resolved without issue:

`collect_abi_reg_uses` collects uses of registers for argument values as they exist when reaching the call, so the register class reported here should be derived from the existing affinity (and my arm for `RegUnit` is incorrect). This explains why `RegClass` is correct in returning `rci`, even if that disagrees with the ABI-requested location. `process_reg_uses` then inserts copies, but I don't entirely understand that logic. I _think_ that by widening back to a register class that contains `unit` in this arm, the worst that could happen is that we decide to copy `Value`, ending its live range with an `Affinity::RegClass(rci)`, which might be wider than the specific affinity the use came from. That would force some extra copies, but I don't think would be incorrect. The value's copy wouldn't have any set location, and would later be colored into the ABI-correct register. In this case, we can come back later and make `RegUse` smarter by making it aware of specific register affinities, and avoiding the `insert_copy` if it would copy to where it needs to go anyway?

I think this all only applies if we need to reload for call arguments, and I'm not really sure how I'd construct a test case for this. Either way, I'll change this to return a register class derived from `unit`, and see if that still works - I'd hope that trying to end the value's live range with an affinity that causes a conflict would at least cause a panic in codegen, so not hitting that even with some fuzzing is a bit perplexing to me.. It could be that for x86 the bug case isn't reachable here; `regclass_for_abi_type` would give GPR, the same as `toprc_for(unit)`, and wouldn't be a conflict.

If the above reads correct to you, I'd like to expand on `collect_abi_reg_uses`'s docs a bit to clarify how it cares about the values' affinities - namely that it's interested in where they currently are, not where the function will eventually want them, and that the function's ABI at this point is not important here.

File: cranelift-codegen/src/regalloc/liveness.rs (hunk 10 above)
Comment (iximeow) [reply cranelift-codegen/src/regalloc/liveness.rs]: :+1: this is now `Affinity::intersect`

File: cranelift-codegen/src/regalloc/reload.rs (hunk 12 above)
Comment (iximeow) [reply cranelift-codegen/src/regalloc/reload.rs]: yes! I did the same with the `call_signature` if above this region as well.

**Comment by iximeow** [comment]
Is there anything I can do to nudge this along? I think now that regalloc is more apt to use caller-save registers these changes are more likely to remove spills and make for faster generated code, in addition to just using fewer `mov`.

**Comment by bnjbvr** [comment]
I'll try to give it a look, since it has some interesting performance implications. Also I'll test it on a broader set of tests (including Spidermonkey testing suite and a few big wasm benchmarks), to see if any issues show up.

**Comment by bnjbvr** [comment]
Fwiw: this passes all of my testing (great job! regalloc is a tricky area in general), and for benchmarks tracked in Embenchen, it doesn't seem to regress runtime perf (outside noise), and even provides speedups of around 2% (and up to 23% for one particular workload). Unfortunately I won't have time to review it this week; will look at it next week unless somebody beat me to it.

**Comment by bjorn3** [comment]
Tried it on cg_clif. Unfortunately it already crashed at the third cg_clif test:

```
thread 'rustc' panicked at 'index out of bounds: the len is 0 but the index is 0', /rustc/aa4e57ca8f18b836bf77923cd0d9ad1390f0110b/src/libcore/slice/mod.rs:2796:10
stack backtrace:
   0: backtrace::backtrace::libunwind::trace
             at /cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.40/src/backtrace/libunwind.rs:88
   1: backtrace::backtrace::trace_unsynchronized
             at /cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.40/src/backtrace/mod.rs:66
   2: std::sys_common::backtrace::_print_fmt
             at src/libstd/sys_common/backtrace.rs:77
   3: <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt
             at src/libstd/sys_common/backtrace.rs:61
   4: core::fmt::write
             at src/libcore/fmt/mod.rs:1028
   5: std::io::Write::write_fmt
             at src/libstd/io/mod.rs:1412
   6: std::sys_common::backtrace::_print
             at src/libstd/sys_common/backtrace.rs:65
   7: std::sys_common::backtrace::print
             at src/libstd/sys_common/backtrace.rs:50
   8: std::panicking::default_hook::{{closure}}
             at src/libstd/panicking.rs:188
   9: std::panicking::default_hook
             at src/libstd/panicking.rs:205
  10: rustc_driver::report_ice
  11: std::panicking::rust_panic_with_hook
             at src/libstd/panicking.rs:468
  12: std::panicking::continue_panic_fmt
             at src/libstd/panicking.rs:373
  13: rust_begin_unwind
             at src/libstd/panicking.rs:302
  14: core::panicking::panic_fmt
             at src/libcore/panicking.rs:139
  15: core::panicking::panic_bounds_check
             at src/libcore/panicking.rs:96
  16: <usize as core::slice::SliceIndex<[T]>>::index
             at /rustc/aa4e57ca8f18b836bf77923cd0d9ad1390f0110b/src/libcore/slice/mod.rs:2796
  17: core::slice::<impl core::ops::index::Index<I> for [T]>::index
             at /rustc/aa4e57ca8f18b836bf77923cd0d9ad1390f0110b/src/libcore/slice/mod.rs:2647
  18: <alloc::vec::Vec<T> as core::ops::index::Index<I>>::index
             at /rustc/aa4e57ca8f18b836bf77923cd0d9ad1390f0110b/src/liballoc/vec.rs:1861
  19: cranelift_codegen::regalloc::reload::Context::find_candidates
             at /home/bjorn/Documenten/cranelift/cranelift-codegen/src/regalloc/reload.rs:410
  20: cranelift_codegen::regalloc::reload::Context::visit_inst
             at /home/bjorn/Documenten/cranelift/cranelift-codegen/src/regalloc/reload.rs:212
  21: cranelift_codegen::regalloc::reload::Context::visit_ebb
             at /home/bjorn/Documenten/cranelift/cranelift-codegen/src/regalloc/reload.rs:133
  22: cranelift_codegen::regalloc::reload::Context::run
             at /home/bjorn/Documenten/cranelift/cranelift-codegen/src/regalloc/reload.rs:118
  23: cranelift_codegen::regalloc::reload::Reload::run
             at /home/bjorn/Documenten/cranelift/cranelift-codegen/src/regalloc/reload.rs:86
  24: cranelift_codegen::regalloc::context::Context::run
             at /home/bjorn/Documenten/cranelift/cranelift-codegen/src/regalloc/context.rs:173
  25: cranelift_codegen::context::Context::regalloc
             at /home/bjorn/Documenten/cranelift/cranelift-codegen/src/context.rs:328
  26: cranelift_codegen::context::Context::compile
             at /home/bjorn/Documenten/cranelift/cranelift-codegen/src/context.rs:157
  27: cranelift_module::module::Module<B>::define_function
             at /home/bjorn/Documenten/cranelift/cranelift-module/src/module.rs:552
```

&lt;placeholder for bugpoint reduced testcase&gt;

**Comment by iximeow** [comment]
I'm waiting for the above bugpoint-reduced testcase to look further but I'm pretty sure I misused `argidx` in a way that's only apparent for indirect calls when the register being called has spilled. `argidx` [here](https://github.com/CraneStation/cranelift/pull/782/files#diff-f45acc8d176cbd7814747ea8724fcc19R410) should check fixed arguments before looking through arguments from the called function's signature.

**Comment by iximeow** [comment]
@bjorn3 fixed the panic you saw in https://github.com/CraneStation/cranelift/pull/782/commits/4d2b6837c0db8e53dcce038cecdd0d649511bbde. The solution, pleasantly, was to delete some code :) Calls have `ReloadCandidate` added for their arguments later, which I'd adjusted for the new register affinity, so checking arguments at all where the panic occurred was an error.

**Comment by bjorn3** [comment]
Maybe %rbx was used in that function, but it isnt after the changes. It is a callee saved register, so I think the abi code adds it as arg when it is used within the body.

**Comment by iximeow** [comment]
Looks like you're right. It's also a fail in a test that doesn't exist at the tip of this branch - CI seems to run with master merged in now, which is cool if a bit surprising.

**Comment by bjorn3** [comment]
I believe Travis CI already merged a PR into master before testing it.

**Comment by iximeow** [comment]
I've rebased this branch on `master` and fixed the conflict with `multi-val-b1.clif`. No more failing CI, yay

**Comment by bjorn3** [comment]
Currently testing this with cg_clif.

**Comment by bjorn3** [comment]
```
Benchmark #1: ./raytracer_cg_clif_affinity
  Time (mean ± σ):     12.905 s ±  0.174 s    [User: 12.891 s, System: 0.012 s]
  Range (min … max):   12.825 s … 13.635 s    20 runs
 
  Warning: Statistical outliers were detected. Consider re-running this benchmark on a quiet PC without any interferences from other programs. It might help to use the '--warmup' or '--prepare' options.
 
Benchmark #2: ./raytracer_cg_clif_no_affinity
  Time (mean ± σ):     13.029 s ±  0.027 s    [User: 13.015 s, System: 0.011 s]
  Range (min … max):   12.971 s … 13.081 s    20 runs
```

The performance difference before and after this PR is very tiny for the `simple-raytracer` benchmark.

All tests for cg_clif pass.

bnjbvr:
Hi @iximeow. We've discussed this in an internal meeting, and our current thinking is that in light of the plan to replace the coloring register allocator by something simpler (see also #1246), we'd like to reduce investing time on the coloring regalloc, and minimize disruption as much as possible (that is, only maintain it, and stop enhancing it). It's been working for a while now, and fuzzing didn't bring up new issues. Taking this PR is a bit risky at this point. Sorry that this decision came so late and that the review didn't happen earlier. Thanks for working on it, though :heart:

iximeow:
Ah that's a bummer. How would you (I'm not sure if I should address this to you or post-hoc to folks in the internal meeting? :grin:) feel about accepting the first of the two commits here - adding the register-specific affinity, just not actually modifying the coloring allocator to use it? I fairly strongly believe that any register allocator we use ought to be able to respect register-specific affinities, and that commit adds the ability to represent them.

bnjbvr:
Good question! I'll take this to our next meeting (Monday), but my initial feeling is that it'd make sense to have it.

bnjbvr:
I've brought it up during our meeting, and we're unsure whether we'll reuse the same data structures or come up with our own, so it's unclear we could easily reuse this, sorry.

Issue #772, opened 2019-05-23
[simplejit] Can't make memory readable+executable on SELinux 


benschau:
Probably not that big of an issue, but for my use case, I have to be able to run on a system that's enforcing SELinux.

# To reproduce
You'll need any system that supports SELinux; I use Fedora 30, which has it enforced by default.
You can check whether SELinux is enabled via `sudo getenforce`. If it isn't enforcing, enable it for this test using `sudo setenforce 1`.

    git clone https://github.com/CraneStation/simplejit-demo (problem can be reproduced with
    any invocation of `Memory::set_readable_and_executable()`)

    cargo run
    
    It will fail with:
    thread 'main' panicked at 'unable to make memory readable+executable: SystemCall(0s 
    { code: 13, kind: PermissionDenied, message: "Permission denied" })', src/libcore/result.rs:997:5

For a more simple test case, you can also try running `simplejit-minimal.rs` and Jonathan Turner's [rustyjit](https://github.com/jonathandturner/rustyjit). They'll both fail with the same kind of issue when SELinux is enforced. 

The problem originates from the region crate ala `cranelift-simplejit/src/memory.rs:123`, which ultimately ends up calling `libc::mprotect()`. That's why `rustyjit` fails as well.

# Workaround
Chris Wellons' [A Jit Compiler Skirmish with SELinux](https://nullprogram.com/blog/2018/11/15/) describes the issue and some of his workarounds pretty well, and I've confirmed that his workarounds do indeed work. In particular, what's happening is that when SELinux is enforced, you can't set execute protections on any page allocated on the heap. 

So, you could change the `rustyjit` program to use the `memmap` crate like in this [gist](https://gist.github.com/benschau/6e0156c8e96fc5502479675cb0bc18d3) I made. Just replace the respective files in `rustyjit` and it'll magically work on SELinux- it uses one of the workarounds that Chris describes using an anonymous memory map to set the required permissions. I'm betting that you could apply something similar to Cranelift, though I don't really know what's desirable.

benschau:
#783

bnjbvr:
I think this was fixed by #783, please reopen if it's not the case.

PR #783, opened 2019-05-29
issue #772: added a memmap replacement to support selinux

benschau:
A fix using [anonymous memory maps](https://crates.io/crates/memmap); locked behind a feature just in case, since it isn't necessary unless you are enforcing SELinux policies.

### DIFF
--- a/cranelift-simplejit/Cargo.toml
+++ b/cranelift-simplejit/Cargo.toml
@@ -17,9 +17,14 @@ region = "2.0.0"
 libc = { version = "0.2.42" }
 errno = "0.2.4"
 target-lexicon = { version = "0.4.0" }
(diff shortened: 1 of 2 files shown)


sunfishcode:
Would it make sense here to just replace `PtrLen` altogether with `MmapMut`? The other fields of `PtrLen` are just the `MmapMut`'s `as_mut_ptr()` and `len()` values. If that works, it would make the code a little simpler, and would also mean we could have one codepath that works for both SELinux and non-SELinux.

benschau:
I'm not entirely sure- I just don't know what situations `MmapMut` wouldn't work in and raw pointers would, so I just put it behind a feature just in case. That being said, I agree it could be made simpler if `MmapMut` just replaces `PtrLen`.

Issue #779, opened 2019-05-28
Only create copy_nop instructions for types for which an encoding exists


bjorn3:
```
thread 'main' panicked at 'assertion failed: ok', cranelift-codegen/src/regalloc/reload.rs:236:21
stack backtrace:
   0: std::sys::unix::backtrace::tracing::imp::unwind_backtrace
             at src/libstd/sys/unix/backtrace/tracing/gcc_s.rs:39
   1: std::sys_common::backtrace::_print
             at src/libstd/sys_common/backtrace.rs:71
   2: std::panicking::default_hook::{{closure}}
             at src/libstd/sys_common/backtrace.rs:59
             at src/libstd/panicking.rs:197
   3: std::panicking::default_hook
             at src/libstd/panicking.rs:211
   4: std::panicking::rust_panic_with_hook
             at src/libstd/panicking.rs:474
   5: std::panicking::begin_panic
             at /rustc/3c235d5600393dfe6c36eeed34042efad8d4f26e/src/libstd/panicking.rs:408
   6: cranelift_codegen::regalloc::reload::Context::visit_inst
             at cranelift-codegen/src/regalloc/reload.rs:236
   7: cranelift_codegen::regalloc::reload::Context::visit_ebb
             at cranelift-codegen/src/regalloc/reload.rs:133
   8: cranelift_codegen::regalloc::reload::Context::run
             at cranelift-codegen/src/regalloc/reload.rs:118
   9: cranelift_codegen::regalloc::reload::Reload::run
             at cranelift-codegen/src/regalloc/reload.rs:86
  10: cranelift_codegen::regalloc::context::Context::run
             at cranelift-codegen/src/regalloc/context.rs:164
  11: cranelift_codegen::context::Context::regalloc
             at cranelift-codegen/src/context.rs:308
  12: cranelift_codegen::context::Context::compile
             at cranelift-codegen/src/context.rs:145
  13: cranelift_codegen::context::Context::compile_and_emit
             at cranelift-codegen/src/context.rs:104
```

Caused by #773

<details>

```
target x86_64

function u0:0(i64, i64, i64) system_v {
    ss0 = explicit_slot 16
    ss1 = explicit_slot 1
    ss2 = explicit_slot 16
    ss3 = explicit_slot 1
    ss4 = explicit_slot 16
    ss5 = explicit_slot 8
    ss6 = explicit_slot 16
    ss7 = explicit_slot 16
    ss8 = explicit_slot 16
[...]

PR #784, opened 2019-05-29
Only create copy_nop instructions for types for which an encoding exi…

julian-seward1:
…sts.  Issue #779.

PR #773 detects, at reload time, `copy` instructions that copy a value from stack
slot back to the same stack slot.  It replaces them with `copy_nop` instructions
that have a null encoding (hence producing no code).

For x86_64, `copy_nop` encodings for the types I64, I32, F64 and F32 are
provided.  Unfortunately the code that detects the redundant copy doesn't
check the type of the copied value, hence leaving itself open to the danger of
creating a `copy_nop` instruction cannot be encoded (which is different from
saying it has a null encoding).

This patch:

* Expands the x86_64 set of `copy_nop` encodings to: I64 I32 I16 I8 F64 and F32

* In `visit_inst()` in `reload.rs`, checks the type of the copied value accordingly.

* Adds comments explaining the above.

### DIFF
--- a/cranelift-codegen/meta-python/isa/x86/encodings.py
+++ b/cranelift-codegen/meta-python/isa/x86/encodings.py
@@ -342,11 +342,23 @@ X86_64.enc(base.copy_special, *r.copysp.rex(0x89, w=1))
 X86_32.enc(base.copy_special, *r.copysp(0x89))
 
 # Stack-slot-to-the-same-stack-slot copy, which is guaranteed to turn
-# into a no-op.
-X86_64.enc(base.copy_nop.i64, r.stacknull, 0)
-X86_64.enc(base.copy_nop.i32, r.stacknull, 0)
-X86_64.enc(base.copy_nop.f64, r.stacknull, 0)
-X86_64.enc(base.copy_nop.f32, r.stacknull, 0)
+# into a no-op.  Ideally we could to make this encoding available for
+# all types, and write `base.copy_nop.any`, but it appears that the
+# controlling type variable must not polymorphic.  So we make do with
+# the following limited set, and guard the generating transformation in
+# regalloc/reload.rs accordingly.
+#
+# The same encoding is generated for both the 64- and 32-bit architectures.
+# Note that we can't use `enc_both` here, because that attempts to create a
+# variant with a REX prefix in the 64-bit-architecture case.  But since
+# there's no actual instruction for the REX prefix to modify the meaning of,
+# it will modify the meaning of whatever instruction happens to follow this
+# one, which is obviously wrong.  Note also that we can and indeed *must*
+# claim that there's a 64-bit encoding for the 32-bit arch case, even though
+# no such single instruction actually exists for the 32-bit arch case.
+for ty in [types.i64, types.i32, types.i16, types.i8, types.f64, types.f32]:
+    X86_64.enc(base.copy_nop.bind(ty), r.stacknull, 0)
+    X86_32.enc(base.copy_nop.bind(ty), r.stacknull, 0)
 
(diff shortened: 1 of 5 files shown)


bjorn3:
It works! Thanks @julian-seward1!

bjorn3:
Could you please add a test like:

```
test compile
target x86_64

function u0:0(i64, i64, i64) system_v {
    sig0 = () system_v
    sig1 = (i16) system_v
    fn1 = u0:94 sig0
    fn2 = u0:95 sig1

ebb0(v0: i64, v1: i64, v2: i64):
    v3 = iconst.i16 0
    jump ebb1(v3)

ebb1(v4: i16):
    call fn1()
    call fn2(v4)
    jump ebb1(v4)
}
```

julian-seward1:
@bjorn3: what does this test check?  What's the rationale?

bjorn3:
It is a reduced reproduction for #779.

cranelift-codegen/meta-python/isa/x86/encodings.py
```diff
@@ -342,9 +342,15 @@ def enc_i32_i64_ld_st(inst, w_bit, recipe, *args, **kwargs):
...
-# into a no-op.
+# into a no-op.  Ideally we could to make this encoding available for
+# all types, and write `base.copy_nop.any`, but it appears that the
+# controlling type variable must not polymorphic.  So we make do with
+# the following limited set, and guard the generating transformation in
+# regalloc/reload.rs accordingly.
 X86_64.enc(base.copy_nop.i64, r.stacknull, 0)
 X86_64.enc(base.copy_nop.i32, r.stacknull, 0)
+X86_64.enc(base.copy_nop.i16, r.stacknull, 0)
+X86_64.enc(base.copy_nop.i8,  r.stacknull, 0)
```
bnjbvr: What about boolean types? Do we need to add an encoding for b1 and b8?

cranelift-codegen/src/regalloc/reload.rs
```diff
@@ -222,15 +222,28 @@ impl<'a> Context<'a> {
...
+                            src_ty.is_int() || src_ty.is_float()
+                        }
+                    }
                     _ => false,
                 };
                 if can_transform {
-                    // Convert the instruction into a `copy_nop`.
+                    // Convert the instruction into a `copy_nop`.  The
+                    // assertion will fail if there is no encoding for
+                    // `copy_nop` at the required type.
```
bnjbvr: I'd remove this comment and instead I'd put a text comment in the assert like `debug_assert!(ok, "copy_nop encoding missing for this type");`

cranelift-codegen/meta-python/isa/x86/encodings.py
julian-seward1: From the point of view of "will CL assert/fail", no, we don't need them, because the transformation done by reload.rs is now limited to types satisfying `is_int` or `is_float`, and all targets (x86_64, x86_32, rv32, rv64) now have encodings for those types.

From the point of view of "are we missing optimisations", maybe.  But it seems unlikely to me, assuming that most boolean values are consumed by an "if" as soon as they are created.  On that basis, I omitted them.  If profiling later shows evidence that we need these cases, then we can add them.  I haven't seen any need for them in the profiling I've done so far.

julian-seward1:
New revision; addresses all review comments.  @bnjbvr, in the loops that create the instruction encodings, I used `base.copy_nop.bind(ty)` to create a version of the instruction of type `ty`.  This seems to work but is mere cargo-culting from examples above.  Do you if it is correct?

PR #786, opened 2019-06-05
Only create copy_nop instructions for types for which an encoding exi…

julian-seward1:
…sts. Issue #779.

PR #773 detects, at reload time, `copy` instructions that copy a value from stack
slot back to the same stack slot.  It replaces them with `copy_nop` instructions
that have a null encoding (hence producing no code).

For x86_64, `copy_nop` encodings for the types I64, I32, F64 and F32 are
provided.  Unfortunately the code that detects the redundant copy doesn't
check the type of the copied value, hence leaving itself open to the danger of
creating a `copy_nop` instruction cannot be encoded (which is different from
saying it has a null encoding).

This patch:

* Expands the x86_64 set of `copy_nop` encodings to: I64 I32 I16 I8 F64 and F32

* Adds encodings for the same for x86_32, rv64 and rv32.

* In `visit_inst()` in `reload.rs`, checks the type of the copied value accordingly.

* Adds comments explaining the above.

### DIFF
--- a/cranelift-codegen/meta-python/isa/riscv/encodings.py
+++ b/cranelift-codegen/meta-python/isa/riscv/encodings.py
@@ -8,7 +8,7 @@ from .defs import RV32, RV64
 from .recipes import OPIMM, OPIMM32, OP, OP32, LUI, BRANCH, JALR, JAL
 from .recipes import LOAD, STORE
 from .recipes import R, Rshamt, Ricmp, Ii, Iz, Iicmp, Iret, Icall, Icopy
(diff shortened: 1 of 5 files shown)


julian-seward1:
@bnjbvr, the travis builds fail on Linux only, for reasons I don't understand.  I can't reproduce the failure locally on Linux.  Any ideas?

bnjbvr:
As discussed on IRC, that's the Python lint that failed. Julian has installed tools and can now reproduce the issue.

PR #788, opened 2019-06-06
[meta] Uniquely number every instruction in the Rust crate

bnjbvr:
This implements a TODO that I left for later: each instruction has to have a precise opcode number that's stable (and that's reused when generating the encodings files). The way it's implemented is via a `PrimaryMap<OpcodeNumber, Instruction>`; I haven't replaced every reference to an Instruction by its InstructionNumber, because:

- it would mean that every time we want to read an Instruction's field, we'd need to have the PrimaryMap around, which would probably make it appear everywhere.
- Instruction is just a thin wrapper on its reference-counted structure containing all its field, so cloning it is cheap.

(Note: I'd be happy to get rid of the Rc<> and use the OpcodeNumber everywhere instead, but I'd be also very happy to postpone this work to after the Rust port is done, since another WIP PR may rely on this in the meanwhile.)

This also simplifies the generation of instructions, and makes `constant_hash::generate_table` more generic by allowing to pass it an `Iterator` instead of a `Vec` as the first argument (since we now pass it a `PrimaryMap` or a `Vec` in gen_settings).

### DIFF
--- a/cranelift-codegen/meta/src/cdsl/instructions.rs
+++ b/cranelift-codegen/meta/src/cdsl/instructions.rs
@@ -1,3 +1,10 @@
+use std::fmt;
+use std::ops;
+use std::rc::Rc;
+use std::slice;
+
+use cranelift_entity::{entity_impl, PrimaryMap};
+
 use crate::cdsl::camel_case;
 use crate::cdsl::formats::{
     FormatField, FormatRegistry, InstructionFormat, InstructionFormatIndex,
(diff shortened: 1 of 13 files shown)


bnjbvr:
@lars-t-hansen, can you have a look, please? If you can't, feel free to forward the review to somebody else, please!

lars-t-hansen:
Forwarded to Julian, as it is my understanding that either he or Anthony has expressed an interest in him working on this :-)

bnjbvr:
This has evolved enough that this is now stale and invalid. Closing.

Issue #789, opened 2019-06-06
Panic in SimpleJITRelocSink::reloc_jt


bjorn3:
```
thread 'rustc' panicked at 'not yet implemented', /home/bjorn/.cargo/git/checkouts/cranelift-af8713b52459e6cd/9094a3b/cranelift-simplejit/src/backend.rs:537:9
stack backtrace:
[...]
   8: std::panicking::begin_panic
             at /rustc/7cdaffd7962c4aae0cadd82baa241901b03f9458/src/libstd/panicking.rs:408
   9: <cranelift_simplejit::backend::SimpleJITRelocSink as cranelift_codegen::binemit::memorysink::RelocSink>::reloc_jt
             at /home/bjorn/.cargo/git/checkouts/cranelift-af8713b52459e6cd/9094a3b/cranelift-simplejit/src/backend.rs:537
[...]
```

<details>

```
test compile
set is_pic
target x86_64-unknown-linux-gnu

function u0:0() system_v {
; symbol _ZN21mini_core_hello_world4main17hdf273f44579fe5f2E
; instance Instance { def: Item(DefId(0:58 ~ mini_core_hello_world[317d]::main[0])), substs: [] }
; sig ([]; c_variadic: false)->()

; ssa {_4: (empty), _1: NOT_SSA, _5: (empty), _2: (empty), _3: (empty), _0: NOT_SSA}
; msg   loc.idx    param    pass mode            ssa flags  ty
; ret    _0      -          NoPass               NOT_SSA    ()

    ss0 = explicit_slot 8 ; _1: mini_core::Option<fn()> size=8 align=8,8
    gv0 = symbol colocated u1:8
    gv1 = symbol colocated u1:9
    gv2 = symbol colocated u1:10
    gv3 = symbol colocated u1:11
    sig0 = (i64) system_v
    sig1 = (i64) system_v
    sig2 = (i64) system_v
    fn0 = u0:2 sig0
    fn1 = colocated u0:1 sig1 ; Instance { def: Item(DefId(1:190 ~ mini_core[8787]::panic[0])), substs: [] }
    fn2 = u0:2 sig2
    jt0 = jump_table [ebb6, ebb3]

                                ebb0:
                                    v0 = stack_addr.i64 ss0
                                    jump ebb1

                                ebb1:
                                    nop 
; _1 = const main::FUNC_REF
@0000                               v1 = global_value.i64 gv0
@0000                               v2 = load.i64 aligned v1
[...]

PR #790, opened 2019-06-07
Handle Reloc::X86PCRelRodata4 in sundry reloc_jt

lars-t-hansen:
See issue #789: this reloc for jump table entries can be ignored by back-ends that don't need to split apart code and its rodata/jump tables.  This patch implements the ignoring for simplejit and faerie.

### DIFF
--- a/cranelift-faerie/src/backend.rs
+++ b/cranelift-faerie/src/backend.rs
@@ -413,7 +413,15 @@ impl<'a> RelocSink for FaerieRelocSink<'a> {
             .expect("faerie relocation error");
     }
 
-    fn reloc_jt(&mut self, _offset: CodeOffset, _reloc: Reloc, _jt: ir::JumpTable) {
-        unimplemented!();
+    fn reloc_jt(&mut self, _offset: CodeOffset, reloc: Reloc, _jt: ir::JumpTable) {
+        match reloc {
+            Reloc::X86PCRelRodata4 => {
+                // Not necessary to record this unless we are going to split apart code and its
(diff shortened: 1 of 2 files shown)


bnjbvr:
Thanks! Since that's using a trait, what do you think of instead providing a default no-op implementation for the `reloc_jt` function, so it behaves as before? Not having a default implementation forces embedders to think about which behavior they'd want, but that's the only advantage I can think of.

bnjbvr:
Oops, didn't see your comment on the issue before I posted that; suggestion is still valid with this implementation (which would also reduce copy/pasting).

lars-t-hansen:
> Thanks! Since that's using a trait, what do you think of instead providing a default no-op implementation for the `reloc_jt` function, so it behaves as before? Not having a default implementation forces embedders to think about which behavior they'd want, but that's the only advantage I can think of.

Obviously this comes down to personal preference, but...  Since the previous implementation had an `unimplemented!()` in every specialization of RelocSink, and not a default method, I would think that the desire here is to actually make the embedders think about the behavior they want.

A default implementation could do nothing for this specific reloc type I guess and then crash for all other reloc types.  I'm not sure this is much of an improvement but I see the appeal.

(My personal preference is no default method, but LMK if you feel strongly about it.)

bnjbvr:
I don't feel too strongly about it, and having embedders make a thoughtful choice sounds safer. The relocation is at least commented, and the two RelocSink implementations can serve as examples, so this looks good to me.

Issue #502, opened 2018-09-07
Make "wasm" and "compile" subcommand output more consistent.


sunfishcode:
clif-util's "wasm" and "compile" subcommands essentially do the same thing, except that "wasm"'s input is a wasm file, and "compile"'s input is a clif file, but then they both do a full compilation. Currently, they print different things:
 - "wasm" has a -v option which enables colorized messages like "Translating... ok", "compile" doesn't
 - "compile" prints out the ".byte" sequence, "wasm" doesn't
 - "compile" runs the capstone disassembler, "wasm" doesn't

and possibly other differences. We should harmonize these two subcommands.

Benjamin-L:
I'm going to try to take a crack at this. It seems like the most obvious way to do this is to just duplicate a bunch of code from the `wasm` subcommand into `compile`. It seems like a possibly better option would be to put all the shared functionality into one place so that it doesn't get duplicated. I'm not sure exactly how to go about this with the current structure, but I'm going to keep poking around and try to figure out how things work.

sunfishcode:
Sounds good! I don't have a specific code organization in mind yet, so I'll be interested to see where this leads.

lars-t-hansen:
I didn't see this ticket before now but I landed a fix a couple weeks ago that adds disassembly functionality to the wasm subcommand, you have to use -D.  It also changed the compile subcommand to required -D to get disassembly.  (It's possible more things are needed here so I'll not do anything about this ticket.)

sunfishcode:
@Benjamin-L Do you have any updates here, or do you mind if we assign this to someone else?

Benjamin-L:
Go ahead and assign it to somebody else. I'd stopped working on this a while ago, sorry about not mentioning it here.

PR #791, opened 2019-06-08
Issue 502 - unify wasm compile output

ash-tamraz:
This is in reference to [issue 502](https://github.com/CraneStation/cranelift/issues/502)

I'm just for feedback on what I have at this point. It doesn't feel like I've done much of anything, but I'm not really sure what else to do to unify these further. I'm wondering if maybe I've missed the point of the issue? Any feedback is appreciated!

### DIFF
--- a/src/compile.rs
+++ b/src/compile.rs
@@ -1,10 +1,11 @@
 //! CLI tool to read Cranelift IR files and compile them into native code.
 
+use crate::clif_shared::{check_translation, check_translation_preamble,
+print_size, report_times, module_code_size};
 use crate::disasm::{print_all, PrintRelocs, PrintStackmaps, PrintTraps};
 use crate::utils::{parse_sets_and_triple, read_to_string};
 use cranelift_codegen::print_errors::pretty_error;
 use cranelift_codegen::settings::FlagsOrIsa;
-use cranelift_codegen::timing;
 use cranelift_codegen::Context;
 use cranelift_reader::{parse_test, ParseOptions};
 use std::path::Path;
@@ -17,7 +18,11 @@ pub fn run(
     flag_report_times: bool,
     flag_set: &[String],
     flag_isa: &str,
+    flag_print_size: bool,
+    flag_check_translation: bool,
+    flag_verbose: bool,
 ) -> Result<(), String> {
+
     let parsed = parse_sets_and_triple(flag_set, flag_isa)?;
 
     for filename in files {
@@ -30,6 +35,9 @@ pub fn run(
             &path.to_path_buf(),
             &name,
             parsed.as_fisa(),
+            flag_print_size,
+            flag_check_translation,
+            flag_verbose,
         )?;
     }
     Ok(())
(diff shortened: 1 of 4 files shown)


src/compile.rs
```diff
@@ -36,15 +55,26 @@ pub fn run(
 }
 
 fn handle_module(
+    _flag_verbose: bool,
```
bjorn3: Please remove leading `_`.

src/compile.rs
```diff
@@ -10,8 +10,26 @@ use cranelift_reader::parse_test;
...
+    ($x: expr, $($tts:tt)*) => {
+        if $x {
+            print!($($tts)*);
+        }
+    }
+}
+
 pub fn run(
     files: Vec<String>,
+    _flag_verbose: bool,
```
bjorn3: Please remove leading `_`.

src/compile.rs
ash-tamraz: When I initially added this code, I got a warning from cargo suggesting that I add the leading `_`. I've removed it and it looks like it's working fine now, though.

src/compile.rs
ash-tamraz: I've removed them.

sunfishcode:
This patch makes "compile" more consistent with "wasm", which is a good start. The other thing we'd like to do here is actually factor out parts of the code which are the same between the two, so that they can stay consistent as things evolve. Does that make sense?

ash-tamraz:
>This patch makes "compile" more consistent with "wasm", which is a good start. The other thing we'd like to do here is actually factor out parts of the code which are the same between the two, so that they can stay consistent as things evolve. Does that make sense?

Definitely! I have two methods for unifying shared functionality between `src/compile.rs` and `src/wasm.rs` in mind: 
- Have `src/clif-util.rs` call a unified module which performs wasm or compile specific operations depending on a parameter indicating the calling subcommand (probably a boolean).
- Leave `src/clif-util.rs` as it is and have `src/compile.rs` and `src/wasm.rs` individually call functions in a module that unifies shared functionality as needed.

Which would you say is the best design for factoring out shared code? Or maybe you have suggestions for better designs for this application?

ash-tamraz:
@sunfishcode I did some quick unification of code shared between `wasm` and `compile`. Let me know if the changes and design match what you had in mind, or if I need to make some changes.

sunfishcode:
That's a good start! One of the next key pieces will be the code around the `compile_and_emit` calls. This is a little tricker, because the code paths aren't exactly the same right now, but wasm's `flag_check_translation` and `flag_print_size` are things that'd be nice to have in the execute path as well, which should make more things in common between the two.

ash-tamraz:
> That's a good start! One of the next key pieces will be the code around the `compile_and_emit` calls. This is a little tricker, because the code paths aren't exactly the same right now, but wasm's `flag_check_translation` and `flag_print_size` are things that'd be nice to have in the execute path as well, which should make more things in common between the two.

Apologies for the delay! Here are some changes reflecting your feedback, finally. (:-P)

I renamed the shared file `clif_shared.rs`, as it was a little less wordy but still communicates the goal (I know the convention is to name files with a `-` as opposed to a `_`, but I couldn't figure out how to indicate to cargo that the `-` should be replaced with a `_` in in `use` statements).

`print_size` and `module_code_size` currently overlap, e.g. `cargo run wasm -v -X --target=x86_64 wasmtests/arith.wat ` triggers both. It shouldn't be difficult to resolve that, though.

alexcrichton:

tschneidereit:
We'll archive this repository in favor of the Wasmtime repository — see Alex's comment about the merge above. I'm closing this PR now in order to have an explanation in a comment before archiving the repo auto-closes all PRs.

Issue #354, opened 2018-05-28
i128 support


sunfishcode:
Rust has an [i128 type](https://doc.rust-lang.org/std/primitive.i128.html), so a rustc backend will need a way to support it. While one option would be to oblige the frontend to lower i128 into i64 operations, Cretonne already has legalization strategies for splitting integers in half, and much of this code ought to work for i128.

The tricky part is that the `iconst` instruction, as well as all `*_imm` instructions, uses `imm64`. One option is to just change these to `Imm128`. I'm inclined to do that, as it keeps i128 consistent with the other integer types. However, most users don't need i128, so it would be good to ensure that doing this doesn't add significant overhead or complexity for users that don't use it.

Another option would be to introduce an `i128const` instruction, and leave the `*_imm` variants unsupported in i128. That wouldn't impact users that don't use it, though it'd be less tidy.

Vurich:
Why not have arbitrary-sized integers, like LLVM?

sunfishcode:
The short-term answer is that Cranelift is focusing on being a codegen-oriented IR, for general-purpose CPUs, and arbitrary-sized integers aren't very relevant there.

Longer term, it depends on how important it ends up being, and whether simpler alternatives would suffice for the things we actually need to do. What's your use case?

Vurich:
Ah yeah, for a codegen IR it makes a lot less sense. The code required to make it generic would be more complexity than it's worth. There are languages with arbitrary-bit-sized integers, like [zig](https://github.com/ziglang/zig) but you can easily generate IR for them using the available integers as primitives.

It might be worth requiring frontends to generate the 128-bit number emulation themselves. I don't believe that there are any x86 or ARM instructions operating on 128-bit numbers, apart from the result of x86 `mulq` (and maybe ARM's equivalent? I don't know ARM) which you could support by having the IR's multiplication instruction on 64-bit numbers return a pair.

ckaran:
As a suggestion, you might want to reserve any tokens that match `i[0-9]+` and `u[0-9]+`.  If you want to expand the IR to include `i128` and `u128` in the future (or something less sane like `i65536`), you'll have the option.  If you want to write code to check that people aren't using these as identifiers, that's fine, but I think just putting it in the docs will be enough (anyone that violates the docs has to deal with the consequences in the future).

sunfishcode:
[...]

PR #795, opened 2019-06-12
I128

bjorn3:
cc #354

This adds a i128 type. While only `iconcat` and `isplit` support them, this makes it easier `cg_clif` to handle 128 bit integers, as the low and high bits are bundled together. `cg_clif` would need a bigger change to special case `i128` as not consisting of just one `Value`. It also makes it easier to inspect the ir generated by `cg_clif`, as there is no need to think about which two values together represent the `i128`.

You can create an `i128` value by using `iconcat` with two `i64` values: first the lsb, then the msb (on little endian)

## TODO

* [x] Fix isplit i128 -> i64,i64

### DIFF
--- a/cranelift-codegen/meta/src/cdsl/instructions.rs
+++ b/cranelift-codegen/meta/src/cdsl/instructions.rs
@@ -79,12 +79,14 @@ impl InstructionGroup {
     }
 }
 
+#[derive(Debug)]
 pub struct PolymorphicInfo {
     pub use_typevar_operand: bool,
     pub ctrl_typevar: TypeVar,
     pub other_typevars: Vec<TypeVar>,
 }
 
+#[derive(Debug)]
 pub struct InstructionContent {
     /// Instruction mnemonic, also becomes opcode name.
     pub name: String,
@@ -139,7 +141,7 @@ pub struct InstructionContent {
     pub writes_cpu_flags: bool,
 }
 
-#[derive(Clone)]
+#[derive(Clone, Debug)]
 pub struct Instruction {
     content: Rc<InstructionContent>,
 }
--- a/cranelift-codegen/meta/src/cdsl/instructions.rs
+++ b/cranelift-codegen/meta/src/cdsl/instructions.rs
@@ -1127,6 +1127,11 @@ fn bind_vector(
     mut value_types: Vec<ValueTypeOrAny>,
 ) -> BoundInstruction {
     let num_lanes = vector_size_in_bits / lane_type.lane_bits();
+    assert!(
+        num_lanes >= 2,
+        "Minimum lane number for bind_vector is 2, found {}.",
+        num_lanes,
+    );
     let vector_type = ValueType::Vector(VectorType::new(lane_type, num_lanes));
     value_types.push(ValueTypeOrAny::ValueType(vector_type));
     verify_polymorphic_binding(&inst, &value_types);
--- a/cranelift-codegen/meta/src/cdsl/instructions.rs
+++ b/cranelift-codegen/meta/src/cdsl/instructions.rs
@@ -14,7 +14,7 @@ use crate::cdsl::type_inference::Constraint;
 use crate::cdsl::types::{LaneType, ReferenceType, ValueType, VectorType};
 use crate::cdsl::typevar::TypeVar;
 
-#[derive(Copy, Clone, PartialEq, Eq, Hash, PartialOrd, Ord)]
+#[derive(Copy, Clone, Debug, PartialEq, Eq, Hash, PartialOrd, Ord)]
 pub struct OpcodeNumber(u32);
 entity_impl!(OpcodeNumber);
 
--- a/cranelift-codegen/meta/src/gen_legalizer.rs
+++ b/cranelift-codegen/meta/src/gen_legalizer.rs
@@ -89,6 +89,8 @@ fn unwrap_inst(
                     fmtln!(fmt, "func.dfg.resolve_aliases(args[{}]),", n);
                 } else if op.is_varargs() {
                     let n = inst.imm_opnums.iter().chain(inst.value_opnums.iter()).max().map(|n| n + 1).unwrap_or(0);
+                    // We need to create a `Vec` here, as using a slice would result in a borrowck
+                    // error later on.
                     fmtln!(fmt, "\
                         args.iter().skip({}).map(|&arg| func.dfg.resolve_aliases(arg)).collect::<Vec<_>>(),\
                     ", n);
@@ -115,6 +117,9 @@ fn unwrap_inst(
             let name = var_pool
                 .get(apply.args[i].maybe_var().expect("vararg without name"))
                 .name;
+
+            // Above name is set to an `Vec` representing the varargs. However it is expected to be
+            // `&[Value]` below, so we borrow it.
             fmtln!(fmt, "let {} = &{};", name, name);
         }
     }
@@ -419,6 +424,8 @@ fn gen_transform<'a>(
         }
 
         if transform.def_pool.get(transform.src).apply.inst.is_branch {
+            // A branch might have been legalized into multiple branches, so we need to recompute
+            // the cfg.
             fmt.line("cfg.recompute_ebb(pos.func, pos.current_ebb().unwrap());");
         }
 
--- a/cranelift-codegen/meta/src/gen_legalizer.rs
+++ b/cranelift-codegen/meta/src/gen_legalizer.rs
@@ -61,10 +61,10 @@ fn unwrap_inst(
             fmtln!(fmt, "{},", field.member);
         }
 
-        if iform.num_value_operands == 1 {
-            fmt.line("arg,");
-        } else if iform.has_value_list || iform.num_value_operands > 1 {
+        if iform.has_value_list || iform.num_value_operands > 1 {
             fmt.line("ref args,");
+        } else if iform.num_value_operands == 1 {
+            fmt.line("arg,");
         }
 
         fmt.line("..");
@@ -87,6 +87,11 @@ fn unwrap_inst(
                 } else if op.is_value() {
                     let n = inst.value_opnums.iter().position(|&i| i == op_num).unwrap();
                     fmtln!(fmt, "func.dfg.resolve_aliases(args[{}]),", n);
+                } else if op.is_varargs() {
+                    let n = inst.imm_opnums.iter().chain(inst.value_opnums.iter()).max().map(|n| n + 1).unwrap_or(0);
+                    fmtln!(fmt, "\
+                        args.iter().skip({}).map(|&arg| func.dfg.resolve_aliases(arg)).collect::<Vec<_>>(),\
+                    ", n);
                 }
             }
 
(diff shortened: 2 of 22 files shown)


bjorn3:
The test fails with:

```
function u0:0(i64 [%rdi], i64 [%rsi]) -> i64 [%rax], i64 [%rdx], i64 [%rcx], i64 [0] fast {
                                ebb0(v0: i64 [%rdi], v1: i64 [%rsi]):
[RexOp1pu_id#b8,%rax]               v2 = iconst.i64 42
[RexOp1pu_id#b8,%rcx]               v3 = iconst.i64 0
[RexOp1rmov#8089]                   regmove v3, %rcx -> %rdx
[RexOp1rmov#8089]                   regmove v0, %rdi -> %rcx
[Op1ret#c3]                         return v2, v3, v0, v1
; error: inst3: ABI expects v1 at stack offset 0, got %rsi

}
```

Can somebody help me with fixing it?

sunfishcode:
The immediate problem is that the x86 backend only supports [3 integer values to be returned in registers](https://github.com/CraneStation/cranelift/blob/master/cranelift-codegen/src/isa/x86/abi.rs#L24), and after splitting the i128 into two, the function would require 4 integer return registers.

We'll need to convert functions with return types like that to use out parameters. One way to do that is to require frontends to lower such code to using out parameters instead -- and ideally add helper routines in cranelift-frontend to make it easy for frontends to do this. Another would be to teach the legalization phase in cranelift-codegen to convert such return types into out parameters.

I have a mild preference for the former, to keep the code in cranelift-codegen as simple as possible, though either would work.

sunfishcode:
For the record, 3 integer types is more than is specified in the System V ABI, but is consistent with [LLVM](https://github.com/llvm/llvm-project/blob/master/llvm/lib/Target/X86/X86CallingConv.td#L217).

bjorn3:
> The immediate problem is that the x86 backend only supports 3 integer values to be returned in registers, and after splitting the i128 into two, the function would require 4 integer return registers.

That explains it. Using an out param when a value returned ByValPair consists of two i128 values should workaround this problem for cg_clif then.

> One way to do that is to require frontends to lower such code to using out parameters instead -- and ideally add helper routines in cranelift-frontend to make it easy for frontends to do this.

When this option is chosen, it would be nice to add a verifier error for more than three ret values on x86.

I want to add some legalizations for loading and storing i128 before this is merged.

bjorn3:
I got the cg_clif side working: https://github.com/bjorn3/rustc_codegen_cranelift/pull/627

bjorn3:
`add` and `sub` are already legalized to 64bit ops by the same code legalizing the 64bit versions to 32bit on i686. There rest of the math instructions are codegened to compiler_builtins calls by cg_clif.

bjorn3:
The linux build timed out during cache unpacking. The macOS build did succeed however.

bjorn3:
Travis is failing during cache (un)packing again.

sunfishcode:
Eek, I don't know what's going on with Travis here. If anyone has any ideas of what we should do here, please share :-}.

ethanhs:
I've run into issues with Travis timing out before, and its been doing it more lately. In general going to Travis and clicking restart build helps, have you tried that?

sunfishcode:
I've now clicked "restart build".

bjorn3:
Travis is green

bjorn3:
@bnjbvr could you review this?

bnjbvr:
Yes, I could get to it; unfortunately I'm a bit in a review debt situation, so there might be many other things before I can come back to this.

bjorn3:
Thanks!

bnjbvr:
Sorry, I realistically don't have the time these days to review this, deferring to Dan.

bjorn3:
Rebased

bjorn3:
I don't get the CI failure locally.

cranelift-codegen/meta/src/isa/x86/encodings.rs
```diff
@@ -1755,7 +1757,7 @@ pub fn define(
     extractlane_mapping.insert(32, (vec![0x66, 0x0f, 0x3a, 0x16], Some(use_sse41_simd))); // PEXTRD
     extractlane_mapping.insert(64, (vec![0x66, 0x0f, 0x3a, 0x16], Some(use_sse41_simd))); // PEXTRQ, only x86_64
 
-    for ty in ValueType::all_lane_types() {
+    for ty in ValueType::all_lane_types().filter(allowed_simd_type) {
```
bjorn3: I assume it was unintentional to support b1 here.

cranelift-codegen/meta/src/gen_legalizer.rs
```diff
@@ -87,6 +87,11 @@ fn unwrap_inst(
                 } else if op.is_value() {
                     let n = inst.value_opnums.iter().position(|&i| i == op_num).unwrap();
                     fmtln!(fmt, "func.dfg.resolve_aliases(args[{}]),", n);
+                } else if op.is_varargs() {
+                    let n = inst.imm_opnums.iter().chain(inst.value_opnums.iter()).max().map(|n| n + 1).unwrap_or(0);
+                    fmtln!(fmt, "\
+                        args.iter().skip({}).map(|&arg| func.dfg.resolve_aliases(arg)).collect::<Vec<_>>(),\
+                    ", n);
```
sunfishcode: I'm sorry I don't have as much context here as I should; can you remind me what this patch needs varargs for?

cranelift-codegen/meta/src/gen_legalizer.rs
```diff
@@ -402,6 +417,11 @@ fn gen_transform<'a>(
             fmt.line("let removed = pos.remove_inst();");
             fmt.line("debug_assert_eq!(removed, inst);");
         }
+
+        if transform.def_pool.get(transform.src).apply.inst.is_branch {
+            fmt.line("cfg.recompute_ebb(pos.func, pos.current_ebb().unwrap());");
+        }
```
sunfishcode: Please add a comment here explaining why this is needed.

cranelift-codegen/src/legalizer/mod.rs
```diff
@@ -36,31 +38,79 @@ use self::heap::expand_heap_addr;
...
+
+        let arg = match pos.func.dfg[inst] {
+            ir::InstructionData::Unary { arg, .. } => pos.func.dfg.resolve_aliases(arg),
+            _ => panic!("Expected isplit: {}", pos.func.dfg.display_inst(inst, None)),
+        };
+
+        match pos.func.dfg.value_def(arg) {
+            ir::ValueDef::Result(inst, _num) => {
+                if let ir::InstructionData::Binary {
+                    opcode, args: _, ..
```
sunfishcode: I believe you can simplify this logic by removing the `args`, since it's unused and you have a `..` already, but also replacing `opcode` with `ir::Opcode::Iconcat`, so that the pattern match only suceeds if it has that opcode, which should simplify the ifs below.

cranelift-codegen/src/legalizer/mod.rs
```diff
@@ -102,16 +159,33 @@ pub fn legalize_function(func: &mut ir::Function, cfg: &mut ControlFlowGraph, is
...
+                LegalizeInstResult::SplitLegalizePending => {
+                    pending_splits.insert(inst);
+                }
             }
         }
     }
 
+    while let Some(ebb) = pos.next_ebb() {
+        split::split_ebb_params(pos.func, cfg, ebb);
+    }
```
sunfishcode: The previous loop is already `while let Some(_ebb) = pos.next_ebb() {`, so this does loop do anything?

cranelift-codegen/src/legalizer/mod.rs
```diff
@@ -498,3 +572,67 @@ fn expand_stack_store(
     mflags.set_aligned();
     pos.func.dfg.replace(inst).store(mflags, val, addr, 0);
 }
+
+/// Split a load into two parts before `iconcat`ing the result together.
+fn narrow_load(
```
sunfishcode: Would it make sense to call this `split_load`?

cranelift-codegen/src/legalizer/mod.rs
```diff
@@ -498,3 +572,67 @@ fn expand_stack_store(
...
+        small_ty,
+        flags,
+        ptr,
+        offset.try_add_i64(8).expect("load offset overflow"),
+    );
+    pos.func.dfg.replace(inst).iconcat(al, ah);
+}
+
+/// Split a store into two parts after `isplit`ing the value.
+fn narrow_store(
```
sunfishcode: Ditto, `split_store`?

cranelift-codegen/src/legalizer/mod.rs
bjorn3: This is in the narrow legalization group.

sunfishcode: Ah, makes sense.

cranelift-codegen/meta/src/isa/x86/encodings.rs
bjorn3: cc @abrown

cranelift-codegen/meta/src/cdsl/instructions.rs
```diff
@@ -1138,6 +1140,11 @@ fn bind_vector(
     mut value_types: Vec<ValueTypeOrAny>,
 ) -> BoundInstruction {
     let num_lanes = vector_size_in_bits / lane_type.lane_bits();
+    assert!(
+        num_lanes >= 2,
```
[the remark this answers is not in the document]
bjorn3: When using `bind_vector` with a single lane, it will emit types like `I128X1` in the final output, which is not a valid type, only `I128` is.

cranelift-codegen/meta/src/gen_legalizer.rs
bjorn3: https://github.com/CraneStation/cranelift/pull/795/files#diff-6263bd0b84702d1f31e46a46ce47b28fR257-R291

Legalizing `brz.i128` and `brnz.i128`.

bjorn3:
> I think the biggest thing I still need to get a better understanding for is pending_splits and SplitLegalizePending.

That is basically to handle the case where an `isplit` gets a value which is created by an instruction which legalizes to an `iconcat`, but where the instruction is part of an ebb which comes later in the layout.

```
ebb1:
    v2, v3 = isplit v1
    trap user0

ebb0(v0: i128):
    v1 = iadd_imm v0, 1
    jump ebb1
```

`v2, v3 = isplit v1` would get be visited first, but `v1 = iadd_imm v0, 1` is not visited yet, so we recors that it needs to be split. When all other instructions are legalized, we can try to split `v2, v3 = isplit v1` again. Which would succeed this time.

cranelift-codegen/src/legalizer/mod.rs
bjorn3: Merged with the above loop.

cranelift-codegen/meta/src/isa/x86/encodings.rs
abrown: Correct, I think I caught it in most places but may have missed it here.

bjorn3:
Rebased

bjorn3:
I can finally move cg_clif back to upstream Cranelift :tada:

bjorn3:
^ https://github.com/bjorn3/rustc_codegen_cranelift/pull/700

Issue #792, opened 2019-06-09
thread 'main' panicked at 'assertion failed: new_info == info'


ash-tamraz:
Running this on Ubuntu 16.04.6 on an x86_64 machine.

While in my fork that is even with `CraneStation:master`, on a branch with no changes, running `cargo run compile -p --target=x86_64 filetests/wasm/i64-compares.clif` produces the output:
```
    Finished dev [unoptimized + debuginfo] target(s) in 0.10s
     Running `target/debug/clif-util compile -p --target=x86_64 filetests/wasm/i64-compares.clif`
thread 'main' panicked at 'assertion failed: new_info == info', cranelift-codegen/src/context.rs:109:9
stack backtrace:
   0: std::sys::unix::backtrace::tracing::imp::unwind_backtrace
             at src/libstd/sys/unix/backtrace/tracing/gcc_s.rs:39
   1: std::sys_common::backtrace::_print
             at src/libstd/sys_common/backtrace.rs:70
   2: std::panicking::default_hook::{{closure}}
             at src/libstd/sys_common/backtrace.rs:58
             at src/libstd/panicking.rs:200
   3: std::panicking::default_hook
             at src/libstd/panicking.rs:215
   4: std::panicking::rust_panic_with_hook
             at src/libstd/panicking.rs:478
   5: std::panicking::begin_panic
             at /rustc/6c2484dc3c532c052f159264e970278d8b77cdc9/src/libstd/panicking.rs:412
   6: cranelift_codegen::context::Context::compile_and_emit
             at cranelift-codegen/src/context.rs:109
   7: clif_util::compile::handle_module
             at src/compile.rs:68
   8: clif_util::compile::run
             at src/compile.rs:26
   9: clif_util::main
             at src/clif-util.rs:248
  10: std::rt::lang_start::{{closure}}
             at /rustc/6c2484dc3c532c052f159264e970278d8b77cdc9/src/libstd/rt.rs:64
  11: std::panicking::try::do_call
             at src/libstd/rt.rs:49
             at src/libstd/panicking.rs:297
  12: __rust_maybe_catch_panic
             at src/libpanic_unwind/lib.rs:87
  13: std::rt::lang_start_internal
             at src/libstd/panicking.rs:276
             at src/libstd/panic.rs:388
             at src/libstd/rt.rs:48
  14: std::rt::lang_start
             at /rustc/6c2484dc3c532c052f159264e970278d8b77cdc9/src/libstd/rt.rs:64
  15: main
[...]

PR #797, opened 2019-06-18
Fix a calculation error for rodata_size in memsink

lars-t-hansen:
The calculation for rodata_size subtraced only the jump table size from the current offset, but had to subtract the sum of all the previous sections, including the code.

### DIFF
--- a/cranelift-codegen/src/binemit/memorysink.rs
+++ b/cranelift-codegen/src/binemit/memorysink.rs
@@ -142,7 +142,7 @@ impl<'a> CodeSink for MemoryCodeSink<'a> {
     }
 
     fn end_codegen(&mut self) {
-        self.info.rodata_size = self.offset() - self.info.jumptables_size;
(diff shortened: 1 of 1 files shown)


lars-t-hansen:
Oh yeah, this is for #792.

bnjbvr:
Can you dup this test into filetests, please, so this serves as a non-regression test?

lars-t-hansen:
@bnjbvr, not quite sure what you have in mind re the tests.  MemoryCodeSink is used only by clifutil and simplejit so it's not clear to me how this fits into filetests.

bnjbvr:
@lars-t-hansen I could trigger the same error by running `clif-util compile` on a lot of random files in filetests (sorry, can't remember which ones! but it happened quite frequently), so we could just copy one of these test cases and replace the testing directive with "compile" I think, and that should reproduce the error. Does it make sense? In any case, happy to take the patch asap.

lars-t-hansen:
Merging this now because I don't want it to linger longer and we have some test coverage, but I'll fix the test case for it as soon as I can next week.

PR #799, opened 2019-06-20
Meta port 4: Finish port to the Rust DSL and remove the Python code

bnjbvr:
Almost done, but at least it now passes all the Cranelift tests.

Before being ready for review:

- [x] make sure this passes the spidermonkey test suite
- [x] as well as a few big wasm programs. (compilation works for Godot Engine, Angry Bots, and Tanks, Embenchen compiles and runs properly)
- [x] this needs to be split into several commits. This is not going to be pretty.
- [x] ~~we need to figure out what to do with the documentation that was auto-generated from Python. (It lives on https://cranelift.readthedocs.io.)~~
- [x] I've removed the Python code that was a starting point for the semantics verification, and it has not been ported over to Rust. Make sure this is ok.
- [x] need to add comments, address a few TODOs, etc.
- [x] remove references to Python in comments.


data-pup:
really cool to see this coming to fruition :smile_cat: :tada:

bnjbvr:
I propose that the documentation be updated afterwards: some of it is still quite valid and useful in my opinion. I don't know if docs.rs would be able to show the documentation that's in the generated files, but if that's the case, we could add doc comments in the generated files, and use that in addition to the static documentation that's in the `rst` files in the docs directory.

syrusakbary:
This is awesome 🎉

bnjbvr:
Here we go, I think I'm done with splitting. For what it's worth, there are still a few large commits like the recipe/encodings for each target ISA. They are partially manually copied (scary! but we have a lot of tests) and partially automatically generated from the python code. So this is ready for review! I've already split the PR into a smaller part, and I can propose other smaller PRs if reviewers would prefer that.

cranelift-codegen/meta/src/isa/riscv/encodings.rs
```diff
@@ -0,0 +1,381 @@
...
+    let ishl_imm = shared.by_name("ishl_imm");
+    let isub = shared.by_name("isub");
+    let jump = shared.by_name("jump");
+    let regmove = shared.by_name("regmove");
+    let spill = shared.by_name("spill");
+    let sshr = shared.by_name("sshr");
+    let sshr_imm = shared.by_name("sshr_imm");
+    let ushr = shared.by_name("ushr");
+    let ushr_imm = shared.by_name("ushr_imm");
+    let return_ = shared.by_name("return");
```
sunfishcode: A similar sequence appears [in legalize.rs](https://github.com/CraneStation/cranelift/blob/master/cranelift-codegen/meta/src/shared/legalize.rs#L43) as well as x86/encodings.rs. And, these appear to be one of the main blockers to splitting up functions like `define` here into smaller pieces, because then you'd have to duplicate these declarations even further.

My best idea so far is to define global static variables in a submodule named `band` and similar, so that code like the `define` function here can import them all with something like `use opcode_names::*`, although static variables come with different problems.

Macros might also work here, but I agree with keeping these things low-magic for now.

I'm curious if you have any other ideas for what we might do in the future here.

cranelift-codegen/meta/src/isa/riscv/recipes.rs
```diff
@@ -0,0 +1,259 @@
+use std::collections::HashMap;
+
+use crate::cdsl::formats::FormatRegistry;
+use crate::cdsl::instructions::Predicate;
```
sunfishcode: In x86/recipes.rs, `instructions::Predicate` is imported as `InstructionPredicate`, so it'd be nice to do that here too to keep the different architectures consistent. In fact, would it make sense to just rename this `Predicate` to `InstructionPredicate` at the source so that we don't need to rename it in the imports?

cranelift-codegen/meta/src/isa/x86/encodings.rs
```diff
@@ -0,0 +1,1569 @@
...
+
+use std::collections::HashMap;
+
+use crate::cdsl::encodings::{Encoding, EncodingBuilder};
+use crate::cdsl::instructions::{
+    BoundInstruction, InstSpec, Instruction, InstructionGroup, Predicate as InstructionPredicate,
+    PredicateNode as InstructionPredicateNode, PredicateRegistry,
+};
+use crate::cdsl::recipes::{EncodingRecipe, EncodingRecipeNumber, Recipes};
+use crate::cdsl::settings::{PredicateNumber as SettingsPredicateNumber, SettingGroup};
```
sunfishcode: `settings::PredicateNumber` gets renamed to `SettingsPredicateNumber` in several places. My own experience is that import renamings like this end up as one-more-thing that I have to remember when reading code and figuring out how all the parts relate to each other, so I have a preference for renaming at the source rather than at the import, however it's not a strong preference.

cranelift-codegen/meta/src/cdsl/encodings.rs
```diff
@@ -0,0 +1,174 @@
...
+    }
+    pub fn to_rust_comment(&self, recipes: &Recipes) -> String {
+        format!("[{}#{:02x}]", recipes[self.recipe].name, self.encbits)
+    }
+}
+
+#[derive(Clone)]
+pub struct Encoding {
+    content: Rc<EncodingContent>,
+}
```
sunfishcode: Instead of a `struct`, could this be a type alias, like `type Encoding = Rc<EncodingContent>;`?

cranelift-codegen/meta/src/isa/riscv/encodings.rs
bnjbvr: Yeah, I realized pretty late I could *probably* have used static data almost everywhere (and defer the cost of finding variables to the compiler itself). I think global variables may have more ownership issues (and require locks), but maybe we can avoid those by defining a very unsafe black box for globals (knowing the meta crate will never do concurrent work... but will it not, though?).

I like the idea of having submodules. At some point I thought it might be even nice to group meta-descriptions by instruction, instead of by ISA. That is, to have the instruction definition, legalizations and encodings in the `band` submodule itself, so everything is nicely contained.

I'd like to avoid macros too, to make debugging easy and not add too much meta-thinking to the meta crate.

cranelift-codegen/meta/src/isa/riscv/recipes.rs
bnjbvr: Good call, I've hesitated for long with this, but your arguments make sense to me. Will rename this and PredicateNumber as well.

bnjbvr:
> I see you crossed off the item about cranelift.readthedocs.io content above; was there a resolution there?

I proposed in another comment (above) to clean up the RST documentation afterwards, because: 1. some of it could be useful as high-level explanations to how Cranelift works; 2. this PR is big enough at this time; 3. I don't know much about how the documentation build process works, and it seemed to interact with the Python code. Does it make sense to defer this?

cranelift-codegen/meta/src/cdsl/encodings.rs
bnjbvr: Nice idea!

bnjbvr:
Addressed all remarks from this first batch, thanks for looking at this PR! I've also taken the liberty to expand more names that were renamed on imports (PredicateMap => InstructionPredicateMap, etc.).

bnjbvr:
Thanks @nbp for all the great reviews!

PR #787, opened 2019-06-06
Add the dyn keyword to trait objects in struct members and parameters

bnjbvr:
Also remove unused result warnings for non-fatal type checks, and actually propagate the result of one of these type checks that was probably not enabled by mistake.

With this PR, Cranelift compiles with Rustc nightly without any warnings.

PR #800, opened 2019-06-21
Specify explicit "dyn" for trait objects

sstangl:
Trait objects without an explicit "dyn" are deprecated.
This greatly reduces the warning spam when building with Rust 1.37.

The syntax is valid for all Rust versions supported by Cranelift:
it was introduced in Rust 1.27, and Cranelift requires at least 1.32.


bnjbvr:
Thanks! But that's https://github.com/CraneStation/cranelift/pull/787 :) Care to review it, please?

PR #801, opened 2019-06-21
Meta port 1

bnjbvr:
Some prerequisites split off the other big PR, that can be reviewed independently.

Selecting many reviewers, yet just getting one review should be sufficient here. Thanks!


cranelift-codegen/meta/src/default_map.rs
```diff
@@ -0,0 +1,28 @@
+use std::collections::HashMap;
+use std::hash::Hash;
+
+/// A standard map with an additional get_or_default method that allows to create default entries
+/// when a key is looked up for the first time.
+pub struct DefaultMap<K, V> {
```
nbp: Note: This new type seems to be unused as part of part 1.

cranelift-codegen/meta/src/cdsl/instructions.rs
```diff
@@ -7,41 +14,51 @@ use crate::cdsl::type_inference::Constraint;
...
+            all_instructions,
+            own_instructions: Vec::new(),
         }
     }
 
     pub fn push(&mut self, builder: InstructionBuilder) {
-        self.instructions.push(builder.build(self.format_registry));
+        // gen_inst wants the opcode number to start from 1.
+        // TODO this can change once we get rid of the Python code.
+        let opcode_number = OpcodeNumber(self.all_instructions.next_key().as_u32() + 1);
```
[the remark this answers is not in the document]
bnjbvr: Thats right, that's because gen_inst wants the opcode number to start from 1. You have a good point that it could be mistaken for the `all_instructions` index, which is not what it should mean. Probably instead I should just save the actual index as the opcode number, and offset it when generating the file.

cranelift-codegen/meta/src/default_map.rs
```diff
@@ -0,0 +1,28 @@
...
+    V: Default,
+{
+    pub fn new() -> Self {
+        Self {
+            map: HashMap::new(),
+        }
+    }
+
+    pub fn get_or_default(&mut self, k: K) -> &mut V {
+        self.map.entry(k).or_insert_with(|| V::default())
```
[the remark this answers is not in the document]
bnjbvr: Yeah, I made this data structure to avoid repeating the default value everywhere the `or_default` method would be used. Augmenting hash map with the trait sounds like a simpler idea, thanks.

cranelift-codegen/meta/src/default_map.rs
bnjbvr: And you're correct; I haven't bother to use a trailing underscore for unused entities, since it would add silly changes to the subsequent PRs. Happy to do that if that's preferable.

nbp: I am fine with the current changes.

bnjbvr:
Adressed nits; waiting for CI to be green before merging.

Issue #796, opened 2019-06-13
EBBs aren't working out all that well


lars-t-hansen:
Cranelift uses extended basic blocks (EBBs) for its flow graphs: these are essentially single-entry multiple-exit linear sequences of instructions, they can be composed of what would normally be multiple basic blocks (BBs), namely, they are traces of BBs through subtrees of the BB flow graph.  BBs are "classical" but EBBs have some theoretical advantages in that there are fewer of them (so less overhead of time and space) and larger (so some non-global optimizations have greater scope to work in) and faster (computing the immediate dominator of the non-first BB of an EBB is trivial).

However, EBBs have some disadvantages.  As BBs are the classical representation of flow graphs, most of the literature uses BBs, and to use classical algorithms directly on EBBs we must perform a translation of the algorithm, or we must reconstruct the BB graph from the EBBs when that translation is not possible.  Furthermore, EBBs have a dicey relationship with loops: an EBB can be partly in a loop and partly outside a loop (if a side exit from the EBB is the back edge), so a loop algorithm must either split the EBB at the exit (doable but not currently done, AIUI) or risk being unsound (probably the current situation).

It turns out that several passes of Cranelift reconstruct the BB graph.  Supposedly (I have this from @sunfishcode) the SSA construction algorithm does this, and my experimental live range splitter must do it, because it needs to reconstruct the SSA after splitting.

The evidence is thus mounting that EBBs are not the best choice for the flow graphs and that we should consider moving to BBs.  This would probably be a large undertaking.

julian-seward1:
You could argue that CL spends most of its time and space dealing with instructions, not ebbs, so the additional overheads of this might be relatively modest.

> This would probably be a large undertaking.

Yes, but to some extent it would be mechanical.  We'd basically go looking for code fragments that say `if insn.is_exit && !insn.is_last_in_ebb`, replace with `if false`, and fold out dead control flow and representation accordingly.

We'd also have to change conditional jump instructions to specify both branch targets, I guess, and add a preferred-fallthrough indicator so we can make explicit the layout decisions that are implicit in the ebb scheme.

nbp:
> to use classical algorithms directly on EBBs we must perform a translation of the algorithm

I agree with this claim, and the overhead is a mental overhead that we have to cover all the cases.
From experience dealing with a Range Analysis which had many corner cases, I would recommend against the increasing complexity of the compiler, as these would later lead to security issues.

[...]

PR #804, opened 2019-06-26
Add verification phase to migrate from EBB to BB.

nbp:
This patch adds a verification executed by `verify_if` which checks that EBB are ending with branches with no interleaved instructions. These checks are enabled only when the environment variable `CRANELIFT_BB` is set.

While checking for branches ending an EBB is not the same as a BB, such as conditional branch + terminator, this is a first step towards moving to BB.

### DIFF
--- a/cranelift-codegen/src/verifier/mod.rs
+++ b/cranelift-codegen/src/verifier/mod.rs
@@ -267,6 +267,8 @@ struct Verifier<'a> {
     expected_cfg: ControlFlowGraph,
     expected_domtree: DominatorTree,
     isa: Option<&'a dyn TargetIsa>,
+    // To be removed when #796 is completed.
+    verify_encodable_as_bb: bool,
 }
 
 impl<'a> Verifier<'a> {
@@ -278,6 +280,7 @@ impl<'a> Verifier<'a> {
             expected_cfg,
             expected_domtree,
             isa: fisa.isa,
+            verify_encodable_as_bb: std::env::var("CRANELIFT_BB").is_ok(),
         }
     }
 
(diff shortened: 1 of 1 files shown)


cranelift-codegen/src/verifier/mod.rs
```diff
@@ -468,6 +468,41 @@ impl<'a> Verifier<'a> {
...
 
+    /// Check that the given EBB can be encoded as a BB, by checking that only
+    /// branching instructions are ending the EBB.
+    fn encodable_as_bb(
+        &self,
+        ebb: Ebb,
+        errors: &mut VerifierErrors,
+    ) -> VerifierStepResult<()> {
+        // Skip this verification if the environment variable is not set.
+        if let Err(_) = std::env::var("CRANELIFT_BB") {
```
bnjbvr: Maybe make this a property of the Verifier itself, so it's not doing a syscall per ebb?

nbp:
This verification pass is part of the work needed for solving the #796 issue.

cranelift-codegen/src/verifier/mod.rs
```diff
@@ -267,6 +267,8 @@ struct Verifier<'a> {
     expected_cfg: ControlFlowGraph,
     expected_domtree: DominatorTree,
     isa: Option<&'a dyn TargetIsa>,
+    // To be rmeoved when #796 is completed.
```
lars-t-hansen: Typo, "rmeoved"

cranelift-codegen/src/verifier/mod.rs
```diff
@@ -468,6 +471,35 @@ impl<'a> Verifier<'a> {
...
+            return Ok(());
+        };
+
+        let dfg = &self.func.dfg;
+        let inst_iter = self.func.layout.ebb_insts(ebb);
+        // Skip non-branching instructions.
+        let inst_iter = inst_iter.skip_while(|&inst| !dfg[inst].opcode().is_branch());
+        // Skip branching instructions.
+        let mut inst_iter = inst_iter.skip_while(|&inst| {
+            dfg[inst].opcode().is_branch() || dfg[inst].opcode().is_terminator()
```
lars-t-hansen: I think you should only accept at most one none-terminating branch and exactly one terminator -- that is pretty close to the final form we need.  The code above will accept multiple conditional exits.

In general even that test is too weak; we want to look for more specific patterns.  For example if the terminating instruction is BrTable you probably don't want to allow a conditional branch before it.  I actually suspect that we should only allow a conditional exit if the terminator is Fallthrough or Jump.  But it's possible that that is too hard to implement for the first iteration, so I won't insist on it.

cranelift-codegen/src/verifier/mod.rs
julian-seward1: Reading env vars doesn't involve a trip into the kernel, at least on Linux .. pointers to them are pushed on the stack before main() is entered.

cranelift-codegen/src/verifier/mod.rs
nbp: I changed the second `skip_while` by a manual iteration relying on the fact that an iterator is expected to return `None` when it finished and continues to return `None` on follow-up calls.

cranelift-codegen/src/verifier/mod.rs
```diff
@@ -468,6 +471,57 @@ impl<'a> Verifier<'a> {
...
+        };
+
+        let dfg = &self.func.dfg;
+        let inst_iter = self.func.layout.ebb_insts(ebb);
+        // Skip non-branching instructions.
+        let mut inst_iter = inst_iter.skip_while(|&inst| !dfg[inst].opcode().is_branch());
+        let br = inst_iter.next();
+        let br_1 = inst_iter.next();
+        let br_2 = inst_iter.next();
+        match (br, br_1, br_2) {
```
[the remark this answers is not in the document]
nbp: I first took this approach of having multiple branches, and found it to be more confusing to edit and follow. Which is why I switched to this match approach where I first manipulate the iterator, and then only do comparisons.

Maybe I can add a comment next to  `(None, _, _)` that we got a block with no branches, such as `ebb0: return`.

cranelift-codegen/src/verifier/mod.rs
```diff
@@ -468,6 +471,57 @@ impl<'a> Verifier<'a> {
...
+                        errors,
+                        end,
+                        "return instruction found after a branching instructions in {}",
+                        ebb
+                    )
+                } else {
+                    Ok(())
+                }
+            }
+            (Some(_), Some(inst), Some(_)) => fatal!(
```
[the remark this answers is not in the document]
nbp: Yes, but here we only know that the first element of the tuple is a branch instruction. The fact of having any instruction after a branch which is not a branching terminator is enough. The second instruction is not a teminator, as it would not have more instructions after it.

bbouvier:
not sure why I was tagged here

PR #805, opened 2019-06-26
Mark the jump_table_entry Instruction as loading

bnjbvr:
The Cranelift instruction `jump_table_entry` loads the jump table entry based on a user-controlled index value that's passed. It's possible to have LICM hoist these instructions above a loop, since they're not considered dangerous in any way. The thing is that the jump table bounds check (`brif` as generated/optimized) is not tied in any way to the `jump_table_entry` instruction, so it's not hoisted at the same time as the jump_table_entry. As a matter of fact, the entry loading may be hoisted (with the user controlled value) before the bounds check is carried out.

This PR marks "jump_table_entry" instructions as loading instructions, since they actually load from memory based on a user-controlled index, thus LICM considers them more dangerous.

### DIFF
--- a/cranelift-codegen/meta-python/base/instructions.py
+++ b/cranelift-codegen/meta-python/base/instructions.py
@@ -164,7 +164,7 @@ jump_table_entry = Instruction(
     Currently, the only type supported is entries which are relative to the
     base of the jump table.
     """,
-    ins=(x, addr, Size, JT), outs=entry)
+    ins=(x, addr, Size, JT), outs=entry, can_load=True)
 
 jump_table_base = Instruction(
     'jump_table_base', r"""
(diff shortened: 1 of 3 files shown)


bnjbvr:
> a control dependence is not tracked correctly

I'd agree with you if `brif` did return the tested value in the case where it doesn't branch, and if this value were used by the `jump_table_entry` instruction (that could be another fix). As it is now, there's no way Cranelift can understand there's a data dependency between the twos.

lars-t-hansen:
That's why I wrote "control dependency" and not "data dependency".

Consider user code along the lines of `if cond { something_that_traps }`.  The trapping instruction is control dependent on the condition even if the condition and the trapping instruction share no operands.  Cranelift knows not to hoist the trapping instruction above the condition because it is flagged as trapping, not because it understands the dependence.  In the same way, it now knows not to hoist the jump table load because it is flagged as loading (hence possibly trapping, though in this case it won't), not because it understands the dependence.  For the limited code reorganization we now do, this is probably OK, and I'm not suggesting we do something more elaborate; but we have to be careful.

PR #811, opened 2019-06-28
allow module environment to parse name section

data-pup:
This allows a `ModuleEnvironment` trait object to parse the name section of a wasm binary!

### DIFF
--- a/cranelift-wasm/src/environ/dummy.rs
+++ b/cranelift-wasm/src/environ/dummy.rs
@@ -17,7 +17,7 @@ use cranelift_codegen::ir::immediates::{Offset32, Uimm64};
 use cranelift_codegen::ir::types::*;
 use cranelift_codegen::ir::{self, InstBuilder};
 use cranelift_codegen::isa::TargetFrontendConfig;
-use cranelift_entity::{EntityRef, PrimaryMap};
+use cranelift_entity::{EntityRef, PrimaryMap, SecondaryMap};
 use std::boxed::Box;
 use std::string::String;
 use std::vec::Vec;
@@ -124,6 +124,9 @@ pub struct DummyEnvironment {
 
     /// Instructs to collect debug data during translation.
     debug_info: bool,
+
+    /// Function names.
+    function_names: SecondaryMap<FuncIndex, String>,
 }
 
 impl DummyEnvironment {
@@ -135,6 +138,7 @@ impl DummyEnvironment {
             func_bytecode_sizes: Vec::new(),
             return_mode,
             debug_info,
+            function_names: SecondaryMap::new(),
         }
     }
 
@@ -151,6 +155,12 @@ impl DummyEnvironment {
     /// Return the number of imported functions within this `DummyEnvironment`.
     pub fn get_num_func_imports(&self) -> usize {
         self.info.imported_funcs.len()
+    }
+
+    /// Return the name of the function, if a name for the function with
+    /// the corresponding index exists.
+    pub fn get_func_name(&self, func_index: FuncIndex) -> Option<&str> {
+        self.function_names.get(func_index).map(String::as_ref)
     }
 }
 
@@ -537,6 +547,11 @@ impl<'data> ModuleEnvironment<'data> for DummyEnvironment {
         };
         self.func_bytecode_sizes.push(body_bytes.len());
         self.info.function_bodies.push(func);
+        Ok(())
+    }
+
+    fn declare_func_name(&mut self, func_index: FuncIndex, name: &'data str) -> WasmResult<()> {
+        self.function_names[func_index] = String::from(name);
         Ok(())
     }
 }
--- a/cranelift-wasm/src/environ/spec.rs
+++ b/cranelift-wasm/src/environ/spec.rs
@@ -468,6 +468,14 @@ pub trait ModuleEnvironment<'data> {
         data: &'data [u8],
     ) -> WasmResult<()>;
 
+    /// Declares the name of a function to the environment.
+    ///
+    /// By default this does nothing, but implementations can use this to read
+    /// the function name subsection of the custom name section if desired.
+    fn declare_func_name(&mut self, _func_index: FuncIndex, _name: &'data str) -> WasmResult<()> {
+        Ok(())
+    }
+
     /// Indicates that a custom section has been found in the wasm file
     fn custom_section(&mut self, name: &'data str, data: &'data [u8]) -> WasmResult<()> {
         drop((name, data));
--- a/cranelift-wasm/src/module_translator.rs
+++ b/cranelift-wasm/src/module_translator.rs
@@ -4,10 +4,10 @@ use crate::environ::{ModuleEnvironment, WasmError, WasmResult};
 use crate::sections_translator::{
     parse_code_section, parse_data_section, parse_element_section, parse_export_section,
     parse_function_section, parse_global_section, parse_import_section, parse_memory_section,
-    parse_start_section, parse_table_section, parse_type_section,
+    parse_name_section, parse_start_section, parse_table_section, parse_type_section,
 };
 use cranelift_codegen::timing;
-use wasmparser::{ModuleReader, SectionCode};
+use wasmparser::{CustomSectionKind, ModuleReader, SectionCode};
 
 /// Translate a sequence of bytes forming a valid Wasm binary into a list of valid Cranelift IR
 /// [`Function`](cranelift_codegen::ir::Function).
@@ -81,6 +81,14 @@ pub fn translate_module<'data>(
                     message: "don't know how to handle the data count section yet",
                     offset: reader.current_position(),
                 });
+            }
+
+            SectionCode::Custom {
+                kind: CustomSectionKind::Name,
+                name: _,
+            } => {
+                let names = section.get_name_section_reader()?;
+                parse_name_section(names, environ)?;
             }
 
             SectionCode::Custom { name, kind: _ } => {
--- a/cranelift-wasm/src/sections_translator.rs
+++ b/cranelift-wasm/src/sections_translator.rs
@@ -12,7 +12,7 @@ use crate::translation_utils::{
     tabletype_to_type, type_to_type, FuncIndex, Global, GlobalIndex, GlobalInit, Memory,
     MemoryIndex, SignatureIndex, Table, TableElementType, TableIndex,
 };
-use crate::wasm_unsupported;
+use crate::{wasm_unsupported, HashMap};
 use core::convert::TryFrom;
 use cranelift_codegen::ir::{self, AbiParam, Signature};
 use cranelift_entity::EntityRef;
@@ -21,8 +21,8 @@ use wasmparser::{
     self, CodeSectionReader, Data, DataKind, DataSectionReader, Element, ElementKind,
     ElementSectionReader, Export, ExportSectionReader, ExternalKind, FuncType,
     FunctionSectionReader, GlobalSectionReader, GlobalType, ImportSectionEntryType,
-    ImportSectionReader, MemorySectionReader, MemoryType, Operator, TableSectionReader,
-    TypeSectionReader,
+    ImportSectionReader, MemorySectionReader, MemoryType, NameSectionReader, Naming, NamingReader,
+    Operator, TableSectionReader, TypeSectionReader,
 };
 
 /// Parses the Type section of the wasm module.
@@ -350,4 +350,48 @@ pub fn parse_data_section<'data>(
     }
 
     Ok(())
+}
+
+/// Parses the Name section of the wasm module.
+pub fn parse_name_section<'data>(
+    mut names: NameSectionReader<'data>,
+    environ: &mut dyn ModuleEnvironment<'data>,
+) -> WasmResult<()> {
+    while let Ok(subsection) = names.read() {
+        match subsection {
+            wasmparser::Name::Function(function_subsection) => {
+                if let Some(function_names) = function_subsection
+                    .get_map()
+                    .ok()
+                    .and_then(parse_function_name_subsection)
+                {
+                    for (index, name) in function_names {
+                        environ.declare_func_name(index, name)?;
+                    }
+                }
+                return Ok(());
+            }
+            wasmparser::Name::Local(_) | wasmparser::Name::Module(_) => {}
+        };
+    }
+    Ok(())
+}
+
+fn parse_function_name_subsection<'data>(
+    mut naming_reader: NamingReader<'data>,
+) -> Option<HashMap<FuncIndex, &str>> {
+    let mut function_names = HashMap::new();
+    for _ in 0..naming_reader.get_count() {
+        let Naming { index, name } = naming_reader.read().ok()?;
+        if function_names
+            .insert(FuncIndex::from_u32(index), name)
+            .is_some()
+        {
+            // If the function index has been previously seen, then we
+            // break out of the loop and early return `None`, because these
+            // should be unique.
+            return None;
+        }
+    }
+    return Some(function_names);
 }
--- a/cranelift-wasm/tests/wasm_testsuite.rs
+++ b/cranelift-wasm/tests/wasm_testsuite.rs
@@ -2,7 +2,7 @@ use cranelift_codegen::isa;
 use cranelift_codegen::print_errors::pretty_verifier_error;
 use cranelift_codegen::settings::{self, Flags};
 use cranelift_codegen::verifier;
-use cranelift_wasm::{translate_module, DummyEnvironment, ReturnMode};
+use cranelift_wasm::{translate_module, DummyEnvironment, FuncIndex, ReturnMode};
 use std::fs;
 use std::fs::File;
 use std::io;
@@ -10,7 +10,7 @@ use std::io::prelude::*;
 use std::path::Path;
 use std::str::FromStr;
 use target_lexicon::triple;
-use wabt::{wat2wasm_with_features, Features};
+use wabt::{wat2wasm_with_features, Features, Wat2Wasm};
 
 #[test]
 fn testsuite() {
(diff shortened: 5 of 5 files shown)


cranelift-wasm/src/module_translator.rs
```diff
@@ -137,6 +137,16 @@ pub fn translate_module<'data>(
...
+
+        if reader.eof() {
+            return Ok(());
+        }
+        section = reader.read()?;
+    }
+
+    if let SectionCode::Custom { name: "name", .. } = section.code {
+        let names = section.get_name_section_reader()?;
+        parse_name_section(names, environ)?;
```
data-pup: The [wasm specification](https://webassembly.github.io/spec/core/appendix/custom.html#name-section) mentions that the name section should come after the data section, so this block is placed here.

cranelift-wasm/src/environ/spec.rs
```diff
@@ -378,4 +378,10 @@ pub trait ModuleEnvironment<'data> {
         offset: usize,
         data: &'data [u8],
     );
+
+    /// Declares the name of a function.
+    ///
+    /// By default this does nothing, but implementations can use this to read
+    /// the name section if desired.
+    fn declare_func_name(&self, _func_index: FuncIndex, _name: &'data str) {}
```
data-pup: By default this does nothing, since the name section is an optional custom section.

cranelift-wasm/src/module_translator.rs
```diff
@@ -137,6 +137,16 @@ pub fn translate_module<'data>(
...
         let data = section.get_data_section_reader()?;
         parse_data_section(data, environ)?;
+
+        if reader.eof() {
+            return Ok(());
+        }
+        section = reader.read()?;
+    }
+
+    if let SectionCode::Custom { name: "name", .. } = section.code {
```
bnjbvr: Can you match on `kind: CustomSectionKind::Name` instead of `name: "name"`?

cranelift-wasm/src/sections_translator.rs
```diff
@@ -338,3 +338,23 @@ pub fn parse_data_section<'data>(
...
 }
+
+/// Parses the Name section of the wasm module.
+pub fn parse_name_section<'data>(
+    names: NameSectionReader<'data>,
+    environ: &mut dyn ModuleEnvironment<'data>,
+) -> WasmResult<()> {
+    for subsection in names {
+        let function_names = match subsection? {
+            wasmparser::Name::Function(f) => f,
```
bnjbvr: Do you wish to handle the module name subsection too, or would you prefer that we did it in a subsequent PR?

cranelift-wasm/src/module_translator.rs
data-pup: Fixed this, good eye!

cranelift-wasm/src/sections_translator.rs
data-pup: I went ahead and added some more changes to handle both the module name and local name subsections, since this PR isn't too large to begin with :slightly_smiling_face:

cranelift-wasm/src/environ/spec.rs
```diff
@@ -378,4 +379,28 @@ pub trait ModuleEnvironment<'data> {
         offset: usize,
         data: &'data [u8],
     );
+
+    /// Declares the name of a module to the environment.
+    ///
+    /// By default this does nothing, but implementations can use this to read
+    /// the module name subsection of the custom name section if desired.
+    fn declare_module_name(&self, _name: &'data str) {}
```
data-pup: Added a method for declaring the module name here, and a method for declaring local names below.

cranelift-wasm/src/module_translator.rs
```diff
@@ -137,6 +137,20 @@ pub fn translate_module<'data>(
...
+
+        if reader.eof() {
+            return Ok(());
+        }
+        section = reader.read()?;
+    }
+
+    if let SectionCode::Custom {
+        kind: CustomSectionKind::Name,
+        ..
```
data-pup: matching on `kind` rather than `name` now.

cranelift-wasm/src/sections_translator.rs
```diff
@@ -338,3 +338,45 @@ pub fn parse_data_section<'data>(
...
+                    for _ in 0..naming_reader.get_count() {
+                        let Naming {
+                            index: local_index,
+                            name,
+                        } = naming_reader.read()?;
+                        environ.declare_local_name(
+                            FuncIndex::from_u32(func_index),
+                            LocalIndex::from_u32(local_index),
+                            name,
+                        );
```
data-pup: Local names are handled in this branch now, and the module name is handled above. Listing these variants' respective `match` arms in ascending order according to their subsection id's in the [wasm spec](https://webassembly.github.io/spec/core/appendix/custom.html#subsections).

cranelift-wasm/src/translation_utils.rs
```diff
@@ -49,6 +49,11 @@ entity_impl!(MemoryIndex);
 pub struct SignatureIndex(u32);
 entity_impl!(SignatureIndex);
 
+/// Index type of local variable inside the WebAssembly module.
+#[derive(Copy, Clone, PartialEq, Eq, Hash, PartialOrd, Ord, Debug)]
+pub struct LocalIndex(u32);
+entity_impl!(LocalIndex);
```
data-pup: This newtype is something I'm slightly hesitant about adding, but it seemed like it was a more consistent choice to have `ModuleEnvironment<'_>::declare_local_name` use a `LocalIndex` newtype rather than a bare `u32` for the local index. Thoughts?

data-pup: **note:**  I'm split the local subsection work out into a separate commit, so if that is overkill I'm happy to revert back to `10f990c` and just handle function names and the module name.

bnjbvr: I think it's nicer to have a small entity type for this, because this documents the `declare_local_name` method better and makes sure we knowingly pass a `LocalIndex` there.

cranelift-wasm/src/sections_translator.rs
```diff
@@ -357,7 +357,26 @@ pub fn parse_name_section<'data>(
                     environ.declare_func_name(FuncIndex::from_u32(index), name);
                 }
             }
-            _ => continue,
+            wasmparser::Name::Local(local_names) => {
+                let mut func_local_reader = local_names.get_function_local_reader()?;
+                for _ in 0..func_local_reader.get_count() {
+                    let func_local_names = func_local_reader.read()?;
+                    let func_index = func_local_names.func_index;
```
bnjbvr: The [wasm spec](https://webassembly.github.io/spec/core/appendix/custom.html#name-maps) says that the `func_index` here (and below `local_index`) should be uniquely defined, i.e. they can't be used twice. Is this done by wasm-parser, or do we need to add our own checks here?

bnjbvr: cc @yurydelendik who might have an answer here. I've looked very quickly at wasmparser.rs, and I don't think it does it. It could also be that the spec is outdated.

yurydelendik: The wasmparser does not do validation of name section, and in most of the cases "like any custom section they may be ignored". Expect the name section to be invalid at any point, e.g. old version of name section format or non-unique names. There is no defined behavior for invalid or incomplete data in the section (possible behaviors are: reject entire section, or recover and use valid portions).

FWIW I know that only the function names entries are used at this point (e.g. LLVM or binaryen)

data-pup:
Sorry about letting this simmer for so long, was out of town! Cycling back to this now, and taking care of rebasing etc.

I wanted to add some test coverage to this, and ended up bumping into a small issue in `module_translator.rs` that I'm currently fixing. Regarding the discussion about validation above:

> Is [validation] done by wasm-parser, or do we need to add our own checks here?

I will add some checks to this so that an error is returned if the name section is not valid. Putting that responsibility on the struct that implements the module environment trait didn't sound very user friendly :sweat:

bnjbvr:
> Sorry about letting this simmer for so long, was out of town! Cycling back to this now, and taking care of rebasing etc.

No worries, take care! And thanks for getting back to this.

cranelift-wasm/tests/wasm_testsuite.rs
```diff
@@ -45,6 +45,35 @@ fn use_fallthrough_return() {
     );
 }
 
+#[test]
+fn use_name_section() {
```
data-pup: Added a test to check that the name section is parsed, and that module/functions/local names are declared. This meant adding a small test fixture, `wasmtests/names.wat`

cranelift-wasm/tests/wasm_testsuite.rs
```diff
@@ -45,6 +45,35 @@ fn use_fallthrough_return() {
...
+    assert_eq!(
+        dummy_environ.get_func_name(FuncIndex::from_u32(0)).unwrap(),
+        "func_name"
+    );
+    assert_eq!(
+        dummy_environ
+            .get_local_name(FuncIndex::from_u32(0), LocalIndex::from_u32(0))
+            .unwrap(),
+        "loc_name"
+    );
```
data-pup: :point_up: wouldn't mind dividing these 3 assertions out into separate tests if needed, but that felt a little noisy for this file, whose main focus is the `testsuite` function.

wasmtests/names.wat
```diff
@@ -0,0 +1,5 @@
+(module $module_name
+  (func $func_name (local $loc_name i32)
+  )
```
data-pup: small wat module that will have a value in each name subsection when compiled to wasm.

cranelift-wasm/src/sections_translator.rs
```diff
@@ -338,3 +339,65 @@ pub fn parse_data_section<'data>(
...
+                let module_name = module_name.get_name()?;
+                environ.declare_module_name(module_name);
+            }
+            wasmparser::Name::Function(function_names) => {
+                let mut naming_reader = function_names.get_map()?;
+                let mut previous_indices: HashSet<u32> = HashSet::new();
+                let mut previous_names: HashSet<&str> = HashSet::new();
+                for _ in 0..naming_reader.get_count() {
+                    let Naming { index, name } = naming_reader.read()?;
+                    if previous_indices.insert(index) && previous_names.insert(name) {
```
data-pup: Here's where the validation previously discussed happens! After initialising two empty hash sets for names/indices, we check that each new `Naming` does not include a name/index that has been seen before. (i.e., `insert` returns true)

data-pup: Tangentially, I hoped to initialise these `previous_*` sets via `HashSet::with_capacity` but `reader.get_count` returns a `usize`. Converting would mean using `try_into().unwrap()`, and introducing the potential for error didn't seem worth the small perf gain.

cranelift-wasm/src/module_translator.rs
```diff
@@ -127,7 +127,6 @@ pub fn translate_module<'data>(
         let code = section.get_code_section_reader()?;
         parse_code_section(code, environ)?;
 
-        reader.skip_custom_sections()?;
```
data-pup: While adding the test cases below, noticed that this line causes us to skip right past the name section if there isn't a data section in the binary.

cranelift-wasm/src/environ/dummy.rs
```diff
@@ -152,6 +165,25 @@ impl DummyEnvironment {
...
+    pub fn get_func_name(&self, func_index: FuncIndex) -> Option<&str> {
+        self.function_names.get(&func_index).map(String::as_ref)
+    }
+
+    /// Return the name of the local variable, if a name for the local with
+    /// the corresponding indices exists.
+    pub fn get_local_name(&self, func_index: FuncIndex, local_index: LocalIndex) -> Option<&str> {
+        self.local_names
+            .get(&(func_index, local_index))
+            .map(String::as_ref)
```
data-pup: Tried `.get(..).as_ref()` in these methods and was running into type errors bc of `Option<&String>` rather than `Option<&str>`. Not sure if I'm forgetting a better way to do this, so I figured these `String::as_ref` calls were worth pointing out :slightly_smiling_face:

cranelift-wasm/src/environ/dummy.rs
```diff
@@ -124,6 +125,15 @@ pub struct DummyEnvironment {
...
     debug_info: bool,
+
+    /// The name of the module.
+    module_name: Option<String>,
+
+    /// Function names.
+    function_names: HashMap<FuncIndex, String>,
+
+    /// Local variable names.
+    local_names: HashMap<(FuncIndex, LocalIndex), String>,
```
data-pup: added these little members to facilitate the `use_name_section` test, hopefully that isn't too much to add?

data-pup:
@bnjbvr this is ready for another round of review, when you are ready :slightly_smiling_face:

bnjbvr:
As an FYI: I am still a bit swamped by my review load. If anybody wants to take over the review (cc @yurydelendik), feel free to do so! Otherwise I'll get back to this in a few days, worst case start of next week.

data-pup:
> As an FYI: I am still a bit swamped by my review load.

That's totally okay! I know review bandwidth is a scarce resource as a maintainer. Happy to wait up on this, and resolve any merge conflicts if they come up in the meantime. :heart_decoration:

yurydelendik:
@data-pup I'm working on the reading of name section but in wasmtime (https://github.com/yurydelendik/wasmtime/tree/fake-dwarf) Can you explain the use case that prompts name section translation upfront? Can we parse on demand (e.g. when trap happens)?

Not knowing a use case:

- the single method is sufficient in `ModuleEnvironment` (instead of three)?
- currently wasm translation fails on invalid name section -- it needs to be ignored or data repaired.

data-pup:
Hi @ yurydelendik! Thanks for taking a look at this :slightly_smiling_face: 

re: usecase, I was aiming to support fastly/lucet#152. When translating the module, it would be nice to have access to the information in the name section for debugging.

> the single method is sufficient in ModuleEnvironment (instead of three)?
> currently wasm translation fails on invalid name section -- it needs to be ignored or data repaired.

One way we could restructure this would be something like this:

```rust
pub trait ModuleEnvironment<'data> {
    // ...
    fn declare_names(
        &mut self, 
        _module_name: Option<String>,
        _function_names: HashMap<FuncIndex, String>,
        _local_names: HashMap<(FuncIndex, LocalIndex), String>
   ) {}
}
```

This could be called after the different name subsections had been parsed, or skipped over if an error was encountered. Do you think that this would be more ideal?

data-pup:
Ah, and regarding this point:

> Can we parse on demand (e.g. when trap happens)?

If we were to parse on demand, what would that look like? That might require reworking this a fair bit, and/or starting over, but I'm not quite familiar enough with all of cranelift's internals to know where I would start with that.

yurydelendik:
> I was aiming to support fastly/lucet#152.

Looking requirements at the above issue, it only expects symbol names for the generated functions. The wasm name sections can be incomplete or contain non-unique names, so it shall be used only as hints for symbol names.

A cranelift data structure need to be more generic than wasm name section and the latter shall not dictate how to designed the former, e.g. maybe we don't need local names since it will not be used.

data-pup:
@yurydelendik This should be ready for another round of review now :slightly_smiling_face:

Regarding previous conversation, I removed the methods for declaring module and local names, and updated the `sections_translator::parse_name_section` function so that it will not return an error in the event of parse errors. Note that this still does return a `WasmResult<()>` so that an error returned by `declare_func_name` isn't swallowed. (see https://github.com/CraneStation/cranelift/pull/886#pullrequestreview-271670122) 

If an error does occur while parsing, the name section will be ignored.

data-pup:
@yurydelendik ping?

#898 might also be looking at in regards to this PR, that changes some of the code for how a module is parsed relative to custom sections :slightly_smiling_face:

cranelift-wasm/src/sections_translator.rs
yurydelendik: >> Do you wish to handle the module name subsection too

> more changes to handle both the module name and local name subsections

While local name subsection is wasm specific (and I think not really useful in general case), not sure how useful module name will be.

data-pup: Makes sense! At this point that code has been removed, and function names are the only information parsed out. :slightly_smiling_face:

cranelift-wasm/tests/wasm_testsuite.rs
```diff
@@ -45,6 +45,28 @@ fn use_fallthrough_return() {
     );
 }
 
+#[test]
+fn use_name_section() {
+    let wat = read_file(Path::new("../wasmtests/names.wat")).expect("error reading wat file");
```
bnjbvr: Light suggestion, feel free to dismiss: since the names.wat test is so small, and there's now test information that's scattered (in both names.wat and here, the "$func_name" name), could we just inline the content of names.wat here instead?

cranelift-wasm/src/environ/dummy.rs
```diff
@@ -124,6 +125,9 @@ pub struct DummyEnvironment {
 
     /// Instructs to collect debug data during translation.
     debug_info: bool,
+
+    /// Function names.
+    function_names: HashMap<FuncIndex, String>,
```
bnjbvr: light suggestion: since FuncIndex is an EntityRef, could this be a SecondaryMap instead, for more packing?

cranelift-wasm/src/sections_translator.rs
```diff
@@ -351,3 +351,46 @@ pub fn parse_data_section<'data>(
...
+            wasmparser::Name::Local(_) | wasmparser::Name::Module(_) => {}
+        };
+    }
+    Ok(())
+}
+
+fn parse_function_name_subsection<'data>(
+    mut naming_reader: NamingReader<'data>,
+) -> Option<HashMap<FuncIndex, &str>> {
+    let mut function_names = HashMap::new();
```
bnjbvr: (could be a SecondaryMap here too)

cranelift-wasm/src/sections_translator.rs
```diff
@@ -351,3 +351,46 @@ pub fn parse_data_section<'data>(
...
+) -> WasmResult<()> {
+    while let Ok(subsection) = names.read() {
+        match subsection {
+            wasmparser::Name::Function(function_subsection) => {
+                if let Some(function_names) = function_subsection
+                    .get_map()
+                    .ok()
+                    .and_then(parse_function_name_subsection)
+                {
+                    for (index, name) in function_names {
```
bnjbvr: Since each subsection may appear at most once per spec, should we have a boolean that checks if this is the first time we've seen such a subsection, and ignore the following function name subsections if we've seen it before?

sunfishcode:
This looks in good shape, though as #898 has now landed, would you mind looking at the merge conflicts? It looks like there's a question about what to do about the trait interface. Should we call the `custom_section` hook for the names section? My rough sense is no, if cranelift-wasm is parsing the name section and calling the `declare_func_name` callback, the consumer doesn't also need to see the name section as a `custom_section`. But either way, we should document it :-).

data-pup:
Just wanted to check in on this issue and say that I will make some time to address the remaining comments and merge-conflicts early this week :slightly_smiling_face: Have been slightly preoccupied, but this should be finished shortly. Thanks for the reviews and comments! :heart_decoration:

cranelift-wasm/src/sections_translator.rs
data-pup: Gave this a whirl, but it feels like `HashMap` _is_ what we need here? I found that because `SecondaryMap` does not expose an `insert` method, there isn't an elegant way to force an early return when a duplicate name is found. Would also be happy to ignore the possibility of duplicate values altogether here, but that seemed like a design decision I'd leave up to you :slightly_smiling_face:

cranelift-wasm/src/sections_translator.rs
data-pup: Since other subsections are not being parsed at the moment, I decided to just use an early return here, since that would have the same effect without the need for an extra variable. If/when other subsections _are_ parsed in the future, figured that would be an easy enough change to make, and this reads cleanly for the time being.

cranelift-wasm/tests/wasm_testsuite.rs
```diff
@@ -31,17 +31,42 @@ fn testsuite() {
...
+    handle_module(data, &flags, ReturnMode::FallthroughReturn);
+}
+
+#[test]
+fn use_name_section() {
+    let wat = r#"
+        (module $module_name
+            (func $func_name (local $loc_name i32)
+            )
+        )"#;
```
data-pup: inlined `names.wat` here. I believe that the raw string literal is _technically_ overkill, but it seems like proper form since we're referencing this as a `[u8]` below in `Wat2Wasm::convert`.

data-pup:
Sorry to force push, but I have addressed the comments above and resolved merge conflicts that arose while this has been in review. :slightly_smiling_face:

cranelift-wasm/src/sections_translator.rs
bnjbvr: Sorry, I wasn't talking about other kinds of subsections, but if there were several names subsections in the same name section. That being said, we're not validating properly the rest of the sections anyway, so it could be something we'd improve in the future; let's not block this PR on this.

PR #813, opened 2019-07-01
BB-like trapcc

nbp:
This patch come as part of #796 work and based on top of #810 , which ensures that we are using a restricted form of EBB.

This patch fixes the `trapz` and `trapnz` legalization by generating basic-blocks which are following the verifier phase which got added in #804, and allow the following test cases to pass when executed with the `CRANELIFT_BB` environment variable:
- filetests/isa/x86/legalize-heaps.clif
- filetests/isa/x86/legalize-memory.clif
- filetests/isa/x86/legalize-tables.clif


cranelift-codegen/src/legalizer/mod.rs
```diff
@@ -124,7 +124,6 @@ pub fn legalize_function(func: &mut ir::Function, cfg: &mut ControlFlowGraph, is
 include!(concat!(env!("OUT_DIR"), "/legalizer.rs"));
 
 /// Custom expansion for conditional trap instructions.
-/// TODO: Add CFG support to the Python patterns so we won't have to do this.
```
bnjbvr: Please leave this comment, we could still add CFG support to the Rust patterns later.

cranelift-codegen/src/legalizer/mod.rs
```diff
@@ -149,29 +148,43 @@ fn expand_cond_trap(
...
     } else {
-        func.dfg.replace(inst).brz(arg, new_ebb, &[]);
+        func.dfg.replace(inst).brz(arg, new_ebb_resume, &[]);
     }
 
+    // Add jump instruction after the inverted branch
     let mut pos = FuncCursor::new(func).after_inst(inst);
     pos.use_srcloc(inst);
+    pos.ins().jump(new_ebb_trap, &[]);
+    // Insert the new label and the unconditional trap terminator.
```
bnjbvr: nit: here and below, can you add a \n before the comment so there's a visual hint that a new block starts here?

cranelift-codegen/src/legalizer/mod.rs
```diff
@@ -149,29 +148,43 @@ fn expand_cond_trap(
...
+    pos.insert_ebb(new_ebb_trap);
```
bnjbvr: Pre-existing: not sure if we should prefer `split_ebb` here. It seems that when `split_ebb` is used, we can then use `recompute_split_ebb` instead of `recompute_ebb`, which might do even less work. No strong preference here, current code is fine by me.

cranelift-codegen/src/legalizer/mod.rs
```diff
@@ -149,29 +148,43 @@ fn expand_cond_trap(
...
+    // Replace trap instruction by the inverted condition.
     if trapz {
-        func.dfg.replace(inst).brnz(arg, new_ebb, &[]);
+        func.dfg.replace(inst).brnz(arg, new_ebb_resume, &[]);
     } else {
-        func.dfg.replace(inst).brz(arg, new_ebb, &[]);
+        func.dfg.replace(inst).brz(arg, new_ebb_resume, &[]);
     }
 
+    // Add jump instruction after the inverted branch
```
bnjbvr: nit: please end comment with a dot.

cranelift-codegen/src/legalizer/mod.rs
```diff
@@ -149,29 +148,43 @@ fn expand_cond_trap(
...
     let mut pos = FuncCursor::new(func).after_inst(inst);
     pos.use_srcloc(inst);
+    pos.ins().jump(new_ebb_trap, &[]);
+    // Insert the new label and the unconditional trap terminator.
+    pos.insert_ebb(new_ebb_trap);
     pos.ins().trap(code);
-    pos.insert_ebb(new_ebb);
+    // Insert the new label and resume the the execution when the trap fails.
```
bnjbvr: nit: `the the`. I think it could even just be `and resume execution when`.

cranelift-codegen/src/legalizer/mod.rs
nbp: `recompute_split_ebb` seems to only exists as part of the dominator code computation. I will leave this code as-is as I do not yet fully understand how the flow graph is stored nor manipulated.

PR #816, opened 2019-07-02
BB like br table

nbp:
This patch come as part of #796 work and based on top of #813 , which ensures that we are using a restricted form of EBB.

This patch fixes the `br_table` legalizations by generating basic-blocks which are following the verifier phase which got added in #804, and allow the following test cases to pass when executed with the `CRANELIFT_BB` environment variable:

- filetests/isa/x86/legalize-br-table.clif
- filetests/legalizer/br_table_cond.clif


lars-t-hansen:
It looks like your other patch (for the verifier change) is also included here, can you remove that?

cranelift-codegen/src/legalizer/mod.rs
```diff
@@ -149,29 +148,43 @@ fn expand_cond_trap(
...
     //     trap
-    //   new_ebb:
     //
+    //   new_ebb_resume:
+    //     ..
     let old_ebb = func.layout.pp_ebb(inst);
-    let new_ebb = func.dfg.make_ebb();
+    let new_ebb_trap = func.dfg.make_ebb();
+    let new_ebb_resume = func.dfg.make_ebb();
+
```
lars-t-hansen: Should be able to use the ebb splitting method to create new_ebb_trap?  I confess I'm not yet an expert on how this code all works, so I'm not 100% sure, but it does seem so.

nbp: If there is an ebb splitting method, I do not know it yet. Where should I look for it?

nbp:
> It looks like your other patch (for the verifier change) is also included here, can you remove that?

Done.
It will cause some nightmare to properly handle these branches locally, but this should simplify the review process.

cranelift-codegen/src/legalizer/mod.rs
nbp: Looking at the code, it seems simpler and more coherent to use `insert_ebb` as the surrounding code is adding instruction. `split_ebb` sounds more adapted to breaking an existing block in 2, without changing any of the instruction.

Maybe I am wrong …

PR #818, opened 2019-07-02
Bump the required rust version to 1.35.

sunfishcode:
Rust 1.34 gets errors like this:

```
$ cargo +1.34.0 build --release
   Compiling cranelift-wasm v0.32.0 (cranelift/cranelift-wasm)
error[E0259]: the name `std` is defined multiple times
  |
  = note: `std` must be defined only once in the type namespace of this module

error: aborting due to previous error

For more information about this error, try `rustc --explain E0259`.
error: Could not compile `cranelift-wasm`.

To learn more, run the command again with --verbose.
```

Updating to Rust 1.35 fixes this.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,7 @@ language: rust
 rust:
     # The oldest version we currently support. See
     # CONTRIBUTING.md#rustc-version-support for details.
-    - 1.34.0
+    - 1.35.0
     - beta
     - nightly
 matrix:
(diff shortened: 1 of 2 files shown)


bnjbvr:
Is there a way to just specify "latest stable version" everywhere? (In text it's not a problem, but in CI could we do that?)

bjorn3:
Yes, but I believe firefox only upgrades rustc some time after a new stable.

nbp:
I think adding a `rust-toolchain` file at the root of the repository might be a way to do that, but I am not sure we want our CI to have silent automatic upgrades, as this could imply that changes which were green in a pull request could go red after merging.

I will suggest pinning the latest stable, and manually bump it from time to time.

Note, the `rust-toolchain` file will cause cargo to pull the corresponding version before building, and as such I am not sure how it might be possible to override it if one want to check with nightly, without modifying this file.

bnjbvr:
> this could imply that changes which were green in a pull request could go red after merging.

This is a general problem on Github sometimes [merge skew](https://bors.tech/essay/2017/02/02/pitch/) and that Bors knows how to attempt. That being said, agreed that since the stable branch is tier-1, it's probably better to keep control over what the version number is.

PR #820, opened 2019-07-03
Meta port 2: more instruction predicates + transform group mapping

bnjbvr:
This starts at 4336356 (part 1 to be merged soon)

Things that might help review:

- f69aae6b894c7c865cf6a6cb0d4af28fd86f0243 adds a mapping between global and local transform groups (= legalization transform groups) indices. That's because all the transform groups are stored within a single array (that's the "global" index into `shared_def.transform_groups`), while when they get used by an ISA, references to transform group refer to a per-ISA array (present in the per-ISA file, e.g. `legalize-x86.rs`). This mapping is used when generating the legalization files, and later when generating encodings (and mapping into the local legalization array when no encodings are found).
- the instruction predicates are unused as of now, but get used in the next commits.
- there's no way / no need to have predicates mixing AND and OR predicates, so I haven't implemented it, and I've made sure to panic when we try to do it.

### DIFF
--- a/cranelift-codegen/meta/src/cdsl/formats.rs
+++ b/cranelift-codegen/meta/src/cdsl/formats.rs
@@ -67,6 +67,20 @@ impl fmt::Display for InstructionFormat {
     }
 }
 
+impl InstructionFormat {
+    pub fn imm_by_name(&self, name: &'static str) -> &FormatField {
+        self.imm_fields
+            .iter()
+            .find(|&field| field.member == name)
+            .unwrap_or_else(|| {
+                panic!(
+                    "unexpected immediate field named {} in instruction format {}",
+                    name, self.name
+                )
+            })
+    }
+}
+
 pub struct InstructionFormatBuilder {
     name: &'static str,
     num_value_operands: usize,
@@ -198,6 +212,14 @@ impl FormatRegistry {
             .sig_to_index
             .get(&sig)
             .expect("unknown InstructionFormat; please define it in shared/formats.rs first")
+    }
+
+    pub fn by_name(&self, name: &str) -> InstructionFormatIndex {
+        self.map
+            .iter()
+            .find(|(_key, value)| value.name == name)
+            .unwrap_or_else(|| panic!("format with name '{}' doesn't exist", name))
+            .0
     }
 
     pub fn get(&self, index: InstructionFormatIndex) -> &InstructionFormat {
--- a/cranelift-codegen/meta/src/cdsl/settings.rs
+++ b/cranelift-codegen/meta/src/cdsl/settings.rs
@@ -148,6 +148,14 @@ impl SettingGroup {
         }
         panic!("Should have found bool setting by name.");
     }
+
+    pub fn predicate_by_name(&self, name: &'static str) -> SettingPredicateNumber {
+        self.predicates
+            .iter()
+            .find(|pred| pred.name == name)
+            .unwrap_or_else(|| panic!("unknown predicate {}", name))
+            .number
+    }
 }
 
 /// This is the basic information needed to track the specific parts of a setting when building
@@ -209,10 +217,12 @@ struct ProtoPredicate {
     node: PredicateNode,
 }
 
+pub type SettingPredicateNumber = u8;
+
 pub struct Predicate {
     pub name: &'static str,
     node: PredicateNode,
-    pub number: u8,
+    pub number: SettingPredicateNumber,
 }
 
 impl Predicate {
--- a/cranelift-codegen/meta/src/cdsl/ast.rs
+++ b/cranelift-codegen/meta/src/cdsl/ast.rs
@@ -440,7 +440,7 @@ impl Apply {
         format!("{}({})", self.inst.name, args)
     }
 
-    fn inst_predicate(
+    pub fn inst_predicate(
         &self,
         format_registry: &FormatRegistry,
         var_pool: &VarPool,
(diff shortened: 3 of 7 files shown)


cranelift-codegen/meta/src/cdsl/formats.rs
```diff
@@ -67,6 +67,20 @@ impl fmt::Display for InstructionFormat {
     }
 }
 
+impl InstructionFormat {
+    pub fn imm_by_name(&self, name: &'static str) -> &FormatField {
+        for field in &self.imm_fields {
+            if field.member == name {
+                return field;
```
nbp: nit: 
```rust
  self.imm_fields
      .iter()
      .find(|field| field.member == name)
      .unwrap_or_else(|| panic!(…))
```

cranelift-codegen/meta/src/cdsl/formats.rs
```diff
@@ -200,6 +214,14 @@ impl FormatRegistry {
             .expect("unknown InstructionFormat; please define it in shared/formats.rs first")
     }
 
+    pub fn by_name(&self, name: &str) -> InstructionFormatIndex {
+        self.map
+            .iter()
+            .find(|(_key, value)| value.name == name)
+            .unwrap()
```
nbp: nit: `unwrap_or_else(|| panic!("ahah, I am not the only one to assume that {} is present, but no!", name))`

cranelift-codegen/meta/src/cdsl/settings.rs
```diff
@@ -148,6 +148,14 @@ impl SettingGroup {
         }
         panic!("Should have found bool setting by name.");
     }
+
+    pub fn predicate_by_name(&self, name: &'static str) -> SettingPredicateNumber {
+        self.predicates
+            .iter()
+            .find(|pred| pred.name == name)
+            .expect(&format!("unknown predicate {}", name))
```
nbp: nit: use `unwrap_or_else(|| panic!(…))` to avoid computing the string at every call.

bnjbvr:
> These changes sounds fine to me, can you give examples of predicates, and transform groups in this PR, for the sake of adhoc documentation.

Sure!

# Legalization, transform groups, transforms

First, some transform groups. These were called "XFormGroup" in the Python code, and they are a group of ... Transform (née XForm). A "transform" is used in the context of legalization. It contains a source pattern (that involves a single CL instruction to be legalized) and a destination pattern: a single or several CL instructions that have the equivalent semantic of the source pattern.

Transforms can either be manually defined in the non-meta crate if they're too complicated: e.g. legalization of `call` instructions, in [meta we define it's a manual transform](https://github.com/CraneStation/cranelift/blob/master/cranelift-codegen/meta/src/shared/legalize.rs#L126) and in [non-meta we define it](https://github.com/CraneStation/cranelift/blob/master/cranelift-codegen/src/legalizer/call.rs#L13).

They can also be defined in the meta crate, if they don't involve the control flow graph. See for instance the legalization of `iadd` when the type is too big for the platform [here](https://github.com/CraneStation/cranelift/blob/master/cranelift-codegen/meta/src/shared/legalize.rs#L197-L206). What this says is that in the `narrow` transform group, whenever we reach legalization, if we identify the source pattern (first argument, here `a = iadd(x, y)`), we expand to the sequence as shown in the second (vector) argument, ie the destination pattern.

More interesting facts about legalization:

- first Cranelift tries to find an encoding in the encoding lists (ENCLISTS in generated source), and if it can't find any, it uses the associated legalization code == the transform group index, thus the transform group to be used.
- Transform groups are associated to a instructions' controlling type: the mapping is done in the meta crate, per ISA, e.g. [here](https://github.com/CraneStation/cranelift/blob/master/cranelift-codegen/meta/src/isa/x86/mod.rs#L33-L50). This mapping means that the group is used if no encodings for an instruction with the given controlling type is found. (Monomorphic means monomorphic instructions = no controlling type in the instruction, e.g. jump; default is used for all the types that are not specified, namely vector types at this time)
- transform groups can be "chained" with one another, say A is chained with B, if no valid transforms are found in A, then we look for a transform in B, until we exhaust the chain.

(Instructions with no valid encoding are sometimes called "ghost" instructions, which you might see in error messages, whenever an instruction should have a encoding but doesn't have any. Some instructions are valid ghost instructions, e.g. fallthrough doesn't generate any code.)

# Predicates

Predicates can be one of two kinds: instruction predicates (e.g. for an iadd_imm encoding, is the immediate within the range of 32 bits values?) or ISA predicates (e.g. is sse4 available, on an x86 machine?). For instance, in part 4 of this series of PR, we define an encoding for FP zero constants, which has an instruction predicate checking if the constant is zero:

```rust
    let f_unary_ieee32 = formats.get(formats.by_name("UnaryIeee32"));
    let is_zero_32_bit_float = InstructionPredicate::new_is_zero_32bit_float(f_unary_ieee32, "imm");
    e.enc32_instp(
        f32const,
        rec_f32imm_z.opcodes(vec![0x0f, 0x57]),
        is_zero_32_bit_float.clone(),
    );
```

(format is defined in `codegen/meta/src/shared/formats`, `imm` refers to the immediate float32 field in this format)

Predicates can be attached to encodings, or to recipes (their low-level counterpart).

Note that if an instruction predicate isn't satisfied, then the encoding list interpreter will move on to the next entry not guarded by this predicate, until it reaches the point when either it's found one encoding, or there are no more next entries for the CL instruction we'd like to encode; that's at this point that legalization using transform groups happens.

PR #822, opened 2019-07-03
BB-like manual legalization for x86 ISA

nbp:
This patch come as part of #796 work, which ensures that we are using a restricted form of EBB.

This patches (to be squashed) fixes the code produced by `srem`, `sdiv` and `fcvt_*` functions, by generating basic-blocks which are following the verifier phase which got added in #804, and allow the following test cases to pass when executed with the CRANELIFT_BB environment variable:

- ./filetests/wasm/i32-arith.clif
- ./filetests/wasm/i64-arith.clif
- ./filetests/wasm/f64-arith.clif
- ./filetests/wasm/f32-arith.clif
- ./filetests/regalloc/global-constraints.clif
- ./filetests/isa/x86/legalize-custom.clif
- ./filetests/isa/x86/legalize-div.clif
- ./filetests/isa/x86/legalize-div-traps.clif
- ./filetests/isa/x86/legalize-fcvt_from_usint-i16.clif


cranelift-codegen/src/isa/x86/enc_tables.rs
```diff
@@ -185,9 +188,11 @@ fn expand_sdivrem(
     // Start by checking for a -1 divisor which needs to be handled specially.
     let is_m1 = pos.ins().ifcmp_imm(y, -1);
     pos.ins().brif(IntCC::Equal, is_m1, minus_one, &[]);
+    pos.ins().jump(nominal, &[]);
```
bjorn3:
```suggestion
    pos.ins().fallthrough(nominal, &[]);
```

bnjbvr:
Just a very small request: can you please the commit message (and maybe this PR's title) because that's not "encoding tables" that are being tweaked here, but manual legalization functions (for fcvt_to_sint, etc.)? (All the functions starting with `expand_` are legalizations too complicated to be encoded in the meta-language and that are defined manually.)

nbp:
> Just a very small request: can you please the commit message […]

I intent to squash these commit into a single commit at the very end. In which case the current commit message should not matter much.

cranelift-codegen/src/isa/x86/enc_tables.rs
nbp: After discussing with @lars-t-hansen , we are actually going to replace all the `fallthrough` by `jump`, as the branch relaxation, which is the very last pass of the compiler is supposed to automagically do that for us.  I will revert this patch locally :/

PR #824, opened 2019-07-05
Moar simple preoptimizations

bnjbvr:
- first commit enables more usage of `_imm` instructions, for some which didn't have it.
- second commit allows folding of binary-with-immediates chains together, whenever it's possible to easily apply the instruction to the immediates.
- third commit pattern matches `(x << N) >> N` and transforms it into an (un)signed-extended move whenever possible.

I need to measure on benchmarks if this has an effect (so far I'm seeing mostly noise, so this is hard to tell).

### DIFF
--- a/cranelift-codegen/src/simple_preopt.rs
+++ b/cranelift-codegen/src/simple_preopt.rs
@@ -13,7 +13,7 @@ use crate::ir::{
     dfg::ValueDef,
     immediates,
     instructions::{Opcode, ValueList},
-    types::{I32, I64},
+    types::{I16, I32, I64, I8},
     DataFlowGraph, Ebb, Function, Inst, InstBuilder, InstructionData, Type, Value,
 };
 use crate::timing;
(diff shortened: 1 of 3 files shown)


bnjbvr:
Added a small commit that also folds `x + 0` into `x`, and for other simple arithmetic operations too. One could argue this should go in the preopt crate, but this is actually a consequence of folding binary operations into binary-imm operations in the first place. I've seen many cases of `x + 0` or `x | 0` in the wild.

bnjbvr:
Picking Lars as a reviewer, but feel free to pass to somebody else.

cranelift-codegen/src/simple_preopt.rs
```diff
@@ -450,76 +452,85 @@ fn do_divrem_transformation(divrem_info: &DivRemByConstInfo, pos: &mut FuncCurso
     }
 }
 
+#[inline]
+fn resolve_imm64_value(dfg: &DataFlowGraph, value: Value) -> Option<immediates::Imm64> {
+    if let ValueDef::Result(iconst_inst, _) = dfg.value_def(value) {
```
lars-t-hansen: Nit: "inst" or "candidate_inst" would be more reasonable since we don't yet know it's an Iconst.

cranelift-codegen/src/simple_preopt.rs
```diff
@@ -450,76 +452,85 @@ fn do_divrem_transformation(divrem_info: &DivRemByConstInfo, pos: &mut FuncCurso
...
+                };
+                let ty = pos.func.dfg.ctrl_typevar(inst);
+                pos.func
+                    .dfg
+                    .replace(inst)
+                    .BinaryImm(new_opcode, ty, imm, args[0]);
+            } else if let Some(imm) = resolve_imm64_value(&pos.func.dfg, args[0]) {
+                let new_opcode = match opcode {
+                    Opcode::Isub => Opcode::IrsubImm,
+                    _ => return,
```
lars-t-hansen: Could/should we also handle imm + x here or do we know that that form does not occur because it's normalized elsewhere?

cranelift-codegen/src/simple_preopt.rs
```diff
@@ -466,6 +466,56 @@ fn resolve_imm64_value(dfg: &DataFlowGraph, value: Value) -> Option<immediates::
...
+            arg: prev_arg,
+            imm: prev_imm,
+        } = &pos.func.dfg[arg_inst]
+        {
+            if imm != *prev_imm {
+                return false;
+            }
+
+            let dest_ty = pos.func.dfg.ctrl_typevar(inst);
+            if dest_ty != pos.func.dfg.ctrl_typevar(arg_inst) || !dest_ty.is_int() {
```
lars-t-hansen: Can dest_ty not be int here, given the conditions under which we're operating (namely, the opcode is a right shift)?

cranelift-codegen/src/simple_preopt.rs
bnjbvr: I'll try it, I don't have any idea if it's normalized elsewhere.

cranelift-codegen/src/simple_preopt.rs
bnjbvr: Yeah, SIMD concerns indeed. I'll let it as is until we have more SIMD features in.

cranelift-codegen/src/simple_preopt.rs
bnjbvr: It has strictly no effect on any of the embenchen benchmarks. But in AngryBots, it's effective for >500 iadd, >100 imul, and one bor. Tanks has the iadd pattern twice. Considering we have this match anyways so adding more caught opcodes should be ~ cheap, I'll add it. Good idea!

PR #828, opened 2019-07-08
Redundant-reload removal pass

julian-seward1:
…nt redundant-fill removal.

### DIFF
--- a/cranelift-codegen/meta/src/isa/x86/encodings.rs
+++ b/cranelift-codegen/meta/src/isa/x86/encodings.rs
@@ -311,6 +311,7 @@ pub fn define(
     let fdiv = shared.by_name("fdiv");
     let ffcmp = shared.by_name("ffcmp");
     let fill = shared.by_name("fill");
+    let fill_nop = shared.by_name("fill_nop");
     let floor = shared.by_name("floor");
     let fmul = shared.by_name("fmul");
     let fpromote = shared.by_name("fpromote");
@@ -423,7 +424,9 @@ pub fn define(
     let rec_fax = r.template("fax");
     let rec_fcmp = r.template("fcmp");
     let rec_fcscc = r.template("fcscc");
+    let rec_ffillnull = r.recipe("ffillnull");
     let rec_ffillSib32 = r.template("ffillSib32");
+    let rec_fillnull = r.recipe("fillnull");
     let rec_fillSib32 = r.template("fillSib32");
     let rec_fld = r.template("fld");
     let rec_fldDisp32 = r.template("fldDisp32");
@@ -864,6 +867,18 @@ pub fn define(
 
     e.enc_i32_i64(fill, rec_fillSib32.opcodes(vec![0x8b]));
     e.enc_i32_i64(regfill, rec_regfill32.opcodes(vec![0x8b]));
+
+    // No-op fills, created by late-stage redundant-fill removal.
+    for &ty in &[I64, I32, I16, I8] {
+        e.enc64_rec(fill_nop.bind(ty), rec_fillnull, 0);
+        e.enc32_rec(fill_nop.bind(ty), rec_fillnull, 0);
+    }
+    e.enc64_rec(fill_nop.bind(B1), rec_fillnull, 0);
+    e.enc32_rec(fill_nop.bind(B1), rec_fillnull, 0);
+    for &ty in &[F64, F32] {
+        e.enc64_rec(fill_nop.bind(ty), rec_ffillnull, 0);
+        e.enc32_rec(fill_nop.bind(ty), rec_ffillnull, 0);
+    }
 
     // Load 32 bits from `b1`, `i8` and `i16` spill slots. See `spill.b1` above.
 
--- a/cranelift-codegen/meta/src/shared/instructions.rs
+++ b/cranelift-codegen/meta/src/shared/instructions.rs
@@ -1163,6 +1163,24 @@ pub fn define(
         .can_load(true),
     );
 
+    ig.push(
+        Inst::new(
+            "fill_nop",
+            r#"
+        This is identical to :inst:`fill`, except it has no encoding, since
+        it is a no-op.
+
+        This instruction is created only during late-stage peephole
+        optimisation, after all registers and stack slots have been assigned.
+        It is used to replace :inst:`fill`s that have been identified as
+        redundant.
+        "#,
+        )
+        .operands_in(vec![x])
+        .operands_out(vec![a])
+        .can_load(true),
+    );
+
     let src = &operand("src", regunit);
     let dst = &operand("dst", regunit);
 
--- a/cranelift-codegen/src/context.rs
+++ b/cranelift-codegen/src/context.rs
@@ -22,6 +22,7 @@ use crate::licm::do_licm;
 use crate::loop_analysis::LoopAnalysis;
 use crate::nan_canonicalization::do_nan_canonicalization;
 use crate::postopt::do_postopt;
+use crate::redundant_reload_remover::RedundantReloadRemover;
 use crate::regalloc;
 use crate::result::CodegenResult;
 use crate::settings::{FlagsOrIsa, OptLevel};
@@ -49,6 +50,9 @@ pub struct Context {
 
     /// Loop analysis of `func`.
     pub loop_analysis: LoopAnalysis,
+
+    /// Redundant-reload remover context.
+    pub redundant_reload_remover: RedundantReloadRemover,
 }
 
 impl Context {
@@ -71,6 +75,7 @@ impl Context {
             domtree: DominatorTree::new(),
             regalloc: regalloc::Context::new(),
             loop_analysis: LoopAnalysis::new(),
+            redundant_reload_remover: RedundantReloadRemover::new(),
         }
     }
 
@@ -81,6 +86,7 @@ impl Context {
         self.domtree.clear();
         self.regalloc.clear();
         self.loop_analysis.clear();
+        self.redundant_reload_remover.clear();
     }
 
     /// Compile the function, and emit machine code into a `Vec<u8>`.
@@ -146,6 +152,7 @@ impl Context {
         self.regalloc(isa)?;
         self.prologue_epilogue(isa)?;
         if isa.flags().opt_level() == OptLevel::Best {
+            self.redundant_reload_remover(isa)?;
             self.shrink_instructions(isa)?;
         }
         self.relax_branches(isa)
@@ -315,6 +322,14 @@ impl Context {
         isa.prologue_epilogue(&mut self.func)?;
         self.verify_if(isa)?;
         self.verify_locations_if(isa)?;
+        Ok(())
+    }
+
+    /// Do redundant-reload removal after allocation of both registers and stack slots.
+    pub fn redundant_reload_remover(&mut self, isa: &dyn TargetIsa) -> CodegenResult<()> {
+        self.redundant_reload_remover
+            .run(isa, &mut self.func, &self.domtree);
+        self.verify_if(isa)?;
         Ok(())
     }
 
--- a/cranelift-codegen/src/redundant_reload_remover.rs
+++ b/cranelift-codegen/src/redundant_reload_remover.rs
@@ -1,0 +1,462 @@
+//! A late-stage redundant-reload remover, that runs after registers have been
+//! allocated and stack slots have been given specific offsets.
+
+use crate::cursor::{Cursor, EncCursor};
+use crate::dominator_tree::DominatorTree;
+use crate::ir::{
+    Ebb, Function, Inst, InstBuilder, InstructionData, Opcode, StackSlotData, Value, ValueLoc,
+};
+use crate::isa::{RegInfo, RegUnit, TargetIsa};
+
+use std::vec::Vec;
+
+/// The remover's state.
+pub struct RedundantReloadRemover {
+    // The running environment used to detect redundancies.  This is reinitialised at the start of
+    // each EBB.
+    env: AvailEnv,
+}
+
+impl RedundantReloadRemover {
+    /// Create a new remover state.
+    pub fn new() -> Self {
+        Self {
+            env: AvailEnv::new(0),
+        }
+    }
+
+    /// Clear the state of the remover.
+    pub fn clear(&mut self) {
+        self.env.map.clear();
+        self.env.size = 0;
+    }
+
+    // Initialise the state of the remover.
+    pub fn init(&mut self, size: usize) {
+        self.clear();
+        for _ in 0..size {
+            self.env.map.push(None);
+        }
+        self.env.size = size;
+    }
+
+    /// Run the remover.
+    pub fn run(&mut self, isa: &dyn TargetIsa, func: &mut Function, domtree: &DominatorTree) {
+        let mut ctx = Context {
+            cur: EncCursor::new(func, isa),
+            reginfo: isa.register_info(),
+            domtree: domtree,
+            state: &mut RedundantReloadRemover::new(),
+        };
+        ctx.run()
+    }
+}
+
+struct Context<'a> {
+    // Current instruction as well as reference to function and ISA.
+    cur: EncCursor<'a>,
+
+    // Cached ISA information.  We save it here to avoid frequent virtual function calls on the
+    // `TargetIsa` trait object.
+    reginfo: RegInfo,
+
+    // References to contextual data structures we need.
+    domtree: &'a DominatorTree,
+
+    // The running state
+    state: &'a mut RedundantReloadRemover,
+}
+
+impl<'a> Context<'a> {
+    fn run(&mut self) {
+        self.cleanup_func();
+    }
+}
+
+// ======== A description of the redundant-fill-removal algorithm ========
+//
+// The simplest incarnation simply detects fills from the same stack slot into the same register,
+// and removes them:
+//
+//  v1 = fill v0
+//  ...
+//  v2 = fill v0
+//
+// The second fill is redundant provided that
+//   (1) v1 and v2 have the same register, and
+//   (2) "..." overwrites neither that register nor v0's stack slot, and
+//   (3) the second fill is dominated by the first.
+// Condition (3) always holds since we're doing this only inside EBBs, but we have to check (1)
+// and (2).
+//
+// In this example, the instruction processing loop below creates an environment entry on
+// encountering "v1 = fill v0", and uses that entry to determine that "v2 = fill v0" is redundant.
+//
+// The loop must also, even in the simplest version, check all instructions for writes to
+// registers and stack slots, and invalidate environment entries accordingly.
+//
+// === First refinement -- implemented
+//
+// The simplest version only observes that a register and stack slot have the same value when it
+// finds a fill.  However, a spill also causes a register and stack slot to have the same value.
+// Hence this first refinement is to also record such spill-created equivalences in the
+// environment, in the hope of increasing our chances of proving later fills redundant.
+//
+// === Second refinement -- implemented
+//
+// For a fill, even if we can't find an exact match in the environment, finding a load of the same
+// spill slot into a different register is still a win, since we can convert the fill into a
+// reg-reg copy.
+//
+// === Other tricks that don't seem to help
+//
+// Tracking the effects of InstructionData::Copy in the environment for the case when both source
+// and destination are registers.  But that didn't cause any more fills to be removed or converted
+// to copies.  It's not clear why.
+//
+// It's also possible for a spill to be detected as redundant, if such a binding already appears
+// in the environment.  This was also implemented, but it was almost entirely ineffective, so was
+// removed again.
+
+// This describes a spill slot in the obvious way.  It also tracks the SSA value name associated
+// with the slot.
+#[derive(Clone, Copy)]
+struct SlotInfo {
+    value: Value,
+    offset: i32,
+    size: u32,
+}
+
+// Find out if the range `[offset, +size)` overlaps with the range in `si`.
+fn any_overlap(si: &SlotInfo, offset: i32, size: u32) -> bool {
+    let a_offset = si.offset as i64;
+    let a_size = si.size as i64;
+    let b_offset = offset as i64;
+    let b_size = size as i64;
+    let no_overlap = a_offset + a_size <= b_offset || b_offset + b_size <= a_offset;
+    !no_overlap
+}
+
+// Find, in `reginfo`, the register bank that `reg` lives in, and return the lower limit and size
+// of the bank.  This is so the caller can conveniently iterate over all `RegUnits` in the bank
+// that `reg` lives in.
+fn find_bank_limits(reginfo: &RegInfo, reg: RegUnit) -> (RegUnit, u16) {
+    if let Some(bank) = reginfo.bank_containing_regunit(reg) {
+        return (bank.first_unit, bank.units);
+    }
+    // We should never get here, since `reg` must come from *some* RegBank.
+    panic!("find_regclass_limits: reg not found");
+}
+
+// The redundant-fill remover processes each EBB independently.  It makes a single forward pass
+// over each EBB, carrying along a running environment, `AvailEnv`, that maps `RegUnits` to an
+// `Option<SlotInfo>`.
+//
+// If the triplet `(regunit, offset, size)` appears in the environment, this indicates that, at
+// this point, `regunit` and the stack slot denoted by `(offset, size)` have the same value.  We
+// also keep track of the SSA `Value` associated with the `reg`, because that is necessary to
+// create `copy` instructions that read `reg`.
+//
+// Invariants for the environment:
+//
+// The environment may have multiple different registers bound to the same stack slot -- that is,
+// `(offset, size)` pair.  That's OK, and reflects the reality that those two registers contain
+// the same value.
+//
+// The `value` fields must however be unique: it isn't allowable to have two entries in the
+// environment that have the same `value` field.  To do so would be to imply that that SSA value
+// has two different locations, since each environment entry is associated with a different
+// register, and so this would imply that the Value resided in both of those registers.  Function
+// `invariants_ok` checks this.
+//
+// It is also the case that the set of `(offset, size)` segments in the environment must be
+// non-overlapping.  `invariants_ok` doesn't check that, though.
+
+struct AvailEnv {
+    map: Vec<Option<SlotInfo>>,
+    size: usize,
+}
+
+impl AvailEnv {
+    // Create a new one.
+    fn new(size: usize) -> Self {
+        let mut env = AvailEnv {
+            map: Vec::<Option<SlotInfo>>::new(),
+            size: size,
+        };
+        env.map.resize(size, None);
+        debug_assert!(env.map.len() == size);
+        env
+    }
+
+    // Debug only: checks (some of) the required AvailEnv invariants.
+    fn invariants_ok(&self) -> bool {
+        if self.map.len() != self.size {
+            return false;
+        }
+        // Check that the SSA names are unique.  This is super lame (quadratic), but it's only
+        // used in debug builds.
+        for i in 0..self.size {
+            if let Some(si) = self.map[i] {
+                for j in i + 1..self.size {
+                    if let Some(sj) = self.map[j] {
+                        if si.value == sj.value {
+                            return false;
+                        }
+                    }
+                }
+            }
+        }
+        true
+    }
+
+    // Invalidates the binding associated with `reg`.  Note that by construction of `AvailEnv`,
+    // `reg` can only be associated with one binding at once.
+    fn invalidate_by_reg(&mut self, reg: RegUnit) {
+        self.map[reg as usize] = None;
+    }
+
+    // Invalidates any binding characterised by `(offset, size)`.  Per discussion of invariants
+    // above, `(offs, size)` can only be associated with one binding at once.
+    fn invalidate_by_offset(&mut self, offset: i32, size: u32) {
+        for i in 0..self.size {
+            if let Some(si) = &self.map[i] {
+                if any_overlap(&si, offset, size) {
+                    self.map[i] = None;
+                }
+            }
+        }
+    }
+
+    // Invalidates all bindings in the enviroment.
+    fn invalidate_all(&mut self) {
+        for i in 0..self.size {
+            self.map[i] = None;
+        }
+    }
+
+    // Updates the environment to track the effect of an `InstructionData::RegMove` instruction.
+    // This moves a value from one register to another, but doesn't create a new SSA name.  Hence
+    // we copy `src`'s entry to `dst` verbatim, without installing a new value in `dst`'s `Value`
+    // field.
+    fn move_reg(&mut self, src: RegUnit, dst: RegUnit) {
+        self.map[dst as usize] = self.map[src as usize];
+        self.map[src as usize] = None;
+    }
+
+    // Does `env` have the exact binding characterised by `(ref, offset, size)` ?
+    fn has_exact_binding(&self, reg: RegUnit, offset: i32, size: u32) -> bool {
+        if let Some(si) = &self.map[reg as usize] {
+            return si.offset == offset && si.size == size;
+        }
+        // No such binding.
+        false
+    }
+
+    // Does `env` have a binding characterised by `(offset, size)` but to a register, let's call
+    // it `reg2`, that isn't `reg`?  If so, return `reg2` and its associated SSA name.  Note that
+    // `reg2` will have the same bank as `reg`.  It is a checked error to call this function with
+    // a binding matching all three of `(reg, offset, size)`.
+    fn has_inexact_binding(
+        &self,
+        reginfo: &RegInfo,
+        reg: RegUnit,
+        offset: i32,
+        size: u32,
+    ) -> Option<(RegUnit, Value)> {
+        // Find the range of RegUnit numbers for the bank that contains `reg`, and use that as our
+        // search space.  This is so as to guarantee that any match is restricted to the same bank
+        // as `reg`.
+        let (first_unit, num_units) = find_bank_limits(reginfo, reg);
+        for reg2 in first_unit..first_unit + num_units {
+            if let Some(si) = &self.map[reg2 as usize] {
+                if si.offset == offset && si.size == size {
+                    if reg2 == reg {
+                        panic!("has_inexact_binding: binding *is* exact!");
+                    }
+                    return Some((reg2, si.value));
+                }
+            }
+        }
+        // No such binding.
+        None
+    }
+
+    // Create the binding `(reg, value, offset, size)` in `env`, and throw away any previous
+    // binding associated with `reg`.
+    fn bind(&mut self, reg: RegUnit, value: Value, offset: i32, size: u32) {
+        debug_assert!(self.map.len() == self.size);
+        self.map[reg as usize] = Some(SlotInfo {
+            value,
+            offset,
+            size,
+        });
+    }
+}
+
+impl<'a> Context<'a> {
+    /// Run the remover.
+    fn cleanup_func(&mut self) {
+        let mut total_regunits = 0;
+        for ru in self.reginfo.banks {
+            total_regunits += ru.units;
+        }
+        // Perform redundant-fill removal.  We can visit the EBBs in any order, since there's no
+        // context carried across EBB boundaries at present.
+        for &ebb in self.domtree.cfg_postorder().iter().rev() {
+            self.cleanup_ebb(ebb, total_regunits);
+        }
+    }
+
+    // Returns the regunit that `v` is allocated to.  Assumes that `v` actually resides in a
+    // register.
+    fn reg_of_value(&self, v: Value) -> RegUnit {
+        match self.cur.func.locations[v] {
+            ValueLoc::Reg(ru) => ru,
+            _ => panic!("reg_of_value: value isn't in a reg"),
+        }
+    }
+
+    // Returns the stack slot that `v` is allocated to.  Assumes that `v` actually resides in a
+    // stack slot.
+    fn slot_of_value(&self, v: Value) -> &StackSlotData {
+        match self.cur.func.locations[v] {
+            ValueLoc::Stack(slot) => &self.cur.func.stack_slots[slot],
+            _ => panic!("slot_of_value: value isn't in a stack slot"),
+        }
+    }
+
+    // Invalidates any binding associated with a regunit that is written by `inst`.
+    fn invalidate_regs_written_by_inst(&mut self, inst: Inst) {
+        let dst_vals: &[Value] = self.cur.func.dfg.inst_results(inst);
+        for v in dst_vals.iter() {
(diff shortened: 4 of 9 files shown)


cranelift-codegen/src/peephole.rs
```diff
@@ -0,0 +1,469 @@
...
+struct AvailEnv {
+    map: Vec<Option<SlotInfo>>,
+    size: usize,
+}
+
+impl AvailEnv {
+    // Create a new one.
+    fn new(size: usize) -> Self {
+        let mut env = AvailEnv {
+            map: Vec::<Option<SlotInfo>>::new(),
```
bjorn3: Can you please store this in the `PeepHole` struct instead and clear it when calling `PeepHole::clear` to reduce the amount of allocations. The purpose of the `Context` is to cache heap allocated memory for use between compilation of two functions.

bjorn3:
What is the performance impact of this on for example wasm compilation? (compile time and runtime)

julian-seward1:
I haven't measured the effect on compile time, but I don't think it will be much.  It's a single forwards pass over each EBB in isolation.  The effect on the generated code, running on a quiet Intel Xeon at 2.6 Ghz, is as follows:
```
                         # loads
              # loads  converted    run time     run time    % of
What          removed  to copies    before, ms   after, ms   before

wasm_box2d.js    2680       1890          1740        1560     89.6
fib.js              0          1          1098        1022     93.0
wasm_lua_binary  3648       3624          8441        7338     86.9
raybench.js      1017        646          3443        3353     97.4
wasm_zlib.c.js   2188       2102          2287        2163     94.6
rust-fannkuch.js   94         42          4362        4098     93.9
```

cranelift-codegen/meta/src/isa/x86/encodings.rs
```diff
@@ -419,6 +420,8 @@ pub fn define(
     let rec_fnaddr8 = r.template("fnaddr8");
     let rec_fregfill32 = r.template("fregfill32");
     let rec_fregspill32 = r.template("fregspill32");
+    let rec_fillnull = r.recipe("fillnull");
+    let rec_ffillnull = r.recipe("ffillnull");
```
bnjbvr: nit: can you keep this list alphabetically ordered, please?

cranelift-codegen/meta/src/isa/x86/encodings.rs
```diff
@@ -838,6 +841,18 @@ pub fn define(
...
     e.enc_i32_i64(regfill, rec_regfill32.opcodes(vec![0x8b]));
 
+    // No-op fills, created by late-stage redundant-fill removal.
+    for &ty in &[I64, I32, I16, I8] {
+        e.enc64_rec(fill_nop.bind(ty), rec_fillnull, 0);
+        e.enc32_rec(fill_nop.bind(ty), rec_fillnull, 0);
+    }
+    e.enc64_rec(fill_nop.bind(B1), rec_fillnull, 0);
+    e.enc32_rec(fill_nop.bind(B1), rec_fillnull, 0);
+    for &ty in &[F64, F32] {
```
bnjbvr: I just saw that there might be a chance we can use `ValueType::all_lane_types` here to iterate on all the types; it will add bindings for other boolean types, in addition to what you already have, which could be a good thing.

cranelift-codegen/meta/src/shared/formats.rs
```diff
@@ -169,6 +169,11 @@ pub fn define(immediates: &OperandKinds, entities: &OperandKinds) -> FormatRegis
             .imm(("src", stack_slot))
             .imm(("dst", regunit)),
     );
+    registry.insert(
+        Builder::new("FillNop")
```
bnjbvr: Do we actually need this new format? It seems to me that it is unused, since the encodings use the `f_unary` format instead. (If that's true, then we'd need the meta crate to warn about unused formats, because they'll generate dead code.)

cranelift-codegen/src/context.rs
```diff
@@ -49,6 +50,9 @@ pub struct Context {
 
     /// Loop analysis of `func`.
     pub loop_analysis: LoopAnalysis,
+
+    /// Peephole optimiser context.
+    pub peephole: Peephole,
```
bnjbvr: As suggested by @lars-t-hansen, can this live in the register allocation context, and can the peephole code be moved within this directory too? It would make sense to do so since the pass is only needed because of the register allocation's shortcomings.

cranelift-codegen/src/peephole.rs
```diff
@@ -0,0 +1,469 @@
...
+// The simplest incarnation simply detects fills from the same stack slot into
+// the same register, and removes them:
+//
+//  v1 = fill v0
+//  ...
+//  v2 = fill v0
+//
+// The second fill is redundant provided that (1) v1 and v2 have the same
+// register, (2) that "..." overwrites neither that register nor v0's stack
+// slot, and (3) that the second fill is dominated by the first.  (3) is
```
bnjbvr: nit: can you make an actual list here:

```
The second fill is redundant provided that:

(1) v1 and v2 etc
(2) "..." overwrites neither" etc
```

cranelift-codegen/src/peephole.rs
```diff
@@ -0,0 +1,469 @@
...
+// determine that "v2 = fill v0" is redundant.
+//
+// The loop must also, even in the simplest version, check all instructions
+// for writes to registers and stack slots, and invalidate environment entries
+// accordingly.
+//
+// === First refinement -- implemented
+//
+// This is to observe that spills also create equivalences between registers
+// and stack slot values, which in turn increases our changes of proving later
```
bnjbvr: nit: `changes` => `chances`

bnjbvr: Also I don't understand this comment and what it means for the below algorithm. Can you reword it, please?

cranelift-codegen/src/peephole.rs
```diff
@@ -0,0 +1,469 @@
...
+// For a fill, even if we can't find an exact match in the environment,
+// finding a load of the same spill slot into a different register is still a
+// win, since we can convert the fill into a reg-reg copy.
+//
+// === Other tricks that don't seem to help
+//
+// I tried tracking the effects of InstructionData::Copy in the environment,
+// for the case when both source and destination are registers.  But that
+// didn't cause any more fills to be removed or converted to copies.  I'm not
+// sure why.
```
bnjbvr: nit: Please don't use the "I" pronoun in comments and keep the quantity of information maximal. Suggestion:

```
Tracking the effects of InstructionData::Copy when both source and destination are registers didn't cause any more fills to be removed or converted to copies.
```

cranelift-codegen/src/peephole.rs
```diff
@@ -0,0 +1,469 @@
...
+//
+// It's also possible for a spill to be detected as redundant, if such a
+// binding already appears in the environment.  I implemented this, but it
+// seems almost entirely ineffective, so I removed it again.
```
bnjbvr: ditto

cranelift-codegen/src/peephole.rs
```diff
@@ -0,0 +1,469 @@
...
+
+// This describes a spill slot in the obvious way.  It also tracks the SSA
+// value name associated with the slot.
+#[derive(Clone, Copy)]
+struct SlotInfo {
+    value: Value,
+    offs: i32,
```
bnjbvr: nit: please use fully-spelled names in structure members: `offset` here.

cranelift-codegen/src/peephole.rs
```diff
@@ -0,0 +1,469 @@
...
+    size: u32,
+}
+
+// Find out if the range `[offs, +size)` overlaps the with range in `si`.
+fn any_overlap(si: &SlotInfo, offs: i32, size: u32) -> bool {
```
bnjbvr: nit: and full names in parameters too: `slot_info`, `offset`

cranelift-codegen/src/peephole.rs
```diff
@@ -0,0 +1,469 @@
...
+// This describes a spill slot in the obvious way.  It also tracks the SSA
+// value name associated with the slot.
+#[derive(Clone, Copy)]
+struct SlotInfo {
+    value: Value,
+    offs: i32,
+    size: u32,
+}
+
+// Find out if the range `[offs, +size)` overlaps the with range in `si`.
```
bnjbvr: What's the `with` range?

cranelift-codegen/src/peephole.rs
```diff
@@ -0,0 +1,469 @@
...
+fn any_overlap(si: &SlotInfo, offs: i32, size: u32) -> bool {
+    let a_offs = si.offs as i64;
+    let a_size = si.size as i64;
+    let b_offs = offs as i64;
+    let b_size = size as i64;
```
bnjbvr: These could all be `u64` instead, right? (Or are offsets negative?)

cranelift-codegen/src/peephole.rs
```diff
@@ -0,0 +1,469 @@
...
+    }
+    // We should never get here, since `reg` must come from *some* RegBank.
+    panic!("find_regclass_limits: reg not found");
+}
+
+// The redundant-fill remover processes each EBB independently.  It makes a
+// single forward pass over each EBB, carrying along a running environment,
+// `AvailEnv`, that maps `RegUnits` to an `Option<SlotInfo>`.
+//
+// If the a triplet `(regunit, offset, size)` appears in the environment, this
```
bnjbvr: nit: `the a`

cranelift-codegen/src/peephole.rs
```diff
@@ -0,0 +1,469 @@
...
+
+impl AvailEnv {
+    // Create a new one.
+    fn new(size: usize) -> Self {
+        let mut env = AvailEnv {
+            map: Vec::<Option<SlotInfo>>::new(),
+            size: size,
+        };
+        for _ in 0..size {
+            env.map.push(None);
```
bnjbvr: Can you use `resize_default` instead?

cranelift-codegen/src/peephole.rs
```diff
@@ -0,0 +1,469 @@
...
+
+    // Updates the environment to track the effect of an
+    // `InstructionData::RegMove` instruction.  This moves a value from one
+    // register to another, but doesn't create a new SSA name.  Hence we copy
+    // `src`'s entry to `dst` verbatim, without installing a new value in
+    // `dst`'s `Value` field.
+    fn move_reg(&mut self, src: RegUnit, dst: RegUnit) {
+        debug_assert!(self.map.len() == self.size);
+        self.map[dst as usize] = self.map[src as usize];
+        self.map[src as usize] = None;
```
bnjbvr: That would break the SSA invariant to keep the value alive, right? Makes me wonder if there are missed opportunities because of this, and if the value stored in the map should be either an Option<SlotInfo> or a (new kind of) alias

cranelift-codegen/src/peephole.rs
```diff
@@ -0,0 +1,469 @@
...
+        banks: &[RegBank],
+        reg: RegUnit,
+        offs: i32,
+        size: u32,
+    ) -> Option<(RegUnit, Value)> {
+        debug_assert!(self.map.len() == self.size);
+        // Find the range of RegUnit numbers for the bank that contains `ref`,
+        // and use that as our search space.  This is so as to guarantee that
+        // any match is restricted to the same bank as `reg`.
+        let (first_unit, n_units) = find_bank_limits(banks, reg);
```
bnjbvr: nit: please fully spell `n_units` into `num_units`

cranelift-codegen/src/peephole.rs
```diff
@@ -0,0 +1,469 @@
...
+    // Does `env` have a binding characterised by `(offs, size)` but to a
+    // register, let's call it `reg2`, that isn't `reg`?  If so, return `reg2`
+    // and its associated SSA name.  Note that `reg2` will have the same bank
+    // as `reg`.  It is a checked error to call this function with a binding
+    // matching all three of `(reg, offs, size)`.
+    fn has_inexact_binding(
+        &self,
+        banks: &[RegBank],
+        reg: RegUnit,
+        offs: i32,
```
bnjbvr: (nit: fully spelled names please)

cranelift-codegen/src/peephole.rs
```diff
@@ -0,0 +1,469 @@
...
+            }
+        }
+        // No such binding.
+        None
+    }
+
+    // Create the binding `(reg, value, offs, size)` in `env`, and throw away
+    // any previous binding associated with `reg`.
+    fn bind(&mut self, reg: RegUnit, value: Value, offs: i32, size: u32) {
+        debug_assert!(self.map.len() == self.size);
```
bnjbvr: This debug assert appears in every function, and I'm not sure what purpose it serves, since Rust will panic either way if you try to access OOB?

cranelift-codegen/src/peephole.rs
```diff
@@ -0,0 +1,469 @@
...
+        self.map[reg as usize] = Some(SlotInfo { value, offs, size });
+    }
+}
+
+impl<'a> Context<'a> {
+    /// Run the post-regalloc peephole optimiser.
+    fn cleanup(&mut self) {
+        let mut tot_regunits = 0;
```
bnjbvr: nit: fully spelled variable names please

cranelift-codegen/src/peephole.rs
```diff
@@ -0,0 +1,469 @@
...
+        for ru in self.reginfo.banks {
```
bnjbvr: nit: fully spelled variable names please

+ `ru` is not correct, it's a register bank I think.

cranelift-codegen/meta/src/shared/instructions.rs
```diff
@@ -1159,6 +1159,27 @@ pub fn define(
...
 
+    ig.push(
+        Inst::new(
+            "fill_nop",
+            r#"
+        This is identical to :inst:`fill`, except it has no encoding, since
+        it is a no-op.
+
+        This instruction is created only during late-stage peephole
+        optimisation, after all registers and stack slots have been assigned.
```
lars-t-hansen: As discussed, suggest "reload optimisation" instead of "peephole optimisation" in order not to muddy the waters more than necessary.

cranelift-codegen/meta/src/shared/instructions.rs
```diff
@@ -1159,6 +1159,27 @@ pub fn define(
...
+        It is used to replace :inst:`fill`s that have been identified as
+        redundant on the basis that a dominating :inst:`fill` or :inst:`spill`
```
lars-t-hansen: I might end this paragraph after "redundant" since the reasons for the redundancy are not interesting here.  As a case in point, I started using the copy_nop in my simple register allocator for reasons unrelated to its original use.

cranelift-codegen/src/context.rs
```diff
@@ -21,6 +21,7 @@ use crate::legalize_function;
 use crate::licm::do_licm;
 use crate::loop_analysis::LoopAnalysis;
 use crate::nan_canonicalization::do_nan_canonicalization;
+use crate::peephole::Peephole;
```
lars-t-hansen: Again, "peephole" not quite what we want; I'll not bring this up any further.

cranelift-codegen/src/context.rs
lars-t-hansen: The discussion yesterday at first suggested that we need to do the reload optimization after prologue/epilogue since the stack layout happens in that phase, but then later we got around to the point that it's just the identity of the stack slot that matters, and that's available sooner, so I too would be interested in seeing further investigation here to move this pass into regalloc.

cranelift-codegen/src/peephole.rs
```diff
@@ -0,0 +1,469 @@
+//! A late-stage peephole optimiser.  Currently only does simple
+//! redundant-fill removal.
+
+use crate::cursor::{Cursor, EncCursor};
+use crate::dominator_tree::DominatorTree;
+use crate::ir::InstructionData;
+use crate::ir::Opcode;
+use crate::ir::StackSlotData;
+use crate::ir::{Ebb, Function, Inst, InstBuilder, Value, ValueLoc};
```
lars-t-hansen: Nit: merge all the lines importing from ir.

cranelift-codegen/src/peephole.rs
```diff
@@ -0,0 +1,469 @@
...
+//
+// === Other tricks that don't seem to help
+//
+// I tried tracking the effects of InstructionData::Copy in the environment,
+// for the case when both source and destination are registers.  But that
+// didn't cause any more fills to be removed or converted to copies.  I'm not
+// sure why.
+//
+// It's also possible for a spill to be detected as redundant, if such a
+// binding already appears in the environment.  I implemented this, but it
```
lars-t-hansen: There really should not be many redundant spills, if any, given how the register allocator assigns locations.  Once a value is flagged as spilled, it is going to be spilled everywhere and is spilled upon creation; since values are immutable, there's never any reason to spill a value again while it is live.  The live range splitter may change this calculus, but should have its own optimizations to avoid redundant spills.

cranelift-codegen/src/peephole.rs
```diff
@@ -0,0 +1,469 @@
...
+    let b_size = size as i64;
+    let no_overlap = a_offs + a_size <= b_offs || b_offs + b_size <= a_offs;
+    !no_overlap
+}
+
+// Find, in `banks`, the register bank that `reg` lives in, and return the
+// lower limit and size of the bank.  This is so the caller can conveniently
+// iterate over all `RegUnits` in the bank that `reg` lives in.
+fn find_bank_limits(banks: &[RegBank], reg: RegUnit) -> (RegUnit, u16) {
+    for rb in banks.iter() {
```
lars-t-hansen: I'm not fond of loops...  I don't know this data structure well enough, but if we can go from a regunit to a regclass to a top-level regclass to a regbank without looping then that would be good.  Worth investigating a little more deeply, but could be followup work, depending on measured performance impact.

cranelift-codegen/src/peephole.rs
```diff
@@ -0,0 +1,469 @@
...
+    // register to another, but doesn't create a new SSA name.  Hence we copy
+    // `src`'s entry to `dst` verbatim, without installing a new value in
+    // `dst`'s `Value` field.
+    fn move_reg(&mut self, src: RegUnit, dst: RegUnit) {
+        debug_assert!(self.map.len() == self.size);
+        self.map[dst as usize] = self.map[src as usize];
+        self.map[src as usize] = None;
+    }
+
+    // Does `env` have the exact binding characterised by `(ref, offs, size)` ?
```
lars-t-hansen: You probably mean 'reg' not 'ref'.

cranelift-codegen/src/peephole.rs
```diff
@@ -0,0 +1,469 @@
...
+                    let src_value = arg;
+                    let slot = self.slot_of_value(results[0]);
+                    let src_reg = self.reg_of_value(*src_value);
+                    if let Some(offs) = slot.offset {
+                        let size = slot.size;
+                        env.invalidate_by_offset(offs, size);
+                        env.bind(src_reg, *src_value, offs, size);
+                    } else {
+                        // A spill insn with no offset.  We shouldn't see any
+                        // such thing at this late stage in the pipeline.
```
bnjbvr: nit: I'd remove this comment, the panic message is clear enough.

cranelift-codegen/src/peephole.rs
```diff
@@ -0,0 +1,469 @@
...
+
+            match &self.cur.func.dfg[inst] {
+                InstructionData::Unary {
+                    opcode: Opcode::Spill,
+                    arg,
+                } => {
+                    // Extract: (src_reg, src_value, offset, size)
+                    // Invalidate: (offset, size)
+                    // Add new: {src_reg -> (src_value, offset, size)}
+                    let results = self.cur.func.dfg.inst_results(inst);
```
bnjbvr: nit: since results is used only once, can you inline it at use?

cranelift-codegen/src/peephole.rs
```diff
@@ -0,0 +1,469 @@
...
+                    }
+                }
+                InstructionData::Unary {
+                    opcode: Opcode::Fill,
+                    arg,
+                } => {
+                    // Extract: (dst_value, dst_regunit, offset, size)
+                    // Invalidate: (offset, size)
+                    // Add new: {dst_regunit -> (dst_value, offset, size)}
+                    let results = self.cur.func.dfg.inst_results(inst);
```
bnjbvr: ditto for results being used only once

cranelift-codegen/src/peephole.rs
```diff
@@ -0,0 +1,469 @@
...
+            match &self.cur.func.dfg[inst] {
+                InstructionData::Unary {
+                    opcode: Opcode::Spill,
+                    arg,
+                } => {
+                    // Extract: (src_reg, src_value, offset, size)
+                    // Invalidate: (offset, size)
+                    // Add new: {src_reg -> (src_value, offset, size)}
+                    let results = self.cur.func.dfg.inst_results(inst);
+                    let src_value = arg;
```
bnjbvr: Instead of this, in the destructuring pattern, you can do:

```
InstructionData::Unary {
  opcode: Opcode::Spill,
  arg: src_value,
}
```

cranelift-codegen/src/peephole.rs
```diff
@@ -0,0 +1,469 @@
...
+                        env.bind(src_reg, *src_value, offs, size);
+                    } else {
+                        // A spill insn with no offset.  We shouldn't see any
+                        // such thing at this late stage in the pipeline.
+                        panic!("cleanup_ebb: spill with no offset");
+                    }
+                }
+                InstructionData::Unary {
+                    opcode: Opcode::Fill,
+                    arg,
```
bnjbvr: Same remark as above with `src_value`

cranelift-codegen/src/peephole.rs
```diff
@@ -0,0 +1,469 @@
...
+    // forwards along the block, using it to detect redundant fills.
+    fn cleanup_ebb(&mut self, ebb: Ebb, tot_regunits: u16) {
+        // A mapping from [0 .. tot_regunits) to `Option<SlotInfo>`
+        let mut env = AvailEnv::new(tot_regunits as usize);
+
+        self.cur.goto_top(ebb);
+        while let Some(inst) = self.cur.next_inst() {
+            debug_assert!(env.invariants_ok(), "cleanup_ebb: env invariants not ok");
+
+            // To avoid difficulties with the borrow checker, do this in two
```
bnjbvr: For the record, which difficulties are you avoiding here? (I'm wondering if, with little extra effort, we could avoid the indirection and cognitive load that `replacement` materializes.)

cranelift-codegen/src/peephole.rs
```diff
@@ -0,0 +1,469 @@
...
+                    } else {
+                        // A fill insn with no offset.
+                        panic!("cleanup_ebb: fill with no offset");
+                    }
+                }
+                InstructionData::RegSpill { .. }
+                | InstructionData::RegFill { .. }
+                | InstructionData::Call { .. }
+                | InstructionData::CallIndirect { .. } => {
+                    // Play safe: forget everything we know.
```
bnjbvr: Why would we need to forget everything after a call or call_indirect? Is that because a call might write on the caller's stack (and thus fill a slot in a way we're not tracking)?

cranelift-codegen/src/peephole.rs
```diff
@@ -0,0 +1,469 @@
...
+        // order, since there's no context carried across EBB boundaries at
+        // present.
+        for &ebb in self.domtree.cfg_postorder().iter().rev() {
+            self.cleanup_ebb(ebb, tot_regunits);
+        }
+    }
+
+    // Returns the regunit that `v` is allocated to.  Assumes that `v`
+    // actually resides in a register.
+    fn reg_of_value(&self, v: Value) -> RegUnit {
```
bnjbvr: In general, I think this is re-doing the work that's done by the RegDiversions data structure. In particular, this function and the function below are just the methods `reg` and `stack` from this data structure. Is there a way the RegDiversions data structure could be reused, to help with tracking?

cranelift-codegen/src/peephole.rs
lars-t-hansen: Borrowing-mutable something that is already borrowed-immutable, is my guess -- I seem to run into this a lot when trying to iterate over the instructions to update them, and have ended up with the same structure, an information-gathering pass and an updating pass.

lars-t-hansen:
/me is a little skeptical about using RegDiversions but if we move this closer to the register allocator so that we can operate on abstract stack slots instead of on stack offset + size then it's certainly a relevant discussion.

julian-seward1:
Thank you both for the extensive reviews!  I'll redo it accordingly.  As it happens, I have found a correctness issue that initial testing didn't uncover, which I can only reproduce by playing around with register-selection heuristics in the solver.  So I'll attend to that first.

cranelift-codegen/src/peephole.rs
julian-seward1: Done.  As a side comment, we should really profile CL with DHAT to get a system-wide picture of the effectiveness, and turnover rate, of its heap usage.

cranelift-codegen/src/peephole.rs
julian-seward1: These can't be `u64`, since offsets can be negative; they are `i32`s.  I cast them to `i64` to do the checks in order to avoid potentially complex range-end / wraparound logic.

cranelift-codegen/src/peephole.rs
julian-seward1: This is quite a subtle point.  `move_reg` is used only for `InstructionData::RegMove`, and the purpose of that instruction is exactly to move a value to a different register without creating a new SSA name.  So it is true that the old register would still have the same value, which we might be able to use later in the EBB to convert a fill into a copy.  But the problem is, doing such a conversion requires having the SSA name associated with the register available.  And in this situation I don't know what it would be -- I assume there would be none.

It might also be the case that regalloc creates `InstructionData::RegMove` exactly because it needs to free up the source register so that it can immediately overwrite it.  In that case, continuing to track the source register as containing a cached fill value is pointless since it'll shortly get overwritten.  I am however merely guessing here, since I don't know the actual purpose of `InstructionData::RegMove`.

cranelift-codegen/src/peephole.rs
julian-seward1: Yes.  I think I did observe one or perhaps two in our "standard 6" set of perf benchmarks.

cranelift-codegen/src/peephole.rs
julian-seward1: I changed it to use `RegInfo::bank_containing_regunit`.  I don't see a way to do it without the (implied) loop.  But as @bnjbvr says, most targets contain only 2 or 3 banks, of which normally the first is the integer registers, and most register references are to integer registers.  So I'd expect almost always to get a hit in the first iteration of the loop.

cranelift-codegen/src/peephole.rs
julian-seward1: @bnjbvr: yes, it's what Lars says.  Regarding cognitive load, I sometimes find it easier to read the style where we first gather information, then transform, than to understand combined analysis and transformation logic -- although that's not the motivating factor here.

cranelift-codegen/src/peephole.rs
julian-seward1: In general, if calls were correctly annotated with all (or, more generally, a superset of) the registers they write, directly or indirectly, then we wouldn't need to special-case them, since the general-case logic at the end of this instruction-handling `match` would handle them.  (This relates to an email conversion about calls that happened early in your/Lars'/my engagement with Cranelift).  However, AFAIK this isn't the case, so the simple-if-suboptimal thing is to forget everything.  Given that few of the hot blocks that I've studied so far contain calls or returns, this doesn't strike me as much of a loss.

I don't *think* that a call can overwrite our own stack slots (that would seem pretty strange to me), but I'm not 100%.  So forgetting everything is also the safe option from this point of view.

cranelift-codegen/meta/src/shared/formats.rs
julian-seward1: Ah, well spotted.  It seems to work fine if I remove it.

cranelift-codegen/src/peephole.rs
```diff
@@ -0,0 +1,469 @@
...
+            env.map.push(None);
+        }
+        debug_assert!(env.map.len() == size);
+        env
+    }
+
+    // Debug only: checks (some of) the required AvailEnv invariants.
+    fn invariants_ok(&self) -> bool {
+        debug_assert!(self.map.len() == self.size);
+        // This is super lame (quadratic), but it's only used in debug builds.
```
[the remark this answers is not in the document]
julian-seward1: I tried putting `#[cfg(debug)]` just before `fn invariants_ok(...`, but this causes the build to fail.  It seems like the function then disappears, but the `debug_assert!` macro use that calls it, doesn't disappear.  Is there a right way to do this?  I didn't find any other `#[cfg(debug)]`s in the tree.

cranelift-codegen/meta/src/isa/x86/encodings.rs
julian-seward1: On investigation, that seems not to be possible, because we need different recipes, `rec_fillnull` vs `rec_ffillnull`, for the int/bool vs float cases.  And those two recipes can't be combined because they reference different register sets (`GPR` vs `FPR`).

julian-seward1:
Thank you all for the extensive comments!  Here's an updated patch.  It
addresses most but not all review comments, and contains a correctness bug
fix.  Currently unresolved/not-done are:

* changing "peephole" to "reload optimisation" and moving it to regalloc.  No
  controversy there, I just didn't do that yet.

* use of `RegDiversions`: as-yet uninvestigated

* moving this closer to regalloc, by working with abstract stack slots instead
  of offset + size.  My preference would be to land this as-is, and then see
  what else can be achieved by extending it to tree shaped regions, before
  refactoring it.

julian-seward1:
Per irc discussion with bbouvier on 30 July 2019, we agreed:

* To change the name from "peephole" to "redundant_reload_remover", but not to move it into the regalloc/ subtree.  This is because it can't be integrated meaningfully into regalloc/context.rs `Context::run()`, since it doesn't run within that context at all -- it runs much later.

* To not investigate the use of `RegDiversions`.  There is general skepticism that this is viable, and their future role is in any case uncertain.

I investigated changing the `(offset, size)` stack-slot representation in `SlotInfo` to use the existing `StackSlotData`, since that holds the same information (and more).  However, `StackSlotData` represents offsets using an `Option<..>` value, and so using it directly in `SlotInfo` in the patch would have the undesirable effect of making it possible to represent meaningless `SlotInfo`s (ones with no offset).  So I left it as-is.

Possible future work is to investigate whether the slot could be denoted purely by the name of the SSA value it currently holds.  Currently I don't know the feasibility of this.  It might also be a useful concept in a future extension of the redundant-reload remover to cover convergent control flow.

julian-seward1:
3 out of the 6 runs failed with some caching-related timeout, but it looks like in all cases, the build and tests ran successfully.

cranelift-codegen/src/redundant_reload_remover.rs
```diff
@@ -0,0 +1,462 @@
...
+impl RedundantReloadRemover {
+    /// Create a new remover state.
+    pub fn new() -> Self {
+        Self {
+            env: AvailEnv::new(0),
+        }
+    }
+
+    /// Clear the state of the remover.
+    pub fn clear(&mut self) {
```
bnjbvr: Since `init` is called for each EBB's beginning, and init calls clear, maybe we can:

- inline the content of clear in init (and make it slightly smarter)
- and remove this public function clear, since it has no effect?

(I think a previous comment suggested to do so, sorry for the back and forth)

Going a bit further, I think this should be in `AvailEnv` itself, since it modifies its attributes, so it could be reused by the AvailEnv's ctor too, and maybe called "reset". Then the run function could call this between the different EBBs.

cranelift-codegen/src/redundant_reload_remover.rs
```diff
@@ -0,0 +1,462 @@
...
+    // References to contextual data structures we need.
+    domtree: &'a DominatorTree,
+
+    // The running state
+    state: &'a mut RedundantReloadRemover,
+}
+
+impl<'a> Context<'a> {
+    fn run(&mut self) {
+        self.cleanup_func();
```
bnjbvr: Instead, can `cleanup_func` be renamed to `run`? (and let's have the whole block be with the rest of Context's implementation)

cranelift-codegen/src/redundant_reload_remover.rs
```diff
@@ -0,0 +1,462 @@
...
+// has two different locations, since each environment entry is associated with a different
+// register, and so this would imply that the Value resided in both of those registers.  Function
+// `invariants_ok` checks this.
+//
+// It is also the case that the set of `(offset, size)` segments in the environment must be
+// non-overlapping.  `invariants_ok` doesn't check that, though.
+
+struct AvailEnv {
+    map: Vec<Option<SlotInfo>>,
+    size: usize,
```
bnjbvr: What's the point of storing the size, since it shouldn't change over time? To avoid calls to the `len` function of the map attribute? (it implies manual syncing between the actual size and the stored size attribute, which tends to make me a bit nervous)

cranelift-codegen/src/redundant_reload_remover.rs
```diff
@@ -0,0 +1,462 @@
...
+            cur: EncCursor::new(func, isa),
+            reginfo: isa.register_info(),
+            domtree: domtree,
+            state: &mut RedundantReloadRemover::new(),
+        };
+        ctx.run()
+    }
+}
+
+struct Context<'a> {
```
bnjbvr: Can you put the struct definition next to the bigger impl block below? (see next comment that suggests to merge the two impl blocks together)

cranelift-codegen/src/redundant_reload_remover.rs
```diff
@@ -0,0 +1,462 @@
...
+// with the slot.
+#[derive(Clone, Copy)]
+struct SlotInfo {
+    value: Value,
+    offset: i32,
+    size: u32,
+}
+
+// Find out if the range `[offset, +size)` overlaps with the range in `si`.
+fn any_overlap(si: &SlotInfo, offset: i32, size: u32) -> bool {
```
bnjbvr: nit: since it's a boolean testing function, I'd prefer it to be either starting with "has" or "is" or be a verb at the singular form: what do you think about `overlaps`?

cranelift-codegen/src/redundant_reload_remover.rs
```diff
@@ -0,0 +1,462 @@
...
+    // Updates the environment to track the effect of an `InstructionData::RegMove` instruction.
+    // This moves a value from one register to another, but doesn't create a new SSA name.  Hence
+    // we copy `src`'s entry to `dst` verbatim, without installing a new value in `dst`'s `Value`
+    // field.
+    fn move_reg(&mut self, src: RegUnit, dst: RegUnit) {
+        self.map[dst as usize] = self.map[src as usize];
+        self.map[src as usize] = None;
+    }
+
+    // Does `env` have the exact binding characterised by `(ref, offset, size)` ?
```
bnjbvr: nit: not "ref" but "reg"

cranelift-codegen/src/redundant_reload_remover.rs
```diff
@@ -0,0 +1,462 @@
...
+            size,
+        });
+    }
+}
+
+impl<'a> Context<'a> {
+    /// Run the remover.
+    fn cleanup_func(&mut self) {
+        let mut total_regunits = 0;
+        for ru in self.reginfo.banks {
```
bnjbvr: nit: it's not clear what the U in "ru" stands for here (if that's for "units", then it's incorrect since we're iterating over register banks). Maybe the right short name is "rb"?

cranelift-codegen/src/redundant_reload_remover.rs
```diff
@@ -0,0 +1,462 @@
...
+                            // update `dst_reg`s binding to make it the same as `reg2`'s, so as to
+                            // maximise the chances of future matches after this instruction.
+                            debug_assert!(reg2 != dst_reg);
+                            replacement = Some((false, val2));
+                            self.state.env.bind(dst_reg, dst_value, offset, size);
+                        } else {
+                            // This fill creates some new binding we don't know about.  Update
+                            // `env` to track it.
+                            self.state.env.invalidate_by_offset(offset, size);
+                            self.state.env.bind(dst_reg, dst_value, offset, size);
```
bnjbvr: nit: this `bind` line is redundant in the three branches of this if/else. Could we move it outside the if/else, please?

cranelift-codegen/src/redundant_reload_remover.rs
```diff
@@ -0,0 +1,462 @@
...
+            debug_assert!(
+                self.state.env.invariants_ok(),
+                "cleanup_ebb: env invariants not ok"
+            );
+
+            // To avoid difficulties with the borrow checker, do this in two stages.  First,
+            // examine the instruction to see if it can deleted or modified, and park the relevant
+            // information in `replacement`.  Update `env` too.  Later, use `replacement` to
+            // actually do the transformation if necessary.
+            let mut replacement: Option<(bool, Value)> = None;
```
bnjbvr: very light suggestion: we could use a small enum instead of the bool, to make it easier to follow. But that's so reduced in scope that it may not be worth it.

cranelift-codegen/src/redundant_reload_remover.rs
```diff
@@ -0,0 +1,462 @@
...
+// Invariants for the environment:
+//
+// The environment may have multiple different registers bound to the same stack slot -- that is,
+// `(offset, size)` pair.  That's OK, and reflects the reality that those two registers contain
+// the same value.
+//
+// The `value` fields must however be unique: it isn't allowable to have two entries in the
+// environment that have the same `value` field.  To do so would be to imply that that SSA value
+// has two different locations, since each environment entry is associated with a different
+// register, and so this would imply that the Value resided in both of those registers.  Function
```
bnjbvr: I am not sure if this is technically true: we're operating after register allocations, which may introduce spurious moves that aren't tied to SSA semantics. Relaxing this invariant would mean that we could **not** have to delete the src binding in a register-to-register move. I agree that's it's useful to have such an invariant for clarity of understanding, but I am wondering if this is required to have it? (the SSA value is only used to transform fills with inexact bindings in reg -> reg2 copies)

julian-seward1: This is an interesting point.  Now that we've established the need for a copy_to_ssa instruction that copies from *any* register to "SSA-world", we could indeed relax this invariant.  However, I suspect it wouldn't give any improvement, because the (only?) reason to divert a value from one register to a different one is that we want to immediately overwrite the source with a new value.

julian-seward1: Actually, doing this removes about 0.5% more loads (statically counting) and doesn't seem to break anything.  So I'll enable it by default.

julian-seward1:
This patch/PR has been abandoned and is replaced by PR906.

PR #830, opened 2019-07-08
Fixes #815: Cast the icmp immediate when simplifying it into an icmp_imm

### LINKED ISSUE #815: iconst + icmp -> icmp_imm incorrect for out of bound values with signed comparison
Eg

```
ebb0(v0: i8):
    v1 = iconst.i8 255 ; -1i8 as u8
    v2 = icmp sge v0, v1; true for v0 s> -1
```

gets converted to

```
ebb0(v0: i8):
    v2 = icmp_imm sge v0, 255 ; never true, as an i8 is never bigger than 127
```

The optimization should be prevented for out of bound integers, or it should convert the `255` to `-1`.

cc https://github.com/bjorn3/rustc_codegen_cranelift/commit/b0ecbf14a23a301183ff194ca97a62e23e2cc6f7

bnjbvr:


cranelift-codegen/src/simple_preopt.rs
```diff
@@ -516,6 +517,27 @@ fn simplify(pos: &mut FuncCursor, inst: Inst) {
                     imm,
                 } = pos.func.dfg[iconst_inst]
                 {
+                    // For signed comparisons, we may need to cast the immediate to the instruction
+                    // result (controlling) type.
+                    let imm = match cond {
+                        IntCC::SignedLessThan
+                        | IntCC::SignedLessThanOrEqual
+                        | IntCC::SignedGreaterThan
+                        | IntCC::SignedGreaterThanOrEqual => {
```
sunfishcode: It's not immediately clear why we need to do this explicit sign-extend step. The `iconst` should already have the same type as the comparison's controlling type. Is it so that `Imm64` values outside the range of that type get normalized? If so, it's unclear why we need to do that for signed and not unsigned here.

bjorn3: See https://github.com/CraneStation/cranelift/issues/815

bnjbvr: Good question: the reason is to do so only for the sign of the immediate, since it's always stored as an i64. When we're generating the instruction, on x86 we use encodings `icscc_ib` or `icscc_id`, which do an actual cast of the immediate from i64 to u8 or u32. So for unsigned immediates this shouldn't be needed.

cranelift-codegen/src/simple_preopt.rs
```diff
@@ -516,6 +517,27 @@ fn simplify(pos: &mut FuncCursor, inst: Inst) {
                     imm,
                 } = pos.func.dfg[iconst_inst]
                 {
+                    // For signed comparisons, we may need to cast the immediate to the instruction
+                    // result (controlling) type.
+                    let imm = match cond {
```
nbp: nit: rename `imm` to `cond_imm` or `new_imm` as you prefer.

cranelift-codegen/src/simple_preopt.rs
```diff
@@ -516,6 +517,27 @@ fn simplify(pos: &mut FuncCursor, inst: Inst) {
...
+                        IntCC::SignedLessThan
+                        | IntCC::SignedLessThanOrEqual
+                        | IntCC::SignedGreaterThan
+                        | IntCC::SignedGreaterThanOrEqual => {
+                            let val: i64 = imm.into();
+                            let new_imm = match pos.func.dfg.ctrl_typevar(inst) {
+                                I8 => (val as i8) as i64,
```
nbp: nit: I am tempted to suggest to move this code either as `impl From<(ir::Type, i64)> for i64` or as a function on the `ir::Type` named `Type::sign_extend(i64) -> i64`.

However, I do not yet know how frequent this sign-extension might be. So feel free to discard this comment if there is no other instances.

bnjbvr:
> Based on Dan's comment , I suspect we might want a more general approach to avoid similar issues.

Right. I can't get to it today, and I'll be out for a couple of weeks, so feel free to steal and close this PR if you'd like to.

bnjbvr:
Closing: we do want a more general approach here, with an extended InstBuilder. (See related issue)

Issue #803, opened 2019-06-25
Implement splat instruction for x86


abrown:
Most of the SIMD instructions defined in the CLIF and WASM specifications are unimplemented. As a first step towards filling those in, this issue tracks adding the `splat` instruction as it is defined in CLIF (https://cranelift.readthedocs.io/en/latest/ir.html#inst-splat) and in WASM (https://github.com/WebAssembly/simd/blob/master/proposals/simd/SIMD.md#create-vector-with-identical-lanes).

abrown:
@sunfishcode, if you have some time it may be helpful to comment on where this headed by looking at https://github.com/bnjbvr/cranelift/pull/1 and https://github.com/bnjbvr/cranelift/pull/3; since I'm in a holding zone until #799 is merged I have been working off of @bnjbvr's repo. It seems to me that this will shake out some bugs and missing features as you will see in my comments to those PRs and, e.g., #802.

PR #833, opened 2019-07-08
Implement splat instruction for x86

abrown:
This is a work in progress and not ready to merge but I am opening the PR for discussion (see #803). I will highlight my top questions in the code below.

[edit] This should be ready for a real review now; I've resolved all of the discussion issues.

### DIFF
--- a/cranelift-codegen/meta/src/gen_legalizer.rs
+++ b/cranelift-codegen/meta/src/gen_legalizer.rs
@@ -51,7 +51,7 @@ fn unwrap_inst(
 
     fmtln!(
         fmt,
-        "let ({}, predicate) = if let ir::InstructionData::{} {{",
+        "let ({}, predicate) = if let crate::ir::InstructionData::{} {{",
         arg_names,
         iform.name
     );
@@ -135,18 +135,16 @@ fn unwrap_inst(
                     .to_comment_string(var_pool)
             ));
 
-            fmt.line("let results = pos.func.dfg.inst_results(inst);");
+            fmt.line("let r = pos.func.dfg.inst_results(inst);");
             for (i, &var_index) in def.defined_vars.iter().enumerate() {
                 let var = var_pool.get(var_index);
-                fmtln!(fmt, "let {} = &results[{}];", var.name, i);
-                if var.has_free_typevar() {
-                    fmtln!(
-                        fmt,
-                        "let typeof_{} = pos.func.dfg.value_type(*{});",
-                        var.name,
-                        var.name
-                    );
-                }
+                fmtln!(fmt, "let {} = &r[{}];", var.name, i);
+                fmtln!(
+                    fmt,
+                    "let typeof_{} = pos.func.dfg.value_type(*{});",
+                    var.name,
+                    var.name
+                );
             }
 
             replace_inst = true;
@@ -187,11 +185,15 @@ fn build_derived_expr(tv: &TypeVar) -> String {
         Some(base) => base,
         None => {
             assert!(tv.name.starts_with("typeof_"));
-            return format!("{}", tv.name);
+            return format!("Some({})", tv.name);
         }
     };
     let base_expr = build_derived_expr(&base.type_var);
-    format!("{}.{}()", base_expr, base.derived_func.name())
+    format!(
+        "{}.map(|t: crate::ir::Type| t.{}())",
+        base_expr,
+        base.derived_func.name()
+    )
 }
 
 /// Emit rust code for the given check.
@@ -221,18 +223,29 @@ fn emit_runtime_typecheck<'a, 'b>(
         Constraint::Eq(tv1, tv2) => {
             fmtln!(
                 fmt,
-                "let predicate = predicate && {} == {};",
+                "let predicate = predicate && match ({}, {}) {{",
                 build_derived_expr(tv1),
                 build_derived_expr(tv2)
             );
+            fmt.indent(|fmt| {
+                fmt.line("(Some(a), Some(b)) => a == b,");
+                fmt.comment("On overflow, constraint doesn\'t apply");
+                fmt.line("_ => false,");
+            });
+            fmtln!(fmt, "};");
         }
         Constraint::WiderOrEq(tv1, tv2) => {
             fmtln!(
                 fmt,
-                "let predicate = predicate && {}.wider_or_equal({});",
+                "let predicate = predicate && match ({}, {}) {{",
                 build_derived_expr(tv1),
                 build_derived_expr(tv2)
             );
+            fmt.indent(|fmt| {
+                fmt.line("(Some(a), Some(b)) => a.wider_or_equal(b),");
+                fmt.comment("On overflow, constraint doesn\'t apply");
+                fmt.line("_ => false,");
+            });
             fmtln!(fmt, "};");
         }
     }
@@ -290,7 +303,8 @@ fn emit_dst_inst(def: &Def, def_pool: &DefPool, var_pool: &VarPool, fmt: &mut Fo
             // Unwrapping would have left the results intact.  Replace the whole instruction.
             fmtln!(
                 fmt,
-                "pos.func.dfg.replace(inst).{};",
+                "let {} = pos.func.dfg.replace(inst).{};",
+                defined_vars,
                 def.apply.rust_builder(&def.defined_vars, var_pool)
             );
 
@@ -406,16 +420,17 @@ fn gen_transform_group<'a>(
     // Function arguments.
     fmtln!(fmt, "pub fn {}(", group.name);
     fmt.indent(|fmt| {
-        fmt.line("inst: ir::Inst,");
-        fmt.line("func: &mut ir::Function,");
-        fmt.line("cfg: &mut ControlFlowGraph,");
-        fmt.line("isa: &dyn TargetIsa,");
+        fmt.line("inst: crate::ir::Inst,");
+        fmt.line("func: &mut crate::ir::Function,");
+        fmt.line("cfg: &mut crate::flowgraph::ControlFlowGraph,");
+        fmt.line("isa: &dyn crate::isa::TargetIsa,");
     });
     fmtln!(fmt, ") -> bool {");
 
     // Function body.
     fmt.indent(|fmt| {
-        fmt.line("use ir::InstBuilder;");
+        fmt.line("use crate::ir::InstBuilder;");
+        fmt.line("use crate::cursor::{Cursor, FuncCursor};");
         fmt.line("let mut pos = FuncCursor::new(func).at_inst(inst);");
         fmt.line("pos.use_srcloc(inst);");
 
--- a/cranelift-codegen/meta/src/isa/x86/encodings.rs
+++ b/cranelift-codegen/meta/src/isa/x86/encodings.rs
@@ -9,7 +9,6 @@ use crate::cdsl::instructions::{
 };
 use crate::cdsl::recipes::{EncodingRecipe, EncodingRecipeNumber, Recipes};
 use crate::cdsl::settings::{SettingGroup, SettingPredicateNumber};
-
 use crate::cdsl::types::ValueType;
 use crate::shared::types::Bool::{B1, B16, B32, B64, B8};
 use crate::shared::types::Float::{F32, F64};
@@ -333,6 +332,7 @@ pub fn define(
     let load_complex = shared.by_name("load_complex");
     let nearest = shared.by_name("nearest");
     let popcnt = shared.by_name("popcnt");
+    let raw_bitcast = shared.by_name("raw_bitcast");
     let regfill = shared.by_name("regfill");
     let regmove = shared.by_name("regmove");
     let regspill = shared.by_name("regspill");
@@ -452,6 +452,7 @@ pub fn define(
     let rec_ldWithIndexDisp8 = r.template("ldWithIndexDisp8");
     let rec_mulx = r.template("mulx");
     let rec_null = r.recipe("null");
+    let rec_null_fpr = r.recipe("null_fpr");
     let rec_pcrel_fnaddr8 = r.template("pcrel_fnaddr8");
     let rec_pcrel_gvaddr8 = r.template("pcrel_gvaddr8");
     let rec_popq = r.template("popq");
@@ -1586,6 +1587,18 @@ pub fn define(
             e.enc32_isap(instruction.clone(), template.clone(), use_sse2);
         }
         e.enc_x86_64_isap(instruction, template, use_sse2);
+    }
+
+    // SIMD bitcast all 128-bit vectors to each other (for legalizing splat.x16x8)
+    for from_type in ValueType::all_lane_types().filter(|t| t.lane_bits() >= 8) {
+        for to_type in ValueType::all_lane_types().filter(|t| t.lane_bits() >= 8 && *t != from_type)
+        {
+            let instruction = raw_bitcast
+                .bind_vector(to_type, 128 / to_type.lane_bits())
+                .bind_vector(from_type, 128 / from_type.lane_bits());
+            e.enc32_rec(instruction.clone(), rec_null_fpr, 0);
+            e.enc64_rec(instruction, rec_null_fpr, 0);
+        }
     }
 
     e
--- a/cranelift-codegen/meta/src/isa/x86/encodings.rs
+++ b/cranelift-codegen/meta/src/isa/x86/encodings.rs
@@ -250,6 +250,17 @@ impl PerCpuModeEncodings {
             self.enc64(inst.clone().bind(I64).bind_any(), template);
         }
     }
+
+    /// Add the same encoding to both X86_32 and X86_64; assumes configuration (e.g. REX, operand binding) has already happened
+    fn enc_32_64_isap(
+        &mut self,
+        inst: BoundInstruction,
+        template: Template,
+        isap: SettingPredicateNumber,
+    ) {
+        self.enc32_isap(inst.clone(), template.clone(), isap);
+        self.enc64_isap(inst, template, isap);
+    }
 }
 
 // Definitions.
@@ -379,6 +390,8 @@ pub fn define(
     let x86_fmax = x86.by_name("x86_fmax");
     let x86_fmin = x86.by_name("x86_fmin");
     let x86_pop = x86.by_name("x86_pop");
+    let x86_pshufd = x86.by_name("x86_pshufd");
+    let x86_pshufb = x86.by_name("x86_pshufb");
     let x86_push = x86.by_name("x86_push");
     let x86_sdivmodx = x86.by_name("x86_sdivmodx");
     let x86_smulx = x86.by_name("x86_smulx");
@@ -462,6 +475,7 @@ pub fn define(
     let rec_pushq = r.template("pushq");
     let rec_ret = r.template("ret");
     let rec_r_ib = r.template("r_ib");
+    let rec_r_ib_unsigned = r.template("r_ib_unsigned");
     let rec_r_id = r.template("r_id");
     let rec_rcmp = r.template("rcmp");
     let rec_rcmp_ib = r.template("rcmp_ib");
@@ -519,6 +533,7 @@ pub fn define(
     let use_lzcnt = settings.predicate_by_name("use_lzcnt");
     let use_bmi1 = settings.predicate_by_name("use_bmi1");
     let use_sse2 = settings.predicate_by_name("use_sse2");
+    let use_ssse3 = settings.predicate_by_name("use_ssse3");
     let use_sse41 = settings.predicate_by_name("use_sse41");
 
     // Definitions.
@@ -1574,6 +1589,28 @@ pub fn define(
     e.enc_both(fcmp.bind(F64), rec_fcscc.opcodes(vec![0x66, 0x0f, 0x2e]));
     e.enc_both(ffcmp.bind(F32), rec_fcmp.opcodes(vec![0x0f, 0x2e]));
     e.enc_both(ffcmp.bind(F64), rec_fcmp.opcodes(vec![0x66, 0x0f, 0x2e]));
+
+    // SIMD splat: before x86 can use vector data, it must be moved to XMM registers; see
+    // legalize.rs for how this is done; once there, x86_pshuf* (below) is used for broadcasting the
+    // value across the register
+
+    // PSHUFB, 8-bit shuffle using two XMM registers
+    for ty in ValueType::all_lane_types().filter(|t| t.lane_bits() == 8) {
+        let number_of_lanes = 128 / ty.lane_bits();
+        let instruction = x86_pshufb.bind_vector(ty, number_of_lanes);
+        let template = rec_fa.nonrex().opcodes(vec![0x66, 0x0f, 0x38, 0x00]);
+        e.enc32_isap(instruction.clone(), template.clone(), use_ssse3);
+        e.enc64_isap(instruction, template, use_ssse3);
+    }
+
+    // PSHUFD, 32-bit shuffle using one XMM register and a u8 immediate
+    for ty in ValueType::all_lane_types().filter(|t| t.lane_bits() == 32) {
+        let number_of_lanes = 128 / ty.lane_bits();
+        let instruction = x86_pshufd.bind_vector(ty, number_of_lanes);
+        let template = rec_r_ib_unsigned.nonrex().opcodes(vec![0x66, 0x0f, 0x70]);
+        e.enc32_isap(instruction.clone(), template.clone(), use_sse2);
+        e.enc64_isap(instruction, template, use_sse2);
+    }
 
     // SIMD scalar_to_vector; this uses MOV to copy the scalar value to an XMM register; according
     // to the Intel manual: "When the destination operand is an XMM register, the source operand is
--- a/cranelift-codegen/meta/src/isa/x86/encodings.rs
+++ b/cranelift-codegen/meta/src/isa/x86/encodings.rs
@@ -10,7 +10,8 @@ use crate::cdsl::instructions::{
 use crate::cdsl::recipes::{EncodingRecipe, EncodingRecipeNumber, Recipes};
 use crate::cdsl::settings::{SettingGroup, SettingPredicateNumber};
 
-use crate::shared::types::Bool::B1;
+use crate::cdsl::types::ValueType;
+use crate::shared::types::Bool::{B1, B16, B32, B64, B8};
 use crate::shared::types::Float::{F32, F64};
 use crate::shared::types::Int::{I16, I32, I64, I8};
 use crate::shared::Definitions as SharedDefinitions;
@@ -340,6 +341,7 @@ pub fn define(
     let rotl_imm = shared.by_name("rotl_imm");
     let rotr = shared.by_name("rotr");
     let rotr_imm = shared.by_name("rotr_imm");
+    let scalar_to_vector = shared.by_name("scalar_to_vector");
     let selectif = shared.by_name("selectif");
     let sextend = shared.by_name("sextend");
     let sload16 = shared.by_name("sload16");
@@ -515,6 +517,7 @@ pub fn define(
     let use_popcnt = settings.predicate_by_name("use_popcnt");
     let use_lzcnt = settings.predicate_by_name("use_lzcnt");
     let use_bmi1 = settings.predicate_by_name("use_bmi1");
+    let use_sse2 = settings.predicate_by_name("use_sse2");
     let use_sse41 = settings.predicate_by_name("use_sse41");
 
     // Definitions.
@@ -603,8 +606,14 @@ pub fn define(
     // Finally, the 0xb8 opcode takes an 8-byte immediate with a REX.W prefix.
     e.enc64(iconst.bind(I64), rec_pu_iq.opcodes(vec![0xb8]).rex().w());
 
-    // Bool constants.
-    e.enc_both(bconst.bind(B1), rec_pu_id_bool.opcodes(vec![0xb8]));
+    // Bool constants (uses MOV)
+    for &ty in &[B1, B8, B16, B32] {
+        e.enc_both(bconst.bind(ty), rec_pu_id_bool.opcodes(vec![0xb8]));
+    }
+    e.enc64(
+        bconst.bind(B64),
+        rec_pu_id_bool.opcodes(vec![0xb8]).rex().w(),
+    );
 
     // Shifts and rotates.
     // Note that the dynamic shift amount is only masked by 5 or 6 bits; the 8-bit
@@ -1564,6 +1573,20 @@ pub fn define(
     e.enc_both(fcmp.bind(F64), rec_fcscc.opcodes(vec![0x66, 0x0f, 0x2e]));
     e.enc_both(ffcmp.bind(F32), rec_fcmp.opcodes(vec![0x0f, 0x2e]));
     e.enc_both(ffcmp.bind(F64), rec_fcmp.opcodes(vec![0x66, 0x0f, 0x2e]));
+
+    // SIMD scalar_to_vector; this uses MOV to copy the scalar value to an XMM register; according
+    // to the Intel manual: "When the destination operand is an XMM register, the source operand is
+    // written to the low doubleword of the register and the regiser is zero-extended to 128 bits."
+    for ty in ValueType::all_lane_types().filter(|t| t.lane_bits() >= 8) {
+        let number_of_lanes = 128 / ty.lane_bits();
+        let instruction = scalar_to_vector.bind_vector(ty, number_of_lanes).bind(ty);
+        let template = rec_frurm.opcodes(vec![0x66, 0x0f, 0x6e]); // MOVD/MOVQ
+        if ty.lane_bits() < 64 {
+            // no 32-bit encodings for 64-bit widths
+            e.enc32_isap(instruction.clone(), template.clone(), use_sse2);
+        }
+        e.enc_x86_64_isap(instruction, template, use_sse2);
+    }
 
     e
 }
--- a/cranelift-codegen/meta/src/isa/x86/encodings.rs
+++ b/cranelift-codegen/meta/src/isa/x86/encodings.rs
@@ -326,6 +326,7 @@ pub fn define(
     let ifcmp_sp = shared.by_name("ifcmp_sp");
     let imul = shared.by_name("imul");
     let indirect_jump_table_br = shared.by_name("indirect_jump_table_br");
+    let insertlane = shared.by_name("insertlane");
     let ireduce = shared.by_name("ireduce");
     let ishl = shared.by_name("ishl");
     let ishl_imm = shared.by_name("ishl_imm");
@@ -476,6 +477,7 @@ pub fn define(
     let rec_ret = r.template("ret");
     let rec_r_ib = r.template("r_ib");
     let rec_r_ib_unsigned = r.template("r_ib_unsigned");
+    let rec_r_ib_unsigned_r = r.template("r_ib_unsigned_r");
     let rec_r_id = r.template("r_id");
     let rec_rcmp = r.template("rcmp");
     let rec_rcmp_ib = r.template("rcmp_ib");
@@ -1624,6 +1626,34 @@ pub fn define(
             e.enc32_isap(instruction.clone(), template.clone(), use_sse2);
         }
         e.enc_x86_64_isap(instruction, template, use_sse2);
+    }
+
+    // SIMD insertlane
+    let mut insertlane_mapping: HashMap<u64, (Vec<u8>, SettingPredicateNumber)> = HashMap::new();
+    insertlane_mapping.insert(8, (vec![0x66, 0x0f, 0x3a, 0x20], use_sse41)); // PINSRB
+    insertlane_mapping.insert(16, (vec![0x66, 0x0f, 0xc4], use_sse2)); // PINSRW
+    insertlane_mapping.insert(32, (vec![0x66, 0x0f, 0x3a, 0x22], use_sse41)); // PINSRD
+    insertlane_mapping.insert(64, (vec![0x66, 0x0f, 0x3a, 0x22], use_sse41)); // PINSRQ, only x86_64
+
+    for ty in ValueType::all_lane_types() {
+        if let Some((opcode, isap)) = insertlane_mapping.get(&ty.lane_bits()) {
+            let number_of_lanes = 128 / ty.lane_bits();
+            let instruction = insertlane.bind_vector(ty, number_of_lanes);
+            let template = rec_r_ib_unsigned_r.opcodes(opcode.clone());
+            if ty.lane_bits() < 64 {
+                e.enc_32_64_isap(instruction, template.nonrex(), isap.clone());
+            } else {
+                // turns out the 64-bit widths have REX/W encodings and only are available on x86_64
+                e.enc64_isap(instruction, template.rex().w(), isap.clone());
+            }
+        }
+    }
+
+    // SIMD bitcast f64 to all 8-bit-lane vectors (for legalizing splat.x8x16); assumes that f64 is stored in an XMM register
+    for ty in ValueType::all_lane_types().filter(|t| t.lane_bits() == 8) {
+        let instruction = bitcast.bind_vector(ty, 16).bind(F64);
+        e.enc32_rec(instruction.clone(), rec_null_fpr, 0);
+        e.enc64_rec(instruction, rec_null_fpr, 0);
     }
 
     // SIMD bitcast all 128-bit vectors to each other (for legalizing splat.x16x8)
--- a/cranelift-codegen/meta/src/isa/x86/legalize.rs
+++ b/cranelift-codegen/meta/src/isa/x86/legalize.rs
@@ -1,7 +1,8 @@
 use crate::cdsl::ast::{var, ExprBuilder, Literal};
 use crate::cdsl::instructions::InstructionGroup;
+use crate::cdsl::types::ValueType;
 use crate::cdsl::xform::TransformGroupBuilder;
-
+use crate::shared::types::Float::F64;
 use crate::shared::types::Int::{I32, I64};
 use crate::shared::Definitions as SharedDefinitions;
 
@@ -19,9 +20,11 @@ pub fn define(shared: &mut SharedDefinitions, x86_instructions: &InstructionGrou
     // List of instructions.
     let insts = &shared.instructions;
     let band = insts.by_name("band");
+    let bitcast = insts.by_name("bitcast");
     let bor = insts.by_name("bor");
     let clz = insts.by_name("clz");
     let ctz = insts.by_name("ctz");
+    let f64const = insts.by_name("f64const");
     let fcmp = insts.by_name("fcmp");
     let fcvt_from_uint = insts.by_name("fcvt_from_uint");
     let fcvt_to_sint = insts.by_name("fcvt_to_sint");
@@ -33,11 +36,15 @@ pub fn define(shared: &mut SharedDefinitions, x86_instructions: &InstructionGrou
     let iadd = insts.by_name("iadd");
     let iconst = insts.by_name("iconst");
     let imul = insts.by_name("imul");
+    let insertlane = insts.by_name("insertlane");
     let isub = insts.by_name("isub");
     let popcnt = insts.by_name("popcnt");
+    let raw_bitcast = insts.by_name("raw_bitcast");
+    let scalar_to_vector = insts.by_name("scalar_to_vector");
     let sdiv = insts.by_name("sdiv");
     let selectif = insts.by_name("selectif");
     let smulhi = insts.by_name("smulhi");
+    let splat = insts.by_name("splat");
     let srem = insts.by_name("srem");
     let udiv = insts.by_name("udiv");
     let umulhi = insts.by_name("umulhi");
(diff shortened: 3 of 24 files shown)


cranelift-codegen/meta/src/isa/x86/encodings.rs
```diff
@@ -59,8 +59,8 @@ impl PerCpuModeEncodings {
         template: Template,
         builder_closure: T,
     ) -> Encoding
-    where
-        T: FnOnce(EncodingBuilder) -> EncodingBuilder,
+        where
+            T: FnOnce(EncodingBuilder) -> EncodingBuilder,
```
abrown: Should I be running `rustfmt` on everything to avoid this?

cranelift-codegen/meta/src/isa/x86/encodings.rs
```diff
@@ -1565,5 +1589,43 @@ pub fn define(
...
+    // TODO legalize 16x8 to two PINSRW (from SSE), then PSHUFD as if it were a 32x4?
+    shuffle_mapping.insert(32, vec![0x66, 0x0f, 0x70]); // PSHUFD from SSE2
+    // TODO legalize 64x2 as two PINSRQ instructions (from SSE 4.1)?
+
+    for ty in LaneTypeIterator::new() {
+        if let Some(opcode) = shuffle_mapping.get(&ty.lane_bits()) {
+            let number_of_lanes = 128 / ty.lane_bits();
+            e.enc32(x86_pshuf.bind_vector(ty, number_of_lanes), rec_r_ib_unsigned.nonrex().opcodes(opcode.clone()));
+            e.enc64(x86_pshuf.bind_vector(ty, number_of_lanes), rec_r_ib_unsigned.nonrex().opcodes(opcode.clone()));
+            // TODO can't use enc_both here because it forces a .rex() encoding
```
abrown: This is the block that performs the `PSHUF` encodings; see the TODO comments for the missing parts or parts that need to be legalized. One question here: when I add a settings predication (e.g. with `enc64_isap`) and then attempt to write a `binemit` test targeting that feature set (e.g. `set use_sse41`), cranelift cannot seem to find an encoding--is there something I'm missing here? Once I figure that out I can add the settings predicates back in.

cranelift-codegen/meta/src/isa/x86/encodings.rs
```diff
@@ -1565,5 +1589,43 @@ pub fn define(
...
+
+    // PINSRD, 32-bit insertion into a lane requiring SSE4.1
+    e.enc_32_64(insertlane.bind_vector(B32, 4), rec_r_ib_unsigned_r.nonrex().opcodes(vec![0x66, 0x0f, 0x3a, 0x22]));
+    e.enc_32_64(insertlane.bind_vector(F32, 4), rec_r_ib_unsigned_r.nonrex().opcodes(vec![0x66, 0x0f, 0x3a, 0x22]));
+    e.enc_32_64(insertlane.bind_vector(I32, 4), rec_r_ib_unsigned_r.nonrex().opcodes(vec![0x66, 0x0f, 0x3a, 0x22]));
+
+    // PINSRQ, 64-bit insertion into a lane requiring SSE4.1 (only on x86_64)
+    e.enc64(insertlane.bind_vector(B64, 4), rec_r_ib_unsigned_r.rex().w().opcodes(vec![0x66, 0x0f, 0x3a, 0x22]));
+    e.enc64(insertlane.bind_vector(F64, 2), rec_r_ib_unsigned_r.rex().w().opcodes(vec![0x66, 0x0f, 0x3a, 0x22]));
+    e.enc64(insertlane.bind_vector(I64, 2), rec_r_ib_unsigned_r.rex().w().opcodes(vec![0x66, 0x0f, 0x3a, 0x22]));
```
abrown: I had to start implementing some of the `insertlane` instructions in order to proceed with my current legalizations.

filetests/isa/x86/pinsr.clif
```diff
@@ -0,0 +1,23 @@
+test binemit
+; set has_sse41
+target x86_64
+
+function %test_pinsr_i32() {
+ebb0:
+[-, %rax]   v0 = iconst.i32 42
+[-, %rbx]   v1 = iconst.i32 99
+[-, %xmm4]  v2 = splat.i32x4 v0
+; TODO it seems weird that insertlane should return anything, the x86 instruction does not so %xmm7 below is meaningless
```
abrown: The CLIF definition of `insertlane` looks like:
```
ig.push(
        Inst::new(
            "insertlane",
            r#"
        Insert ``y`` as lane ``Idx`` in x.

        The lane index, ``Idx``, is an immediate value, not an SSA value. It
        must indicate a valid lane index for the type of ``x``.
        "#,
        )
        .operands_in(vec![x, Idx, y])
        .operands_out(vec![a]),
    );
```
Why even have an `out` operand? If it is needed (e.g. other ISAs?), then it would seem like `a` and `x` are supposed to map to the same register--then how do I indicate this in my code?

cranelift-codegen/meta/src/isa/x86/legalize.rs
```diff
@@ -290,4 +295,61 @@ pub fn define(shared: &mut SharedDefinitions, x86_instructions: &InstructionGrou
...
+        ],
+    );
+
+//    narrow.legalize(
+//        def!(y = splat.I64x2(x)), // TODO this is not possible yet
+//        vec![
+//            def!(a = bitcast(x)), // move into the lowest 64 bits of an XMM register
+//            def!(y = insertlane(a, lane_one, x)), // move into the highest 64 bits of the same XMM register
+//        ],
+//    );
```
abrown: The majority of my questions are about this unfinished block of code:
 - I started using `bitcast` as a way to move data into the XMM registers; this may be confusing to someone unfamiliar with cranelift. Should this be done differently?
 - the `def!` and `def_rhs!` macros in `cranelift-codegen/meta/src/cdsl/ast.rs` will not allow me to specify vector types--should those macros be changed or is there a different way to do this?
 - is there a way to do the legalizations across all N-bit types (e.g. F16, I16, B16) to avoid repeated code?

cranelift-codegen/meta/src/isa/x86/encodings.rs
abrown: Travis to the rescue: https://travis-ci.org/CraneStation/cranelift/jobs/556043281#L1455--`cargo fmt --all`.

sunfishcode: Yes :-). For example, I have my editor set up to run `rustfmt` automatically, so it happens automatically.

cranelift-codegen/meta/src/cdsl/type_inference.rs
```diff
@@ -364,7 +364,6 @@ impl TypeEnvironment {
 
             // Sanity check: translated constraints should refer only to real variables.
             for arg in constraint.typevar_args() {
-                assert!(vars_tv.contains(arg));
```
bjorn3: Why was this assert removed?

cranelift-codegen/meta/src/isa/x86/encodings.rs
sunfishcode: Currently, tests handle this by selecting an explicit CPU, like this: `target x86_64 haswell`. That should hopefully unblock you for now, though ideally we should add a better way to handle architecture flags like `use_sse41`.

cranelift-codegen/meta/src/isa/x86/encodings.rs
abrown: 👍

cranelift-codegen/meta/src/gen_legalizer.rs
```diff
@@ -134,6 +134,19 @@ fn unwrap_inst(
                     .get(var_pool.get(def.defined_vars[0]).dst_def.unwrap())
                     .to_comment_string(var_pool)
             ));
+
+            fmt.line("let r = pos.func.dfg.inst_results(inst);");
+            for (i, &var_index) in def.defined_vars.iter().enumerate() {
+                let var = var_pool.get(var_index);
+                fmtln!(fmt, "let {} = &r[{}];", var.name, i);
```
bjorn3: Is this necessary given that https://github.com/CraneStation/cranelift/pull/833/files#diff-a685da9c9934e511050d2fefc936f786R54 already gets vars.

cranelift-codegen/meta/src/gen_legalizer.rs
```diff
@@ -134,6 +134,19 @@ fn unwrap_inst(
...
+                fmtln!(
+                    fmt,
+                    "let typeof_{} = pos.func.dfg.value_type(*{});",
```
bjorn3: This is similar to https://github.com/CraneStation/cranelift/pull/833/files#diff-a685da9c9934e511050d2fefc936f786R114

cranelift-codegen/meta/src/cdsl/type_inference.rs
abrown: @bnjbvr? (he made that commit)

cranelift-codegen/meta/src/gen_legalizer.rs
abrown: Another one for @bnjbvr

cranelift-codegen/meta/src/gen_legalizer.rs
abrown: Another one for @bnjbvr

cranelift-codegen/meta/src/isa/x86/legalize.rs
sunfishcode: The key thing with `bitcast` is that the source and destination have to have the same size. The `x86_pshuf` instruction defined elsewhere in this patch states that its input is a vector, so as written, this code is using a `bitcast` to convert between a scalar and a vector with a different size. We should probably add a new instruction named something like `scalar_to_vector` which converts a scalar value into a vector value, filling the other lanes with zeros, as `movq`/`movd` do.

sunfishcode: For legalizations across F16/I16/B16, we can take advantage of the fact that this meta code is just a Rust program, so we can create a utility routine that takes a `LaneType` argument and defines the legalization with that type.

sunfishcode: For the `def!` and `def_rhs!` macros, can you say more about what problem you're hitting?

filetests/isa/x86/pinsr.clif
sunfishcode: It has an `out` operand because Cranelift IR uses SSA form, and the value after the insert is different from the value before, so we need to represent it with a different `Value`, even if it'll use the same register. To tell the register allocator that the output register is the same as one of the input registers, we can "tie" the output to one of the inputs. See the `"rr"` recipe in cranelift-codegen/meta/src/isa/x86/recipes.rs for example.

cranelift-codegen/meta/src/isa/x86/legalize.rs
abrown: - re: `bitcast`: I think what you're saying makes sense; I checked and I think `mov` to an XMM register does zero out the top bits--I'll make the new instruction and call it `scalar_to_vector`. Should it be x86-only?
- re: `def!`: I am trying to tie specific legalizations to specific types; e.g. a splat to an 8x16 is a different sequence of instructions than 16x8 than 32x4, etc. This is rather non-uniform (and I am open to other ways of doing this) but I don't see another way right now because of things like:
    - `pshuf` for 8-bit lanes requiring the shuffle mask in a register (instead of as an immediate),
    - `pshuf` for 64-bit is not existent, presumably because two `pinsr` instructions do the same thing
    - `pshuf` for 16-bit only works on MM registers
But the first problem is that I can't even bind vector types to instructions using `def!`; I don't think Rust's macros can receive something like `splat.I32x4` and know how to split at the `x` (but I may be wrong about this--still new to Rust macros).

filetests/isa/x86/pinsr.clif
abrown: Ah, I see! `0` is pointing to the first `gpr`?

cranelift-codegen/meta/src/isa/x86/encodings.rs
abrown: But I see code like `set is_pic` in `filetests`; how come `is_pic` can be modified but I can't modify the other settings from the tests?

filetests/isa/x86/pinsr.clif
abrown: I tried this out for the `insertlane` recipe:
```
        let format = formats.get(f_insert_lane);
        recipes.add_template_recipe(
            EncodingRecipeBuilder::new("r_ib_unsigned_r", f_insert_lane, 2)
                .operands_in(vec![fpr, gpr])
                .operands_out(vec![0])
                .inst_predicate(InstructionPredicate::new_is_unsigned_int(
                    format, "lane", 8, 0,
                ))
                .emit(
                    r#"
                    {{PUT_OP}}(bits, rex2(in_reg1, in_reg0), sink);
                    modrm_rr(in_reg1, in_reg0, sink);
                    let imm:i64 = lane.into();
                    sink.put1(imm as u8);
                "#,
                ),
        );
```
However, when I do this the `filetests/isa/x86/pinsr.clif` test now fails:
```
FAIL filetests/isa/x86/pinsr.clif: binemit(%test_pinsr_b8):
No matching encodings for v3 = insertlane.b8x16 v2, 10, v1 in [Mp3r_ib_unsigned_r#d20]
```
I don't understand why an encoding shouldn't be found. Is there some syntax I'm missing?
```
[-, %rax]   v0 = iconst.i32 42
[-, %rbx]   v1 = iconst.i32 99
[-, %xmm4]  v2 = splat.i32x4 v0
[-, %xmm4]  v3 = insertlane v2, 2, v1    ; bin: 66 0f 3a 22 e3 02
```

cranelift-codegen/meta/src/isa/x86/legalize.rs
abrown: An update on the last statement "I don't think Rust's macros can receive something like `splat.I32x4` and know how to split at the `x`": I think this may be possible with [procedural macros](https://doc.rust-lang.org/reference/procedural-macros.html) but it would make `def!` a bit more complex and add a `proc_macro` dependency.

abrown:
@sunfishcode, one other thought: is there a way to guard this feature under some flag until we can properly test that all of the encodings are correct by compiling and running some WASM code?

cranelift-codegen/meta/src/isa/x86/legalize.rs
sunfishcode: I think `scalar_to_vector` can be target-independent.

For the `def!` issue, would an option be to add a new `def_vec!` macro which takes additional arguments which provide the vector element and type? That'd be a little redundant, but would allow us to avoid a proc_macro dependency for now. @bnjbvr Do you have any ideas?

cranelift-codegen/meta/src/isa/x86/legalize.rs
```diff
@@ -290,4 +297,80 @@ pub fn define(shared: &mut SharedDefinitions, x86_instructions: &InstructionGrou
...
+    //                def!(d = x86_pshuf(c, broadcast_immediate_u8)), // broadcast the bytes in the XMM register with PSHUF
+    //                expr_bitcast_x16x8, // def!(y = bitcast.ty#8(d)), // pretend this is an X16x8 again, this should translate to a NOOP
+    //            ],
+    //        );
+    //    }
+
+    // SIMD splat: 32-bits
+    for ty in LaneTypeIterator::new().filter(|t| t.lane_bits() == 32) {
+        narrow.legalize(
+            def!(y = splat.ty#4(x)),
```
abrown: I'm open to suggestions but this (i.e. `T#N`) seemed like the closest way to clearly show that we are doing a SIMD binding for this instruction; Rust macros won't see `x` in `I32x4` as a separate token so we need a different syntax.

cranelift-codegen/meta/src/isa/x86/recipes.rs
```diff
@@ -788,6 +790,48 @@ pub fn define<'shared>(
         );
     }
 
+    // XX /r ib with 8-bit unsigned immediate (e.g. for pshuf)
+    {
+        let format = formats.get(f_extract_lane);
+        recipes.add_template_recipe(
+            EncodingRecipeBuilder::new("r_ib_unsigned", f_extract_lane, 2)
```
abrown: I'm not sure about the naming of these templates: `r == register`? `ib == immediate byte`?

cranelift-codegen/meta/src/isa/x86/encodings.rs
sunfishcode: Ah, I forgot that you can set a flag on the target line too: `target x86_64 use_sse41=true`. That's the proper way to do this.

cranelift-codegen/meta/src/isa/x86/legalize.rs
```diff
@@ -290,4 +297,80 @@ pub fn define(shared: &mut SharedDefinitions, x86_instructions: &InstructionGrou
...
+    //                .assign_to(vec![y.clone().into()]);
+    //
+    //        narrow.legalize(
+    //            expr, // def!(y = splat.X16x8(x)),
+    //            vec![
+    //                def!(a = x86_scalar_to_vector(x)), // translate to an x86 MOV to get the value in the lowest 16 bits of an XMM register
+    //                def!(b = insertlane(a, lane_one, x)), // insert the value again but in the next lowest 16 bits
+    //                expr_bitcast_i32x4, // def!(c = bitcast.I32#4(b)), // pretend this is an I32x4 so we can use PSHUFD, this should translate to a NOOP
+    //                def!(d = x86_pshuf(c, broadcast_immediate_u8)), // broadcast the bytes in the XMM register with PSHUF
+    //                expr_bitcast_x16x8, // def!(y = bitcast.ty#8(d)), // pretend this is an X16x8 again, this should translate to a NOOP
```
abrown: Since there is no good instruction for shuffling 16-bit lanes in x86 (`PSHUFW` seems to only work for MM registers) my plan was to fill up the bottom 32 bits, "cast" to a 32x4, shuffle with the 32x4 PSHUFD, and then "cast" back. I can't get the above to work in cranelift due to some typing issues. Is this type of "cast" possible? What is the best way to tell cranelift to cast for SSA purposes but emit no instructions?

sunfishcode:
> @sunfishcode, one other thought: is there a way to guard this feature under some flag until we can properly test that all of the encodings are correct by compiling and running some WASM code?

We could use the `enable_simd` setting for this. It's defaulted to true right now, but we could set it to false while the SIMD code is under development to prevent frontends from accidentally using it.

cranelift-codegen/meta/src/isa/x86/legalize.rs
sunfishcode: Another option for 16-bit lane shuffling is to use `pshufb`, when `SSSE3` is available.

`bitcast` ought to be the way to do this, and emit zero instructions when possible. If there are problems using `bitcast` for this, they're bugs and we should fix them :-).

cranelift-codegen/meta/src/isa/x86/legalize.rs
bnjbvr: Re: macro: yes, adding new syntax with special chars could be the way; I too would like to avoid procedural macros, and try to keep the macro as simple as possible. Another way is to bind your instructions for the vector type before using the macro; I've done this when the instruction would need to be bound twice, for instance, so there should be a few examples in the code too.

cranelift-codegen/meta/src/cdsl/type_inference.rs
bnjbvr: @bjorn3 This has landed separately in another PR, see [commit message](https://github.com/CraneStation/cranelift/commit/137b13815372c0b61b8de42e7b481a2a8f8b10bd).

cranelift-codegen/meta/src/gen_legalizer.rs
bnjbvr: @bjorn3 these are different things. The variables defined here are used to be able to reason on the types of result variables. The else branch did generate these result variables and corresponding typeof variables, but the `if` branch didn't, for no good reasons. (It also landed in a separate PR, see [commit](https://github.com/CraneStation/cranelift/commit/6a2a5e4dde73c16256c484252e77d804cfc55676)).

cranelift-codegen/meta/src/gen_legalizer.rs
bnjbvr: (answered above)

filetests/isa/x86/pinsr.clif
bnjbvr: > Ah, I see! 0 is pointing to the first gpr?

0 is pointing to the first element in the list of input constraints, so in this case, the first is `fpr`.

bnjbvr: > I don't understand why an encoding shouldn't be found.

There's no clear reason just looking at the code; I'd advise to use a debugger and step in like we did before, to see what goes wrong (either no encodings are found in the encodings list, or the encodings constraints aren't satisfied).

cranelift-codegen/meta/src/isa/x86/recipes.rs
bnjbvr: Yes, sounds consistent with what's in there.

cranelift-codegen/meta/src/isa/x86/legalize.rs
bnjbvr: For what it's worth, for 16-bit lane splat, in Spidermonkey, we used to duplicate the content of the low 8 bits to the low 16 bits (with a shift and OR), and then do a 8-bit splat, if SSSE3 is not available. This sounds like what you want to achieve here with a 32-bit splat.

Re: `bitcast`, approving what Dan says. In addition to this, you might need to add a few encodings for some bound types, with a recipe that emits no code (like the `null` recipe for instance, but might need a new one for FPR / vector types).

cranelift-codegen/meta/src/isa/x86/legalize.rs
abrown: I'm removing it; @bnjbvr's suggestion to just bind the instruction before the `def!` is more straightforward.

cranelift-codegen/meta/src/isa/x86/legalize.rs
abrown: 👍 I'm removing the extra macro case and doing the binding prior to using the instruction in `def!`

cranelift-codegen/meta/src/isa/x86/legalize.rs
abrown: Thanks for the help on this; I implemented it with the 32-bit PSHUFD (supported in SSE2) because I think it can be done with 3 instructions (MOV + PINSRD + PSHUFD) but I can switch over to PSHUFB if that doesn't work. Let me comment over on the new version of this; there's a test failing...

filetests/isa/x86/legalize-splat.clif
```diff
@@ -0,0 +1,72 @@
...
+}
+
+; check:   ebb0:
+; nextln:     v0 = bconst.b16 true
+; nextln:     v2 = x86_scalar_to_vector.b16x8 v0
+; nextln:     v3 = insertlane v2, 1, v0
+; nextln:     v4 = bitcast.i32x4 v3
+; nextln:     v5 = x86_pshufd v4, 0
+; nextln:     v1 = bitcast.b16x8 v5
+; nextln:     return v1
```
abrown: All of the other blocks in this file seem to work but this one fails with:
```
FAIL filetests/isa/x86/legalize-splat.clif: compile(%test_splat_b16):
function %test_splat_b16() -> b16x8 [%xmm0] baldrdash {
                                ebb0:
[-]                                 v0 = bconst.b16 true
[RexMp2frurm#56e]                   v2 = x86_scalar_to_vector.b16x8 v0
[Mp2r_ib_unsigned_r#5c4]            v3 = insertlane v2, 1, v0
[null_fpr#00]                       v4 = bitcast.i32x4 v3
; error: inst5: arg 0 (v3) with type b16x8 failed to satisfy type set ValueTypeSet { lanes: BitSet(511), ints: BitSet(120), floats: BitSet(96), bools: BitSet(0) }

[Mp2r_ib_unsigned#570]              v5 = x86_pshufd v4, 0
[null_fpr#00]                       v1 = bitcast.b16x8 v5
; error: inst1: has an invalid controlling type b16x8

[Op1ret#c3]                         return v1
}

; 2 verifier errors detected (see above). Compilation aborted.
```

I'm pretty sure I've created the right bitcasts at https://github.com/CraneStation/cranelift/pull/833/files#diff-d8dea7e9b6533024c2c26fbb8d88867bR1654 and following; the legalization for this is at https://github.com/CraneStation/cranelift/pull/833/files#diff-4643787d63e13b9a6b517d0bebb29decR346.

abrown: Thanks for the help, @sunfishcode! Should be fixed now with `raw_bitcast`

filetests/isa/x86/pinsr.clif
abrown: Finally fixed this one.

abrown:
@bnjbvr, @sunfishcode, @bjorn3: this should be ready for a real review now. Do I need to rebase and squash or will you squash when you merge from here?

bnjbvr:
Nice. I would personally like it if there was a way to squash some of your commits together, so:

- each represent a unit of work
- that properly compiles and passes tests

This might be hard to do after the fact, but in some cases it's possible to do so, which would help make the review much simpler. Does it make sense to you to try to do so?

cranelift-codegen/meta/src/isa/x86/settings.rs
```diff
@@ -3,6 +3,9 @@ use crate::cdsl::settings::{PredicateNode, SettingGroup, SettingGroupBuilder};
 pub fn define(shared: &SettingGroup) -> SettingGroup {
     let mut settings = SettingGroupBuilder::new("x86");
 
+    // CPUID.01H:ECX
```
bjorn3: *EDX?

abrown: 👍

cranelift-codegen/meta/src/shared/instructions.rs
```diff
@@ -2630,6 +2640,22 @@ pub fn define(
...
 
+    let s = &operand_doc("s", Scalar, "A scalar value");
+    let a = &operand_doc("a", TxN, "A vector value (i.e. held in an XMM register)");
+
+    ig.push(
+        Inst::new(
+            "scalar_to_vector",
+            r#"
+    Scalar To Vector -- move a value out of a scalar register and into a vector
+    register; the scalar should be moved to the lowest-order bits of the vector
```
bjorn3: s/should/will

This will be read mostly by people writing a clif ir producer

cranelift-codegen/meta/src/shared/instructions.rs
```diff
@@ -2630,6 +2640,22 @@ pub fn define(
...
+    register and any higher bits should be zeroed.
```
bjorn3: Same

cranelift-codegen/meta/src/shared/instructions.rs
```diff
@@ -106,6 +106,16 @@ pub fn define(
             .build(),
     );
 
+    let Scalar = &TypeVar::new(
+        "scalar",
+        "Any scalar value that can be used as a lane in a vector",
+        TypeSetBuilder::new()
+            .bools(8..64)
```
bjorn3:
```suggestion
            .bools(Interval::All)
```

cranelift-codegen/meta/src/shared/instructions.rs
```diff
@@ -106,6 +106,16 @@ pub fn define(
...
+            .ints(8..64)
```
bjorn3:
```suggestion
            .ints(Interval::All)
```

cranelift-codegen/meta/src/shared/instructions.rs
```diff
@@ -106,6 +106,16 @@ pub fn define(
...
+            .floats(32..64)
```
bjorn3:
```suggestion
            .floats(Interval::All)
```

cranelift-codegen/src/isa/x86/abi.rs
```diff
@@ -69,10 +69,11 @@ impl ArgAssigner for Args {
...
         let ty = arg.value_type;
 
-        // Check for a legal type.
-        // We don't support SIMD yet, so break all vectors down.
+        // Vectors should stay in vector registers TODO perhaps this should be conditional on some ISA feature (e.g. SSE) and if not present do a ValueConversion
         if ty.is_vector() {
-            return ValueConversion::VectorSplit.into();
+            let reg = FPR.unit(self.fpr_used);
+            self.fpr_used += 1;
+            return ArgumentLoc::Reg(reg).into();
```
abrown: Someone should probably look at this to make sure I'm doing this right but, also, it seems like this should only happen when `enable_simd=true`--I don't see any reference to settings nearby that I can access, though.

abrown:
@bjorn3's comments are addressed (do you have more?) and I rebased the commits to something that should make more sense in a git history, running `test-all.sh` after each. I do have one pending question about argument assignment (above) if anyone has an opinion.

cranelift-codegen/meta/src/isa/x86/encodings.rs
```diff
@@ -1565,5 +1589,80 @@ pub fn define(
     e.enc_both(ffcmp.bind(F32), rec_fcmp.opcodes(vec![0x0f, 0x2e]));
     e.enc_both(ffcmp.bind(F64), rec_fcmp.opcodes(vec![0x66, 0x0f, 0x2e]));
 
+    // SIMD splat: before x86 can use vector data, it must be moved to XMM registers; see
+    // legalize.rs for how this is done; once there, x86_pshuf* (below) is used for broadcasting the
+    // value across the register
+
+    // PSHUFB, 8-bit shuffle using two XMM registers
+    for ty in LaneTypeIterator::new().filter(|t| t.lane_bits() == 8) {
+        let number_of_lanes = 128 / ty.lane_bits();
```
bjorn3: `lane_bits` is always 8 here right?

cranelift-codegen/meta/src/isa/x86/encodings.rs
```diff
@@ -1565,5 +1589,80 @@ pub fn define(
...
+        let instruction = x86_pshufb.bind_vector(ty, number_of_lanes);
+        let template = rec_fa.nonrex().opcodes(vec![0x66, 0x0f, 0x38, 00]);
+        e.enc32_isap(instruction.clone(), template.clone(), use_ssse3);
+        e.enc64_isap(instruction, template, use_ssse3);
+    }
+
+    // PSHUFD, 32-bit shuffle using one XMM register and a u8 immediate
+    for ty in LaneTypeIterator::new().filter(|t| t.lane_bits() == 32) {
+        let number_of_lanes = 128 / ty.lane_bits();
```
bjorn3: And 32 here?

cranelift-codegen/meta/src/isa/x86/legalize.rs
```diff
@@ -290,4 +301,84 @@ pub fn define(shared: &mut SharedDefinitions, x86_instructions: &InstructionGrou
...
+        r#"
+    Legalize instructions by narrowing.
+
+    Use x86-specific instructions if needed."#,
+    )
+    .isa("x86")
+    .chain_with(shared.transform_groups.by_name("narrow").id);
+
+    // SIMD
+    let broadcast_immediate_u8 = Literal::constant(uimm8, 0x00);
```
bjorn3: `let uimm8_zero`?

cranelift-codegen/meta/src/isa/x86/legalize.rs
```diff
@@ -290,4 +301,84 @@ pub fn define(shared: &mut SharedDefinitions, x86_instructions: &InstructionGrou
...
+    let broadcast_immediate_f64 = Literal::constant(ieee64, 0x00);
```
bjorn3: `let ieee64_0x00`?

cranelift-codegen/meta/src/isa/x86/legalize.rs
```diff
@@ -290,4 +301,84 @@ pub fn define(shared: &mut SharedDefinitions, x86_instructions: &InstructionGrou
...
+    let lane_one = Literal::constant(uimm8, 0x01);
```
bjorn3: `let uimm8_one`?

cranelift-codegen/meta/src/isa/x86/legalize.rs
```diff
@@ -290,4 +301,84 @@ pub fn define(shared: &mut SharedDefinitions, x86_instructions: &InstructionGrou
...
+    let b = var("b");
+    let c = var("c");
+    let d = var("d");
+
+    // SIMD splat: 8-bits
+    for ty in LaneTypeIterator::new().filter(|t| t.lane_bits() == 8) {
+        let splat_x8x16 = splat.bind_vector(ty, 128 / ty.lane_bits());
```
bjorn3: `lane_bits` is always 8 here right?

cranelift-codegen/meta/src/isa/x86/legalize.rs
```diff
@@ -290,4 +301,84 @@ pub fn define(shared: &mut SharedDefinitions, x86_instructions: &InstructionGrou
...
+        let bitcast_f64_to_x8x16 = bitcast.bind_vector(ty, 128 / ty.lane_bits()).bind(F64);
```
bjorn3: Same

cranelift-codegen/meta/src/isa/x86/legalize.rs
```diff
@@ -290,4 +301,84 @@ pub fn define(shared: &mut SharedDefinitions, x86_instructions: &InstructionGrou
...
+                def!(b = f64const(broadcast_immediate_f64)), // zero out a different XMM register; the shuffle mask for moving the lowest byte to all other byte lanes is 0x0
+                def!(c = bitcast_f64_to_x8x16(b)), // no instruction emitted; informs the SSA that the 0 in b can be used as a vector of this type
+                def!(y = x86_pshufb(a, c)), // PSHUFB takes two XMM operands, one of which is a shuffle mask (i.e. b)
+            ],
+        );
+    }
+
+    // SIMD splat: 16-bits
+    for ty in LaneTypeIterator::new().filter(|t| t.lane_bits() == 16) {
+        let splat_x16x8 = splat.bind_vector(ty, 128 / ty.lane_bits());
```
bjorn3: Same for 16 here and below.

cranelift-codegen/meta/src/shared/instructions.rs
```diff
@@ -2630,6 +2634,44 @@ pub fn define(
         .operands_out(vec![a]),
     );
 
+    let x = &operand("x", Any);
+    let a = &operand_doc("a", AnyTo, "Bits of `x` reinterpreted");
+
+    ig.push(
+        Inst::new(
+            "raw_bitcast",
```
bjorn3: Maybe name it `simd_bitcast`?

cranelift-codegen/meta/src/cdsl/types.rs
```diff
@@ -299,7 +299,7 @@ pub struct LaneTypeIterator {
 
 impl LaneTypeIterator {
     /// Create a new lane type iterator.
-    fn new() -> Self {
+    pub fn new() -> Self {
```
bjorn3: You can use `ValueType::all_lane_types()` as call sites instead of making this public.

cranelift-codegen/meta/src/isa/x86/encodings.rs
```diff
@@ -1565,5 +1589,80 @@ pub fn define(
...
 
+    // SIMD splat: before x86 can use vector data, it must be moved to XMM registers; see
+    // legalize.rs for how this is done; once there, x86_pshuf* (below) is used for broadcasting the
+    // value across the register
+
+    // PSHUFB, 8-bit shuffle using two XMM registers
+    for ty in LaneTypeIterator::new().filter(|t| t.lane_bits() == 8) {
+        let number_of_lanes = 128 / ty.lane_bits();
+        let instruction = x86_pshufb.bind_vector(ty, number_of_lanes);
+        let template = rec_fa.nonrex().opcodes(vec![0x66, 0x0f, 0x38, 00]);
```
bjorn3:
```suggestion
        let template = rec_fa.nonrex().opcodes(vec![0x66, 0x0f, 0x38, 0x00]);
```

cranelift-codegen/meta/src/isa/x86/legalize.rs
bjorn3: Also what does the `_x8x16` part mean?

cranelift-codegen/meta/src/isa/x86/recipes.rs
```diff
@@ -367,11 +367,13 @@ pub fn define<'shared>(
     let f_call = formats.by_name("Call");
     let f_call_indirect = formats.by_name("CallIndirect");
     let f_copy_special = formats.by_name("CopySpecial");
+    let f_extract_lane = formats.by_name("ExtractLane"); // TODO this would preferably retrieve a BinaryImm8 format but because formats are compared structurally and ExtractLane has the same structure this is impossible--if we rename ExtractLane, it may even impact parsing
```
bjorn3: ???

cranelift-codegen/meta/src/shared/instructions.rs
```diff
@@ -2630,6 +2634,44 @@ pub fn define(
...
+            r#"
+        Cast the bits in `x` as a different type of the same bit width.
+
+        This instruction does not change the data's representation but allows
+        data in registers to be used as different types, e.g. an i32x4 as a 
+        b8x16. The only constraint on the result `a` is that it can be 
+        `raw_bitcast` back to the original type. Also, in a raw_bitcast between
+        vector types with the same number of lanes, the value of each result 
+        lane is a raw_bitcast of the corresponding operand lane. TODO there is 
+        currently no mechanism for enforcing the bit width constraint.
```
bjorn3: Maybe special case in the legalizer? It has more instructions special cased.

cranelift-codegen/src/ir/immediates.rs
```diff
@@ -738,6 +738,12 @@ impl From<f64> for Ieee64 {
     }
 }
 
+impl From<i32> for Ieee64 {
```
bjorn3: I was confused by this over at https://github.com/CraneStation/cranelift/pull/833#discussion_r302852200. I thought this did `f64::from_bits`, not `as f64`.

cranelift-codegen/meta/src/isa/x86/legalize.rs
abrown: In my next set of commits for `extractlane` I have a commit that changes `bind_vector` so that only the lane type is passed; insided `bind_vector` the number of lanes is calculated based on 128-bit registers. The way I see it, if we have to have different vector sizes later we can change that but for now having the calculation for the number of lanes in one place is better than many places and also better than manually setting the number of lanes at each call site--less opportunity for mistakes. Why don't we fix this in the next PR?

cranelift-codegen/meta/src/isa/x86/legalize.rs
abrown: I want to make clear what the types of the bitcast are and I needed some way to represent `{B|I|F}8x16}`. The first `x` is confusing, I guess; what do you suggest?

cranelift-codegen/meta/src/shared/instructions.rs
abrown: In talking to @sunfishcode I got the impression that this is not a SIMD-only instruction; presumably it could be used on scalar types to convert one SSA type to another without changing the bits.

cranelift-codegen/meta/src/isa/x86/recipes.rs
abrown: That format is useful for more than the ExtractLane instruction; it should probably be named `BinaryImm8` so it can be used for multiple instructions without confusion. But I'm not sure I can change this name without breaking something--what do you think?

cranelift-codegen/meta/src/isa/x86/legalize.rs
bjorn3: `any8x16` or `a8x16`

cranelift-codegen/meta/src/shared/instructions.rs
abrown: I think I mean a constraint mechanism on `TypeVar`: like there eventually should be a way to create a two `TypeVar`s, A and B, in this file and constrain B so that it can be a different type than A but it must have the same number of bits. Am I understanding this right, @sunfishcode?

cranelift-codegen/meta/src/shared/instructions.rs
bjorn3: The difference between this and `bitcast` is the ability to change lane amount right?

cranelift-codegen/meta/src/isa/x86/legalize.rs
abrown: yeah, that is better 😄

cranelift-codegen/meta/src/shared/instructions.rs
abrown: I don't think so: I originally implemented all of this behavior under `bitcast` (and I was able to change the number of lanes just fine) but @sunfishcode wanted me to create a new instruction because `bitcast` has the idea of changing the bits and what I was doing was more a true cast--no bit changes.

cranelift-codegen/meta/src/isa/x86/legalize.rs
abrown: No, this is a good find. I'm pretty sure I actually should be using `f64::from_bits` since the whole point of this is to fill up an XMM register with zeroes (there may be a better way to do this but I saw this approach already emitted the XORPS I needed at https://github.com/CraneStation/cranelift/blob/master/cranelift-codegen/meta/src/isa/x86/encodings.rs#L1421-L1438). Let me fix the `From` below.

cranelift-codegen/src/ir/immediates.rs
abrown: I'm changing this to use `f64::from_bits`

cranelift-codegen/meta/src/isa/x86/recipes.rs
bjorn3: As far as I know you can pick any name.

abrown: I was actually just trying to change the name and apparently the name affects all types of files, here's an incomplete list (because I can't get it to build yet):
```
 cranelift-codegen/meta/src/isa/riscv/recipes.rs |  8 ++++----
 cranelift-codegen/meta/src/isa/x86/recipes.rs   | 26 +++++++++++++-------------
 cranelift-codegen/meta/src/shared/formats.rs    |  4 ++--
 cranelift-codegen/src/postopt.rs                |  2 +-
 cranelift-codegen/src/simple_preopt.rs          |  6 +++---
 cranelift-codegen/src/verifier/mod.rs           |  4 ++--
 cranelift-codegen/src/write.rs                  |  4 ++--
 cranelift-reader/src/parser.rs                  | 10 +++++-----
```
It's tied into a bunch of other code so I would prefer leaving this comment in and doing this separately once a few people agree we should change the name.

cranelift-codegen/src/isa/x86/abi.rs
abrown: Pass in `settings` to `assign()`; pass in one big SettingsGroup if possible, otherwise pass in both a shared and ISA-specific SettingsGroup. Still need to split if SIMD is not enabled.

cranelift-codegen/meta/src/shared/instructions.rs
abrown: @sunfishcode to open a new issue for a mechanism to do this.

cranelift-codegen/src/isa/x86/abi.rs
```diff
@@ -39,10 +41,20 @@ struct Args {
     fpr_used: usize,
     offset: u32,
     call_conv: CallConv,
+    shared_flags: shared_settings::Flags,
+    #[allow(dead_code)]
+    isa_flags: isa_settings::Flags,
```
abrown: @sunfishcode I added both sets of flags as we discussed; `isa_flags` is not yet used so I suppressed the warning. As I looked at it more, I realized I didn't need to implement this setting-passing for the other ISAs because the default behavior there is correct: they should split vectors because there is no SIMD implementation yet.

cranelift-codegen/meta/src/shared/instructions.rs
sunfishcode: I've now filed https://github.com/CraneStation/cranelift/issues/854 to track this.

cranelift-codegen/meta/src/shared/instructions.rs
sunfishcode: Right. It's not just about changing the lane amount; it's also about the fact that boolean types in cranelift are opaque, and we don't want to expose their bit patterns. There are operations we want to do on such types, such as shuffling the lanes, which don't ultimately expose their bit patterns, but it's difficult to express that in the type system. So I have a slight preference for `raw_bitcast` here.

cranelift-codegen/src/isa/x86/abi.rs
sunfishcode: Makes sense. I do think we'll ultimately have a need for the settings here, so this looks fine for now.

PR #838, opened 2019-07-09
[WIP] Don't cache target dir on nightly to prevent old build artifacts from accumulating

### LINKED ISSUE #831: [travis] Don't cache cargo target dir on nightly
Caching the target dir on nightly will cause an accumulation of cached files, as the cargo for one nightly doesn't delete files created by another nightly. This is I think the reason nightly builds on travis take ~50min before timeout instead of `30min like the other channels. It will be necessary to delete all caches afterwards.

bjorn3:
cc #831

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -45,6 +45,9 @@ before_script:
     - rustup update
     - rustfmt +stable --version
 script: ./test-all.sh
+before_cache:
+    # Don't cache target dir on nightly to prevent old build artifacts from accumulating. (#831)
+    - '[[ "$TRAVIS_RUST_VERSION" == "nightly" ]] && rm -rf target/'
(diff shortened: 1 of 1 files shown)


bjorn3:
This seems to be the first PR with the linux nightly job not timing out. The beta linux job did however. @bnjbvr could you please delete all caches once?

bnjbvr:
I've emptied all the caches and restarted beta builds.

bjorn3:
Thanks! All jobs passed!

bnjbvr:
Thanks @bjorn3 ! Actually didn't realize this was still open, and since we've moved to Azure pipelines (which don't seem to show any caching issues), we can close this now.

PR #842, opened 2019-07-10
[docs] Update documentation after meta-python removal

bnjbvr:
This helps making the documentation a bit less useless:

- for instructions in the meta crate, prefer to use the full documentation string instead of only the first line. This will make the generated `docs.rs` documentation more helpful, and that's the content that was in the `ir.rst` documentation source file.
- add links to the documentation for codegen and the meta crate, adds TODO items telling that doc can be outdated, updates copyright license.
- remove all the magic Python stuff that was happening when generating documentation, now that the Python meta module is dead.


bnjbvr:
Updated: having the whole documentation for instructions meant that code blocks (that is, extra indented blocks) were interpreted as doc tests, while containing clif examples :clown_face:.

bnjbvr:
> I wonder if it makes sense to think in terms of eventually migrating all the documentation to rustdoc / docs.rs, rather than sphinx / readthedocs.

Totally agree; scattering docs would be counter-productive.

PR #843, opened 2019-07-10
Change Postion's default method into one from Default trait

MarkMcCaskey:
Hello!

Apologies about the trivial commit; just happened to see this code and thought I'd send a PR.  As far as I'm aware, there are no downsides to using the `Default` trait here (`Position::default()` still works), and it's more 'standard' in my mind.

from a quick grep there are other instances of `default` methods, if they're also just calling default, then we can do the same thing more widely.  If that's something you all are interested in I can add more to this PR tonight

### DIFF
--- a/cranelift-codegen/src/ir/function.rs
+++ b/cranelift-codegen/src/ir/function.rs
@@ -222,22 +222,13 @@ impl Function {
 }
 
 /// Additional annotations for function display.
+#[derive(Default)]
 pub struct DisplayFunctionAnnotations<'a> {
     /// Enable ISA annotations.
     pub isa: Option<&'a dyn TargetIsa>,
 
     /// Enable value labels annotations.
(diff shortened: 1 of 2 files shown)


bnjbvr:
Thanks for this PR! It doesn't make the current code easier or harder to read, but it is more concise, so I'm in favor of taking it. If there are other places where you think we could use more `Default::default()`, can you add them to this PR? I'll happily review it. Thanks!

MarkMcCaskey:
Yeah, I think it's less surprising and more idiomatic/flexible, too, to use the Default trait when implementing a default method.

Turns out the rest of the code was actually very good about using the `Default` trait!  Most implemented it manually, but that's fine.  There was an instance or two where it could have been derived automatically, but the code as-is was more expressive (for example using a constant with the name `INVALID`), so I didn't want to change it.

PR #844, opened 2019-07-12
Split edges to have a block to add regmove & copy instructions.

nbp:
This work is part of #796 and aim at creating space in which `copy` and `regmove` instruction can be inserted by the register allocator. Creating BB instead of EBB require that we have space to insert these new instructions as we can no longer insert any instruction between a conditional branch and a terminator branch.

So far this patch fixes the issue which are about inserting `copy` instructions, but the coloring still attempt to insert `regmove` instruction before the terminator due to the liveness analysis and live ranges which are trying to keep all mutations limited to a single EBB, and not transmitting these changes to the next one, even if it only has a single predecessor.

~~Draft: The test suite should be updated to match these new inserted blocks.~~


nbp:
This branch now contains the changes to forward the diversion to the next EBB, as long as the next EBB has only one unique predecessor. This prevent the `undivert` phase of the coloring to insert `regmove` instruction between a conditional branch and a terminator instruction. Now `regmove` are inserted as needed in following EBB or before a jump to an EBB which has more than one predecessor.

I think the relation used to determine the unique predecessor implemented in the current patch is bogus and will not properly handle infinite loops. I will see if I can replace it. Apart from this relation the rest of the code should remain as-is.

In the mean time, this code passes all test cases except 3, which I will fix as soon as I fix the previous relation.

nbp:
With the following changes, all Cranelift test cases passes with the environment variable `CRANELIFT_BB` set, as well as without.

lars-t-hansen:
Additional test cases that must be run (assuming you've only run the ones in the cranelift dir):
- embenchen tests with "-c cranelift" which checks that the correct output is produced
- all spidermonkey wasm tests
- ideally also compile + run tanks, zen garden, angry bots

cranelift-codegen/src/verifier/locations.rs
```diff
@@ -49,11 +53,19 @@ impl<'a> LocationVerifier<'a> {
     fn check_constraints(&self, errors: &mut VerifierErrors) -> VerifierStepResult<()> {
         let dfg = &self.func.dfg;
         let mut divert = RegDiversions::new();
+        let mut entry_divert = EntryRegDiversions::new();
 
         for ebb in self.func.layout.ebbs() {
```
lars-t-hansen: More a question to myself than anything, since I'm not sure about this: Is this an adequate iteration order after splitting edges?  The edge splitter inserts new blocks near the end of the ebb list, so the layout will not be compatible with the dominator tree, say.

cranelift-codegen/src/regalloc/branch_splitting.rs
```diff
@@ -0,0 +1,228 @@
...
+}
+
+impl<'a> Context<'a> {
+    fn run(&mut self) {
+        // Any ebb order will do.
+        self.topo.reset(self.cur.func.layout.ebbs());
+        while let Some(ebb) = self.topo.next(&self.cur.func.layout, self.domtree) {
+            // Branches can only be at the last or second to last position in an
+            // extended basic block.
+            self.cur.goto_last_inst(ebb);
```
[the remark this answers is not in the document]
nbp: So far we have not yet settled on how many a branch-group would include, and so far having a single conditional branch seems to be the simplest approach. We can refine this later on as we add more branches.

cranelift-codegen/src/regalloc/branch_splitting.rs
```diff
@@ -0,0 +1,228 @@
...
+                .dfg
+                .inst_variable_args(inst)
+                .iter()
+                .map(|x| *x)
+                .collect();
+
+            // Create the block the branch will jump to.
+            let new_ebb = self.make_empty_ebb();
+
+            // Remove the arguments from the branch and make it jump to the new block.
```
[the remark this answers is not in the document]
nbp: @bnjbvr or @sunfishcode  any idea if there is a way to only patch the jump target without changing the rest of the isntruction. My understanding is that we have to call the builder once more and provide all the inputs. Maybe we can have an instruction builder which takes the default from one instruction, and allow us to patch it after, before building it?

cranelift-codegen/src/regalloc/coloring.rs
```diff
@@ -183,6 +193,51 @@ impl<'a> Context<'a> {
...
+                        }
+                        SingleDest(ebb, _) => ebb,
+                    };
+
+                    // We have a single branch with a single target, and an EBB
+                    // which immediate dominator is the branch instruction. Thus
+                    // the target EBB has a single predecessor.
+                    if self.cfg.pred_iter(target).count() == 1 {
+                        // TODO: Assert that the entry is only set once.
+                        // Transfer the diversion to the next EBB.
```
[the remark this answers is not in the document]
nbp: I am not sure if there is an easy way to assert this information, without keeping an extra mapping. Or adding an extra Option on top of the RegDiversions, which would be quite weird as the output of the `.get` function.

cranelift-codegen/src/regalloc/coloring.rs
```diff
@@ -183,6 +193,51 @@ impl<'a> Context<'a> {
...
+                        self.entry_divert[target] = self.divert.iter().collect();
+                        debug!(
+                            "Set entry-diversion for {} to\n      {}",
+                            target,
+                            self.entry_divert[target].display(&self.reginfo)
+                        );
+                        self.divert.clear();
```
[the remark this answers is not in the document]
nbp: it is not, however, I do not understand how EBB are allowed to end with unresolved diversions.

cranelift-codegen/src/regalloc/diversion.rs
```diff
@@ -38,10 +40,14 @@ impl Diversion {
 }
 
 /// Keep track of diversions in an EBB.
+#[derive(Clone)]
 pub struct RegDiversions {
     current: FxHashMap<Value, Diversion>,
```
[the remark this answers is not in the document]
nbp: Honestly, if the example I am seeing are representative of the real world, having a hashmap is probably overkill as we are only going to have a handful of diversions.

cranelift-codegen/src/verifier/locations.rs
nbp: the branch splitting only adds blocks right after the current one (for the terminator), or almost at the end (for conditional branches).

The regalloc will follow the reversed post-order, which should work fine for forwarding diversion to blocks which have a single predecessors.

This algorithm should not have any issues with new blocks added by branch splitting but will wrongly report errors on blocks with a single predecessors which were already present in the CFG before the branch splitting. We should change this to either the topo-order or the reversed post-order.

cranelift-codegen/src/regalloc/branch_splitting.rs
```diff
@@ -0,0 +1,228 @@
...
+    // References to contextual data structures we need.
+    domtree: &'a mut DominatorTree,
+    topo: &'a mut TopoOrder,
+    cfg: &'a mut ControlFlowGraph,
+}
+
+impl<'a> Context<'a> {
+    fn run(&mut self) {
+        // Any ebb order will do.
+        self.topo.reset(self.cur.func.layout.ebbs());
```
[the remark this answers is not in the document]
nbp: Unfortunately no, because the `self.cur` is used in a mutable way later on, and it is not possible to have a reference to `func` as the cursor already has a reference to it.

However, we do not need to iterate over the newly generated blocks, so I replaced this code by `self.domtree.cfg_postorder().iter().rev()`, the reverse is not needed but would change the order of the generated blocks at the end.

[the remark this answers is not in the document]
nbp: hum … never mind, I thought I could but apparently not, as we do not always compute the `postorder` vector when we compute the dominator tree.  So without cloning the topological order is the only one I know.  Except maybe an integer iteration over the `self.cur.layout.ebbs()`

nbp:
> * embenchen tests with "-c cranelift" which checks that the correct output is produced
> * all spidermonkey wasm tests
> * ideally also compile + run tanks, zen garden, angry bots

I will go and test these. However, I will note that if developing Cranelift requires compiling Firefox to check if we can make changes, then there is a problem with Cranelift test suite, which is lacking test cases.

nbp:
Running SpiderMonkey test suite report the following error messages: `unexpected table jump after a conditional branch`, which is strange as I would have expected the legalisation of `br_table` to already handle this issue. I will investigate.

nbp:
The previous case is not related to table jump, just an incorrect error message. Most of these issues are coming from having constants inserted after the branches.

This problem would likely go away as soon as the `FunctionBuilder` is fixed to prevent generating code which does not comply with the BB representation.

cranelift-codegen/src/regalloc/coloring.rs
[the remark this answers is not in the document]
nbp: @julian-seward1 Do you have any idea what might be going on? And if this is a bug/feature?

cranelift-codegen/src/regalloc/coloring.rs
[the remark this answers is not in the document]
nbp: I switched to using `SparseMap` instead, such that I can distinguish the empty entry diversions case from the non-empty case.  Consequently, the code is slightly more complex as I have to store an extra `Ebb` as a key of the Value of the `SparseMap`. Thus I can now replace this TODO with a `debug_assert!(!….constains_key(…))`

`SparseMap` interface might be better if the copy of the Key in a tuple were to be done implicitly.

sstangl:
With #867 and this patch in the repo simultaneously, this is the output from jit-tests, run via `python jit-test/jit_test.py --args="--wasm-compiler=cranelift" dbg64/js/src/js wasm`:

```
FAILURES:
    wasm/bench/wasm_box2d.js
    --test-wasm-await-tier2 wasm/bench/wasm_box2d.js
    --wasm-gc wasm/gc/disabled.js
    wasm/gc/disabled.js
TIMEOUTS:
```

Before those patches, with just basic block assertions, there were 54 failures. Looking good :+1:

cranelift-codegen/src/regalloc/branch_splitting.rs
```diff
@@ -0,0 +1,223 @@
...
+        }
+    }
+
+    fn visit_conditional_branch(&mut self, inst: Inst, opcode: Opcode) {
+        let target = match self.cur.func.dfg[inst] {
+            InstructionData::Branch { destination, .. }
+            | InstructionData::BranchIcmp { destination, .. }
+            | InstructionData::BranchInt { destination, .. }
+            | InstructionData::BranchFloat { destination, .. } => destination,
+            _ => panic!("Unexpected instruction in classify_branch"),
```
sstangl: Nit: not named "classify_branch"

cranelift-codegen/src/regalloc/branch_splitting.rs
```diff
@@ -0,0 +1,223 @@
...
+        };
+
+        // If there are any parameters, split the edge.
+        if self.should_split_edge(inst, target) {
```
sstangl: Pretty silly nit: instead of indenting the whole (long!) function, can just be:

```rust
if !self.should_split_edge(inst, target) {
    return;
}
```

cranelift-codegen/src/regalloc/branch_splitting.rs
```diff
@@ -0,0 +1,223 @@
...
+
+    /// Returns whether we should introduce a new branch.
+    fn should_split_edge(&self, _branch: Inst, target: Ebb) -> bool {
+        // We should split the edge if the target has any parameters.
+        if self.cur.func.dfg.ebb_params(target).len() > 0 {
+            return true;
+        };
+
+        // Or, if the target is has more than one block reaching it.
+        if self.cfg.pred_iter(target).count() != 1 {
```
sstangl: `Iterator.count()` can be O(n), because it consumes the iterator. To check whether there is more than one element, it can do something like:

```rust
if let Some(_) = self.cfg.pred_iter(target).skip(1).next() {
````

File: cranelift-codegen/src/regalloc/branch_splitting.rs [hunk 11]
```diff
@@ -0,0 +1,223 @@
...
+                .map(|x| *x)
+                .collect();
+
+            // Create the block the branch will jump to.
+            let new_ebb = self.make_empty_ebb();
+
+            // Remove the arguments from the branch and make it jump to the new block.
+            match opcode {
+                Opcode::Brz => {
+                    let val = self.cur.func.dfg.inst_args(inst)[0];
```
sstangl: Common to all branches: can be LICM'd in the code :)

cranelift-codegen/src/regalloc/branch_splitting.rs
```diff
@@ -0,0 +1,223 @@
...
+        }
+
+        // If blocks were added the cfg and domtree are inconsistent and must be recomputed.
+        if self.new_blocks {
+            self.cfg.compute(&self.cur.func);
+            self.domtree.compute(&self.cur.func, self.cfg);
+        }
+    }
+
+    fn visit_conditional_branch(&mut self, inst: Inst, opcode: Opcode) {
```
sstangl: Probably is worth renaming `inst: Inst` to `branch: Inst` for readability: I got lost a bit in the code beneath, and forgot what kind of instruction `inst` is supposed to be.

cranelift-codegen/src/regalloc/branch_splitting.rs
```diff
@@ -0,0 +1,223 @@
...
+                    self.cur.func.dfg.replace(inst).brz(val, new_ebb, &[]);
+                }
+                Opcode::Brnz => {
+                    let val = self.cur.func.dfg.inst_args(inst)[0];
+                    self.cur.func.dfg.replace(inst).brnz(val, new_ebb, &[]);
+                }
+                Opcode::BrIcmp => {
+                    if let InstructionData::BranchIcmp { cond, .. } = self.cur.func.dfg[inst] {
+                        let x = self.cur.func.dfg.inst_args(inst)[0];
+                        let y = self.cur.func.dfg.inst_args(inst)[0];
```
sstangl: I'm not sure why this is correct. It looks suspicous. Might be worth a comment?

cranelift-codegen/src/regalloc/branch_splitting.rs
```diff
@@ -0,0 +1,223 @@
...
+            let new_ebb = self.make_empty_ebb();
+
+            // Remove the arguments from the branch and make it jump to the new block.
+            match opcode {
+                Opcode::Brz => {
+                    let val = self.cur.func.dfg.inst_args(inst)[0];
+                    self.cur.func.dfg.replace(inst).brz(val, new_ebb, &[]);
+                }
+                Opcode::Brnz => {
+                    let val = self.cur.func.dfg.inst_args(inst)[0];
```
sstangl: This code keeps using `self.cur.func.dfg` repeatedly, but it's making the lines very long. That causes `cargo fmt` to get really excited and split a one-liner across five lines, which makes this function way longer than it should be.

It would read better if we had `let dfg = &self.cur.func.dfg;` or something to that effect at the top.

cranelift-codegen/src/regalloc/branch_splitting.rs
bnjbvr: I suspect this should be 1 in the second line, yeah.

cranelift-codegen/src/regalloc/branch_splitting.rs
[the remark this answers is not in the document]
bnjbvr: Can you get a mutable reference on `self.cur.func.dfg[inst]`, match on it that it's one of the expected formats (the different formats ought to be grouped in a single case, since we're only interested in the destination ebb, if it has the same name for every format), and then just modify the destination ebb?

bjorn3:
Would it be an idea to use BB before regalloc and EBB after, to prevent degrading quality of the output asm? I assume most algorithms implemented by following a paper are run before regalloc, so using EBB after regalloc shouldn't cause problems.

julian-seward1:
Regalloc is pretty much the end of the line -- afaics there are no passes after regalloc that would benefit from seeing Ebbs rather than Bbs.  (But I could be wrong.)  Do you have any particular transformations in mind?

bjorn3:
> afaics there are no passes after regalloc that would benefit from seeing Ebbs rather than Bbs

I mean that using ebbs after regalloc would make this PR unnecessary. This PR reduces the output asm quality a lot: https://github.com/CraneStation/cranelift/issues/873#issuecomment-515599245.

nbp:
@bjorn3 , My understanding is that @sstangl is working on #873 and will fix this issue before we can even merge these changes.

sstangl:
Yeah, I'm specifically working on addressing codegen regressions caused by this patch splitting lots of critical edges.

cranelift-codegen/src/regalloc/branch_splitting.rs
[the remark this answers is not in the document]
nbp: I think I found a generic way to do that for any branch, based on the assumption that the variable arguments would always be the Ebb parameters. We can already replace the `inst_data.branchdestination_mut()`, but for the arguments I think we have to rely on `inst_data.take_value_list()` and `inst_data.put_value_list(…, …)`.

nbp:
I removed the changes related to the register diversions and the coloring from this pull request, such that this pull request is only about making room for the coloring to later add the `regmove` where needed.

I will open another pull request on top of this one and #883 to handle the coloring and the verification phase associated to it.

cranelift-codegen/src/regalloc/branch_splitting.rs
```diff
@@ -0,0 +1,179 @@
+//! Split the outgoing edges of conditional branches that pass parameters.
+//!
+//! The `minimal` and `tree` register allocators require this. One of the reason for splitting edges
+//! is to be able to insert `copy` and `regmove` instructions between a conditional branch and the
+//! following terminator.
```
lars-t-hansen: The comment about the "minimal" and "tree" allocators should be removed.

cranelift-codegen/src/regalloc/branch_splitting.rs
```diff
@@ -0,0 +1,223 @@
...
+    topo: &'a mut TopoOrder,
+    cfg: &'a mut ControlFlowGraph,
+}
+
+impl<'a> Context<'a> {
+    fn run(&mut self) {
+        // Any ebb order will do.
+        self.topo.reset(self.cur.func.layout.ebbs());
+        while let Some(ebb) = self.topo.next(&self.cur.func.layout, self.domtree) {
+            // Branches can only be at the last or second to last position in an extended basic
```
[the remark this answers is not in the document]
nbp: Let's do that in a follow-up patch.

cranelift-codegen/src/regalloc/branch_splitting.rs
```diff
@@ -0,0 +1,178 @@
...
+            self.cur.insert_ebb(new_ebb);
+
+            // Reset the cursor to point to new terminator of the old ebb.
+            self.cur.goto_inst(jump);
+        }
+    }
+
+    // A new ebb must be inserted before the last ebb because the last ebb may have a
+    // fallthrough_return and can't have anything after it.
+    fn make_empty_ebb(&mut self) -> Ebb {
```
bjorn3: Can this be done only when `fallthrough_return` exists?

cranelift-codegen/src/regalloc/branch_splitting.rs
```diff
@@ -0,0 +1,178 @@
...
+        let new_ebb = self.cur.func.dfg.make_ebb();
+        let last_ebb = self.cur.layout().last_ebb().unwrap();
+        self.cur.layout_mut().insert_ebb(new_ebb, last_ebb);
+        self.has_new_blocks = true;
+        new_ebb
+    }
+
+    /// Returns whether we should introduce a new branch.
+    fn should_split_edge(&self, _branch: Inst, target: Ebb) -> bool {
```
bjorn3: Remove this param?

cranelift-codegen/src/regalloc/branch_splitting.rs
nbp: You mean, inserting it at the end if the branch instruction is not a `fallthrough_return`? This sounds indeed better to me. I will do that.

cranelift-codegen/src/regalloc/branch_splitting.rs
```diff
@@ -34,6 +35,11 @@ struct Context<'a> {
     /// True if new blocks were inserted.
     has_new_blocks: bool,
 
+    /// Record whether newly inserted empty blocks should be inserted last, or
+    /// before the last, to avoid disturbing the expected control flow of
+    /// `fallthroug_return` statements.
+    has_fallthrough_return: Option<bool>,
```
bjorn3: Maybe explain that `None` means not yet calculated, while `Some` means the inner bool determines if there is an `has_fallthrough_return`.

nbp: Thank, I added 2 statements to this comment to clarify that the Option wraps the computed value.

PR #845, opened 2019-07-12
[codegen] Fold uextend.i64.i32 into a no-op whenever possible on x86

bnjbvr:
Still a WIP. This removes spurious moves (generated by uextend.i64.i32) whenever the previous instruction did zero the upper 32 bits for us.

Remaining things to do:

- [ ] Fix constraints of the `null` recipe being not satisfied.
- [ ] Make a whitelist of (instruction, encodings) that are acceptable, so it handles most of the cases where this optimization is useful.
- [ ] Once #943 is merged, add back the uextend in the wasm generation of ExtractLaneU, and fold it back in this new peephole optimization pass.


lars-t-hansen:
This is a fine stopgap solution.  In general however we want something table-driven (ie a machine-generated automaton) that runs during machine code generation and acts on the actual encodings emitted.  Such an automaton can likely subsume other optimizations, eg the shrinking pass.

bnjbvr:
I'll need to re-work this a bit, and think about it (maybe publicly as part of an issue). Closing since there are no good reasons to keep this WIP PR open for too long.

PR #847, opened 2019-07-12
Add support for some serde serialization

Mrowqa:
Needed for wasmtime compilation cache.

### DIFF
--- a/cranelift-codegen/Cargo.toml
+++ b/cranelift-codegen/Cargo.toml
@@ -20,6 +20,7 @@ failure_derive = { version = "0.1.1", default-features = false }
 hashmap_core = { version = "0.1.9", optional = true }
 target-lexicon = { version = "0.4.0", default-features = false }
 log = { version = "0.4.6", default-features = false }
+serde = { version = "1.0.94", features = ["derive"], optional = true }
 # It is a goal of the cranelift-codegen crate to have minimal external dependencies.
 # Please don't add any unless they are essential to the task of creating binary
 # machine code. Integration tests that need external dependencies can be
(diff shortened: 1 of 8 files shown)


cranelift-codegen/Cargo.toml
```diff
@@ -58,6 +59,9 @@ arm32 = []
 arm64 = []
 riscv = []
 
+# For dependent crates that want to serialize some parts of cranelift
+cranelift-serde = ["serde"]
```
sunfishcode: Would you mind renaming this to `enable-serde`, so that it isn't confused with the `cranelift-serde` crate?

cranelift-entity/src/map.rs
```diff
@@ -56,11 +56,21 @@ where
         }
     }
 
+    /// Returns the number of elements in the underlying vector.
+    pub fn len(&self) -> usize {
+        self.elems.len()
+    }
```
sunfishcode: Instead of using `.len()`, can the code that would use this use `.keys().size_hint()` instead?

cranelift-codegen/Cargo.toml
Mrowqa: Will do.

cranelift-entity/src/map.rs
Mrowqa: Well, the exact length should be known for the serializer API. But I'll check whether currently used serializer requires the size up front.

Mrowqa: It's not required. I have removed the `.len()` method.

Mrowqa: I haven't noticed it, since cache system fails quietly. The size information is required.

sunfishcode: Ok, then let's go ahead and add `len()` back then. But please add a comment mentioning that the length of a `SecondaryMap` is not necessarily the same as the length of the corresponding `PrimaryMap`.

sunfishcode:
Looks good!

PR #848, opened 2019-07-12
Do not import libc on windows

Mrowqa:
There are warnings about unused imports on Windows.

### DIFF
--- a/cranelift-simplejit/src/memory.rs
+++ b/cranelift-simplejit/src/memory.rs
@@ -1,8 +1,7 @@
 #[cfg(not(feature = "selinux-fix"))]
 use errno;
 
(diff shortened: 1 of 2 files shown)


cranelift-simplejit/src/memory.rs
```diff
@@ -2,6 +2,7 @@
 use errno;
 
 #[cfg(not(feature = "selinux-fix"))]
+#[cfg(not(windows))]
```
sunfishcode: Would it work to combine this with the previous line, as `#[cfg(not(any(feature = "selinux-fix", windows)))]`?

Mrowqa: I considered it in the first place, but current approach seemed to me a little bit more readable. Will fix it.

PR #855, opened 2019-07-17
Implement extractlane instruction for x86

abrown:
The primary commit for this is 9785535 but the following three are improvements from comments to #833 or suggested by the IDE (9a036ec).

### DIFF
--- a/cranelift-codegen/meta/src/cdsl/instructions.rs
+++ b/cranelift-codegen/meta/src/cdsl/instructions.rs
@@ -181,8 +181,17 @@ impl Instruction {
         bind_ref(self.clone(), Some(reference_type.into()), Vec::new())
     }
 
-    pub fn bind_vector(&self, lane_type: impl Into<LaneType>, num_lanes: u64) -> BoundInstruction {
-        bind_vector(self.clone(), lane_type.into(), num_lanes, Vec::new())
+    pub fn bind_vector_from_lane(
+        &self,
+        lane_type: impl Into<LaneType>,
+        vector_size_in_bits: u64,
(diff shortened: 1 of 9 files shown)


abrown:
@sunfishcode can you take a look at this?

cranelift-codegen/meta/src/cdsl/instructions.rs
```diff
@@ -177,8 +177,17 @@ impl Instruction {
         bind(self.clone(), Some(lane_type.into()), Vec::new())
     }
 
-    pub fn bind_vector(&self, lane_type: impl Into<LaneType>, num_lanes: u64) -> BoundInstruction {
-        bind_vector(self.clone(), lane_type.into(), num_lanes, Vec::new())
+    pub fn bind_vector_from_lane(
+        &self,
+        lane_type: impl Into<LaneType>,
+        vector_size_in_bits: u64,
```
abrown: @sunfishcode, this is how I parameterized binding of vectors like I was talking about in https://github.com/CraneStation/cranelift/pull/868#discussion_r307072094.

abrown:
As I was playing with `clif-util run` from #872 I realized that the recipe for `extractlane` could not be re-used for `x86_pushd`--the output operand for `extractlane` must be a `gpr` and the order of the registers passed to `modrm_rr` must be flipped. I fixed the tests to catch this and rebased the fixes in.

abrown:
@arunetm, here is an `extractlane` implementation for x86.

abrown:
@sunfishcode this one is also rebased and ready for review

PR #861, opened 2019-07-22
cranelift-wasm hooks to instrument wasm operators

iximeow:
In some circumstances we'd like to insert instructions, and update translator-internal state, based on the wasm operations being translated. This adds a pair of functions to `FuncEnvironment` to support that (and adjusts `func_translator` to call them appropriately)

### DIFF
--- a/cranelift-wasm/src/func_translator.rs
+++ b/cranelift-wasm/src/func_translator.rs
@@ -206,7 +206,9 @@ fn parse_function_body<FE: FuncEnvironment + ?Sized>(
     while !state.control_stack.is_empty() {
         builder.set_srcloc(cur_srcloc(&reader));
         let op = reader.read_operator()?;
-        translate_operator(op, builder, state, environ)?;
+        environ.before_translate_operator(&op, builder)?;
+        translate_operator(&op, builder, state, environ)?;
+        environ.after_translate_operator(&op, builder)?;
     }
 
     // The final `End` operator left us in the exit block where we need to manually add a return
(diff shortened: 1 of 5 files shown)


cranelift-wasm/src/func_translator.rs
```diff
@@ -206,7 +206,9 @@ fn parse_function_body<FE: FuncEnvironment + ?Sized>(
     while !state.control_stack.is_empty() {
         builder.set_srcloc(cur_srcloc(&reader));
         let op = reader.read_operator()?;
-        translate_operator(op, builder, state, environ)?;
+        environ.before_translate_operator(&op, builder)?;
+        translate_operator(&op, builder, state, environ)?;
```
iximeow: `op` being used in `{before,after}_translate_operator` is what necessitates the bulk of this PR, by changing to `&op` here.

cranelift-wasm/src/state.rs
```diff
@@ -130,8 +130,13 @@ impl ControlStackFrame {
 /// - The depth of the two unreachable control blocks stacks, that are manipulated when translating
 ///   unreachable code;
 pub struct TranslationState {
+    /// A stack of values corresponding to the active values in the input wasm function at this
```
iximeow: I was initially thinking we'd want to expose the function's corresponding `TranslationState` (in case a user of cranelift-wasm wanted to add/modify values or the control stack), but that ended up not being necessary in how I'm using this right now. `#[deny(missing_docs)]` got me to fill in docs here and I figure they don't hurt to keep around :)

It seems like tweaking `TranslationState` would be a very typical `{before,after}_translate_operator` use case, so passing along `TranslationState` might make sense even if not strictly necessary in my use right now

PR #862, opened 2019-07-22
Generate basic blocks for Wasm `br_if` statements.

sstangl:
This is a very simple transformation. Most of the work was just getting a pipeline set up so that I could actually check that the changes were working. I included some other commits, which also make error messages significantly more useful.

This changes:
```
...
@00b3 [RexOp2urm_noflags#4b6,%rcx]  v13 = bint.i32 v12
@00b6 [RexOp1r_ib#4083,%rcx]        v15 = band_imm v13, 1
@00b8 [RexOp1tjccb#74]              brz v15, ebb3 *****
@00bc [RexOp1pu_id#b8,%rcx]         v18 = iconst.i32 7
@00be [RexOp1umr#89,%rdx]           v29 = uextend.i64 v7
@00be [RexOp1ld#808b,%rbx]          v30 = load.i64 notrap aligned readonly v1
@00be [RexOp1stWithIndexDisp8#89]   store_complex v18, v30+v29+12
@00c1 [Op1jmpb#eb]                  jump ebb2
...
```

Into:
```
...
@00c9 [RexOp2urm_noflags#4b6,%rcx]  v13 = bint.i32 v12
@00d4 [RexOp1r_ib#4083,%rcx]        v15 = band_imm v13, 1
@00de [RexOp1tjccb#74]              brz v15, ebb3
@00de [-]                           fallthrough ebb4

                                ebb4:
@00e0 [RexOp1pu_id#b8,%rcx]         v18 = iconst.i32 7
@00e8 [RexOp1umr#89,%rdx]           v29 = uextend.i64 v7
@00e8 [RexOp1ld#808b,%rbx]          v30 = load.i64 notrap aligned readonly v1
@00e8 [RexOp1stWithIndexDisp8#89]   store_complex v18, v30+v29+12
@00eb [Op1jmpb#eb]                  jump ebb2
...
```

Should be very easy to review. I kept it feature-gated because we haven't determined the effect on the various codegen passes yet.

### DIFF
--- a/cranelift-frontend/src/frontend.rs
+++ b/cranelift-frontend/src/frontend.rs
@@ -479,15 +479,18 @@ impl<'a> FunctionBuilder<'a> {
             "all blocks should be filled before dropping a FunctionBuilder"
         );
 
-        // Check that all blocks are valid basic blocks.
+        // In debug mode, check that all blocks are valid basic blocks.
         #[cfg(feature = "basic-blocks")]
-        debug_assert!(
-            self.func_ctx
-                .ebbs
-                .keys()
-                .all(|ebb| self.func.is_ebb_basic(ebb).is_ok()),
-            "all blocks should be encodable as basic blocks"
-        );
+        #[cfg(debug_assertions)]
+        {
+            // Iterate manually to provide more helpful error messages.
+            for ebb in self.func_ctx.ebbs.keys() {
+                if let Err((inst, _msg)) = self.func.is_ebb_basic(ebb) {
+                    let inst_str = self.func.dfg.display_inst(inst, None);
+                    panic!("{} failed basic block invariants on {}", ebb, inst_str);
+                }
+            }
+        }
 
         // Clear the state (but preserve the allocated buffers) in preparation
         // for translation another function.
--- a/cranelift-frontend/src/frontend.rs
+++ b/cranelift-frontend/src/frontend.rs
@@ -41,8 +41,17 @@ pub struct FunctionBuilder<'a> {
 
 #[derive(Clone, Default)]
 struct EbbData {
-    filled: bool,
+    /// An Ebb is "pristine" iff no instructions have been added since the last
+    /// call to `switch_to_block()`.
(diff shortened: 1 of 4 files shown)


nbp:
Just looking at the given example …

@lars-t-hansen explicitly stated that we should ban `fallthrough` until the really end of the pipeline, as we might introduce new block in the middle.

cranelift-frontend/src/frontend.rs
```diff
@@ -41,8 +41,17 @@ pub struct FunctionBuilder<'a> {
...
-    filled: bool,
+    /// An Ebb is "pristine" iff no instructions have been added since the last
+    /// call to `switch_to_block()`.
     pristine: bool,
+
+    /// An Ebb is "filled" iff a terminator instruction has been inserted since
+    /// the last call to `switch_to_block()`.
+    ///
+    /// A filled block cannot be pristine.
+    filled: bool,
```
nbp: follow-up:
If a filled block cannot be pristine, it sound to me that these 2 flag can represent a state which is not allowed to exists.
Shouldn't an `enum` better describe this state.

```rust
enum EbbState {
  UnChanged,
  Changed,
  ChangedAndTerminated,
}
```

sstangl:
> Just looking at the given example …
> 
> @lars-t-hansen explicitly stated that we should ban `fallthrough` until the really end of the pipeline, as we might introduce new block in the middle.

Yeah, the output there is after all pipelines have run. It's emitted as a `jump`, and the last pass turns that into a `fallthrough`.

cranelift-frontend/src/frontend.rs
sstangl: That's a great observation! I'll look into that. I suspect the enum would work and be clearer.

sstangl: `is_pristine()` and `is_filled()` appears to be part of the public API for `cranelift-frontend`, so cleaning that up is a bit of a bigger thing. I'll file an issue for tracking.

PR #866, opened 2019-07-24
Skip non-branching instructions when using basic blocks.

nbp:
This changes are replacing the iteration over all instructions by an iteration over the last 2 instructions of basic blocks, when `basic-blocks` feature is enabled.

I am confident this change is likely improving performances. However, until the front-end is completely fixed to generate valid basic blocks, we have no way to check the performances of this code on benchmarks.

I tried running the test suite, but it takes less than a second, which does not give me much confidence.

### DIFF
--- a/cranelift-codegen/src/cfg_printer.rs
+++ b/cranelift-codegen/src/cfg_printer.rs
@@ -42,7 +42,7 @@ impl<'a> CFGPrinter<'a> {
         for ebb in &self.func.layout {
             write!(w, "    {} [shape=record, label=\"{{{}", ebb, ebb)?;
             // Add all outgoing branch instructions to the label.
-            for inst in self.func.layout.ebb_insts(ebb) {
+            for inst in self.func.layout.ebb_likely_branches(ebb) {
                 let idata = &self.func.dfg[inst];
                 match idata.analyze_branch(&self.func.dfg.value_lists) {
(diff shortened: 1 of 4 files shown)


sstangl:
Might this be a good time to introduce the formal notion of a BranchGroup?

Mostly I'm worried that this 2-branch invariant is being hardcoded too far away from the Verifier code that ensures it. It would be good for those two places to share some vocabulary.

bnjbvr:
@nbp Hi, what's the status of this PR? No pressure here, just trying to understand where we're at, and if we'd want to open an issue describing the work that's done and that's remaining, instead of keeping an open WIP PR.

bnjbvr:
@nbp ping? (see previous comment)

nbp:
I think this PR was quite complete at the time, it was only a draft as it depend on the "new" bb.

eqrion:
> This looks good to me. @nbp do you have time to rebase this? If not I can take this and rebase it.

This will require a change now that #1363 has been merged, as the feature gate has been removed.

eqrion:
> > This looks good to me. @nbp do you have time to rebase this? If not I can take this and rebase it.
> 
> This will require a change now that #1363 has been merged, as the feature gate has been removed.

I'm going to work on this after #1365 is merged. I don't this commit will apply at all after then, so I'll probably end up rewriting it and attributing @nbp.

PR #868, opened 2019-07-24
Implement vconst instruction for x86

abrown:
This PR also adds a simple constant pool implementation based on the discussion in https://github.com/abrown/cranelift/pull/2. Some form of constant pool is necessary because vector constants of 128-bits (the current cranelift vector size) must be emitted separately--in `rodata` memory--for the x86 instructions to move them in to the XMM registers.

This PR depends on commits waiting to be merged in #855.


cranelift-codegen/src/ir/entities.rs
```diff
@@ -209,6 +227,8 @@ pub enum AnyEntity {
     StackSlot(StackSlot),
     /// A Global value.
     GlobalValue(GlobalValue),
+    /// A constant.
+    Constant(Constant),
```
sunfishcode: I'll grant that this isn't clearly documented, but right now `AnyEntity` is only used for those entities which show up in clif IR, and it's mostly used for reporting errors where the user is expected to understand the index value. With constants, the syntax here makes the indices implicit. Do you actually need `AnyEntity::Constant`, or is this just something that looked needed?

cranelift-codegen/src/ir/function.rs
```diff
@@ -110,6 +110,7 @@ impl Function {
         self.encodings.clear();
         self.locations.clear();
         self.offsets.clear();
+        self.jt_offsets.clear();
```
sunfishcode: Wow, good catch. Yes, this is one of the downsides of the current scheme of clearing data structures between functions :-/.

cranelift-codegen/meta/src/cdsl/instructions.rs
```diff
@@ -1081,9 +1081,9 @@ fn bind(
 fn bind_vector(
     inst: Instruction,
     lane_type: LaneType,
-    num_lanes: u64,
     mut value_types: Vec<ValueTypeOrAny>,
 ) -> BoundInstruction {
+    let num_lanes = 128 / lane_type.lane_bits();
```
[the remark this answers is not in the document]
abrown: Well, I was doing the `128 / lane_bits` calculation all over the place so at least now it is in one place. I could replace `128` with a static variable?

cranelift-codegen/src/ir/entities.rs
abrown: Just following the documentation at line 215: `A reference to any of the entities defined in this module`. Perhaps I should change this to `A reference to any of the entities defined in this module that can appear in CLIF IR`?

abrown: ...and I just checked `./test-all.sh` and I don't see any failures when I remove `Constant` from `AnyEntity`

cranelift-codegen/meta/src/cdsl/instructions.rs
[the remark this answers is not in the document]
sunfishcode: I guess, it was one thing when it was in the x86 code, close to the places that will need changes anyway when we do AVX2. Here, it's in platform-independent code, so it feels more out of place.

Would it help to add a `num_lanes` function, next to `lane_bits`, where you could pass in 128 and it returns that width divided by the `lane_bits()` value?

Alternatively, maybe we could keep the code as you have it here, and just rename `bind_vector` to `bind_vector128` or so, so that it's clear on the caller side that this is for 128-bit vectors. Does that make sense?

cranelift-codegen/src/ir/entities.rs
sunfishcode: Yes, if you could make that doc change, that sounds good.

cranelift-codegen/meta/src/cdsl/instructions.rs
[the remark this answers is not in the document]
abrown: I decided for a variation on the first option: I kept the calculation where it is but added a parameter `vector_size_in_bits` to `bind_vector_from_lane` which should future-proof this. That change is really a part of #855 (a smaller PR than this constant-related PR) so I will add a comment there about this.

cranelift-codegen/src/ir/entities.rs
abrown: Done in the rebase.

cranelift-codegen/src/ir/function.rs
bjorn3: Maybe add a macro to automatically create the `clear` functions from the struct def?

abrown:
After talking to @yurydelendik (thanks for the help!) and debugging the execution of these instructions in wasmtime, I realized that the relocation offsets I initially implemented for constants were not going to work. Not understanding how wasmtime relocated jump tables, I mirrored the jump table could as much as I could. After talking to @yurydelendik I realized that wasmtime is currently not relocating the jump tables from where they are emitted in `MemorySink` (i.e. right after the function body). Thus, the offsets computed against `%rip` in cranelift's codegen are not modified by wasmtime and must be correct when they are emitted by cranelift. Understanding this I then modified the constant pool code to have correct offsets (not offsets from the beginning of the constant pool but offsets from the beginning of the function) in `relaxation.rs`, just like the jump tables are doing. Now I can run SIMD examples in wasmtime such as:

```
(module
  (func $test_splat (result i32)
    i32.const 42
    i32x4.splat
    i32x4.extract_lane 0
  )

  (func $test_insert_lane (result i32)
      v128.const i64x2 0 0
      i32.const 99
      i32x4.replace_lane 1
      i32x4.extract_lane 1
  )

  (func $test_const (result i32)
    v128.const i32x4 1 2 3 4
    i32x4.extract_lane 3
  )

  ;; exporting the function allows us to call it with `wasmtime --invoke=...`
  (export "test_splat" (func $test_splat))
  (export "test_insert_lane" (func $test_insert_lane))
  (export "test_const" (func $test_const))
)
```

by running the following:

```
cargo +nightly build && \
  wat2wasm --enable-simd --debug-names ../simd.wat -o ../simd.wasm && \
  target/debug/wasmtime ../simd.wasm -g --enable-simd --invoke=test_insert_lane
```

The breaking change from this PR that wasmtime (and other cranelift users) will have to absorb is the new `RelocSink` method I added--`reloc_constant`. For the reasons I describe above, however, I currently implement this as an empty method in wasmtime. That being the case, perhaps `RelocSink::reloc_constant` is currently unnecessary? Or at the very least, perhaps the parameters should change from `code_offset: CodeOffset, reloc: Reloc, constant_offset: ConstantOffset` (where `constant_offset` is the offset from the beginning of the function to the constant value) to `code_offset: CodeOffset, reloc: Reloc, constant: Constant` (where `constant` is an entity that allows the user to look up the constant inside the `ConstantPool` implementation).

cranelift-wasm/src/code_translator.rs
```diff
@@ -917,30 +917,46 @@ pub fn translate_operator<FE: FuncEnvironment + ?Sized>(
         | Operator::TableSize { .. } => {
             return Err(WasmError::Unsupported("proposed bulk memory operators"));
         }
+        Operator::V128Const { value } => {
+            let constant = builder.func.dfg.constants.insert(value.bytes().to_vec());
+            state.push1(builder.ins().vconst(I32X4, constant)) // TODO need to decide what to do about I32X4 here; the WASM spec has V128
```
abrown: @bnjbvr and I talked about this but probably others need to weigh in (@sunfishcode?): the SIMD spec talks about a `v128.const` instruction but we do not have anything exactly equivalent to `v128` in cranelift. Some options:
 - use a default (e.g. see `I32X4` here) and then legalize later to raw_bitcast this value to the next instruction that uses it
 - use `BVType` (see https://github.com/CraneStation/cranelift/blob/2b1cfea1634c58d22b880eb082c748350041c3f0/cranelift-codegen/meta/src/cdsl/types.rs#L395-L399)
 - use `I128` from #795 with some legalization
 - add a new `V128` type

cranelift-codegen/src/ir/immediates.rs
```diff
@@ -316,15 +316,15 @@ impl FromStr for Uimm128 {
             Err("Expected a hexadecimal string, e.g. 0x1234")
         } else if s.len() % 2 != 0 {
             Err("Hexadecimal string must have an even number of digits")
-        } else if s.len() > 32 {
+        } else if s.len() > 34 {
```
bjorn3: ???

abrown: 2 characters for `0x` plus 32 characters (16 x 2) for the data bytes

bjorn3: Got it.

abrown:
@arunetm, here is a `const` implementation for x86.

cranelift-wasm/src/code_translator.rs
abrown: From talking to @sunfishcode: the current best decision is to add bitcasts from-to `I8x16`.

abrown: This is resolved by 
d76a85e

abrown:
@sunfishcode, @bnjbvr: as we discussed, I added code in cranelift-wasm to keep track of the values from `v128.const` that are defaulting to `I8x16`; immediately before the value is used (in the few implemented SIMD instructions) I `raw_bitcast` the value to the appropriate type if necessary.

abrown:
@sunfishcode rebased this as well and is ready for review; as a reminder, we will have to update `RelocSink` in wasmtime when we publish this version of cranelift and update in wasmtime (see https://github.com/CraneStation/wasmtime/pull/238).

cranelift-wasm/src/code_translator.rs
```diff
@@ -1287,3 +1306,441 @@ fn translate_br_if_args(
...
+        | Operator::I64AtomicRmw16UXchg { .. }
+        | Operator::I64AtomicRmw32UXchg { .. }
+        | Operator::I32AtomicRmwCmpxchg { .. }
+        | Operator::I64AtomicRmwCmpxchg { .. }
+        | Operator::I32AtomicRmw8UCmpxchg { .. }
+        | Operator::I32AtomicRmw16UCmpxchg { .. }
+        | Operator::I64AtomicRmw8UCmpxchg { .. }
+        | Operator::I64AtomicRmw16UCmpxchg { .. }
+        | Operator::I64AtomicRmw32UCmpxchg { .. } => unimplemented!(
+            "Currently only the SIMD instructions are translated to their return type"
```
sunfishcode: Would it work to use a `_` pattern here, so that we don't have to name all the non-SIMD operators explicitly here? Also, it'd be convenient to include the a `{:?}` with the operator in the unimplemented message here, so that someone can quickly see what instruction is missing.

cranelift-wasm/src/state.rs
```diff
@@ -179,6 +179,11 @@ pub struct TranslationState {
     // `FuncEnvironment::make_direct_func()`.
     // Stores both the function reference and the number of WebAssembly arguments
     functions: HashMap<FuncIndex, (ir::FuncRef, usize)>,
+
+    // Track values that are introduced with SIMD's v128.const instruction; because cranelift has no
+    // v128 type, we type these values as I8x16 and use this mapping to raw_bitcast to the correct
+    // type once the value is used
+    constants: HashMap<Value, bool>,
```
sunfishcode: Does the same problem apply to `load`? Would it work to just unconditionally bitcast any SIMD value that doesn't have the desired SIMD type?

cranelift-codegen/meta/src/isa/x86/encodings.rs
```diff
@@ -1694,13 +1746,20 @@ pub fn define(
...
-                .bind_vector(to_type, 128 / to_type.lane_bits())
-                .bind_vector(from_type, 128 / from_type.lane_bits());
+                .bind_vector_from_lane(to_type, sse_vector_size)
+                .bind_vector_from_lane(from_type, sse_vector_size);
             e.enc32_rec(instruction.clone(), rec_null_fpr, 0);
             e.enc64_rec(instruction, rec_null_fpr, 0);
         }
     }
 
+    // SIMD vconst using MOVUPS
```
sunfishcode: Can you also comment that we use `MOVUPS` because we don't currently have a way to align the constant pools?

cranelift-wasm/src/state.rs
abrown: Yeah, in my work-ahead branch I basically started doing just that--when functions return a `v128` then the original type of any SIMD instruction is lost and must be re-cast to the expected type when it is next used. I will remove this commit and add the code I have in my other branch.

cranelift-codegen/src/binemit/relaxation.rs
```diff
@@ -135,7 +136,11 @@ pub fn relax_branches(
     let jumptables_size = offset - jumptables;
     let rodata = offset;
 
-    // TODO: Once we have constant pools we'll do some processing here to update offset.
+    let constants: Vec<Constant> = func.dfg.constants.iter().map(|(h, _)| h.clone()).collect();
+    for constant_handle in constants {
+        func.dfg.constants.set_offset(constant_handle, offset);
+        offset += func.dfg.constants.get(constant_handle).len() as u32
+    }
```
sunfishcode: This took a bit of unpacking. We iterate over `func.dfg.constants`, do a `.map` to discard the values and just get the keys, and we have to `.collect()` them because we later want to mutate the values (which hold the values we just discarded), which we look up using the keys.

Would it be possible to simplify this by:
 - give `ConstantPool` an `entries_mut()` function which returns the same thing as `iter()` but hands out the full `&mut ConstantPoolEntry` as values
 - give `ConstantPoolEntry` its own `set_offset` and `len` functions, which are the two things that `relax_branches` here needs to do
?

cranelift-codegen/src/binemit/relaxation.rs
```diff
@@ -135,7 +136,11 @@ pub fn relax_branches(
     let jumptables_size = offset - jumptables;
     let rodata = offset;
 
-    // TODO: Once we have constant pools we'll do some processing here to update offset.
+    let constants: Vec<Constant> = func.dfg.constants.iter().map(|(h, _)| h.clone()).collect();
+    for constant_handle in constants {
+        func.dfg.constants.set_offset(constant_handle, offset);
+        offset += func.dfg.constants.get(constant_handle).len() as u32
```
sunfishcode: We should use `try_from()` to convert to `u32` here. It may not be possible to create a 4 GiB constant today, but if anyone ever could, silent wraparound here would be a subtle bug.

cranelift-codegen/src/ir/constant.rs
```diff
@@ -0,0 +1,203 @@
...
+    fn new(data: ConstantData) -> Self {
+        ConstantPoolEntry { data, offset: None }
+    }
+}
+
+/// Maintains the mapping between a constant handle (i.e. [Constant](ir::entities::Constant)) and
+/// its constant data (i.e. [ConstantData](ir::constant::ConstantData))
+#[derive(Clone)]
+pub struct ConstantPool {
+    // this mapping maintains the insertion order as long as Constants are created with sequentially increasing integers
```
sunfishcode: When a comment forms a substantial sentence, please capitalize it and end it with a period.

cranelift-codegen/src/ir/constant.rs
```diff
@@ -0,0 +1,203 @@
...
+    handles_to_values: BTreeMap<Constant, ConstantPoolEntry>,
+    // this mapping is unordered (no need for lexicographic ordering) but allows us to map constant data back to handles
```
sunfishcode: ditto :)

PR #781, opened 2019-05-28
Stackmaps

sunfishcode:
This is the continuation of https://github.com/CraneStation/cranelift/pull/440, rebased and updated to work with Cranelift master.

It contains the basic components of [reference types](https://github.com/WebAssembly/reference-types) support and stackmaps:
 - reference types
 - `isnull` and `null` instructions
 - stackmap instructions indicating places to emit stack maps
 - a stackmaps pass to add stackmap instructions for live reference variables: cranelift-codegen/src/regalloc/stackmaps.rs

PR #871, opened 2019-07-25
Add reference types to Cranelift

c27kwan:


cranelift-codegen/meta/src/cdsl/types.rs
```diff
@@ -31,6 +32,7 @@ static _RUST_NAME_PREFIX: &'static str = "ir::types::";
 pub enum ValueType {
     BV(BVType),
     Lane(LaneType),
+    Reference(ReferenceType),
```
bjorn3: Maybe put this in `LaneType` instead.

c27kwan: We decided not to put reference types as a LaneType because LaneTypes are vectorizable while references are not. We need a logical separation for parts such as gen_types.rs:33 which generates vectors by iterating through all LaneTypes.

cranelift-simplejit/src/backend.rs
```diff
@@ -492,13 +510,13 @@ fn lookup_with_dlsym(name: &str) -> *const u8 {
 
         for handle in &handles {
             let addr = winapi::um::libloaderapi::GetProcAddress(*handle, c_str_ptr);
-            if addr.is_null() {
+            if addr.isnull() {
```
philipc: All three of these `is_null` changes seem wrong to me. Note that these are only in code used for windows builds.

sunfishcode: Good catch; these are likely from a search-and-replace that happened quite a while ago to rename the wasm instruction, but these aren't talking about the wasm instruction, so we should change these back.

sunfishcode:
Looks good! Other than the issue @philipc pointed out, this is in good shape!

cranelift-simplejit/src/backend.rs
c27kwan: Yeah, good catch. The change was carried over from #781. Removed in the latest commit.

cranelift-codegen/meta/src/shared/instructions.rs
```diff
@@ -1366,7 +1366,7 @@ pub fn define(
         Several Argument Types.
```
bjorn3: This doesnt make sense as description of what `stack_map` is.

cranelift-codegen/meta/src/cdsl/typevar.rs
```diff
@@ -406,6 +410,7 @@ pub struct TypeSet {
     pub ints: NumSet,
     pub floats: NumSet,
     pub bools: NumSet,
+    pub references: NumSet,
```
bnjbvr: (light suggestion, feel free to ignore if you think it's best at the moment): I think it would be nice to use "refs" instead of "references", the same way we use "ints" instead of "integers".

cranelift-codegen/meta/src/shared/types.rs
```diff
@@ -145,6 +145,38 @@ impl Iterator for FlagIterator {
     }
 }
 
+#[derive(Debug, Clone, Copy, Eq, PartialEq, Hash)]
+pub enum Reference {
+    /// 32-bit reference.
+    R32 = 32,
+    /// 64-bit int.
```
bnjbvr: nit: not "int" but "reference"

cranelift-codegen/src/ir/instructions.rs
```diff
@@ -652,6 +656,7 @@ mod tests {
             ints: BitSet8::from_range(4, 7),
             floats: BitSet8::from_range(0, 0),
             bools: BitSet8::from_range(3, 7),
+            refs: BitSet8::from_range(4, 7),
```
bnjbvr: Can you add assertions on what you expect below, next to the rest of the assertions?

cranelift-codegen/src/ir/instructions.rs
```diff
@@ -668,6 +673,7 @@ mod tests {
             ints: BitSet8::from_range(0, 0),
             floats: BitSet8::from_range(5, 7),
             bools: BitSet8::from_range(3, 7),
+            refs: BitSet8::from_range(0, 0),
```
bnjbvr: ditto (and a few times below as well)

cranelift-codegen/src/ir/types.rs
```diff
@@ -106,8 +106,8 @@ impl Type {
         self.replace_lanes(match self.lane_type() {
             B8 | I8 => B8,
             B16 | I16 => B16,
-            B32 | I32 | F32 => B32,
-            B64 | I64 | F64 => B64,
+            B32 | I32 | F32 | R32 => B32,
+            B64 | I64 | F64 | R64 => B64,
```
bnjbvr: We should probably assert there aren't ref types in here, instead (otherwise, this means we're trying to resolve operand constraints for a derived type variable, which base contains ref types, and derivation is as_bool: we're trying to convert ref types to bools!).

bnjbvr:
Also, wanted to point out that you can give attribution to several authors on Github commits with [this feature](https://help.github.com/en/articles/creating-a-commit-with-multiple-authors), if you're in the mood for doing so.

cranelift-codegen/meta/src/cdsl/types.rs
```diff
@@ -515,3 +532,84 @@ impl Iterator for SpecialTypeIterator {
         }
     }
 }
+
+/// Reference type is scalar type, but not lane type
```
bjorn3:
```suggestion
/// A reference type is a scalar type, but not a lane type.
```

cranelift-codegen/meta/src/shared/instructions.rs
```diff
@@ -1318,6 +1353,26 @@ pub fn define(
...
+        "Variable number of args for Stackmap",
+    );
+
+    ig.push(
+        Inst::new(
+            "savepoint",
+            r#"
+        Several Argument Types.
+
+        This instruction will provide the values that are live at some point
```
bnjbvr: nit: not all the values, should only be reference values.

cranelift-codegen/meta/src/shared/instructions.rs
```diff
@@ -1318,6 +1353,26 @@ pub fn define(
...
 
+    let N = &operand_doc(
+        "args",
+        variable_args,
+        "Variable number of args for Stackmap",
+    );
+
+    ig.push(
+        Inst::new(
+            "savepoint",
```
bnjbvr: I am slightly wondering if there is a typo here and if this name was supposed to be "safepoint", since it's usually more used in the compiler literature for such an instruction.

cranelift-codegen/meta/src/shared/instructions.rs
```diff
@@ -2581,6 +2636,23 @@ pub fn define(
         .operands_out(vec![a]),
     );
 
+    let a = &operand("a", b1);
+    let x = &operand("x", Ref);
+
+    ig.push(
+        Inst::new(
+            "isnull",
```
bnjbvr: light preference to name this `is_null`, as we have other instances of names like that ("load_complex", etc.).

cranelift-codegen/src/verifier/mod.rs
```diff
@@ -1672,9 +1672,12 @@ impl<'a> Verifier<'a> {
         // Instructions with side effects are not allowed to be ghost instructions.
         let opcode = self.func.dfg[inst].opcode();
 
-        // The `fallthrough` and `fallthrough_return` instructions are marked as terminators and
-        // branches, but they are not required to have an encoding.
-        if opcode == Opcode::Fallthrough || opcode == Opcode::FallthroughReturn {
+        // The `fallthrough`, `fallthrough_return`, and `stackmap` instructions are not required
```
bnjbvr: nit: this comment seems to refer to the instruction's previous name.

cranelift-simplejit/src/memory.rs
```diff
@@ -94,7 +94,7 @@ impl PtrLen {
                 PAGE_READWRITE,
             )
         };
-        if !ptr.is_null() {
+        if !ptr.isnull() {
```
bnjbvr: Probably bad search-and-replace here too.

cranelift-simplejit/src/backend.rs
```diff
@@ -492,13 +492,13 @@ fn lookup_with_dlsym(name: &str) -> *const u8 {
...
                 continue;
             }
             return addr as *const u8;
         }
 
-        let msg = if handles[1].is_null() {
+        let msg = if handles[1].isnull() {
```
bnjbvr: (bad search-and-replace here too; it'd be nice that if there's been new changes cancelling those, they'd be included in this same commit instead, to make each commit atomic and testable independently from the others, which is quite useful when using `git bisect` for instance.)

cranelift-codegen/meta/src/isa/x86/recipes.rs
```diff
@@ -2795,6 +2853,25 @@ pub fn define<'shared>(
                     "#,
                 ),
         );
+
+        recipes.add_template_recipe(
+            EncodingRecipeBuilder::new("icscc_id_ref", f_unary, 5 + 3)
```
bnjbvr: Since icscc_i{b,d}_ref are unguarded by predicates, at least one of the associated encodings has to be unreachable in the encodings interpreter. Can you check this (= only one of the two recipes is actually used), and if that's true, remove this one? (which takes more memory)

cranelift-codegen/meta/src/isa/x86/encodings.rs
```diff
@@ -1668,5 +1706,36 @@ pub fn define(
...
+    // Sign-extended 32-bit immediate.
+    e.enc64(
+        null.bind_ref(R64),
+        rec_u_id_ref.rex().opcodes(vec![0xc7]).rrr(0).w(),
+    );
+
+    // Finally, the 0xb8 opcode takes an 8-byte immediate with a REX.W prefix.
+    e.enc64(
+        null.bind_ref(R64),
+        rec_pu_iq_ref.opcodes(vec![0xb8]).rex().w(),
```
bnjbvr: There are 4 different encodings for `null.r64`, which are all unguarded (at the recipe or encoding level), so I think some of them are unreachable. Could we remove the largest ones?

cranelift-codegen/meta/src/isa/x86/encodings.rs
```diff
@@ -1668,5 +1706,36 @@ pub fn define(
         }
     }
 
+    // Reference type instructions
+
+    // Null references implemented as iconst 0
+    e.enc32(null.bind_ref(R32), rec_pu_id_ref.opcodes(vec![0xb8]));
+
+    e.enc64(null.bind_ref(R32), rec_pu_id_ref.rex().opcodes(vec![0xb8]));
+    e.enc64(null.bind_ref(R32), rec_pu_id_ref.opcodes(vec![0xb8]));
```
bnjbvr: Hmm, do we want to support references of 32-bits on a 64-bit platform, in general? If so, I think the stack map records should contain the size of the reference value too. (cc @sunfishcode)

bnjbvr:
Ok, I looked at the last commit, which resolves some of the comments I've opened :) If possible, it'd be nice to play a bit with `git rebase` and include the changes from the last commit within the commit that introduced them. (The [git-absorb](https://github.com/tummychow/git-absorb) addon could be useful to help you!)

Great work overall. Two open-ended questions:

- do we want to support 32-bits references on 64-bits platforms, and conversely?
- is there any additional work that's needed in Spidermonkey / Baldrdash to support ref types with the Cranelift backend for wasm?

Thanks!

cranelift-codegen/meta/src/gen_binemit.rs
```diff
@@ -18,10 +18,11 @@ fn gen_recipe(formats: &FormatRegistry, recipe: &EncodingRecipe, fmt: &mut Forma
     let inst_format = formats.get(recipe.format);
     let num_value_ops = inst_format.num_value_operands;
 
-    let want_args = recipe.operands_in.iter().any(|c| match c {
-        OperandConstraint::RegClass(_) | OperandConstraint::Stack(_) => true,
-        OperandConstraint::FixedReg(_) | OperandConstraint::TiedInput(_) => false,
-    });
+    let want_args = inst_format.has_value_list
```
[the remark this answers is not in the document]
c27kwan: In the generated file, we don't have access to the instruction data's `args` when `want_args` is set to false. One of the new instructions added in this change, `savepoint`, needs to access `args` because it tells us the live variables at that point in the program. `savepoint` is a `MultiAry`instruction, but that is insufficient for `want_args` to be set to true since all other `MultiAry` instructions don't need `args`.

This change will generate more dead code than necessary since a lot of instructions have a value list but don't need access to `args`. The Rust compiler should be able to recognise and eliminate dead code, so it's not a major issue. Ideally, there is an alternative way that generates less dead code. Feel free to suggest an alternative or I can add a TODO comment above this change.

cranelift-codegen/src/binemit/stackmap.rs
```diff
@@ -0,0 +1,124 @@
...
+                    _ => false,
+                };
+            }
+        }
+
+        // SpiderMonkey stackmap structure:
+        // <trap reg dump> + <general spill> + <frame> + <inbound args>
+        // Bit vector goes from lower addresses to higher addresses
+
+        // TODO: Get trap register layout from Spidermonkey and prepend to bitvector below.
```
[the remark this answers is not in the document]
c27kwan: We had an in-person meeting about what the `Stackmap` metadata format should be at the last all hands. We want to keep both Spidermonkey and Cranelift modular and minimize the inter-language communication. In the end, I believe @julian-seward1 suggested to pass the trap register dump layout format from SpiderMonkey to Cranelift and to prepend. That way, when we pass the metadata back to Spidermonkey, it's just a bitvector. I believe the rationale was to imitate what we did with Ion/Baldr to keep it consistent. 

@sunfishcode and I discussed the plan for this change in person. In order to have working reference types in Cranelift, we need to make changes in Spidermonkey. Instead of trying to do two patches at the same time, we've decided to have the Spidermonkey patch come in another time and get the first steps towards having reference types in Cranelift merged first. 

Iirc, Dan's plan is to have someone else (who is familiar with Spidermonkey) pick up the remaining work. I left the TODO comment there as a reminder of that this change is incomplete and what remains to be done. I can remove the TODO comment if we don't want it?

c27kwan:
@bnjbvr About `trapnf`: When writing the test files and mocking the trap interrupt that happens in function prologues, I was not able to use the `trap` instruction since it must always be a terminal. Presumably, we use a different kind of trap instruction that is non-terminal (e.g. `trapif` ) when we build the function prologue. After a discussion with @sunfishcode, it didn't seem right to only support traps that will terminate the program or conditional traps . So we added `trapnf` for general purpose traps that don't need to be a terminal in a block.

cranelift-codegen/meta/src/isa/x86/encodings.rs
c27kwan: Maybe we shouldn't? We use a 1 bit = 1 word mapping when we create the `Stackmap` metadata. This is an implicit assumption that references are always 1 word long. If we choose to support 32-bits references, we would need to make them not word-aligned.

cranelift-codegen/meta/src/gen_binemit.rs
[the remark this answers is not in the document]
bnjbvr: Good to know, thanks; can you add a comment that it's for multi-ary instructions, please?

cranelift-codegen/src/binemit/stackmap.rs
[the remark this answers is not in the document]
bnjbvr: Sounds good to me for the plan; we can keep the comment here, even if I think we shouldn't mimic Spidermonkey's bit vector layout here, and have something higher-level for Cranelift's purpose.

bnjbvr:
@c27kwan Thanks for the explanation for `trapnf`, sounds good to me.

cranelift-codegen/src/binemit/stackmap.rs
```diff
@@ -0,0 +1,124 @@
...
+
+        for _i in 0..num_words {
+            vec.push(false);
+        }
+
+        // Frame (includes spills and inbound args)
+        for (ss, ssd) in stack.iter() {
+            if live_ref_in_stack_slot.contains(&ss) {
+                // Assumption: greater magnitude of offset imply higher address
+                let index = (((ssd.offset.unwrap().abs() as u32) / word_size) - 1) as usize;
```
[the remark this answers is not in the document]
bnjbvr: That makes sense. Removing the size of the stack slot data before dividing sounds clearer to me too, thanks for suggesting this.

cranelift-codegen/meta/src/shared/instructions.rs
c27kwan: Hm, @sunfishcode made a commit a while ago specifically to rename the instruction from `is_null` to `isnull`. (https://github.com/CraneStation/cranelift/pull/781/commits/66489aeb5bf2f964f95ece0be7988ca4ac65e5be)

It is probably because it shadowed the other method we have in our codebase (which led to the bad search-and-replace). I think the use of the underscore is inconsistent within Cranelift IR since we also have instructions like `trapif`, `extractlane`, and `debugtrap`. I don't mind changing it back to `is_null`, but after maybe @sunfishcode tell us why he chose to exclude the underscore?

cranelift-codegen/src/binemit/memorysink.rs
```diff
@@ -149,6 +154,12 @@ impl<'a> CodeSink for MemoryCodeSink<'a> {
         self.info.rodata_size = self.offset() - (self.info.jumptables_size + self.info.code_size);
         self.info.total_size = self.offset();
     }
+
+    fn add_stackmap(&mut self, val_list: &[Value], func: &Function, isa: &dyn TargetIsa) {
+        let ofs = self.offset();
+        let sm = Stackmap::from_values(&val_list, func, isa);
```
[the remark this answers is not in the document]
c27kwan: I'll expand `sm` to `stackmap`, but `ofs` is more consistent with the rest of the impl for `CodeSink`. We use `ofs` inside of `reloc_ebb` , `reloc_external`, and `reloc_jt`. In general, the variable names in this impl are very short.

cranelift-faerie/src/backend.rs
```diff
@@ -141,6 +143,8 @@ impl Backend for FaerieBackend {
         total_size: u32,
     ) -> ModuleResult<FaerieCompiledFunction> {
         let mut code: Vec<u8> = vec![0; total_size as usize];
+        // TODO: Replace this with FaerieStackmapSink once it is implemented.
```
philipc: What do you see as the possible uses for this in the faerie backend?

cranelift-codegen/src/binemit/stackmap.rs
```diff
@@ -0,0 +1,124 @@
...
+    pub fn from_values(
+        args: &[ir::entities::Value],
+        func: &ir::Function,
+        isa: &dyn TargetIsa,
+    ) -> Self {
+        let stack = &func.stack_slots;
+        let loc = &func.locations;
+        let mut live_ref_in_stack_slot: std::collections::HashSet<ir::StackSlot> =
+            std::collections::HashSet::new();
+        // References can be in registers, but live registers values are pushed onto the stack before calls and traps.
```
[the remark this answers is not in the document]
c27kwan: I'm not sure what you're referring to, so maybe @julian-seward1 can help answer this.

As far as this change is concerned, I'm imitating the way `Stackmap`s are constructed in Spidermonkey -- which focuses only on the stack. The reason seems to be that registers are already pushed onto the stack by the time we construct the `Stackmap`. Is that also true for Cranelift?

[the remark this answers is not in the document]
bnjbvr: I've discussed with Julian in private. My main worry that if a register contained a Ref, and it was spilled but reused after a safepoint, it could contain some stall ref value (if the GC decided to relocate the value, e.g. in the case of a generational GC).

In Spidermonkey, it is safe to do so because we have our own invariants that wasm stack maps don't need to worry about registers at all. In general for Cranelift, I don't think it's the case, and later (not in this PR), we should implement register maps too. So the implementation is perfect for our needs here. Can you add a comment that we should maybe worry about registers later, please? Thanks!

cranelift-codegen/meta/src/shared/instructions.rs
c27kwan: Yes, this was a very consistent misspelling. Fixed in the latest commit. :)

cranelift-simplejit/src/backend.rs
c27kwan: I've decided to squash all the commits together. I initially separated it into smaller commits so that it would be easier to follow the changes, but it probably makes more sense it have it be one commit overall for merging.

cranelift-codegen/meta/src/shared/settings.rs
```diff
@@ -92,6 +92,18 @@ pub fn define() -> SettingGroup {
         true,
     );
 
+    settings.add_bool(
+        "enable_stackmap_instruction",
```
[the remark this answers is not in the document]
c27kwan: I renamed it to `enable_safepoints` instead.

cranelift-codegen/meta/src/isa/x86/recipes.rs
c27kwan: Talked to @sunfishcode and we decided to create a new recipe `is_zero` that uses less memory. Both icscc_i{b,d}_ref are gone now. :)

cranelift-codegen/meta/src/isa/x86/encodings.rs
c27kwan: Redundant encodings have now been removed.

cranelift-codegen/meta/src/isa/x86/encodings.rs
c27kwan: Talked to @sunfishcode in person. We decided to remove the lines that support 32-bit references on 64-bit platform. It's easier to add them back later if we wish. For now, we don't want that to be a part of this change.

cranelift-codegen/meta/src/cdsl/typevar.rs
```diff
@@ -462,10 +473,12 @@ impl TypeSet {
         let mut copy = self.clone();
         copy.ints = NumSet::new();
         copy.floats = NumSet::new();
+        copy.references = NumSet::new();
         copy.bitvecs = NumSet::new();
         if (&self.lanes - &num_set![1]).len() > 0 {
             copy.bools = &self.ints | &self.floats;
             copy.bools = &copy.bools | &self.bools;
+            copy.bools = &copy.bools | &self.references;
```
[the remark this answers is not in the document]
c27kwan: You're right. The ref-> bool logic is removed in the latest commit.

cranelift-codegen/meta/src/shared/instructions.rs
sunfishcode: I don't recall a strong reason. I'm ok adding the underscore back.

cranelift-faerie/src/backend.rs
sunfishcode: There are use cases for compiling WebAssembly AOT into native binaries, such as to run on small devices where compilation would be prohibitively slow. In such contexts, we'll need the same stackmap support we need for JIT use cases.

philipc: I see, so this will just need to collect the info similar to what is done with traps?

sunfishcode: Yep.

cranelift-codegen/src/regalloc/context.rs
```diff
@@ -192,6 +193,12 @@ impl Context {
         self.coloring
             .run(isa, func, domtree, &mut self.liveness, &mut self.tracker);
 
+        // This function runs after register allocation has taken
+        // place, meaning values have locations assigned already
+        if isa.flags().enable_stackmap_instruction() {
```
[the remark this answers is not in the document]
c27kwan: In the latest push, we prevent the `safepoint` instruction from being used outside of `emit_stackmap` and to panic if reference types are used when safepoints are disabled.

cranelift-codegen/src/binemit/stackmap.rs
```diff
@@ -0,0 +1,122 @@
...
+    pub fn from_values(
+        args: &[ir::entities::Value],
+        func: &ir::Function,
+        isa: &dyn TargetIsa,
+    ) -> Self {
+        let loc = &func.locations;
+        let mut live_ref_in_stack_slot = std::collections::HashSet::new();
+        // References can be in registers, and live registers values are pushed onto the stack before calls and traps.
+        // TODO: Implement register maps. If a register containing a reference is spilled and reused after a safepoint,
+        // it could contain a stall reference value if the garbage collector relocated the value.
```
lars-t-hansen: nit: "stall" => "stale"

filetests/safepoint/basic.clif
```diff
@@ -0,0 +1,55 @@
...
+; nextln:   v10 = copy v0
+; nextln:   jump ebb1(v10)
+; nextln: 
+; nextln: ebb1(v3: i32 [%rax]):
+; nextln:   v8 = iconst.i32 1
+; nextln:   v4 = isub v8, v3
+; nextln:   jump ebb2(v4)
+; nextln: 
+; nextln: ebb2(v5: i32 [%rcx]):
+; nextln:   safepoint v1, v2
```
lars-t-hansen: Out of curiosity, why is v5 not saved here?  It seems live.

bjorn3: I think it only saves references, not other values. For a GC only references matter anyway.

c27kwan: That is correct -- the`safepoint` instruction only saves the live reference values at that point.

cranelift-codegen/meta/src/shared/instructions.rs
```diff
@@ -394,6 +404,17 @@ pub fn define(
         .can_trap(true),
     );
 
+    ig.push(
+        Inst::new(
+            "trapnf",
+            r#"
+        A non-fatal trap.
```
sunfishcode: "trapnf" is pretty similar to "trapnz", such that it could be confused for a kind of conditional trap. The phrase "trap on non-fatal" comes to mind, though of course that doesn't make sense ;-). How about we rename this to "resumable_trap"?

Also, it'd be good to put some of your explanation of what a non-fatal trap is above in the documentation here.

cranelift-codegen/meta/src/shared/instructions.rs
```diff
@@ -85,6 +85,15 @@ pub fn define(
         TypeSetBuilder::new().ints(32..64).build(),
     );
 
+    let Ref = &TypeVar::new(
+        "Ref",
+        "A scalar reference type",
+        TypeSetBuilder::new()
+            .refs(Interval::All)
+            .simd_lanes(Interval::All)
```
bnjbvr: I think this should be removed, otherwise this allows vector of reference types.

filetests/safepoint/basic.clif
lars-t-hansen: > That is correct -- the`safepoint` instruction only saves the live reference values at that point.

Duh, of course.  I've been so concerned about liveness tracking in general that I didn't even think about how gc safepoints are different :-/

PR #872, opened 2019-07-26
Add ability to run CLIF IR using `clif-util run [-v] {file}*`

abrown:
This PR allows users to execute CLIF functions from the command line by running `clif-util run [-v] {file}*`. It will look for functions with a signature like `() -> bool` that have a `run:` comment after them, compile them with the host machine's `TargetIsa`, and succeed if the function returns `true`. If the function returns `false` this utility keeps track of the number of failures much like `clif-util test`. I hoped to use this utility to verify that sequences of instructions (e.g. my SIMD changes) actually execute correctly on a host machine.

### DIFF
--- a/src/run.rs
+++ b/src/run.rs
@@ -1,0 +1,188 @@
+//! CLI tool to compile Cranelift IR files to native code in memory and execute them.
+
+use crate::utils::read_to_string;
+use core::mem;
+use cranelift_codegen::binemit::{Reloc, Stackmap};
+use cranelift_codegen::ir::{ExternalName, Function, JumpTable};
+use cranelift_codegen::isa::TargetIsa;
+use cranelift_codegen::Context;
+use cranelift_codegen::{binemit, ir};
+use cranelift_native::builder as host_isa_builder;
+use cranelift_reader::{parse_test, Details, IsaSpec};
+use mmap::{MapOption, MemoryMap};
+use region;
+use region::Protection;
+use std::path::PathBuf;
+use walkdir::{DirEntry, WalkDir};
+
+pub fn run(files: Vec<String>, flag_print: bool) -> Result<(), String> {
+    let mut total = 0;
+    let mut errors = 0;
+    for file in iterate_files(files) {
+        total += 1;
+        match run_single_file(&file) {
+            Ok(_) => {
+                if flag_print {
+                    println!("{}", file.to_string_lossy());
+                }
+            }
+            Err(e) => {
+                if flag_print {
+                    println!("{}: {}", file.to_string_lossy(), e);
+                }
+                errors += 1;
+            }
+        }
+    }
+
+    if flag_print {
+        match total {
+            0 => println!("0 files"),
+            1 => println!("1 file"),
+            n => println!("{} files", n),
+        }
+    }
+
+    match errors {
+        0 => Ok(()),
+        1 => Err(String::from("1 failure")),
+        n => Err(format!("{} failures", n)),
+    }
+}
+
+/// Iterate over all of the files passed as arguments, recursively iterating through directories
+fn iterate_files(files: Vec<String>) -> impl Iterator<Item = PathBuf> {
+    files
+        .into_iter()
+        .flat_map(WalkDir::new)
+        .filter(|f| match f {
+            Ok(d) => !is_hidden(d) && !is_directory(d),
+            Err(e) => {
+                println!("Unable to read file: {}", e);
+                false
+            }
+        })
+        .map(|f| {
+            f.expect("This should not happen: we have already filtered out the errors")
+                .into_path()
+        })
+}
+fn is_hidden(d: &DirEntry) -> bool {
+    d.file_name().to_str().map_or(false, |s| s.starts_with("."))
+}
+fn is_directory(d: &DirEntry) -> bool {
+    d.file_type().is_dir()
+}
+
+/// Run all functions in a file that are succeeded by "run:" comments
+fn run_single_file(path: &PathBuf) -> Result<(), String> {
+    let file_contents = read_to_string(&path).map_err(|e| e.to_string())?;
+    run_file_contents(file_contents)
+}
+
+/// Main body of `run_single_file` separated for testing
+fn run_file_contents(file_contents: String) -> Result<(), String> {
+    let test_file = parse_test(&file_contents, None, None).map_err(|e| e.to_string())?;
+    let isa = create_target_isa(test_file.isa_spec)?;
+    for (func, Details { comments, .. }) in test_file.functions {
+        if comments.iter().filter(|c| c.text.contains("run")).count() > 0 {
+            run_single_function(func, isa.as_ref())?
+        }
+    }
+    Ok(())
+}
+
+/// Build an ISA based on the current machine running this code (the host)
+fn create_target_isa(isa_spec: IsaSpec) -> Result<Box<dyn TargetIsa>, String> {
+    if let IsaSpec::None(flags) = isa_spec {
+        // build an ISA for the current machine
+        let builder = host_isa_builder()?;
+        Ok(builder.finish(flags))
+    } else {
+        Err(String::from("A target ISA was specified in the file but should not have been--only the host ISA can be used for running CLIF files"))?
+    }
+}
+
+/// Compile and execute a single function, expecting a boolean to be returned; a 'true' value is
+/// interpreted as a successful test execution and mapped to Ok whereas a 'false' value is
+/// interpreted as a failed test and mapped to Err.
+fn run_single_function(func: Function, isa: &dyn TargetIsa) -> Result<(), String> {
+    if !(func.signature.params.is_empty()
+        && func.signature.returns.len() == 1
+        && func.signature.returns.first().unwrap().value_type.is_bool())
+    {
+        return Err(String::from(
+            "Functions must have a signature like: () -> boolean",
+        ));
+    }
+    // TODO observed segfaults with return types of b64
+
+    // set up the context
+    let mut context = Context::new();
+    context.func = func;
+
+    // compile and encode the result to machine code
+    let relocs = &mut NullRelocSink {};
+    let traps = &mut NullTrapSink {};
+    let stackmaps = &mut NullStackmapSink {};
+    let code_info = context.compile(isa).map_err(|e| e.to_string())?;
+    let code_page = MemoryMap::new(code_info.total_size as usize, &[MapOption::MapWritable])
+        .map_err(|e| e.to_string())?;
+    let callable_fn: fn() -> bool = unsafe {
+        context.emit_to_memory(isa, code_page.data(), relocs, traps, stackmaps);
+        region::protect(code_page.data(), code_page.len(), Protection::ReadExecute)
+            .map_err(|e| e.to_string())?;
+        mem::transmute(code_page.data())
+    };
+
+    // execute
+    match callable_fn() {
+        true => Ok(()),
+        false => Err(format!("Failed: {}", context.func.name.to_string())),
+    }
+}
+
+/// Relocation helper
+struct NullRelocSink {}
+
+impl binemit::RelocSink for NullRelocSink {
+    fn reloc_ebb(&mut self, _: u32, _: Reloc, _: u32) {}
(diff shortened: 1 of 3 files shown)


src/run.rs
```diff
@@ -0,0 +1,171 @@
...
+    let mut relocs = NullRelocSink {};
+    let mut traps = NullTrapSink {};
+    let mut mem = vec![];
+    context
+        .compile_and_emit(isa, &mut mem, &mut relocs, &mut traps)
+        .map_err(|err| pretty_error(&context.func, Some(isa), err))?;
+
+    // make the machine code executable in Rust
+    let callable_fn: fn() -> bool = unsafe {
+        protect(mem.as_mut_ptr(), mem.len(), Protection::ReadWriteExecute).expect("..."); // this segfaults if it is only ReadExecute, a bactrace in `crashdumpctl debug` shows the explosion during deallocation at the end of this block
```
bjorn3: You should `mmap` pages, instead of `mprotect`ing the heap. The later is UB. Also why not use `cranelift-simplejit`?

src/run.rs
```diff
@@ -0,0 +1,171 @@
...
+        0 => Ok(()),
+        1 => Err(String::from("1 failure")),
+        n => Err(format!("{} failures", n)),
+    }
+}
+
+/// Build an ISA based on the current machine running this code (the host)
+fn create_target_isa() -> Box<dyn TargetIsa> {
+    let settings_builder = settings::builder();
+    // b.set("opt_level", "fastest");
```
bjorn3: Please get the flags from the clif file and maybe from the cmdline too.

src/run.rs
```diff
@@ -0,0 +1,171 @@
...
+    host_isa_builder()
+        .expect("A builder could not be created")
+        .finish(flags)
+}
+
+/// Iterate over all of the files passed as arguments. TODO eventually this should recurse into
+/// subdirectories, ignore non-CLIF files, ignore hidden files, etc.
+fn iterate_files(files: Vec<String>) -> Box<dyn Iterator<Item = PathBuf>> {
+    let iterator = files.into_iter().map(PathBuf::from);
+    Box::new(iterator)
```
bjorn3: Please use `impl Iterator<...>` as return type instead of boxing it up.

src/run.rs
```diff
@@ -0,0 +1,171 @@
...
+use region::{protect, Protection};
+use std::path::PathBuf;
+
+pub fn run(files: Vec<String>, flag_print: bool) -> Result<(), String> {
+    let isa = create_target_isa();
+    let mut total = 0;
+    let mut errors = 0;
+    for file in iterate_files(files) {
+        total += 1;
+        let file_name = get_file_name(&file)?;
```
bjorn3: Please use the `path` directly instead to prevent ambiguity when multiple dirs contain the same filename.

src/run.rs
```diff
@@ -0,0 +1,171 @@
...
+/// Compile and execute a single function, expecting a boolean to be returned; a 'true' value is
+/// interpreted as a successful test execution and mapped to Ok whereas a 'false' value is
+/// interpreted as a failed test and mapped to Err.
+fn run_single_function(func: Function, isa: &dyn TargetIsa) -> Result<(), String> {
+    if !(func.signature.params.is_empty()
+        && func.signature.returns.len() == 1
+        && func.signature.returns.first().unwrap().value_type.is_bool())
+    {
+        return Err(String::from(
+            "Functions must have a signature like: () -> boolean",
```
bjorn3: Maybe allow`() -> ()` too, for tests which should not crash.

src/run.rs
abrown: It wasn't exactly clear to me how to take a CLIF file and compile it with SimpleJIT; most of the clif-util code seems to interact directly with `Context` so I had some examples there. Also, in talking to @nbp he pointed me to his [HolyJIT](https://github.com/nbp/holyjit) project which was another helpful example that directly compiles using `Context`. In that project he does `mmap` in some memory to receive the emitted bytes so I can follow what he did there.

Before I continue, though, is there some way I can use the pre-existing SimpleJIT to do what I'm doing now with the CLIF test files?

abrown: @bjorn3 ok, I think the should be resolved in the latest push.

bnjbvr:
Interesting idea. I remember @sunfishcode was saying we should use SimpleJIT for testing the runtime behavior and correctness of Cranelift in the past; we just never got to it. I think doing it directly with clif-util makes sense. I've made a list of things that came to mind while reading this code, and could be addressed either in this PR or in subsequent ones.

I'd be worried that testing only specific signatures could mean a lot of missing test coverage for calls to functions with other signatures; but then having to deal with the Rust ABI is a bit unfortunate. Having functions that return only booleans, means that we can use a helper function to test that a call works, but we stay in Cranelift jit territory (and if there are consistent errors within Cranelift, they couldn't be caught). Which might be good enough for a start.

- [ ] I wonder if we could return more complicated types, like `i64` or a simd 128-bit vector type, that would be used in assertions in comments; we'd have comments like `assert_returns i64 42`.

This would be good to have in automation too, but we'd need to limit running the runtime tests to automation host architectures that are supported by Cranelift. And by not having much control over Travis CI, not sure we can run on e.g. ARM64 in the future. Worth nothing at least.

- [ ] Use this in automation.

Also to be entirely useful, I wonder if this shouldn't run the function in another process, in case the called function causes a segfault / other kind of hard crash. Otherwise, this means the first function that crashes causes the rest of the functions to *not* run, which is unfortunate.

- [ ] Handle segfaults one way or another.

nbp:
For your information, I have been discussing this approach with @abrown over IRC, and suggested to use a single and only signature for function calls as a feature.

If you want to check the generated code for making a function call with a given ABI, this can already be done without the need for executing the code, with ordinary tests using `; check:` and `; nextln:` and even `; bin:`.

So executing the code is just a matter of being able to call a single function. What I initially suggested was to have a `; run: %foo(1, 2, 3) == 51` which would generate a tiny function, compiled with Cranelift, with the signature `() -> bool` and which calls `%foo` with the expected argument and ABI.

I think the approach taken here goes in the right direction. So, even if it feels incomplete on many aspects, we should probably not block this work with too many wishes.

- About checking complex type content, this is something which we can add vocabulary for as we improve this system. With this patch we are able to call a `() -> bool` function. Generating such a function for calling other signature and asserting other properties sounds like incremental future work.
- About running other architecture, this is something we have done for SpiderMonkey testing already, by importing the simulators of other architectures (arm32, arm64, mips*). Another approach would be to rely on `qemu` and generate binaries.
- About running in different processes, I think this would be premature at the moment, but a nice to have feature that I would probably not implement my-self until it pains me enough. Thus we should not block this feature on it.

bnjbvr:
Yeah yeah, I wasn't suggesting to mandate all of this before merging. More thinking about the running in different processes part had me realize that it'd be nice to have this, in two modes: 1. for running the tests in CI where each test is run in its own process (to not crash the test process itself), 2. for debugging tests, we would probably prefer something as is right now, that is, a test (or collection of tests) running in the test process itself, so it's easier to run within gdb/rr etc.

abrown:
I'm all for `fork`-ing to avoid crashes, the `run: %foo(...) == ...` syntax @nbp is talking about, and wiring this up to the automation so that we always know that some CLIF actually runs (especially the x86 SIMD stuff 😄)--I do think all of those could be future PRs, though. 

I am interested in using SimpleJIT here but someone will have to help me out. I looked at it for a moment but how to use it was obscured by the module-level stuff--I really just needed `Context::compile` and `Context::emit_to_memory` (https://github.com/CraneStation/cranelift/pull/872/files#diff-8643ebb587e63cc619cd28e00be6a7e5R127-R131).

bnjbvr:
The more I think about it, the more unclear it is that we can actually use SimpleJIT for this, especially considering how simple the code in the `run` pass is, so this looks fine as is to me in the current state. Thanks for doing this!

abrown:
@sunfishcode rebased this on current master; should be ready to review.

src/run.rs
```diff
@@ -0,0 +1,188 @@
...
+        })
+        .map(|f| {
+            f.expect("This should not happen: we have already filtered out the errors")
+                .into_path()
+        })
+}
+fn is_hidden(d: &DirEntry) -> bool {
+    d.file_name().to_str().map_or(false, |s| s.starts_with("."))
+}
+fn is_directory(d: &DirEntry) -> bool {
```
bnjbvr: light suggestion: since these two are only used once, mind to inline their body where they're called, please?

src/run.rs
```diff
@@ -0,0 +1,188 @@
...
+        1 => Err(String::from("1 failure")),
+        n => Err(format!("{} failures", n)),
+    }
+}
+
+/// Iterate over all of the files passed as arguments, recursively iterating through directories
+fn iterate_files(files: Vec<String>) -> impl Iterator<Item = PathBuf> {
+    files
+        .into_iter()
+        .flat_map(WalkDir::new)
```
bnjbvr: This is *lovely*. (Could require uniquing the names, if some path trees actually intersect.)

src/run.rs
```diff
@@ -0,0 +1,188 @@
...
+    let file_contents = read_to_string(&path).map_err(|e| e.to_string())?;
+    run_file_contents(file_contents)
+}
+
+/// Main body of `run_single_file` separated for testing
+fn run_file_contents(file_contents: String) -> Result<(), String> {
+    let test_file = parse_test(&file_contents, None, None).map_err(|e| e.to_string())?;
+    let isa = create_target_isa(test_file.isa_spec)?;
+    for (func, Details { comments, .. }) in test_file.functions {
+        if comments.iter().filter(|c| c.text.contains("run")).count() > 0 {
```
bnjbvr: nit: can you use the `Iterator::any` function here instead?

src/run.rs
bnjbvr: It's been discussed outside the code, and it's a good design that's overall more global than what () -> () allows, so that looks fine to me.

src/run.rs
```diff
@@ -0,0 +1,188 @@
...
+    }
+}
+
+/// Compile and execute a single function, expecting a boolean to be returned; a 'true' value is
+/// interpreted as a successful test execution and mapped to Ok whereas a 'false' value is
+/// interpreted as a failed test and mapped to Err.
+fn run_single_function(func: Function, isa: &dyn TargetIsa) -> Result<(), String> {
+    if !(func.signature.params.is_empty()
+        && func.signature.returns.len() == 1
+        && func.signature.returns.first().unwrap().value_type.is_bool())
```
bnjbvr: (Actually a question:) would this allow returning one bool vector? If so, maybe we should make this more restrictive?

src/run.rs
```diff
@@ -0,0 +1,188 @@
...
+    {
+        return Err(String::from(
+            "Functions must have a signature like: () -> boolean",
+        ));
+    }
+    // TODO observed segfaults with return types of b64
```
bnjbvr: nit: remaining TODO in code; either we can open a follow-up issue, or it should be handled :)

src/run.rs
```diff
@@ -0,0 +1,188 @@
...
+
+    // execute
+    match callable_fn() {
+        true => Ok(()),
+        false => Err(format!("Failed: {}", context.func.name.to_string())),
+    }
+}
+
+/// Relocation helper
+struct NullRelocSink {}
```
bnjbvr: All these null sinks already exist in codegen::binemit (or elsewhere for the other sinks), could they be imported instead?

src/run.rs
abrown: No, I just checked and `assert!(!cranelift_codegen::ir::types::B32X4.is_bool())` is a successful test.

src/run.rs
abrown: To avoid running twice? Or we could just run things twice... 😄

src/run.rs
abrown: Ok, opened https://github.com/CraneStation/cranelift/issues/911; may need some help figuring this one out.

abrown:
@bnjbvr, I've resolved the issues above and pushed to https://github.com/abrown/cranelift/tree/clif-filetests-test-run so that we can merge as a part of #890.

PR #879, opened 2019-07-29
Stateful translate operator callbacks

iximeow:
It turned out very quickly that we did need `TranslationState` after all


pchickey:
I don't think we actually want to expose `TranslationState` here, that exposes way too much about how this library is implemented and would make it very difficult to make any sort of guarantee of API stability. I talked with ixi about making a wrapper struct that exposes just some bits of the `TranslationState` that should be pretty stable. In Lucet, the only information we actually need from `TranslationState` is whether the present opcode is reachable.

PR #882, opened 2019-08-01
Preopt: use replaced arg after having replaced BinaryImm

iximeow:
when replacing BinaryImm, we use the prior arg, but later use the arg that was replaced when writing an alias if we can determine the new op is actually equivalent to a simple copy

this was found with `strlen("a")` returning 2 in some wasm downstream, but I reduced that down to the test case here.

I was surprised to see in writing the test case that runners for cases in `filetests` have a lot of variance - should `simple_preopt` and `preopt` be merged together? the cases in `preopt` ought to fail under the full `preopt` run, but are only tested with respect to only some of the constant folding done in `simple_preopt`

### DIFF
--- a/filetests/simple_preopt/simplify.clif
+++ b/filetests/simple_preopt/simplify.clif
@@ -278,3 +278,16 @@ ebb0:
 ; nextln:     v2 = sextend.i64 v3
 ; nextln:     return v2
 ; nextln: }
+
+function %add_imm_fold(i32) -> i32 {
+ebb0(v0: i32):
+  v1 = iadd_imm v0, 42
+  v2 = iadd_imm v1, -42
+  return v2
+}
+; sameln: function %add_imm_fold(i32)
+; nextln: ebb0(v0: i32):
+; nextln:    v2 -> v0
+; nextln:    v1 = iadd_imm v0, 42
+; nextln:    nop
+; nextln:    return v2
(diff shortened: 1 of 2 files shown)


filetests/simple_preopt/simplify.clif
```diff
@@ -278,3 +278,16 @@ ebb0:
...
+
+function %add_imm_fold(i32) -> i32 {
+ebb0(v0: i32):
+  v1 = iadd_imm v0, 42
+  v2 = iadd_imm v1, -42
+  return v2
+}
+; sameln: function %add_imm_fold(i32)
+; nextln: ebb0(v0: i32):
+; nextln:    v2 -> v0
```
iximeow: currently cranelift turns `add_imm_fold` into:
```
ebb(v0: i32):
  v1 = iadd_imm v0, 42
  v2 -> v1
  nop
  return v2
```

which is definitely not what the function should compute :grin:

bnjbvr:
> I was surprised to see in writing the test case that runners for cases in filetests have a lot of variance - should simple_preopt and preopt be merged together? the cases in preopt ought to fail under the full preopt run, but are only tested with respect to only some of the constant folding done in simple_preopt

Note that `preopt` may relate to the `cranelift-preopt` crate while `simple_preopt` is in `cranelift-codegen`. The latter is aimed at simple CLIF generators which don't run many optimization passes: we assume compilers like LLVM won't need it (in the context of LLVM compiling wasm that's translated to CLIF), and that's why it's in a different crate. `simple_preopt` contains optimizations that are beneficial even in the case of the CLIF emitter having run optimizations. (Arguably, some of these folding could be carried out by the `cranelift-wasm` crate itself, or even further, the `InstBuilder` / `frontend` should do these behind the curtain on the user's behalf.)

So it seems unlikely that we merge these tests (which would also imply to share the code), because their goals are different. Does it make sense?

PR #883, opened 2019-08-01
Add EntryRegDiversions to record diversions for each block entry.

nbp:
This patch is a pre-requisite for #844 , which introduces `regalloc::EntryRegDiversions` as part of the `ir::Function`.

Technically we do not require `EntryRegDiversions`, but not having them implies that we would always visit all the blocks in a topological order inspired by the reverse post-order, in order to garantee that the `RegDiversions` for each basic block is recorded before the block is visited.

However, this is not possible as the order of the generated code is expected to be the same as the code within Cranelift, as implies the fact that we have `fallthrough` instructions.

Thus #844 would be changed to record the `RegDiversions` at each basic block entry on the `ir::Function::entry_diversions` field, using the `save_for_ebb` function. The `at_ebb` is used to restore the state of the `RegDiversions` as propagated by the predecessor, and the `check_ebb_entry` would be used in a mirror location as `save_for_ebb` but in the locations verifier.

Doing these changes as part of the diversions logic is the least intrusive implementation I found so far, as it only requires to replace `divert.clear()` calls by `divert.at_ebb(&func.entry_diversions, ebb)` calls.

### DIFF
--- a/cranelift-codegen/src/ir/function.rs
+++ b/cranelift-codegen/src/ir/function.rs
@@ -14,7 +14,7 @@ use crate::ir::{
 use crate::ir::{EbbOffsets, InstEncodings, SourceLocs, StackSlots, ValueLocations};
 use crate::ir::{JumpTableOffsets, JumpTables};
 use crate::isa::{CallConv, EncInfo, Encoding, Legalize, TargetIsa};
-use crate::regalloc::RegDiversions;
+use crate::regalloc::{EntryRegDiversions, RegDiversions};
 use crate::value_label::ValueLabelsRanges;
 use crate::write::write_function;
 use core::fmt;
@@ -62,6 +62,12 @@ pub struct Function {
     /// Location assigned to every value.
     pub locations: ValueLocations,
 
+    /// Non-default locations assigned to value at the entry of basic blocks.
+    ///
+    /// At the entry of each basic block, we might have values which are not in their default
+    /// ValueLocation. This field records these register-to-register moves as Diversions.
+    pub entry_diversions: EntryRegDiversions,
+
     /// Code offsets of the EBB headers.
     ///
     /// This information is only transiently available after the `binemit::relax_branches` function
@@ -94,6 +100,7 @@ impl Function {
             layout: Layout::new(),
             encodings: SecondaryMap::new(),
             locations: SecondaryMap::new(),
+            entry_diversions: EntryRegDiversions::new(),
             offsets: SecondaryMap::new(),
             jt_offsets: SecondaryMap::new(),
             srclocs: SecondaryMap::new(),
@@ -112,6 +119,7 @@ impl Function {
         self.layout.clear();
         self.encodings.clear();
         self.locations.clear();
+        self.entry_diversions.clear();
         self.offsets.clear();
         self.jt_offsets.clear();
         self.srclocs.clear();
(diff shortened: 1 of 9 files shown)


cranelift-codegen/src/ir/function.rs
```diff
@@ -62,6 +62,12 @@ pub struct Function {
     /// Location assigned to every value.
     pub locations: ValueLocations,
 
+    /// Non-default locations assigned to value at the entry of basic blocks.
+    ///
+    /// At the entry of each basic block, we might have values which are not in their default
+    /// ValueLocation. This field record these register-to-register move as Diversions.
```
sstangl: Ultra-nit: "records" and "moves"

cranelift-codegen/src/ir/function.rs
```diff
@@ -197,10 +205,12 @@ impl Function {
             !self.offsets.is_empty(),
             "Code layout must be computed first"
         );
+        let mut divert = RegDiversions::new();
+        divert.at_ebb(&self.entry_diversions, ebb);
```
sstangl: Maybe make a wrapper `RegDiversions::new_at()`?

cranelift-codegen/src/regalloc/diversion.rs
```diff
@@ -163,9 +177,94 @@ impl RegDiversions {
         self.current.remove(&value).map(|d| d.to)
     }
 
+    /// Reset the current diversion for the entry of an `Ebb`, based on the recorded content
+    /// available on `func.entry_diversions`.
+    pub fn at_ebb(&mut self, entry_diversions: &EntryRegDiversions, ebb: Ebb) {
+        self.clear();
+        match entry_diversions.map.get(ebb) {
```
sstangl: This should be `if let Some(entry_divert) = entry_diversions.map.get(ebb) {`.

cranelift-codegen/src/regalloc/diversion.rs
```diff
@@ -163,9 +177,94 @@ impl RegDiversions {
         self.current.remove(&value).map(|d| d.to)
     }
 
+    /// Reset the current diversion for the entry of an `Ebb`, based on the recorded content
+    /// available on `func.entry_diversions`.
+    pub fn at_ebb(&mut self, entry_diversions: &EntryRegDiversions, ebb: Ebb) {
```
sstangl: I'm not such a fan of this name, because it's not clear what it's doing.

It seems to: 1) clear the current diversions. 2) put the new ebb's entry diversions into the current map.

Maybe it should be called `switch_to()`, `switch_ebb()`, something like that?

cranelift-codegen/src/regalloc/diversion.rs
```diff
@@ -163,9 +177,94 @@ impl RegDiversions {
         self.current.remove(&value).map(|d| d.to)
     }
 
+    /// Reset the current diversion for the entry of an `Ebb`, based on the recorded content
+    /// available on `func.entry_diversions`.
```
sstangl: This comment is kind of literal. Maybe better: "Resets the state of the current diversion to the recorded EntryDiversions on the given ebb`.

cranelift-codegen/src/ir/function.rs
nbp: Other parts of the code have these separated in 2, a `RegDiversions::new()` out-side a loop, and the `at_ebb` call within the loop. I think this code is more idiomatic if it remains separated in 2 statements.

cranelift-codegen/src/regalloc/diversion.rs
nbp: Cursors have a function named `at_top(mut self, ebb: ir::Ebb) -> Self`.
@sstangl would it be ok, if I were to rename this function `at_top`, to match with cursor naming?

nbp: Not having any answer, I will merge these changes open a follow-up to do this renaming later.

nbp: I opened #958 to address this issue.

PR #884, opened 2019-08-01
Use divert apply everywhere outside the coloring phase.

nbp:
This is a minor change, which make use use `divert.regmove`, `divert.regspill` and `divert.regfill` only within the coloring phase of the compiler. All other phases now rely on `divert.apply`.

This is a side-effect of a change that I initially made and reverted where I attempted to split `RegDiversions` in 2, one meant to record the `EntryRegDiversions` added in #883, and another one meant to read the content of it.

The difference between the coloring phase and others, is that the coloring phase will add the `regmove` instructions, whereas others are only following what is already present in the DFG.

### DIFF
--- a/cranelift-codegen/meta/src/gen_binemit.rs
+++ b/cranelift-codegen/meta/src/gen_binemit.rs
@@ -32,7 +32,7 @@ fn gen_recipe(formats: &FormatRegistry, recipe: &EncodingRecipe, fmt: &mut Forma
     let is_regmove = ["RegMove", "RegSpill", "RegFill"].contains(&inst_format.name);
 
     // Unpack the instruction data.
-    fmtln!(fmt, "if let InstructionData::{} {{", inst_format.name);
+    fmtln!(fmt, "if let &InstructionData::{} {{", inst_format.name);
     fmt.indent(|fmt| {
         fmt.line("opcode,");
         for f in &inst_format.imm_fields {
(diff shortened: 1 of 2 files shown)


nbp:
Note, this change is not mandatory nor blocking any other changes.

cranelift-codegen/meta/src/gen_binemit.rs
```diff
@@ -47,7 +47,7 @@ fn gen_recipe(formats: &FormatRegistry, recipe: &EncodingRecipe, fmt: &mut Forma
         }
         fmt.line("..");
 
-        fmt.outdented_line("} = func.dfg[inst] {");
+        fmt.outdented_line("} = data {");
```
bnjbvr: ... and here too.

cranelift-codegen/meta/src/gen_binemit.rs
```diff
@@ -186,7 +186,8 @@ fn gen_isa(formats: &FormatRegistry, isa_name: &str, recipes: &Recipes, fmt: &mu
     fmt.indent(|fmt| {
         fmt.line("let encoding = func.encodings[inst];");
         fmt.line("let bits = encoding.bits();");
-        fmt.line("match func.encodings[inst].recipe() {");
+        fmt.line("let data = &func.dfg[inst];");
```
bnjbvr: Nice. Maybe rename this `inst_data` as in the other patch...

PR #885, opened 2019-08-01
Simplify jump table instructions and add missing conversion

bnjbvr:
This makes non-legalized jump table instructions operate on operands with
pointer-sized types. This means we need to extend smaller types into the
pointer-sized operand, when the two don't match.

This has the nice benefit to remove the controlling type variable from `jump_table_entry`, and now all the variables involved have the same type (both the input operands and the returned value).

@sunfishcode, assigning you for review since we discussed this offline. Thanks!

### DIFF
--- a/cranelift-codegen/meta/src/shared/instructions.rs
+++ b/cranelift-codegen/meta/src/shared/instructions.rs
@@ -269,15 +269,9 @@ pub fn define(
         .is_branch(true),
     );
 
+    // The index into the br_table can be any type; legalizer will convert it to the right type.
     let x = &operand_doc("x", iB, "index into jump table");
-
-    let Entry = &TypeVar::new(
-        "Entry",
-        "A scalar integer type",
-        TypeSetBuilder::new().ints(Interval::All).build(),
-    );
-
-    let entry = &operand_doc("entry", Entry, "entry of jump table");
+    let entry = &operand_doc("entry", iAddr, "entry of jump table");
     let JT = &operand("JT", jump_table);
 
     ig.push(
@@ -305,6 +299,9 @@ pub fn define(
         .is_branch(true),
     );
 
+    // These are the instructions which br_table legalizes to: they perform address computations,
+    // using pointer-sized integers, so their type variables are more constrained.
+    let x = &operand_doc("x", iAddr, "index into jump table");
     let Size = &operand_doc("Size", uimm8, "Size in bytes");
 
     ig.push(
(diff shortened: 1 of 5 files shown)


cranelift-codegen/meta/src/shared/instructions.rs
```diff
@@ -305,6 +299,9 @@ pub fn define(
         .is_branch(true),
     );
 
+    // These are the instructions which br_table legalizes to: they operate on real pointers, so
+    // their type variables are more constrained.
```
sunfishcode: The word "real" here is a little confusing. And technically the index isn't a pointer; it's just a pointer-sized integer. How about saying something like "they perform address computations, using pointer-sized integers, so their type variables are more constrained"?

filetests/isa/x86/legalize-br-table.clif
```diff
@@ -16,8 +16,9 @@ ebb0(v0: i64):
 ; nextln:    brif uge $oob, ebb2
 ; nextln:    fallthrough $(inb=$EBB)
 ; check:   $inb:
+; nextln:    $(final_idx=$V) = uextend.i64 $idx
 ; nextln:    $(base=$V) = jump_table_base.i64 jt0
-; nextln:    $(rel_addr=$V) = jump_table_entry.i64 $idx, $base, 4, jt0
+; nextln:    $(rel_addr=$V) = jump_table_entry $final_idx, $base, 4, jt0
```
bjorn3: Does this test still work with 64bit ints on i686?

bnjbvr: Line 3 in this file defines the target: target x86_64, so it should be fine, right?

bjorn3: I mean if the target is set to `x86`, does it still work. Otherwise `cranelift_frontend::Switch` is broken for 64bit ints on 32bit archs.

bnjbvr: I think it *should* fail, since there's no way the compiler can decide what's the right thing to do, on a 32-bit platform, when seeing a 64-bits integer passed for a 32-bit operation. Now the question is how violentlym it fails... I think it will say there's no encoding for the instructions `br_table` is legalized to, and not panic at least. (There are likely many other failures on x86 32 bits, since it's untested at the moment.)

bjorn3: Can you please change

https://github.com/CraneStation/cranelift/blob/05d22df6cbf98a072d7525cff4599d930af142a7/cranelift-frontend/src/switch.rs#L225

So that `cranelift_frontend::Switch` keeps working for 64bit integers on 32bit archs?

bnjbvr: I'd like to help, really, but I'm afraid I can't do this: there's no sensible reasonable default for what to do in this case. When an i64 is outside the uint32 range and is passed as the discriminant value of a br_table, we can either silently cut the high bits (above the uint32_max value), or we should signal an error because we're not doing what the developer might expect. It's up to the developer to do the right thing here.

PR #886, opened 2019-08-01
Return a `WasmResult` from `ModuleEnvironment` methods

acfoltzer:
This method in particular needs to accommodate failure because any table index other than zero is
currently invalid.

It's possible that other `ModuleEnvironment` methods would benefit from fallibility, but I didn't want to add too much noise to this PR without being sure. If y'all would prefer doing them all at once, let me know which methods would benefit and I can handle the changes.

**Update**: Per discussion in this thread, the result type has been changed for most of the other `ModuleEnvironment` functions, and along the way some panics have been transformed into `Err`s.

### DIFF
--- a/cranelift-wasm/src/code_translator.rs
+++ b/cranelift-wasm/src/code_translator.rs
@@ -23,12 +23,13 @@
 //! That is why `translate_function_body` takes an object having the `WasmRuntime` trait as
 //! argument.
 use super::{hash_map, HashMap};
-use crate::environ::{FuncEnvironment, GlobalVariable, ReturnMode, WasmError, WasmResult};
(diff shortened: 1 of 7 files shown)


bnjbvr:
Thanks for opening this PR; I certainly understand the motivation to add a Result here (since I've adapted the FuncEnvironment to return Results as well for almost every function as defined in the trait). I agree that it might be nice to extend to other methods too for consistency, especially those which handle a table or a memory (in the MVP version, there can be only one). Are you willing to do these changes here too, or should we defer this? Thanks!

acfoltzer:
:+1: Yeah, I can definitely handle those. I just wanted to make sure that all the changes wouldn't be too unwelcome.

PR #887, opened 2019-08-01
WIP: Add fold_redundant_jumps() pass to the branch relaxation phase.

sstangl:
This adds a `fold_redundant_jumps()` pass to the branch relaxation phase. Nicolas' basic-blocks regalloc patch splits critical edges: if those edges are unused by register allocation, the blocks can be folded back up. Not doing so produces weird codegen where you jump to a jump to a jump.

For example:

Before:
```
Disassembly of 22 bytes:
   0:	55                   	push	rbp
   1:	48 89 e5             	mov	rbp, rsp
   4:	b8 08 00 00 00       	mov	eax, 8
   9:	83 ff 05             	cmp	edi, 5
   c:	7f 04                	jg	0x12
   e:	89 f8                	mov	eax, edi
  10:	eb 02                	jmp	0x14
  12:	eb fc                	jmp	0x10
  14:	5d                   	pop	rbp
  15:	c3                   	ret
```

After:

```
   0:	55                   	push	rbp
   1:	48 89 e5             	mov	rbp, rsp
   4:	b8 08 00 00 00       	mov	eax, 8
   9:	83 ff 05             	cmp	edi, 5
   c:	7f 04                	jg	0x12
   e:	89 f8                	mov	eax, edi
  10:	eb 00                	jmp	0x12
  12:	5d                   	pop	rbp
  13:	c3                   	ret
```

This patch isn't done. It can fold more, and also I'm not sure if anything needs to be done about ebb parameters -- this executes after regalloc, which should already have resolved that, in theory.

### DIFF
--- a/cranelift-codegen/src/binemit/relaxation.rs
+++ b/cranelift-codegen/src/binemit/relaxation.rs
@@ -29,7 +29,9 @@
 
 use crate::binemit::{CodeInfo, CodeOffset};
 use crate::cursor::{Cursor, FuncCursor};
-use crate::ir::{Function, InstructionData, Opcode};
+use crate::dominator_tree::DominatorTree;
+use crate::flowgraph::ControlFlowGraph;
+use crate::ir::{Ebb, Function, Inst, InstructionData, Opcode, Value, ValueList};
 use crate::isa::{EncInfo, TargetIsa};
 use crate::iterators::IteratorExtras;
 use crate::regalloc::RegDiversions;
@@ -40,7 +42,12 @@ use log::debug;
 /// Relax branches and compute the final layout of EBB headers in `func`.
 ///
 /// Fill in the `func.offsets` table so the function is ready for binary emission.
-pub fn relax_branches(func: &mut Function, isa: &dyn TargetIsa) -> CodegenResult<CodeInfo> {
+pub fn relax_branches(
+    func: &mut Function,
+    cfg: &mut ControlFlowGraph,
+    domtree: &mut DominatorTree,
+    isa: &dyn TargetIsa,
+) -> CodegenResult<CodeInfo> {
     let _tt = timing::relax_branches();
 
     let encinfo = isa.encoding_info();
@@ -49,7 +56,10 @@ pub fn relax_branches(func: &mut Function, isa: &dyn TargetIsa) -> CodegenResult
     func.offsets.clear();
     func.offsets.resize(func.dfg.num_ebbs());
 
-    // Start by inserting fall through instructions.
+    // Start by removing redundant jumps.
+    fold_redundant_jumps(func, cfg, domtree);
+
+    // Convert jumps to fallthrough instructions where possible.
     fallthroughs(func);
 
     let mut offset = 0;
@@ -79,7 +89,6 @@ pub fn relax_branches(func: &mut Function, isa: &dyn TargetIsa) -> CodegenResult
         let mut cur = FuncCursor::new(func);
         while let Some(ebb) = cur.next_ebb() {
             divert.clear();
-
             // Record the offset for `ebb` and make sure we iterate until offsets are stable.
             if cur.func.offsets[ebb] != offset {
                 cur.func.offsets[ebb] = offset;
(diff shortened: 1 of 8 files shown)


julian-seward1:
Can this lead to unreachable blocks (the shorted-out ones) being left in the CFG, or are they removed too?

sstangl:
> Can this lead to unreachable blocks (the shorted-out ones) being left in the CFG, or are they removed too?

It can: they're removed too. The CFG is cleaned up and the DominatorTree is recalculated.

sstangl:
It's done! Some tests are still failing (they just need to be updated), but it's good enough to undergo review.

This wound up being quite a bit more complicated than I had hoped, because ebb parameters need translation. The code is logical and I tried to leave helpful commentary, but it definitely isn't a super-simple phase anymore!

The program in the original comment now outputs as:

```
Disassembly of 18 bytes:
   0:	55                   	push	rbp
   1:	48 89 e5             	mov	rbp, rsp
   4:	b8 08 00 00 00       	mov	eax, 8
   9:	83 ff 05             	cmp	edi, 5
   c:	7f 02                	jg	0x10
   e:	89 f8                	mov	eax, edi
  10:	5d                   	pop	rbp
  11:	c3                   	ret	
```

...which is exactly the same as before nbp's regalloc patch. That should eliminate the perf regression in the generated code, and we should be able to land basic-blocks soon.

cranelift-codegen/src/binemit/relaxation.rs
```diff
@@ -134,6 +143,145 @@ pub fn relax_branches(func: &mut Function, isa: &dyn TargetIsa) -> CodegenResult
...
+    }
+
+    // Now we need to fix up first_inst's ebb parameters to match second_inst's,
+    // without changing the branch-specific arguments.
+    //
+    // The intermediary block is allowed to reference any SSA value that dominates it,
+    // but that SSA value may not necessarily also dominate the instruction that's
+    // being patched.
+
+    // Get the arguments and parameters passed by the first jump.
```
nbp: nit: by the first *branch*.

The first branch is not necessary a jump instruction as opposed to the second one.

cranelift-codegen/src/binemit/relaxation.rs
```diff
@@ -134,6 +143,145 @@ pub fn relax_branches(func: &mut Function, isa: &dyn TargetIsa) -> CodegenResult
...
+    let (_, second_params) = func.dfg[second_inst]
+        .arguments(&func.dfg.value_lists)
+        .split_at(num_fixed);
+    let mut second_params = second_params.to_vec(); // Clone for rewriting below.
+
+    // For each parameter passed by the second jump, if any of those parameters
+    // was a block parameter, rewrite it to refer to the value that the first jump
+    // passed in its parameters. Otherwise, make sure it dominates first_inst.
+    //
+    // For example: if we `ebb0: jump ebb1(v1)` to `ebb1(v2): jump ebb2 v2`,
```
nbp: nit: `ebb1(v2): jump ebb2(v2)`  (missing parenthesis)

cranelift-codegen/src/binemit/relaxation.rs
```diff
@@ -134,6 +143,145 @@ pub fn relax_branches(func: &mut Function, isa: &dyn TargetIsa) -> CodegenResult
...
+    // we want to rewrite the original jump to `jump ebb2(v1)`.
+    // Note that v2 does not dominate the first jump instruction.
```
nbp: nit: Remove this last line. Even if it were to dominate by being a loop header, it would not have the expected value.

cranelift-codegen/src/binemit/relaxation.rs
```diff
@@ -134,6 +143,145 @@ pub fn relax_branches(func: &mut Function, isa: &dyn TargetIsa) -> CodegenResult
...
+        } else {
+            // Check whether the parameter value dominates first_inst.
+            let dominates_branch = match func.dfg.value_def(*value) {
+                ValueDef::Result(inst, _) => domtree.dominates(inst, first_inst, &func.layout),
+                ValueDef::Param(ebb, _) => domtree.dominates(ebb, first_inst, &func.layout),
+            };
+
+            // The second_inst passes a parameter that doesn't dominate first_inst,
+            // so first_inst is unable to pass that parameter itself.
+            if !dominates_branch {
```
nbp: I do not see how this condition would ever be possible if the `first_dest` is indeed just a `jump %second_dest` branch.
As any value which is not live from the ebb parameters of the first destination would necessarily comes from the predecessor of the first instruction.

I do not understand why this else branch is not a no-op. Do you have an example?
Otherwise I suggest to remove this else branch.

sstangl: I think you're right. I was trying to make the code safe against crazy CFGs. But you're correct: if this transformation were invalid, then the CFG would have originally been invalid, and we don't need to handle invalid CFGs.

PR #888, opened 2019-08-01
Serialize ValueLabel and StackSlots

yurydelendik:
While rebasing https://github.com/CraneStation/wasmtime/pull/63, found that ValueLabel and StackSlots are not marked for serialization -- this patch addresses that.

/cc @Mrowqa

### DIFF
--- a/cranelift-codegen/src/ir/mod.rs
+++ b/cranelift-codegen/src/ir/mod.rs
@@ -23,6 +23,9 @@ mod trapcode;
 pub mod types;
 mod valueloc;
 
+#[cfg(feature = "enable-serde")]
+use serde::{Deserialize, Serialize};
+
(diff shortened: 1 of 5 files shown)


cranelift-codegen/src/ir/mod.rs
```diff
@@ -74,6 +74,7 @@ pub type SourceLocs = SecondaryMap<Inst, SourceLoc>;
 
 /// Marked with a label value.
 #[derive(Copy, Clone, PartialEq, Eq, Hash)]
+#[cfg_attr(feature = "enable-serde", derive(serde::Serialize, serde::Deserialize))]
```
Mrowqa: My personal taste would be avoiding the `serde::` by adding the extra `use` declaration. I'm not sure about our code conventions though.

PR #890, opened 2019-08-02
Add `test run` to cranelift-filetests to allow executing CLIF

abrown:
This depends on the change in #872; adds the ability to start running CLIF as a part of the build automation. As mentioned there, future commits will add a way (e.g. `fork`) to avoid crashing the test process if a test fails and a more flexible expression syntax for asserting results (e.g. `run: fn(...) == ...`).

### DIFF
--- a/cranelift-filetests/Cargo.toml
+++ b/cranelift-filetests/Cargo.toml
@@ -11,9 +11,12 @@ edition = "2018"
 
 [dependencies]
 cranelift-codegen = { path = "../cranelift-codegen", version = "0.40.0", features = ["testing_hooks"] }
+cranelift-native = { path = "../cranelift-native", version = "0.40.0" }
 cranelift-reader = { path = "../cranelift-reader", version = "0.40.0" }
 cranelift-preopt = { path = "../cranelift-preopt", version = "0.40.0" }
 file-per-thread-logger = "0.1.2"
 filecheck = "0.4.0"
-num_cpus = "1.8.0"
 log = "0.4.6"
+mmap = "0.1.1"
+num_cpus = "1.8.0"
+region = "2.1.2"
--- a/filetests/isa/x86/run-const.clif
+++ b/filetests/isa/x86/run-const.clif
@@ -1,0 +1,11 @@
+test run
+
+function %test_compare_i32() -> b1 {
+ebb0:
+    v0 = iconst.i32 42
+    v1 = iconst.i32 42
+    v2 = icmp eq v0, v1
+    return v2
+}
+
+; run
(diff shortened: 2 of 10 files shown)


abrown:
Rebased; ready for review (@bnjbvr? @sunfishcode?)

filetests/isa/x86/run-const.clif
```diff
@@ -0,0 +1,11 @@
+test run
+
+function %test_compare_i32() -> b1 system_v {
```
bnjbvr: Is the Rust calling convention also system_v on windows machines? Maybe we should filter out tests that specify a calling convention different from the native one? (could be a follow-up)

bjorn3: > Is the Rust calling convention also system_v on windows environments?

It is based on the native C abi I believe.

bnjbvr: Right, so this is Windows fastcall. (We can take care of this in a follow-up issue, but I'd prefer it to be fixed quickly so there are no strange test failures for Windows users)

cranelift-filetests/Cargo.toml
```diff
@@ -11,9 +11,12 @@ edition = "2018"
...
+cranelift-native = { path = "../cranelift-native", version = "0.38.0" }
 cranelift-reader = { path = "../cranelift-reader", version = "0.38.0" }
 cranelift-preopt = { path = "../cranelift-preopt", version = "0.38.0" }
 file-per-thread-logger = "0.1.2"
 filecheck = "0.4.0"
-num_cpus = "1.8.0"
 log = "0.4.6"
+mmap = "0.1.1"
+num_cpus = "1.8.0"
+region = "2.1.2"
```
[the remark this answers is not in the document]
abrown: If we squash these will be in the right place...

filetests/isa/x86/run-const.clif
abrown: @bnjbvr and I talked about this some more and added the following check to avoid user error in specifying the calling convention: https://github.com/CraneStation/cranelift/pull/890/files#diff-f08bdadf8ad754c1568e3caa1d27a2f8R50-R57. Unfortunately we couldn't see an easy way to tell `parse_test` that we want a default other than `CallConv::Fast`, hence the second half of that expression. (I see it set here: https://github.com/CraneStation/cranelift/blob/master/cranelift-reader/src/parser.rs#L1004).

bnjbvr:
There are new conflicts because of Cargo updates, can you please manually rebase @abrown ?

abrown:
@bnjbvr should be good to go now

bnjbvr:
Great work, thank you!

PR #892, opened 2019-08-05
Adds the libcall_call_conv setting and use it for libcall calls expansion

bnjbvr:
This adds up a new way to tell which calling convention should be used when expanding instruction opcodes into libcall calls. Most of the time this should match the default calling convention, but in the case of Baldrdash, which uses special thunks for libcall functions, we need to use the Baldrdash calling convention too.

This tries to be more generic by allowing more calling conventions for other libcalls; I *think* this could be useful (for emulators? for JITs which want to use a calling convention different to the host's), but I am a bit unsure about this. So maybe we could do something simpler here, and instead have a boolean setting "enable baldrdash callconv for libcalls" (but that's very ad-hoc). Opinions welcome.

### DIFF
--- a/cranelift-codegen/meta/src/shared/settings.rs
+++ b/cranelift-codegen/meta/src/shared/settings.rs
@@ -94,6 +94,29 @@ pub fn define() -> SettingGroup {
 
     // Settings specific to the `baldrdash` calling convention.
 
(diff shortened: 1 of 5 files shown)


sunfishcode:
Makes sense!

PR #894, opened 2019-08-06
Bump version to 0.38.0

yurydelendik:


yurydelendik:
0.38.0 is alive at crates.io

PR #898, opened 2019-08-13
Add a custom section hook to `ModuleEnvironment`

alexcrichton:
This commit adds a hook to the `ModuleEnvironment` trait to learn when a
custom section in a wasm file is read. This hook can in theory be used
to parse and handle custom sections as they appear in the wasm file
without having to re-iterate over the wasm file after cranelift has
already parsed the wasm file.

The `translate_module` function is now less strict in that it doesn't
require sections to be in a particular order, but it's figured that the
wasm file is already validated elsewhere to verify the section order.

### DIFF
--- a/cranelift-wasm/src/environ/spec.rs
+++ b/cranelift-wasm/src/environ/spec.rs
@@ -456,4 +456,14 @@ pub trait ModuleEnvironment<'data> {
         offset: usize,
         data: &'data [u8],
     ) -> WasmResult<()>;
+
+    /// Indicates that a custom section has been found in the wasm file
+    fn custom_section(
+        &mut self,
+        name: &'data str,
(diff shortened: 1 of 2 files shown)


bnjbvr:
Note that Spidermonkey is able to do validation while translating to the backend compiler's IR, so if the Cranelift backends want to be competitive with Spidermonkey in the future, it'll need to do so, and the less-restrictive checking of sections order part will have to be reverted. (Right now, Cranelift-in-Spidermonkey validates before compiling, which is sub-optimal but needed for gathering module's metadata.)

bnjbvr:
Was the section ordering an issue in any ways? It seems this commit undid some useful work that will need to be re-implemented in the future, as stated in a previous comment.

Issue #729, opened 2019-04-07
Map to DWARF register numbers


philipc:
We need a way to map cranelift registers to DWARF registers.

I think this should be part of `RegInfo`. LLVM stores this map as an array of register pairs, and performs lookups using `lower_bound`. I'm not sure if cranelift should do the same, or use an array indexed by `RegUnit`.

Should we wait until after `RegInfo` generation is converted from python before doing this?

bnjbvr:
> Should we wait until after RegInfo generation is converted from python before doing this?

For what it's worth, RegInfo generation has already been moved to Rust, and it is being used when building the meta crate. (The Python definitions remain in place, because some other Python code relies on the module)

iximeow:
I've added a `DwarfRegMapper` in #902 that lives in `cranelift-module`. I don't think it should be part of `RegInfo` in particular because this mapping to DWARF registers is for an off-by-default codegen toggle (`.eh_frame`/`.debug_frame`) and will change target to target - specifically thinking about Windows and DWARF numbering x86 registers differently, here. The exact way an output would want to handle cranelift-produced layout information to yield usable unwind data seems beyond `cranelift-codegen`.

If `DwarfRegMapper` seems good, cool! If it doesn't, I'm happy to keep changing things around.

philipc:
Keeping it in the same location as the `gimli` dependency seems fine.

philipc:
Although, the advantage of putting it in `RegInfo` (or `RegBank`) is you wouldn't need multiple string comparisons for every `translate_reg` call.

iximeow:
I was hoping that since everything involved is `&'static str` it'd be turned into pointer comparisons, though I hadn't bothered to actually check that it does.

PR #902, opened 2019-08-15
FrameLayoutChange-based .eh_frame support

iximeow:
This teaches `cranelift-faerie` how to write `.eh_frame` sections that are appropriate with respect to call frames for functions that produce layout information.

This builds quite directly on https://github.com/CraneStation/cranelift/pull/679, which is still open, and whose changes are alongside this as a result. For the time being, my changes on top of @yurydelendik 's PR are best seen as a diff between [our branches](https://github.com/yurydelendik/cranelift/compare/frame-layout...iximeow:framechange_eh_frame_support?expand=1).

Being dependent on an already-floating PR, I want to mostly call attention to the points that:
* frame change information is useful for handling unwinding as @bjorn3 mentioned [here](https://github.com/CraneStation/cranelift/pull/679#issuecomment-510117825)
* consumers of Cranelift (in this case, Lucet) probably want to [unwind through Cranelift-generated code](https://github.com/CraneStation/cranelift/pull/679)
* #679 is good enough to build on to the above ends, and the changes to produce .eh_frame and support unwinding through Cranelift-generated code are fairly small.

I do have some comments in-line in case any readers do take a closer look at the changes here, but I'm not sure it makes sense to sweat the details until #679  is figured out, since it seems like we've bitrotted with respect to today's `master` anyway.

### DIFF
--- a/cranelift-faerie/src/backend.rs
+++ b/cranelift-faerie/src/backend.rs
@@ -16,8 +16,8 @@ use faerie;
 use std::fs::File;
 use target_lexicon::Triple;
 
-use gimli::write::Address;
-use gimli::write::FrameDescriptionEntry;
+use gimli::write::{Address, FrameDescriptionEntry};
+use gimli::{DW_EH_PE_pcrel, DW_EH_PE_sdata4};
 
 #[derive(Debug)]
 /// Setting to enable collection of traps. Setting this to `Enabled` in
@@ -114,17 +114,16 @@ impl<'a> gimli::write::Writer for FaerieDebugSink<'a> {
     fn write_eh_pointer(
         &mut self,
         address: Address,
-        _eh_pe: gimli::DwEhPe,
+        eh_pe: gimli::DwEhPe,
         size: u8,
     ) -> gimli::write::Result<()> {
-        self.write_address(address, size)
-    }
-
-    fn write_address(&mut self, address: Address, size: u8) -> gimli::write::Result<()> {
         match address {
             Address::Constant(val) => self.write_udata(val, size),
             Address::Symbol { symbol, addend } => {
-                assert!(addend == 0);
+                assert_eq!(addend, 0);
+
+                assert_eq!(eh_pe.format(), DW_EH_PE_sdata4, "faerie backend currently only supports PC-relative 4-byte offsets for DWARF pointers.");
+                assert_eq!(eh_pe.application(), DW_EH_PE_pcrel, "faerie backend currently only supports PC-relative 4-byte offsets for DWARF pointers.");
 
                 let name = self.functions[symbol].as_str();
 
--- a/cranelift-faerie/src/backend.rs
+++ b/cranelift-faerie/src/backend.rs
@@ -6,16 +6,20 @@ use anyhow::Error;
 use cranelift_codegen::binemit::{
     Addend, CodeOffset, NullStackmapSink, NullTrapSink, Reloc, RelocSink, Stackmap, StackmapSink,
 };
-use cranelift_codegen::isa::TargetIsa;
-use cranelift_codegen::{self, binemit, ir};
+use cranelift_codegen::isa::{RegInfo, RegUnit, TargetIsa};
+use cranelift_codegen::{self, binemit, ir, isa};
 use cranelift_module::{
     Backend, DataContext, DataDescription, DataId, FuncId, Init, Linkage, ModuleError,
     ModuleNamespace, ModuleResult,
 };
 use faerie;
 use std::fs::File;
-use std::io::Cursor;
 use target_lexicon::Triple;
+
+use gimli::write::Address;
+use gimli::write::CallFrameInstruction;
+use gimli::write::CommonInformationEntry;
+use gimli::write::FrameDescriptionEntry;
 
 #[derive(Debug)]
 /// Setting to enable collection of traps. Setting this to `Enabled` in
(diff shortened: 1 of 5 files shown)


cranelift-faerie/src/backend.rs
```diff
@@ -74,9 +79,164 @@ pub struct FaerieBackend {
...
 
+struct FrameSink {
+    // we need to retain function names to hand out usize identifiers for FDE addresses,
+    // which are then used to look up function names again for relocations, when `write_address` is
+    // called.
+    fn_names: Vec<String>,
+    table: gimli::write::FrameTable,
+}
+
+impl FrameSink {
```
iximeow: `FrameSink` and most of the machinery here could probably live in `cranelift-module`, and only require users in `cranelift-faerie` and `cranelift-simplejit` to provide a `gimli::write::Writer` impl to write addresses to functions.

cranelift-faerie/src/backend.rs
```diff
@@ -87,6 +247,174 @@ impl FaerieCompiledFunction {
...
+    }
+
+    /// Translate a Cranelift `RegUnit` to its matching `Register` for DWARF use.
+    ///
+    /// panics if `reg` cannot be translated - the requested debug information would be
+    /// unencodable.
+    pub fn translate_reg(&self, reg: RegUnit) -> gimli::Register {
+        match self.isa.name() {
+            "x86" => {
+                const X86_GP_REG_MAP: [u16; 16] = [
```
iximeow: `cargo-fmt` ate the formatting on the comments, oops.

cranelift-faerie/src/backend.rs
```diff
@@ -87,6 +247,174 @@ impl FaerieCompiledFunction {
...
+    cfa_def_reg: Option<isa::RegUnit>,
+    cfa_def_offset: Option<isize>,
+}
+
+struct DwarfRegMapper<'a> {
+    isa: &'a Box<dyn TargetIsa>,
+    reg_info: RegInfo,
+}
+
+impl<'a> DwarfRegMapper<'a> {
```
iximeow: I mentioned this on irc a few days ago but `DwarfRegMapper` could arguably live in cranelift-codegen as it's fundamentally an `int->int` function, and doesn't _need_ to return a `gimli::Register`. This and `CFIEncoder` might also make sense in `cranelift-module` as they are, pulling in a `gimli` dependency there and not in `cranelift-codegen`. @yurydelendik would `cranelift-module` be a useful place to share this logic translating `FrameLocationChange` to CFI instructions for your use in wasmtime?

philipc: Issue #729 is about the register mapping.

cranelift-faerie/src/backend.rs
```diff
@@ -74,9 +79,164 @@ pub struct FaerieBackend {
...
+    pub fn cie_for<'a>(
+        &mut self,
+        initial_state: &[ir::FrameLayoutChange],
+        reg_mapper: &'a DwarfRegMapper,
+    ) -> (gimli::write::CieId, CFIEncoder<'a>) {
+        let mut cie = CommonInformationEntry::new(
+            gimli::Encoding {
+                format: gimli::Format::Dwarf32,
+                version: 1,
+                address_size: 4,
```
philipc: This probably should be 8. Not sure if it will be used for the data you are writing though.

cranelift-faerie/src/backend.rs
```diff
@@ -74,9 +79,164 @@ pub struct FaerieBackend {
...
+    ) -> gimli::write::Result<()> {
+        // we only support PC-relative 4byte signed offsets for eh_frame pointers currently. Other
+        // encodings may be permissible, but aren't seen even by gcc/clang/etc, and have not been
+        // tested. Currently, relocations used for addresses expect to be relocating four bytes,
+        // PC-relative, and larger pointer sizes would require selection of other relocation types.
+        assert!(eh_pe.0 == 0x1b);
+
+        // if size is not 4, then the size indicated by `eh_pe` doesn't match with the pointer
+        // we're trying to encode. That's a logical bug, possibly in gimli?
+        assert!(size == 4);
```
philipc: This assert is wrong. The `size` parameter is only used for absolute pointer encodings.

cranelift-faerie/src/backend.rs
```diff
@@ -74,9 +79,164 @@ pub struct FaerieBackend {
...
+
+        self.write_address(address, size)
```
philipc: And so this should be passing a fixed 4, since we've already checked that this is using `DW_EH_PE_sdata4`.

cranelift-faerie/src/backend.rs
bjorn3: It should only be 8 on x86_64.

cranelift-faerie/src/backend.rs
```diff
@@ -74,9 +79,164 @@ pub struct FaerieBackend {
...
+        &mut self,
+        address: Address,
+        eh_pe: gimli::DwEhPe,
+        size: u8,
+    ) -> gimli::write::Result<()> {
+        // we only support PC-relative 4byte signed offsets for eh_frame pointers currently. Other
+        // encodings may be permissible, but aren't seen even by gcc/clang/etc, and have not been
+        // tested. Currently, relocations used for addresses expect to be relocating four bytes,
+        // PC-relative, and larger pointer sizes would require selection of other relocation types.
+        assert!(eh_pe.0 == 0x1b);
```
bjorn3: Can you compare `eh_pe` against `DW_EH_PE_sdata4` instead?

bjorn3 (commented):
:tada:

Next step will be running user code during cleanup.

cranelift-faerie/src/backend.rs
iximeow: this'd be `DW_EH_PE_pcrel | DW_EH_PE_sdata4`, but yeah definitely

cranelift-faerie/src/backend.rs
iximeow: I had definitely misunderstood the point of the `address_size` field here to be "size of an encoded pointer", rather than "size of a pointer for the target ISA". I'll make this pick the right size for the target.

4 and the asserts later about address size/pointer encoding, are a relic of trying to match CIE bytes I'd known work, in tracking down a crash in `libgcc`. In reality, I eventually figured out the issue I'd had is that `libgcc` doesn't handle escaped length fields in 64-bit DWARF, so I just tried `absptr` encodings and it still works fine! :tada:

I ought to add a note about why this specifically uses `Dwarf32` regardless of architecture - in other places I at least wanted to at least comment about architecture-specific magic values, so `format` deserves a `Do not change this!` mention.

Thinking through my confusion with the meaning of this and the `write_eh_pointer` argument, it seems like `write_eh_pointer` ought to not take `address_size` as a parameter. Since it only matters for absolute pointer encodings, I think it could work better to provide `address_size` as a field of an `absptr` variant for `DwEhPe`? Then it would be clear just from the function signature that the implementer is to write an address matching the encoding, and that `gimli` is not making assumptions about the number of bytes that would be written via `address_size`. If that sounds interesting, I can sketch out a PR to gimli to try that.

cranelift-faerie/src/backend.rs
```diff
@@ -74,9 +79,164 @@ pub struct FaerieBackend {
...
+            gimli::Encoding {
+                format: gimli::Format::Dwarf32,
+                version: 1,
+                address_size: 4,
+            },
+            // code alignment factor. Is this right for non-x86_64 ISAs? Probably could be 2 or 4
+            // elsewhere.
+            0x01,
+            // data alignment factor. Same question for non-x86_64 ISAs.
+            -0x08,
```
bjorn3: Why negative?

cranelift-faerie/src/backend.rs
```diff
@@ -74,9 +79,164 @@ pub struct FaerieBackend {
...
+            // ISA-specific, column for the return address (may be a register, may not)
+            reg_mapper.return_address(),
+        );
+
+        cie.fde_address_encoding = gimli::DwEhPe(0x1b);
```
bjorn3: Please use a constant provided by gimli here.

cranelift-faerie/src/backend.rs
```diff
@@ -74,9 +79,164 @@ pub struct FaerieBackend {
...
+                let reloc = faerie::artifact::Reloc::Raw {
+                    reloc: goblin::elf::reloc::R_X86_64_PC32,
+                    addend: 0,
+                };
+
+                self.artifact
+                    .link_with(
+                        faerie::Link {
+                            to: name,
+                            from: ".eh_frame",
```
bjorn3: Is this `__eh_frame` for machO?

cranelift-faerie/src/backend.rs
```diff
@@ -87,6 +247,174 @@ impl FaerieCompiledFunction {
...
+    /// Translate a Cranelift `RegUnit` to its matching `Register` for DWARF use.
+    ///
+    /// panics if `reg` cannot be translated - the requested debug information would be
+    /// unencodable.
+    pub fn translate_reg(&self, reg: RegUnit) -> gimli::Register {
+        match self.isa.name() {
+            "x86" => {
+                const X86_GP_REG_MAP: [u16; 16] = [
+                    // cranelift rax == 0 -> dwarf rax == 0
+                    0, // cranelift rcx == 1 -> dwarf rcx == 2
```
bjorn3: Please use [`gimli::X86_64`](https://docs.rs/gimli/0.19.0/gimli/struct.X86_64.html)

cranelift-faerie/src/backend.rs
```diff
@@ -87,6 +247,174 @@ impl FaerieCompiledFunction {
...
+            }
+        }
+    }
+
+    /// Get the DWARF location describing the call frame's return address.
+    ///
+    /// panics if that location is unknown - the requested debug information would be unencodable.
+    pub fn return_address(&self) -> gimli::Register {
+        match self.isa.name() {
+            "x86" => gimli::Register(0x10),
```
bjorn3: Same here

bjorn3 (changes requested):

cranelift-faerie/src/backend.rs
```diff
@@ -179,6 +508,46 @@ impl Backend for FaerieBackend {
...
+                    .into_iter()
+                    .flat_map(|(addr, change)| encoder.translate(&change).map(|inst| (addr, inst)));
+
+                for (addr, inst) in fde_insts.into_iter() {
+                    fd_entry.add_instruction(addr, inst);
+                }
+
+                frame_sink.add_fde(cie, fd_entry);
+            } else {
+                // we have a frame sink to write .eh_frames into, but are not collecting debug
```
bjorn3: Please panic here to make it easier to debug in that case.

cranelift-faerie/src/backend.rs
iximeow: That's a good question, I'll at least see what other toolchains name it. I also need to use a machO-appropriate relocation.

cranelift-faerie/src/backend.rs
iximeow: didn't know that existed! excellent.

cranelift-faerie/src/backend.rs
iximeow: I realized this morning that there's a case where we will reach this arm but not be a bug: if the user of cranelift has a per-function `nounwind`-like attribute, a `nounwind` function wouldn't need unwind information collected, even though there still would be a frame sink for other functions. Right now it happens to be that Lucet is all-or-nothing on unwind info, but as a rustc backend for example I could see some function being opt-out.

Given that, would you rather a panic to identify the case for later discussion, or should I discard the `else` entirely?

bjorn3: > but as a rustc backend for example I could see some function being opt-out.

I believe unwinding enabled/disabled is done per crate (using `-Cpanic=abort`), not per function.

cranelift-faerie/src/backend.rs
philipc: The `Encoding::address_size` field is used for all DWARF, where the normal format of an address is an unencoded pointer. `.eh_frame` is different from DWARF in that it encodes pointers, but even then it can't be the encoded pointer size because it is possible to use multiple encodings of different sizes.

`DwEhPe` is used to represent the constants, so it can't be changed. The `size` parameter to `write_eh_pointer` is documented as only being used for absolute pointers. I'm not sure we can do better than that. Maybe renaming the parameter to `absolute_address_size` would be worthwhile.

We may want to add a check in gimli to ensure you don't write `.eh_frame` with `Dwarf64`, since the LSB requires 32-bit length fields.

cranelift-faerie/src/backend.rs
philipc: Also, this shouldn't call `write_address`, since `write_address` must only be used for absolute pointers. If you want to share code between the two then put it in another function.

cranelift-faerie/src/backend.rs
iximeow: I've moved this to `cranelift-module` and noted some reasoning in that issue. I'll leave this unresolved until we decide here or there what to do with `DwarfRegMapper`.

cranelift-faerie/src/backend.rs
iximeow: I've updated the comment in source to reflect why, but reiterated here: -8 is mostly-arbitrary. I believe on x86 the number _could_ be just about anything, but -8 is what other toolchains use. By using the same parameters in our CIEs, linkers can reuse CIEs and save some space. Other tools picking -8 is likely because it results in data offsets in CFI programs that match stack slots, counting from 0 at the start of the call frame.

cranelift-faerie/src/backend.rs
iximeow: I didn't particularly care about sharing code, I just misunderstood which of `write_eh_pointer` and `write_address` were supposed to do what.

cranelift-faerie/src/backend.rs
iximeow: This assert is now gone.

cranelift-faerie/src/backend.rs
philipc: LLVM uses the stack slot size, and makes it negative if the stack grows down. It doesn't matter as long is it is a factor, and matching other tools makes sense.

cranelift-faerie/src/backend.rs
philipc: rustc has the unstable `#[unwind(allowed/aborts)]` to control per function, as well I think it has different requirements for `extern "C"`, but I don't know the details of that and it is under ongoing discussion.

cranelift-module/src/dwarf.rs
```diff
@@ -0,0 +1,295 @@
...
+
+const PC_SDATA4: u8 = DW_EH_PE_pcrel.0 | DW_EH_PE_sdata4.0;
+
+/// FrameSink maintains state for and an interface to construct DWARF frame information for a
+/// cranelift-produced module.
+pub struct FrameSink {
+    // we need to retain function names to hand out usize identifiers for FDE addresses,
+    // which are then used to look up function names again for relocations, when `write_address` is
+    // called.
+    fn_names: Vec<String>,
```
philipc: `ModuleContents` does already contain something like this, not sure if it is worth trying to reuse that, but maybe `FuncId::as_u32` can be the identifier.

cranelift-faerie/src/backend.rs
yurydelendik: >  @yurydelendik would cranelift-module be a useful place to share this logic translating FrameLocationChange to CFI instructions for your use in wasmtime?

@iximeow yes, and in fact, I would like to have entire cratelift-module change for other reasons.

alexcrichton:

iximeow:
Closing this with the hope that either [#1181](https://github.com/bytecodealliance/wasmtime/issues/1181) will result in a change that supplants PR, or that this will be significantly different when [#1181](https://github.com/bytecodealliance/wasmtime/issues/1181) is resolved.

PR #904, opened 2019-08-16
Use aliasing instead of copying in simple_preopt

bnjbvr:
As discussed by emails, this replaces copies in simple_preopt to use more aliasing instead. There's another `replace_with_aliases` method in the DFG data structure, but it doesn't exactly do what we want (it replaces the results of one instruction by the results of another). Arguably, this could be put in the DFG data structure too, with a different name.

### DIFF
--- a/cranelift-codegen/src/simple_preopt.rs
+++ b/cranelift-codegen/src/simple_preopt.rs
@@ -18,6 +18,20 @@ use crate::ir::{
 };
 use crate::timing;
 
+#[inline]
+/// Replaces the result_index nth result of instruction inst to an alias of the given value, and
+/// replaces the instruction with a nop.
+fn replace_with_alias(dfg: &mut DataFlowGraph, inst: Inst, result_index: usize, value: Value) {
+    // Replace the result value by an alias.
+    let results = dfg.detach_results(inst);
+    debug_assert!(results.len(&dfg.value_lists) > result_index);
+    let result = results.get(result_index, &dfg.value_lists).unwrap();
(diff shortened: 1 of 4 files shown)


cranelift-codegen/src/simple_preopt.rs
```diff
@@ -18,6 +18,20 @@ use crate::ir::{
...
 
+#[inline]
+/// Replaces the result_index nth result of instruction inst to an alias of the given value, and
+/// replaces the instruction with a nop.
+fn replace_with_alias(dfg: &mut DataFlowGraph, inst: Inst, result_index: usize, value: Value) {
+    // Replace the result value by an alias.
+    let results = dfg.detach_results(inst);
+    debug_assert!(results.len(&dfg.value_lists) > result_index);
+    let result = results.get(result_index, &dfg.value_lists).unwrap();
+    dfg.change_to_alias(result, value);
```
lars-t-hansen: Is this code actually correct if there's morethan one result in the original results list?  The code you lifted this from asserts that there's only one result.  (Not a rhetorical question; not sure what the correct answer is.)  To simplify analysis one could rename this as replace_single_result_with_alias (indeed replace_result_with_alias would be a better name) and assert that there's exactly one result.

bnjbvr: Uhh, there's indeed something fishy. I'd hoped we could call the function twice when we have two results we want to replace, but then it means detaching results twice, which will probably panic. Will open a follow-up.

PR #914, opened 2019-08-20
Make fold_redundant_jumps() feature-gated on basic-blocks

sstangl:
There are bugs in `fold_redundant_jumps` (https://bugzilla.mozilla.org/show_bug.cgi?id=1573792). They don't seem easy to address, because they involve rewriting instructions after register allocation.

Because `fold_redundant_jumps()` really only exists to clean up regalloc cruft for the `basic-blocks` feature, we can feature-gate it on `basic-blocks` to unblock the rest of the team.


bnjbvr:
LGTM, it now needs to revert the reverting patch, though.

sstangl:
Changed to revert the reverting patch. Sorry about the lag! I was hopeful that it would be easier to fix.

PR #919, opened 2019-08-21
Fix segfault due to b64 encoding

### LINKED ISSUE #911: Returning b64 results in a segmentation fault
As a part of https://github.com/CraneStation/cranelift/pull/890 I added support for running CLIF directly as a filetest or through `clif-util run`; the current idea is that functions would return a boolean that indicates whether the function ran successfully or not. This works with the other boolean types, but a test like the following results in a segmentation fault:
```
test run
function %test_b64() -> b64 {
ebb0:
    v0 = bconst.b64 true
    return v0
}
; run
```
```
$ target/debug/clif-util run filetests/isa/x86/run-const.clif 
Segmentation fault (core dumped)
$ coredumpctl dump
           PID: 32159 (clif-util)
           UID: 1000 (abrown)
           GID: 1000 (abrown)
        Signal: 11 (SEGV)
     Timestamp: Tue 2019-08-20 09:15:19 PDT (1min 36s ago)
  Command Line: target/debug/clif-util run filetests/isa/x86/run-const.clif
    Executable: /home/abrown/Code/cranelift/target/debug/clif-util
 Control Group: /user.slice/user-1000.slice/session-c3.scope
          Unit: session-c3.scope
         Slice: user-1000.slice
       Session: c3
     Owner UID: 1000 (abrown)
       Boot ID: 118deca6420a48b69f1c2a455c2500f9
    Machine ID: e9d29f116e2f4d09baf478e99d04fbff
      Hostname: abrown-desk.amr.corp.intel.com
       Storage: /var/lib/systemd/coredump/core.clif-util.1000.118deca6420a48b69f1c2a455c2500f9.32159.1566317719000000.lz4
       Message: Process 32159 (clif-util) of user 1000 dumped core.
                
                Stack trace of thread 32159:
                #0  0x00007fec72edc00f n/a (ld-linux-x86-64.so.2)
                #1  0x00005642de402114 n/a (/home/abrown/Code/cranelift/target/debug/clif-util)
```
I am running a Linux kernel 5.1.18 on Fedora 30 and would expect the `system_v` calling convention to be the default here. I may need some help figuring out why `b64` does not work here.

abrown:
Prior to this patch, bconst.b64 encoded its instruction with a 32-bit immediate that caused improper decoding of the MOV instruction; fixes #911.


awortman-fastly:
can we avoid the 64-bit mov entirely here? since the immediate is either 0 or 1, it doesn't seem like we _need_ to specify a 64bit destination register, and just rely on zero-extension to do its thing. I _think_ that means an equally correct fix would be just removing `.rex().w()` from the recipe?

abrown:
@awortman-fastly, that makes sense to me and the test passes; let me try it that way and see if @bnjbvr has a strong opinion one way or another. If we don't want to use the 32-bit encoding we can revert 56f0733; otherwise we can squash when we merge.

bnjbvr:
@awortman-fastly @abrown It's an elegant solution indeed! Unfortunately, since there's no REX variant with this solution, it means it restricts the possible output register to the 8 first registers in 64-bit mode. I think a proper fix is a mix of both: first the 8 bytes recipe with REX.W, and then after this the 32-bit immediate recipe. (REX must come first just so it's selected first by register allocation, with the fewer constraints on register assignment, and then the relaxation phase may try to use the 4 bytes recipe if that fits the operand constraints.)
Does it make sense?

awortman-fastly:
Ah, I'd misunderstood how `rex()` and `w()` affect the recipe :) The `W` bit is really what I meant isn't necessary here, and thought that `rex()` got an accessor to that, rather than indicating that REX is an acceptable prefix for the recipe. Since `r8d`-`r15d` get the same zero-extension, we just need to not emit the `W` prefix right? f.ex `41b901000000` encodes `mov r9d, 1`, where the `W` really adds nothing extra (but more zeroes)

(sorry for the back and forth on your small change @abrown!)

filetests/isa/x86/bconst-run.clif
```diff
@@ -0,0 +1,9 @@
+test run
+
+; this verifies that returning b64 immediates does not result in a segmentation fault, see https://github.com/CraneStation/cranelift/issues/911
+function %test_b64() -> b64 {
+ebb0:
+    v0 = bconst.b64 true
```
bjorn3: Maybe use a binemit test instead. This may succeed even if the encoding is wrong.

abrown:
I don't see an encoding for a `REX + 8B /r` in the Intel manual; I only see  `REX.W + 8B /r` (for `MOV r64,r/m64`) and `8B /r` (for `MOV r32,r/m32`). Leaving the `W` bit unset in the `REX` byte doesn't make sense to me because `W` means "64 bit operand size" and that is what we would be trying to do with that encoding; the non-REX encoding _would_ try to do a 32-bit operand. But I may be missing something from your comment? Why avoid the `W` bit?

bjorn3:
> Why avoid the W bit?

So that the immediate can be 32bit. Implicit zero extension should take care of writing the right 64bit value into the register if I understand it correctly.

abrown:
Yeah, I think in the case of just wanting a 32-bit operand that makes sense, but just encode it as `B8 + rd id` (I used the wrong one above)--no `REX` and therefore no `W` set. What I'm saying is that if we need a 64-bit version as @bnjbvr mentioned then it doesn't make sense to me to just have `REX B8 + rd io`--that doesn't exist in the Intel manual. It only makes sense to me to have `REX.W B8 + rd io`.

abrown:
I do see this comment "Use of the REX.R prefix permits access to additional registers (R8-R15)" so perhaps that is what @awortman-fastly meant?

awortman-fastly:
I think the manual only lists `REX.W + XX` encodings to indicate that 64bit operands are valid for the instruction (eg `REX.W + 88` isn't present) - the lack of a pairing there doesn't mean that `REX.B B8` would be forbidden.

Section 3.1.1.1, `Opcode Column in the Instruction Summary Table (Instructions without VEX Prefix)` is the one that applies here, since the encoding is `+rd io` as noted in MOV encodings, where the table describes REX.B selecting R8-R15 in place of the lower general-purpose registers for `+rd` encoding instructions

abrown:
I think I see what you mean; you're aiming to make more registers available, right? (Agreed that the instruction table may not have all pairings). I do see the following on Table B-15 (page 2590) and perhaps you are trying to use the second of these?

```
__MOV - Move Data__
...
| immediate to register | 0100 000B : 1100 011w : 11 000 reg : imm |
| immediate32 to qwordregister (zero extend) | 0100 100B 1100 0111 : 11 000 qwordreg : imm32 |
| immediate to register (alternate encoding) | 0100 000B : 1011 w reg : imm |
| immediate64 to qwordregister (alternate encoding) | 0100 100B 1011 1000 reg : imm64 |
...
```

However, I don't see a good way to set the B bit on the REX byte in cranelift (not that that can't be fixed).

abrown:
Never mind, I just don't understand cranelift yet: I can see that the REX.B bit is correctly set for the current 64-bit immediate implementation.

abrown:
I can also verify that when I try to encode bconst.b64 without the `W` bit set and a 32-bit boolean (i.e. `e.enc64(bconst.bind(B64), rec_pu_id_bool.opcodes(vec![0xb8]).rex());`) the `run` test still passes and I see an output encoding of `41 ba 00000001` in binemit.

awortman-fastly:
REX.B is selected based on the operand register when the selected recipe is actually executed - that's handled by [`rex1`](https://github.com/CraneStation/cranelift/blob/39f3fd024ca02414b4904d2667aa7925043f026d/cranelift-codegen/src/isa/x86/binemit.rs#L32-L38) in [`pu_id_bool`](https://github.com/CraneStation/cranelift/blob/8977442cb8df2264d24848965b78f77447a6719e/cranelift-codegen/meta/src/isa/x86/recipes.rs#L897-L910) and the same elsewhere. `rex2` would do the same REX bit selection as appropriate for ModRM-encoded operands, which is why it's not used here (but is elsewhere). So just `.rex()` is sufficient to indicate that a REX prefix is wanted with the recipe (I'm pretty sure? @bnjbvr definitely knows that better than me), and permits the possibility of `REX.B`, but REX alone has no effect on the 16/32/64bit opcodes.

re. which encoding line, I meant the third one in fact! The fourth is what we get with `REX.W B8`, and the third line as a 32-bit mov would have `w=1` for `1011 1RRR` or `0xb8` to `0xbf`. `REX.B` would select the upper registers if `rex1` is given a register that would necessitate the extra bit (r8-r15).

filetests/isa/x86/bconst-run.clif
abrown: I added a binemit test as well.

abrown:
@awortman-fastly, thanks; I think we are on the same page now. I implemented exactly that for the comment above and the emitted `41 ba ...` for register `%r10` matches what you describe (`REX.B == 0100 0001` and `RRR` == `010`) . Let me push it and @bnjbvr can comment.

filetests/isa/x86/bconst-binemit.clif
```diff
@@ -0,0 +1,8 @@
+test binemit
+target x86_64
+
+function %test_b64() -> b64 {
+ebb0:
+[-, %r10]   v0 = bconst.b64 true    ; bin: 41 ba 00000001
```
bnjbvr: Nice! Can you move this in binary64.clif, please? And add one using rax as the output register please?

abrown: Should I rename `bconst-run.clif` as well?

bnjbvr: Yeah, or even feel free to create a subdirectory "run" if you want to (I do not mind either way).

PR #920, opened 2019-08-21
[meta] Generate doc comments for the encodings tables;

bnjbvr:
A simple PR to explain in doc comments what some of the encodings generated code does.


bnjbvr:
Good idea, I've added this!

PR #921, opened 2019-08-21
[codegen] Extend InstBuilder with advanced custom behaviors;

bnjbvr:
This experiments with including more custom behaviors in the InstBuilder, by having all the existing methods be renamed to `default_{INSTNAME}`, and the actual `{INSTNAME}` method just call into it. This way, we can override the `{INSTNAME}` method and add custom behaviors. This PR implements checks to not emit dumb code for iadd/iadd_imm anymore, as an example of what can be done. (This could help with fixing some other bugs, since this has been requested somewhere else, and we could also remove many of the simple_preopt plain replacements.)

@sunfishcode this is far from being ready to land, but does this approach look reasonable to you, before I keep on working on it? (There's apparently no way in Rust to have a default impl for a trait method, and use this default impl from an impl that implements this method, hence the "default_" method trick.)


cranelift-codegen/meta/src/gen_inst.rs
```diff
@@ -1025,7 +1027,46 @@ fn gen_inst_builder(inst: &Instruction, format: &InstructionFormat, fmt: &mut Fo
...
+    let proto = format!(
+        "{}{}({}) -> {}",
+        inst.snake_name(),
+        tmpl,
+        args.join(", "),
+        rtype
+    );
+    fmt.doc_comment(&inst.doc);
+    fmt.line("#[allow(non_snake_case)]");
+    fmtln!(fmt, "fn {} {{", proto);
```
bjorn3: Can you please mark it `#[inline]`?

cranelift-codegen/src/ir/inst_builder.rs
```diff
@@ -0,0 +1,140 @@
...
+/// Any type implementing `InstBuilderBase` gets all the `InstBuilder` methods for free, with
+/// some special enhancements for free.
+impl<'f, T: InstBuilderBase<'f>> InstBuilder<'f> for T {
+    fn iadd(self, x: ir::Value, y: ir::Value) -> Value {
+        let dfg = self.data_flow_graph();
+        let ty = dfg.value_type(x);
+        if ty == dfg.value_type(y) {
+            // See if one of the operands is a constant immediate.
+            if let Some(y_imm) = try_unwrap_iconst(dfg, y) {
+                // Otherwise, we can emit iadd_imm.
```
bjorn3: s/otherwise/then

bnjbvr:
> Are you planning to const fold iadd + load/store to load_complex/store_complex?

Yeah, it could be implemented there too (checking if the ISA supports complex operands first). This is really a proof-of-concept, since most of the imm variants aren't handled in this PR.

bnjbvr:
@sunfishcode ping?

bnjbvr:
Closing: it should be a draft PR but there's no clear way to mark it now as such, unless I'm missing something. Also, I will open an issue to discuss what this is solving, alternative ways, problems this could cause (infinite recursion in InstBuilder, or iloops in the isel interpreter), etc.

PR #927, opened 2019-08-22
cranelift-wasm: bump `wasmparser` version to `0.37.0`

YaronWittenstein:


YaronWittenstein:
@bnjbvr thank you!
I think this change requires a version bump of `cranelift-wasm` too right?

PR #930, opened 2019-08-23
[codegen] Make scalar_to_vector's output type a lane of its input type

bnjbvr:
This makes the scalar_to_vector instruction type sane: it expects the input value to be a lane of the output value. Also removed the reference to XMM registers, because this is a shared instruction for all platforms.

cc @abrown FYI


bnjbvr:
Thanks for the review!

Issue #908, opened 2019-08-19
[codegen] Use roundss/roundsd when available for Ceil/Floor/Trunc/Nearest


bnjbvr:
Currently, the ceil/floor/trunc/nearest instructions, [as defined](https://github.com/CraneStation/cranelift/blob/master/cranelift-codegen/meta/src/shared/instructions.rs), don't have a legal machine representation on any platform. They're thus handled by the [libcall](https://github.com/CraneStation/cranelift/blob/master/cranelift-codegen/src/ir/libcall.rs#L80-L84) legalization, which will do its magic to replace them with a function call.

On x86, though, we can do better when we have SSE4.1 support: use the [roundsd](https://www.felixcloutier.com/x86/roundsd) for f64 / roundss for f32 instruction, with the right rounding mode (each of them matches one of the four instructions).

This means adding new encodings and recipes in the meta language for these instructions: cargo-culting other instructions' encodings and recipes for this should do. Note there's going to be a need for an encoding predicate, to make sure SSE4.1 or later is enabled.

bnjbvr:
I've been let known that it's actually supposed to be implemented, since it's there in our encodings generation: https://github.com/CraneStation/cranelift/blob/master/cranelift-codegen/meta/src/isa/x86/encodings.rs#L1607-L1619 .

It doesn't seem to be picked up, when I run the following clif test and look at the generated assembly with `cargo run compile -dD /tmp/a.clif`:

```
target x86_64 has_sse41=true

function %ceil() -> f32 {
ebb0:
    v1 = f32const 0.0
    v2 = ceil v1
    return v2
}
```

This generates the following assembly:

```
Disassembly of 25 bytes:
   0:	40 55                	push	rbp
   2:	48 89 e5             	mov	rbp, rsp
   5:	40 0f 57 c0          	xorps	xmm0, xmm0
   9:	48 b8 00 00 00 00 00 00 00 00	
				movabs	rax, 0
  13:	40 ff d0             	call	rax
  16:	40 5d                	pop	rbp
  18:	c3                   	ret
```

The movabs is supposed to be relocated to the address of the libcall's call later on, and there's a call instruction proving this uses a libcall call. So the encoding isn't picked up, while it's supposed to be available.

ryzokuken:
After debugging, we found out that this was infact fixed and regressed a month ago through https://github.com/Cranestation/cranelift/commit/66fb5e2d3b90074924ea537bd82df49c4f51aad1.

[...]

PR #931, opened 2019-08-23
Use roundss/roundsd when available for Ceil/Floor/Trunc/Nearest

ryzokuken:
Fixes: https://github.com/CraneStation/cranelift/issues/908

/cc @bnjbvr @wingo @caitp


bnjbvr:
Thanks! As said on IRC, a few more changes are needed in the encodings and/or recipes files, to update the selection of settings according to whether they're related to SIMD instructions or not.

cranelift-codegen/meta/src/isa/x86/encodings.rs
```diff
@@ -585,6 +585,8 @@ pub fn define(
     let use_bmi1 = settings.predicate_by_name("use_bmi1");
     let use_ssse3 = settings.predicate_by_name("use_ssse3");
```
bnjbvr: Can you remove this, please? It's now unused.

PR #934, opened 2019-08-23
clif-wasm version bump 0.40.01 

YaronWittenstein:
`clif-wasm` version bump to `0.40.01` (because of `wasmparser` version bump to `0.37.0`)

I wasn't sure what new version number to set though

### DIFF
--- a/cranelift-wasm/Cargo.toml
+++ b/cranelift-wasm/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "cranelift-wasm"
-version = "0.40.0"
+version = "0.40.01"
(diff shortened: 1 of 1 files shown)


bnjbvr:
@sunfishcode may have more opinions here, usually we update all the crates' versions at once, with minor. Also CI apparently complains that 01 isn't a valid version number.

YaronWittenstein:
> @sunfishcode may have more opinions here, usually we update all the crates' versions at once, with minor. Also CI apparently complains that 01 isn't a valid version number.

I can assist with bumping all crates versions to the number you'll ask me to

Issue #758, opened 2019-04-30
[simplejit] Alignment constraints should be inferred from TargetMachine


bnjbvr:
Follow-up of https://github.com/CraneStation/cranelift/pull/721:

```
const EXECUTABLE_DATA_ALIGNMENT: u8 = 0x10;
const WRITABLE_DATA_ALIGNMENT: u8 = 0x8;
```

> These two numbers appear somewhat magical. Why 0x10 and 0x8? If these are meant to be target-specific, we should have TargetMachine provide them.

(Note: they might even be compiler- or OS- dependent.)

bnjbvr:
@philipc would you be interested to take a stab at this, and arrange the read alignment constraint btw?

philipc:
Not right now, but maybe later. It's something I would have to learn about. e.g. I didn't know all of https://llvm.org/docs/Frontend/PerformanceTips.html#when-to-specify-alignment; does cranelift already take data alignment into consideration during codegen?

bnjbvr:
I'd need to check more thoroughly, but I don't think so, since x64 CPUs just do a slower read for unaligned reads, for non-SIMD data, if I remember correctly. On ARM, this is quite another story...

PR #937, opened 2019-08-25
Add cranelift-object

### LINKED ISSUE #761: Alternative object file backend
I have an initial implementation of COFF object file support currently living in an [`object-write`](https://github.com/philipc/object-write/blob/master/src/coff.rs) repository. This is an alternative implementation to faerie, and it presents a lower level API than faerie. This API requires the user to define sections, symbols, and relocations, rather than functions and data objects. Currently `object-write` also supports writing ELF, but not yet Mach-O.

I also have an initial cranelift module [`Backend` implementation](https://github.com/philipc/cranelift/tree/object/cranelift-object/src) that uses `object-write`. Compared to the faerie module backend, I think this implementation is still fairly straight forward, and the lower level API is more flexible.

I also have investigated changing `rustc_codegen_cranelift` to use this, and it can successfully compile and link a simple program using the MSVC linker. It appears that at least the wrong calling convention is being used though, so it currently segfaults. This can also use the ELF support in `object-write`, and that passes the current `rustc_codegen_cranelift` tests.

This is all still in development, but I'm interested in early feedback about the direction this is going. I wrote `object-write` for other purposes, but I think it's a good fit for cranelift too, and may be useful for #511.

philipc:
Fixes #761

### DIFF
--- a/cranelift-object/src/backend.rs
+++ b/cranelift-object/src/backend.rs
@@ -1,0 +1,584 @@
+//! Defines `ObjectBackend`.
+
+use crate::traps::{ObjectTrapSink, ObjectTrapSite};
+use cranelift_codegen::binemit::{
+    Addend, CodeOffset, NullStackmapSink, NullTrapSink, Reloc, RelocSink,
+};
+use cranelift_codegen::entity::SecondaryMap;
+use cranelift_codegen::isa::TargetIsa;
+use cranelift_codegen::{self, binemit, ir};
+use cranelift_module::{
+    Backend, DataContext, DataDescription, DataId, FuncId, Init, Linkage, ModuleNamespace,
+    ModuleResult,
+};
+use object::write::{Object, Relocation, SectionId, StandardSection, Symbol, SymbolId};
+use object::{RelocationEncoding, RelocationKind, SymbolKind, SymbolScope};
+use std::collections::HashMap;
+use target_lexicon::PointerWidth;
+
+#[derive(Debug)]
+/// Setting to enable collection of traps. Setting this to `Enabled` in
+/// `ObjectBuilder` means that `ObjectProduct` will contains trap sites.
+pub enum ObjectTrapCollection {
+    /// `ObjectProduct::traps` will be empty
+    Disabled,
+    /// `ObjectProduct::traps` will contain trap sites
+    Enabled,
+}
+
+/// A builder for `ObjectBackend`.
+pub struct ObjectBuilder {
+    isa: Box<dyn TargetIsa>,
+    name: String,
+    collect_traps: ObjectTrapCollection,
+    libcall_names: Box<dyn Fn(ir::LibCall) -> String>,
+    function_alignment: u64,
+}
+
+impl ObjectBuilder {
+    /// Create a new `ObjectBuilder` using the given Cranelift target, that
+    /// can be passed to
+    /// [`Module::new`](cranelift_module/struct.Module.html#method.new].
+    ///
+    /// `collect_traps` setting determines whether trap information is collected in the
+    /// `ObjectProduct`.
+    ///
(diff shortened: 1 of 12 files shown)


cranelift-object/src/backend.rs
```diff
@@ -0,0 +1,561 @@
...
+        _writable: bool,
+        _align: Option<u8>,
+    ) {
+        let scope = match linkage {
+            Linkage::Import => SymbolScope::Unknown,
+            Linkage::Local => SymbolScope::Compilation,
+            Linkage::Export | Linkage::Preemptible => SymbolScope::Dynamic,
+        };
+        // TODO: this matches rustc_codegen_cranelift, but may be wrong.
+        let weak = linkage == Linkage::Preemptible;
```
bjorn3: Maybe extract this and scope to a helper function to prevent deduplication.

cranelift-object/src/backend.rs
```diff
@@ -0,0 +1,561 @@
...
+                    &mut reloc_sink,
+                    &mut trap_sink,
+                    &mut stackmap_sink,
+                )
+            };
+        }
+
+        let symbol = self.functions[func_id].unwrap();
+        let section = self.object.section_id(StandardSection::Text);
+        let offset = self.object.add_symbol_data(symbol, section, &code, 1);
```
bjorn3: Faerie uses an alignment of 0x10.

cranelift-object/src/backend.rs
```diff
@@ -0,0 +1,561 @@
...
+        let section = self.object.section_id(if writable {
+            StandardSection::Data
+        } else if relocs.is_empty() {
+            StandardSection::ReadOnlyData
+        } else {
+            StandardSection::ReadOnlyDataWithRel
+        });
+        let offset =
+            self.object
+                .add_symbol_data(symbol, section, &data, u64::from(align.unwrap_or(1)));
```
bjorn3: Faerie uses a min alignment of 0x8 when it is writable.

cranelift-object/src/backend.rs
philipc: Right, but it's a magic number with no documentation as to why it was selected. I prefer to stick to the minimum without good reason. My opinion is that this should be something the caller is required to select, or at least it should be an architecture dependent value (see #758). Same for the data alignment.

bjorn3: `cranelift-object` doesn't provide any way to select the code alignment.

philipc: I think I could change `cranelift-module` to allow that.

philipc: I added an `ObjectBuilder` setting since you'll probably want the same alignment for all functions.

philipc:
I'll make a new object release and update this PR once https://github.com/gimli-rs/object/pull/127 is sorted.

bnjbvr:
(Please don't add merge commits to PRs, if that's possible. We've been trying really hard to have a linear history, and to rebase everything on top of previous work, so as to avoid merge commits and make bisection trivial. It's a small price to pay off for better debugging.)

philipc:
When I rebase, should I squash https://github.com/CraneStation/cranelift/pull/937/commits/38d0289b67c0d0d8a637ac74f83166ff99acafd6 so that the history is bisectable?

bnjbvr:
Ideally, commits would be atomic, do one thing and keep on passing all the tests (that's really best effort, compared to the strong requirement of not having merge commits). If this matches this criteria, it's fine to keep it separated in my opinion. (We should probably agree on these conventions and put them in the CONTRIBUTING.md file at some point.)

philipc:
CONTRIBUTING.md actually does pretty much say that already :)

philipc:
Rebased and squashed.

sunfishcode:
Thanks!

PR #940, opened 2019-08-26
Switch from mmap to memmap to support Windows

tschneidereit:
#890 introduced a dependency on the `mmap` crate, which doesn't support Windows. A quick investigation showed mmap is not actively maintained, so [people mostly use the `memmap` crate](https://github.com/cgag/loc/issues/13#issuecomment-256178954). This PR switches to that, and seems to work without issues.

### DIFF
--- a/cranelift-filetests/src/function_runner.rs
+++ b/cranelift-filetests/src/function_runner.rs
@@ -4,7 +4,7 @@ use cranelift_codegen::ir::Function;
 use cranelift_codegen::isa::{CallConv, TargetIsa};
 use cranelift_codegen::{settings, Context};
 use cranelift_native::builder as host_isa_builder;
-use mmap::{MapOption, MemoryMap};
+use memmap::MmapMut;
 use region;
 use region::Protection;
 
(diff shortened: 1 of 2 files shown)


cranelift-filetests/src/function_runner.rs
```diff
@@ -67,19 +67,23 @@ impl FunctionRunner {
...
-                code_page.data(),
+                code_page.as_mut_ptr(),
                 relocs,
                 traps,
                 stackmaps,
             );
-            region::protect(code_page.data(), code_page.len(), Protection::ReadExecute)
-                .map_err(|e| e.to_string())?;
-            mem::transmute(code_page.data())
+            region::protect(
```
bjorn3: Please use `let code_page = code_page.make_exec().map_err(|e| e.to_string())?;` instead. Maybe switch [`cranelift-simplejit`](https://github.com/CraneStation/cranelift/blob/master/cranelift-simplejit/src/memory.rs) to that method too.

bnjbvr: @tschneidereit said on IRC he'd prefer this to be done as a follow-up; will open a new issue for this.

PR #943, opened 2019-08-26
Improve extractlane and replacelane on x86

abrown:
Since cranelift stores float values in XMM registers there is no need for any register movement when inserting/extracting a float into/from a vector, which are also stored in XMM registers. This change depends on #868 and adds several x86-specific instructions in order to legalize the cranelift instructions to x86.


abrown:
@bnjbvr rebased to fix conflicts.

cranelift-codegen/src/isa/x86/enc_tables.rs
```diff
@@ -893,3 +894,63 @@ fn expand_fcvt_to_uint_sat(
...
+                };
+                // then we must bitcast to the right type
+                pos.ins()
+                    .with_result(original_result)
+                    .raw_bitcast(value_type.lane_type(), shuffled);
+            } else {
+                // remove the extractlane instruction, leaving the float where it is
+                pos.ins()
+                    .with_result(original_result)
+                    .raw_bitcast(value_type.lane_type(), arg);
```
bnjbvr: Maybe control flow could be slightly simplified here: both branches end up with a raw_bitcast, so the `if` could yield the raw_bitcast argument instead?

filetests/isa/x86/extractlane-run.clif
```diff
@@ -0,0 +1,31 @@
+test run
+set enable_simd
+
+function %test_extractlane_b8() -> b8 {
+ebb0:
+    v0 = bconst.b8 true
+    v1 = splat.b8x16 v0
+    v2 = extractlane v1, 10
```
bnjbvr: Hmm, even if it reads the first lane, this test is correct. Is there a way we could use a bconst.b8x16 above?

filetests/isa/x86/extractlane-run.clif
```diff
@@ -0,0 +1,31 @@
...
+    v1 = extractlane v0, 1
+    v2 = icmp_imm eq v1, 2
+    return v2
+}
+; run
+
+function %test_extractlane_f32() -> b1 {
+ebb0:
+    v0 = f32const 0x42.42
+    v1 = splat.f32x4 v0
```
bnjbvr: Can we use a vconst.f32x4 instead and only put the float value in the one lane that is read later? (otherwise even if it read the wrong lane, the test would be correct.)

cranelift-codegen/meta/src/isa/x86/instructions.rs
```diff
@@ -290,5 +290,22 @@ pub fn define(
...
+    let x = &operand("x", TxN);
+    let a = &operand("a", &TxN.lane_of());
+
+    ig.push(
+        Inst::new(
+            "x86_pextr",
+            r#"
+        Extract lane ``Idx`` from ``x``.
+        The lane index, ``Idx``, is an immediate value, not an SSA value. It
+        must indicate a valid lane index for the type of ``x``.
```
bnjbvr: Can you precise how this behaves with respect to the upper bits in the result (zero extension), please?

cranelift-codegen/meta/src/isa/x86/encodings.rs
```diff
@@ -1730,12 +1741,16 @@ pub fn define(
     // written to the low doubleword of the register and the regiser is zero-extended to 128 bits."
     for ty in ValueType::all_lane_types().filter(|t| t.lane_bits() >= 8) {
         let instruction = scalar_to_vector.bind_vector_from_lane(ty, sse_vector_size);
-        let template = rec_frurm.opcodes(vec![0x66, 0x0f, 0x6e]); // MOVD/MOVQ
-        if ty.lane_bits() < 64 {
-            // no 32-bit encodings for 64-bit widths
-            e.enc32(instruction.clone(), template.clone());
+        if ty.is_float() {
+            e.enc_32_64_rec(instruction, rec_null_fpr, 0);
```
bnjbvr: If I do:

```
v0 = f32const 0x42.42
v1 = splat.f32x4 v0
v2 = extractlane 0 v1
v3 = scalar_to_vector v2
```

Then my understanding is that, because of the previous commit and this change, the high bits of v3 will be non-zeroes, in contrary to what the comment of scalar_to_vector suggests. Am I understanding correctly? If so, I think this code should be reworked, and it'd be nice to add a test.

cranelift-codegen/meta/src/isa/x86/instructions.rs
```diff
@@ -307,5 +307,74 @@ pub fn define(
...
+    ig.push(
+        Inst::new(
+            "x86_pinsr",
+            r#"
+        Insert ``y`` into ``x`` at lane ``Idx``.
+        The lane index, ``Idx``, is an immediate value, not an SSA value. It
+        must indicate a valid lane index for the type of ``x``.
+        "#,
+        )
+        .operands_in(vec![x, Idx, y])
```
bnjbvr: Similarly to what you've done for insertps below, should this be restricted to "Int x N" input / output operands?

cranelift-codegen/meta/src/isa/x86/instructions.rs
```diff
@@ -307,5 +307,74 @@ pub fn define(
...
+    let x = &operand("x", FxN);
+    let y = &operand_doc("y", &FxN.lane_of(), "New lane value");
+    let a = &operand("a", FxN);
+
+    ig.push(
+        Inst::new(
+            "x86_insertps",
+            r#"
+        Insert a lane of ``y`` into ``x`` at using ``Idx`` to encode both which lane the value is 
+        extracted from  and which it is inserted to. This is similar to x86_pinsr but inserts
```
bnjbvr: nit: double space between "from" and "and".

cranelift-codegen/meta/src/isa/x86/instructions.rs
```diff
@@ -307,5 +307,74 @@ pub fn define(
...
+
+    let x = &operand("x", FxN);
+    let y = &operand("y", FxN);
+    let a = &operand("a", FxN);
+
+    ig.push(
+        Inst::new(
+            "x86_movsd",
+            r#"
+        Move the low 64 bits of the float vector ``y`` to the low 64 bits of float vector ``x``
```
bnjbvr: Isn't this equivalent to a sequence of copy + raw_bitcast from SIMD vector to scalar? If so, for the sake of keeping the instruction set maximally minimal, could we not introduce this new instruction?

cranelift-codegen/src/isa/x86/enc_tables.rs
```diff
@@ -954,3 +954,64 @@ fn convert_extractlane(
...
+        args: [vector, replacement],
+        lane,
+    } = pos.func.dfg[inst]
+    {
+        // remove the original extractlane instruction
+        pos.remove_inst();
+
+        // extract the original result value and use it (e.g. with_result) as the final result value of our replaced sequences
+        let original_results = pos.func.dfg.detach_results(inst);
+        let original_result = original_results.get(0, &pos.func.dfg.value_lists).unwrap();
```
bnjbvr: Same suggestion about reusing replace().

cranelift-codegen/src/isa/x86/enc_tables.rs
```diff
@@ -954,3 +954,64 @@ fn convert_extractlane(
...
+
+        let value_type = pos.func.dfg.value_type(vector);
+        if value_type.lane_type().is_float() {
+            // floats are already in XMM registers and can stay there
+            match value_type {
+                F32X4 => {
+                    assert!(lane > 0 && lane <= 3);
```
bnjbvr: Is there anything in the code base that disallow lanes out of that range by throwing an error? (i'd expect so in the verifier) If not, this is something we need (in another PR if you prefer).

cranelift-codegen/meta/src/isa/x86/encodings.rs
```diff
@@ -1774,13 +1778,34 @@ pub fn define(
         }
     }
 
+    // for legalizing insertlane with floats, INSERTPS from SSE4.1
+    {
+        let instruction = x86_insertps.bind_vector_from_lane(F32, sse_vector_size);
+        let template = rec_fa_ib.nonrex().opcodes(vec![0x66, 0x0f, 0x3a, 0x21]);
+        e.enc_32_64_maybe_isap(instruction, template, Some(use_sse41_simd));
```
bnjbvr: We also need encodings when we don't have SSE4.1 here (maybe as a follow-up? Because I just realized that many other vector instructions might be in the same situation).

filetests/isa/x86/insertlane-run.clif
```diff
@@ -0,0 +1,46 @@
+test run
+set enable_simd
+
+function %test_insertlane_b8() -> b8 {
+ebb0:
+    v0 = bconst.b8 false
+    v1 = bconst.b8 true
+    v2 = splat.b8x16 v0
+    v3 = insertlane v2, 10, v1
+    v4 = extractlane v3, 10
```
bnjbvr: This test is dependent on splat and extractlane working. Can we try to reduce the dependency by only depending on vconst, and use vconst as the input and expected values, instead?

cranelift-codegen/meta/src/isa/x86/encodings.rs
abrown: Yeah, good point; I think I added this because when legalizing splat I needed to get ints/booleans into an XMM register but I realized that floats already were in the right place. Perhaps the scalar_to_vector definition should change to not specify that the higher bits are zeroed?

filetests/isa/x86/insertlane-run.clif
abrown: Well, I can't compare vconsts yet but I am working on it in a separate branch: why don't I change the initial `splat` to a `vconst` and add a TODO to replace the extractlane + comparison with a direct vector comparison?

cranelift-codegen/meta/src/isa/x86/encodings.rs
abrown: Yeah, I would consider that as a follow-up and perhaps even argue that the effort to add alternate encodings may not be worth the pay-off: how many pre-2006 x86 CPUs will be running cranelift code?

cranelift-codegen/src/lib.rs
```diff
@@ -48,6 +48,8 @@ extern crate alloc as std;
 #[cfg(feature = "std")]
 #[macro_use]
 extern crate std;
+#[macro_use]
+extern crate log;
```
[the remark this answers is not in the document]
abrown: I'm fine either way; @sunfishcode was helping me debug and this is what we did... I'll change it to the more explicit `use log::debug` unless someone tells me otherwise.

filetests/isa/x86/extractlane-run.clif
```diff
@@ -0,0 +1,31 @@
...
+    v2 = extractlane v1, 3
+    v10 = f32const 0x42.42  ; TODO this should not be necessary, v0 should be re-usable
```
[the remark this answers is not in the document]
abrown: I figured out what was wrong here and a following commit resolves this.

cranelift-codegen/meta/src/isa/x86/encodings.rs
bnjbvr: It's a good question: it would introduce more nondeterminism, which wouldn't be confusing if it's properly documented.

One question is, if we decided to do this, what would be the difference between insertlane 0 and scalar_to_vector? I think they'd do the same thing, conceptually, so we could:

- either remove scalar_to_vector, and require to use insertlane 0 instead. If people want to have the higher bits zeroed, they can generate a 0 constant first.
- or keep scalar_to_vector, and it would do the zeroing on behalf of the user.

For the sake of having a minimal IR, and if having a different instruction doesn't bring any new optimization opportunity (I can't think of any brought by scalar_to_vector with zeroing semantics), I think removing scalar_to_vector would be fine. @sunfishcode, any opinions here?

cranelift-codegen/meta/src/isa/x86/encodings.rs
```diff
@@ -1760,16 +1760,16 @@ pub fn define(
...
 
     // SIMD extractlane
-    let mut extractlane_mapping: HashMap<u64, (Vec<u8>, Option<SettingPredicateNumber>)> =
+    let mut x86_pextr_mapping: HashMap<u64, (Vec<u8>, Option<SettingPredicateNumber>)> =
         HashMap::new();
-    extractlane_mapping.insert(8, (vec![0x66, 0x0f, 0x3a, 0x14], Some(use_sse41_simd))); // PEXTRB
-    extractlane_mapping.insert(16, (vec![0x66, 0x0f, 0xc5], None)); // PEXTRW from zSSE2, SSE4.1 has a PEXTRW that can move to reg/m16 but the opcode is four bytes
-    extractlane_mapping.insert(32, (vec![0x66, 0x0f, 0x3a, 0x16], Some(use_sse41_simd))); // PEXTRD
-    extractlane_mapping.insert(64, (vec![0x66, 0x0f, 0x3a, 0x16], Some(use_sse41_simd))); // PEXTRQ, only x86_64
+    x86_pextr_mapping.insert(8, (vec![0x66, 0x0f, 0x3a, 0x14], Some(use_sse41))); // PEXTRB
```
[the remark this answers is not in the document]
abrown: Yeah, the rebase fixes this; this is an outdated commit.

cranelift-codegen/src/isa/x86/enc_tables.rs
```diff
@@ -893,3 +894,63 @@ fn expand_fcvt_to_uint_sat(
...
+    } = pos.func.dfg[inst]
+    {
+        // remove the original extractlane instruction
+        pos.remove_inst();
+
+        // extract the original result value and use it (e.g. with_result) as the final result value of our replaced sequences
+        let original_results = pos.func.dfg.detach_results(inst);
+        let original_result = original_results.get(0, &pos.func.dfg.value_lists).unwrap();
+
+        // NOTE: the following legalization assumes that the upper bits of the XMM register do not need to be zeroed during extractlane
```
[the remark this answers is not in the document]
abrown: So add something like the following to `extractlane`?

```
Extract lane ``Idx`` from ``x``. Note that the upper bits of ``a`` may or may not be zeroed depeding on the ISA but the type system should prevent using ``a`` as anything other than the extracted value.
```

filetests/isa/x86/extractlane-run.clif
abrown: Yeah, I wrote a lot of this before I had worked much on vconst parsing (bda45f4ac) so unfortunately I had to use what I had. I'll change these as I see them; thanks for pointing them out.

cranelift-codegen/meta/src/isa/x86/instructions.rs
abrown: What do you think of my suggestion in https://github.com/CraneStation/cranelift/pull/943#discussion_r321371969?

cranelift-codegen/meta/src/isa/x86/instructions.rs
abrown: But this could be used for either integers or booleans, right?

cranelift-codegen/src/isa/x86/enc_tables.rs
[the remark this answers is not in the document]
bnjbvr: Sounds good!

cranelift-codegen/meta/src/isa/x86/instructions.rs
bnjbvr: That's right; is TxN only limited to int or bools? (if so please ignore me)

cranelift-codegen/src/isa/x86/enc_tables.rs
abrown: I know that at least `wasmparser` is doing those checks here: https://github.com/yurydelendik/wasmparser.rs/blob/master/src/binary_reader.rs#L1256-L1265.

cranelift-codegen/meta/src/isa/x86/instructions.rs
abrown: I think it's ints/floats/bools; you're saying I should create an `IBxN` just for bools/ints then since that's all this instruction should handle?

cranelift-codegen/src/isa/x86/enc_tables.rs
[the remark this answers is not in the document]
abrown: Ok, let me add that.

cranelift-codegen/meta/src/isa/x86/instructions.rs
bnjbvr: Yes, it would help catch more type errors.

cranelift-codegen/meta/src/isa/x86/instructions.rs
abrown: I think we could do this but I'm worried that the semantics might become confusing:

 - `copy` for f32/f64 is encoded as `MOVAPS` which moves all 128 bits from one register to another
 - `x86_movsd` is encoded as `MOVSD` which only moves the bottom 64 bits from one register to another, leaving the higher bits unchanged as long as the source is a register and not a memory location; this effectively inserts an f64 into lane 0 of a vector, which is how I'm using it to legalize `insertlane`

We could replace the encoding of `copy.f64` to use `MOVSD` (and `copy.f32` could use `MOVSS`) instead of `MOVAPS` but this leaves the upper bits unchanged, something which is not clearly reflected in the `copy` instruction documentation. This theoretically is fine as long as we maintain the assumption that the type system will protect us from uncleared high bits. One additional consideration: `MOVSD/MOVSS` are 3-byte opcodes whereas `MOVAPS` is a 2-byte opcode. Thoughts?

bjorn3: > but this leaves the upper bits unchanged, something which is not clearly reflected in the copy instruction documentation. This theoretically is fine as long as we maintain the assumption that the type system will protect us from uncleared high bits.

For 8 and 16 bit values the higher bits can be anything too, because `ireduce` is a no-op.

cranelift-codegen/meta/src/isa/x86/encodings.rs
sunfishcode: Yeah, I agree that scalar_to_vector zeroing the high lanes has overlap with insertlane.

What if make `scalar_to_vector` leave nondeterministic values in the high lanes, but document it as a low-level legalization instruction, and that frontends should generally prefer `insertlane 0`? Nondeterminism is worth avoiding when we can, but there are a variety of situations where it's useful to be able to get a scalar value into a vector register, where one knows that subsequent operations won't care about the high lanes of the vector, and the extra zeroing would be needless overhead.

abrown: Ok, I changed the documentation of `scalar_to_vector`, to look like:
```
    Scalar To Vector -- move a value out of a scalar register and into a vector register; the 
    scalar will be moved to the lowest-order bits of the vector register. Note that this 
    instruction is intended as a low-level legalization instruction and frontends should prefer 
    insertlane; on certain architectures, scalar_to_vector may zero the highest-order bits for some
    types (e.g. integers) but not for others (e.g. floats).
```
See https://github.com/CraneStation/cranelift/pull/943/commits/b2e9e09761182781a2ec0c2b50a23aa463ef22e6#diff-7c1b843a5d2e8c61f75e0d350b3f3914R2774-R2778.

cranelift-codegen/meta/src/isa/x86/instructions.rs
abrown: Sounds like @bjorn3 is ok with changing `copy.f32/f64` to `MOVSD/MOVSS`; @bnjbvr?

abrown: Shoot... I implemented this change and then realized that cranelift's `copy` does not accept a destination operand so I can't control which XMM register to move the lower bits to. I'm leaving `x86_movsd` in for now.

I had another thought that might resolve the underlying issue of adding all these new x86-specific instructions. If I can get https://github.com/abrown/cranelift/pull/3 working, we then have the ability to express the following without any new instructions:

```
e.enc...(insertlane.bind(F64).bind(lane0), [recipe and opcodes for MOVSD]);
e.enc...(insertlane.bind(F64).bind(lane1), [recipe and opcodes for MOVLHPS]);
```

I guess that's more motivation to get that PR working. Should we leave `x86_movsd` in for now and refactor later?

cranelift-codegen/meta/src/isa/x86/encodings.rs
bnjbvr: Chances are there'll be some of them, if you count in the Firefox user base. I'll try to get data on this. At the moment, support for SSE4.1 support can be enabled or disabled, so if we wanted to make it a strong requirement, we'd need to make this choice more clearer up front (by removing the setting, among other things).

cranelift-codegen/meta/src/isa/x86/instructions.rs
bnjbvr: Sounds good to me, yeah. Thanks for trying!

cranelift-codegen/src/isa/x86/enc_tables.rs
bnjbvr: It'd be nice to open a follow-up to make sure that we actually throw an error in the verifier's code, if that happens; we may have other producers than cranelift-wasm.

filetests/isa/x86/insertlane-run.clif
bnjbvr: Yes, that's great, thanks.

cranelift-codegen/meta/src/shared/instructions.rs
```diff
@@ -1537,7 +1537,9 @@ pub(crate) fn define(
         Extract lane ``Idx`` from ``x``.
 
         The lane index, ``Idx``, is an immediate value, not an SSA value. It
-        must indicate a valid lane index for the type of ``x``.
+        must indicate a valid lane index for the type of ``x``. Note that the upper bits of ``a``
+        may or may not be zeroed depeding on the ISA but the type system should prevent using
```
bnjbvr: nit: depending

cranelift-codegen/meta/src/isa/x86/encodings.rs
abrown: Ok, recorded in https://github.com/CraneStation/cranelift/issues/1015.

cranelift-codegen/src/isa/x86/enc_tables.rs
abrown: Opened https://github.com/CraneStation/cranelift/issues/1016.

cranelift-codegen/meta/src/isa/x86/encodings.rs
```diff
@@ -1810,23 +1810,38 @@ pub(crate) fn define(
...
+                    .bind_vector_from_lane(to_type, sse_vector_size)
+                    .bind_vector_from_lane(from_type, sse_vector_size),
+            );
+        }
+    }
+
+    // SIMD raw bitcast floats to vector (and back); assumes that floats are already stored in an XMM register
+    for float_type in &[F32, F64] {
+        for lane_type in ValueType::all_lane_types().filter(allowed_simd_type) {
+            e.enc_32_64_rec(
```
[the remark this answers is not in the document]
abrown: Good find; I fixed and a few other test issues by using `git rebase -i --exec ./test-all.sh master`.

PR #944, opened 2019-08-27
Continuous fuzzing

bookmoons:
Proposing to enable continuous fuzzing of the good fuzzing targets. I see this was mentioned in #306 as a possible future improvement.

This patch runs both existing targets on [Fuzzit](https://fuzzit.dev/). There's [a build](https://travis-ci.org/bookmoons/cranelift/builds/577307862) under my Travis account I'm expecting to succeed.

The PR build will fail due to missing an API key. Setup is like this:
* In Fuzzit create targets `cranelift-translate-module` `cranelift-reader-parse`
* In Fuzzit settings grab an API key. In repo settings in [Travis](https://travis-ci.org/) paste it to envvar `FUZZIT_API_KEY`.

Thank you for considering.

### DIFF
--- a/.azure-pipelines.yml
+++ b/.azure-pipelines.yml
@@ -97,6 +97,30 @@ jobs:
         RUST_BACKTRACE: 1
       condition: and(succeeded(), eq(variables['toolchain'], 'nightly'))
 
+- job: Fuzz_regression
+  displayName: Fuzz regression
+  pool:
+    vmImage: "ubuntu-16.04"
+  variables:
+    toolchain: nightly
+  steps:
+    - template: ci/azure-install-rust.yml
+    - bash: cargo install cargo-fuzz
+    - bash: ci/fuzzit.sh local-regression
+
+- job: Fuzz
+  condition: ne(variables['Build.Reason'], 'PullRequest')
+  pool:
+    vmImage: "ubuntu-16.04"
+  variables:
+    toolchain: nightly
+  steps:
+    - template: ci/azure-install-rust.yml
+    - bash: cargo install cargo-fuzz
+    - bash: ci/fuzzit.sh fuzzing
+      env:
+        FUZZIT_API_KEY: $(FUZZIT_API_KEY)
+
 - job: Build
   strategy:
     matrix:
--- a/README.md
+++ b/README.md
@@ -8,6 +8,7 @@ into executable machine code.
 
 [![Documentation Status](https://readthedocs.org/projects/cranelift/badge/?version=latest)](https://cranelift.readthedocs.io/en/latest/?badge=latest)
 [![Travis Status](https://travis-ci.org/CraneStation/cranelift.svg?branch=master)](https://travis-ci.org/CraneStation/cranelift)
+[![Fuzzit Status](https://app.fuzzit.dev/badge?org_id=CraneStation)](https://app.fuzzit.dev/orgs/CraneStation/dashboard)
 [![Gitter chat](https://badges.gitter.im/CraneStation/CraneStation.svg)](https://gitter.im/CraneStation/Lobby)
 ![Minimum rustc 1.37](https://img.shields.io/badge/rustc-1.37+-green.svg)
 
--- a/ci/fuzzit.sh
+++ b/ci/fuzzit.sh
@@ -1,0 +1,38 @@
+#!/bin/bash
+set -xe
+
+# Validate arguments
+if [ "$#" -ne 1 ]; then
+    cat << EOF
+Usage: $0 <type>
+
+Types are:
+local-regression - Run corpus and past crashes locally to catch regressions.
+fuzzing - Submit for long run fuzzing on Fuzzit.
+EOF
+    exit 1
+fi
+
+# Configure
+set -xe
+NAME=cranelift
+TYPE=$1
+FUZZIT_VERSION=2.4.46
+
+# Setup
+if [[ ! -f fuzzit || ! `./fuzzit --version` =~ $FUZZIT_VERSION$ ]]; then
+    wget -q -O fuzzit https://github.com/fuzzitdev/fuzzit/releases/download/v$FUZZIT_VERSION/fuzzit_Linux_x86_64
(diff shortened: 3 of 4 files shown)


bjorn3:
> INFO: -max_len is not provided; libFuzzer will not generate inputs larger than 4096 bytes
>
> INFO: A corpus is not provided, starting from an empty corpus
>
> #2	INITED cov: 1482 ft: 1332 corp: 1/1b lim: 4 exec/s: 0 rss: 35Mb
>
> #2	DONE   cov: 1482 ft: 1332 corp: 1/1b lim: 4 exec/s: 0 rss: 35Mb

Am I reading this correctly as: tried one input smaller than 4096 bytes?

bjorn3:
> The binary gets submitted to Fuzzit and does a long term fuzzing run.

Is there anywhere I can see the fuzzing job?

bjorn3:
You have to login :(

bjorn3:
I think the last one just ran out of memory. I reduced it to:

```
function %a(){
ebb477777777:
}
```

which will attempt to allocate 477777777 ebbs. Maybe we should add a reader config to deny such pathological inputs.

bjorn3:
```
$ cargo +nightly fuzz run fuzz_translate_module artifact
[...]
INFO: Seed: 1619635779
INFO: Loaded 1 modules   (86922 guards): 86922 [0x560dea246488, 0x560dea29b2b0), 
fuzz/target/x86_64-unknown-linux-gnu/debug/fuzz_translate_module: Running 1 inputs 1 time(s) each.
Running: artifact
Executed artifact in 22 ms
***
*** NOTE: fuzzing was not performed, you have only
***       executed the target code on a fixed set of inputs.
***
```

I don't seem to be able to trigger the crash for `translate-module`.

bnjbvr:
Thanks! More fuzzing is always great. We've now moved over to Azure pipelines, is there a chance you could update your PR, please?

bookmoons:
Will do! Thanks for looking at it. Just did that for wasmer so I have an idea what it takes.

bookmoons:
This is updated. I see fuzzing runs happening in Fuzzit. There's a successful test build:
https://dev.azure.com/bookmoons/cranelift/_build/results?buildId=29

Fuzzit have kindly set up an org that I've preconfigured with targets. If you'd like to [sign in](https://fuzzit.dev/) to make an account I can get you added to the org.

Setup through Azure is now like this:
* In Fuzzit settings grab an API key.
* In Azure Pipelines, Edit the pipeline, open Variables, and paste the key to a FUZZIT_API_KEY variable. Make it secret to prevent exposure.

bookmoons:
I had to update the reader_parse target. d7550024a41d47a8222563a12825e210ef598bce modifies the targeted function. I updated to use the new signature.

bookmoons:
Made these changes. I kept the target update separate. Put the type option docs into the usage message.

tschneidereit:
This looks great to me, thank you so much, @bookmoons!

I'm traveling right now, and won't be able to thoroughly investigate the setup fully, unfortunately. @alexcrichton, could you take a look and see if this setup seems good? Perhaps also compared to what kind of fuzzing is done for Rust?

bookmoons:
I'm seeing a little bug in the fuzz regression stage. Just tested and it's showing on this setup. Probably worth holding off merging until it's resolved.

Reported it in https://github.com/fuzzitdev/fuzzit/issues/38

bookmoons:
Upgrading the Fuzzit client resolved that. This is ready for review again. There's a successful test build:
https://dev.azure.com/bookmoons/cranelift/_build/results?buildId=36

The latest seems to require an API key, so the PR build will fail until that's in.

alexcrichton:
While I haven't done much continuous fuzzing myself in Rust yet, this seems good to me!

Is there a way that API keys aren't required on PRs? Or is there perhaps a readonly key available for PRs and the read/write key could only be used for branches?

jfoote:
👋 I am glad to see this! @bookmoons do you happen to know where reports will be sent when the fuzzer detects an issue? I took a quick look at the fuzzit.dev docs but it was not obvious to me.

bookmoons:
Thanks a lot guys.

I think the reports go out to everyone on the account. There's a list of admins for the org, so if anyone who should be administering wants to sign up I can get you all added.

> Is there a way that API keys aren't required on PRs? Or is there perhaps a readonly key available for PRs and the read/write key could only be used for branches?

I'm going ask Fuzzit about this. The recommendation was that you make the corpus public so it can run in PRs without a key. This seems to break that.

I see in settings you can make a readonly key, so maybe that's the right track.

bookmoons:
> I think the reports go out to everyone on the account. There's a list of admins for the org, so if anyone who should be administering wants to sign up I can get you all added.

When they get sent they go out by email, to the address on the GitHub account. Then you can pop in and download the crashing input.

bookmoons:
The latest Fuzzit client brings back PR fuzzing. Made the upgrade to get that working again.

tschneidereit:
Ah, I see. @bookmoons, can you add me to the org on Fuzzit? I'll revert the PR for now, because it [causes the build to fail](https://dev.azure.com/CraneStation/Cranelift/_build/results?buildId=865&view=logs&j=dd99581b-5d42-545b-15d8-31c1cc56df2c&t=451589ee-97bd-5276-593b-f468b34c6a93&l=372).

bookmoons:
Got you added!

https://app.fuzzit.dev/orgs/cranelift/settings

bookmoons:
If it helps I could open a new PR.

pchickey:
@bookmoons can you add me to the org as well? Thanks!

bookmoons:
Will do @pchickey.

Fuzzit doesn't seem to be finding an account. If you'd like to [sign in](https://fuzzit.dev/) to create one I can get you added.

tschneidereit:
Thanks, @bookmoons! I just updated the API key.

It does indeed seem like we'll have to have a new PR, so could you open one?

bookmoons:
Opened #1042 to replace this.

Issue #935, opened 2019-08-23
clif-file should support cfg macros.


nbp:
In #844 , the clif files output changes whether we have or not the feature flag for enabling basic-blocks. Ideally these test case should be merged with the changes, but we should not break the output when the feature flags are disabled.

The idea would be to change clif file syntax to include a `#[cfg(…)]` syntax to the directives of clif files.

bjorn3:
#945 implemented per test file cfg.

PR #945, opened 2019-08-27
Add feature flags to test files.

nbp:
Cranelift can be compiled with feature flags which can change its output. To
accomodate changes of output related to feature flags, test file can now include
`feature "..."` and `feature ! "..."` directives in the preamble of the test
file.

The test runner would skip the test if the flag does not match the expectation
of the test case.

This PR should fix the issue #935 .

### DIFF
--- a/cranelift-filetests/src/runone.rs
+++ b/cranelift-filetests/src/runone.rs
@@ -8,8 +8,7 @@ use cranelift_codegen::print_errors::pretty_verifier_error;
 use cranelift_codegen::settings::Flags;
 use cranelift_codegen::timing;
 use cranelift_codegen::verify_function;
-use cranelift_reader::IsaSpec;
-use cranelift_reader::{parse_test, ParseOptions};
+use cranelift_reader::{parse_test, Feature, IsaSpec, ParseOptions};
 use log::info;
 use std::borrow::Cow;
 use std::fs;
@@ -52,6 +51,31 @@ pub fn run(path: &Path, passes: Option<&[String]>, target: Option<&str>) -> Test
             return Err(e.to_string());
         }
     };
+
+    for feature in testfile.features.iter() {
+        let (flag, test_expect) = match feature {
+            Feature::With(name) => (name, true),
+            Feature::Without(name) => (name, false),
+        };
+        let cranelift_has = match flag {
+            // Add any cranelift feature flag here, and make sure that it is forwarded to the
+            // cranelift-filetest crate in the top-level Cargo.toml.
+            &"basic-blocks" => cfg!(feature = "basic-blocks"),
+            _ => {
+                return Err(format!(
+                    r#"{:?}: Unknown feature flag named "{}""#,
+                    path, flag
+                ))
+            }
+        };
+        if cranelift_has != test_expect {
+            println!(
+                r#"skipping test {:?}: non-matching feature flag "{}""#,
+                path, flag
+            );
+            return Ok(started.elapsed());
+        }
+    }
 
     if testfile.functions.is_empty() {
         return Err("no functions found".to_string());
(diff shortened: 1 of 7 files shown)


cranelift-filetests/src/runone.rs
```diff
@@ -48,6 +48,31 @@ pub fn run(path: &Path, passes: Option<&[String]>, target: Option<&str>) -> Test
...
     };
 
+    for feature in testfile.features.iter() {
+        let (flag, test_expect) = match feature {
+            Feature::Has(name) => (name, true),
+            Feature::HasNo(name) => (name, false),
+        };
+        let cranelift_has = match flag {
+            // Add any cranelift feature flag here, and make sure that it is forwarded to the
+            // cranelift-filetest crate in the top-level Crago.toml.
```
bnjbvr: nit: Cargo.toml

cranelift-filetests/src/runone.rs
```diff
@@ -48,6 +48,31 @@ pub fn run(path: &Path, passes: Option<&[String]>, target: Option<&str>) -> Test
...
+            &"basic-blocks" => cfg!(feature = "basic-blocks"),
+            _ => {
+                return Err(format!(
+                    r#"{:?}: Unexpected feature flag named "{}""#,
```
bnjbvr: Light suggestion, maybe "unknown" instead of "unexpected"?

cranelift-reader/src/testfile.rs
```diff
@@ -55,3 +57,17 @@ pub struct Comment<'a> {
...
+/// This represent the expectation of the test case. Before running any of the
+/// functions of the test file, the feature set should be compared with the
+/// feature set used to compile Cranelift. If there is any differences, then the
+/// test file should be skipped.
+#[derive(PartialEq, Eq, Debug)]
+pub enum Feature<'a> {
+    /// `feature "..."` lines
+    Has(&'a str),
+    /// `feature ! "..."` lines.
+    HasNo(&'a str),
```
bnjbvr: Light suggestion to rename these two "With" and "Without", maybe? Or alternatively, "Want" / "DontWant"?

cranelift-reader/src/testfile.rs
```diff
@@ -55,3 +57,17 @@ pub struct Comment<'a> {
     /// Text of the comment, including the leading `;`.
     pub text: &'a str,
 }
+
+/// A cranelift feature in a test file preamble.
+///
+/// This represent the expectation of the test case. Before running any of the
```
bnjbvr: nit: represents

cranelift-reader/src/lexer.rs
```diff
@@ -399,6 +401,27 @@ impl<'a> Lexer<'a> {
...
+        assert_eq!(self.lookahead, Some('"'));
+
+        while let Some(c) = self.next_ch() {
+            if c == '"' || c == '\n' {
+                break;
+            }
+        }
+
+        let end = self.pos;
+        if self.lookahead != Some('"') {
```
bnjbvr: If we exited the previous loop because c = `\n`, this will yield an error, I think, which is undesired since this says this can handle multi-line strings. If so, we should remove the `|| c == '\n'` part of the if check in the while above. Can you add a test scanning a string containing a new line, please?

nbp:
@bnjbvr : 
> I wonder if the string after "feature" could just be an identifier instead

I wondered the same, but our feature flag does have a dash symbol which is not part of the allowed symbols in characters. Thus I went for a quoted string, as flag are always manipulated as quoted string.

cranelift-filetests/src/runone.rs
nbp: I agree, replacing by "Unknown" is better.

cranelift-reader/src/lexer.rs
nbp: Done.

cranelift-reader/src/testfile.rs
nbp: I renamed these to "With" and "Without". Thanks.

PR #949, opened 2019-08-27
Remove calling convention on FunctionRunner test

abrown:
For this test to run on Windows, the explicit calling convention must be removed so that the parser is free to use the host's calling convention.

### DIFF
--- a/cranelift-filetests/src/function_runner.rs
+++ b/cranelift-filetests/src/function_runner.rs
@@ -99,7 +99,9 @@ mod test {
     #[test]
     fn nop() {
         let code = String::from(
-            "function %test() -> b8 system_v {
+            "
+            test run
+            function %test() -> b8 {
             ebb0:
                 nop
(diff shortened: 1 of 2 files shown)


tschneidereit:
@abrown that unfortunately [still didn't fix it](https://dev.azure.com/tschneidereit/Cranelift/_build/results?buildId=102) :(

bnjbvr:
Yeah, this test is duplicated in clif-util, src/run.rs.

@tschneidereit Can you try with the additional commit in https://github.com/bnjbvr/cranelift/tree/wintest please?

bnjbvr:
> Do you want to add your commit for the other test to this PR so we can keep them together?

I found a way to add my commit on top of yours, using Github's online editor. [Test run](https://dev.azure.com/tschneidereit/Cranelift/_build/results?buildId=104) shows that Windows test pass, with these two commits!

PR #950, opened 2019-08-27
WIP: Re-fold basic-block split edges in regalloc

sstangl:
This patch unregresses the extra jumps in the codegen introduced by basic-block branch splitting.

This is a simpler version of the fold-redundant-branches code that is much easier to get correct. The issue with the previous pass was that some ebbs were functioning as phi nodes in the CFG, and those blocks were not actually redundant. This patch only considers blocks that were added specifically by regalloc (in nbp's branch-splitting patch), therefore none of the blocks are effective phis, and they're formed predictably.

This is a WIP patch. If this looks good, I'll muddle through data structures to clean it up. In particular, deleting an ebb means that we need to update the Liveness struct -- currently, it just recomputes the entire data structure, but it's possible that I could update the bforest to support deletion. Although with trees, that's fairly error-prone.

At this point, just soliciting comments.

### DIFF
--- a/cranelift-codegen/src/regalloc/branch_folding.rs
+++ b/cranelift-codegen/src/regalloc/branch_folding.rs
@@ -1,0 +1,131 @@
+//! Folds redundant branches created by branch_splitting.
+//!
+//! The branch_splitting phase attempts to split possible critical edges so
+//! that the register allocator always has a unique location per conditional branch
+//! in which to insert register moves.
+//!
+//! If the register allocator decides to not use any of those blocks, they can
+//! be removed from the flowgraph.
+#![cfg(feature = "basic-blocks")]
+
+use std::vec::Vec;
+
+use crate::dominator_tree::DominatorTree;
+use crate::flowgraph::{BasicBlock, ControlFlowGraph};
+use crate::ir::{Ebb, Function, Inst, Opcode, ValueList};
+
+/// Folds an ebb if it is redundant in the CFG.
+/// Returns whether folding was performed (which invalidates the domtree).
+fn try_fold_redundant_ebb(
+    func: &mut Function,
+    cfg: &mut ControlFlowGraph,
+    ebb: Ebb
+) -> bool {
+    // Get branch information for the middle block.
+    // This block was created by regalloc, and must end with a jump or fallthrough.
+    let ebb_jump: Inst = func.layout.last_inst(ebb).expect("terminator");
+    let ebb_dest: Ebb = func.dfg[ebb_jump].branch_destination().expect("single dest");
+
+    // If there's more than just a jump in the block, the block isn't redundant.
+    if func.layout.prev_inst(ebb_jump).is_some() {
+        return false;
+    }
+
+    // Get branch information for the parent block.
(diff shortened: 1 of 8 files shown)


julian-seward1:
@sstangl: [..] it's possible that I could update the bforest to support deletion [..]

Just to note .. it seems at least moderately likely that I'll have to get my head around the bforest machinery as part of my (so far ineffective) attempts to reduce CL's run time.  If that happens, I could maybe hack up the changes you need.

bnjbvr:
Just as an aside, could you please git rebase on top of master? This PR includes many unrelated commits, making it hard to see what's going on.

nbp:
I am going to merge the branch-splitting branch soon. Otherwise, you could target the branch-splitting branch on my repository to reduce the diff, my branch recently merged from master and should not have a larger diff.

cranelift-codegen/src/regalloc/branch_splitting.rs
```diff
@@ -14,31 +14,33 @@ use crate::ir::{Ebb, Function, Inst, InstBuilder, InstructionData, Opcode, Value
...
+/// Split possible critical edges. Returns a list of added blocks.
 pub fn run(
-    isa: &TargetIsa,
+    isa: &dyn TargetIsa,
     func: &mut Function,
     cfg: &mut ControlFlowGraph,
     domtree: &mut DominatorTree,
     topo: &mut TopoOrder,
-) {
+) -> Vec<Ebb> {
```
nbp: Looking at the rest of the Cranelift phases, usually it seems that these values are provided at a `&mut` as argument of the run command instead of returned from the `run` function. Then the `clear` function of the register allocator is used to clean the content of the vector, to be reused without reallocation for the next compiled function.

cranelift-codegen/src/regalloc/context.rs
```diff
@@ -199,6 +200,15 @@ impl Context {
         self.coloring
             .run(isa, func, domtree, &mut self.liveness, &mut self.tracker);
 
+        // Pass: undo branch splitting where the inserted blocks were not
+        // used by the register allocator.
+        #[cfg(feature = "basic-blocks")]
+        branch_folding::run(func, cfg, domtree, new_blocks);
```
nbp: I still think this belongs to the branch relaxation part of the code. Doing so as part of the branch relaxation would avoid recomputing the liveness information.

cranelift-codegen/src/regalloc/branch_folding.rs
```diff
@@ -0,0 +1,131 @@
...
+        .constraints()
+        .num_fixed_value_arguments();
+    let (_, ebb_jump_params) = func.dfg[ebb_jump]
+        .arguments(&func.dfg.value_lists)
+        .split_at(num_fixed);
+
+    // Pass the ebb_jump_params in the parent_jump_params.
+    // No rewriting is needed, because the edge ebb is parameterless.
+    debug_assert!(_parent_jump_params.len() == 0);
+    debug_assert!(func.dfg.num_ebb_params(ebb) == 0);
```
nbp: From my understanding this is the only thing to be added to the version which is currently merged. Just excluding any blocks with parameters would avoid the issue we have with folding empty branches without even looking at the branch splitting case.

nbp:
Close the pull-request as it is replaced by #972.

PR #957, opened 2019-08-30
Legalize i64.const by breaking it into two i32.const

bnjbvr:
So my current thinking about isplit/iconcat is that both should not appear post-legalization, and every isplit must have an iconcat input, so the iconcat results are used instead. This seems to make sense if isplit/iconcat are just type system artifacts.

So trying a very simple program on x86 32-bits, that is generating an i64 constant, I implemented this legalization. This could be done in the meta language if there was literal support in the legalize patterns (cc @abrown).

Now it's funny because it interacts with f64const generation on 32 bits (so the test will fail, hence the WIP status of this PR). f64const are generated by reproducing the bit pattern in an i64, then bitcasting this to a f64 register. This is wrong on 32 bits platforms, where instead we should probably legalize it into two GPRs and then bitcast this to a f64. Or implement real constant pools for f32/f64, which would be even better.

### DIFF
--- a/cranelift-codegen/src/legalizer/mod.rs
+++ b/cranelift-codegen/src/legalizer/mod.rs
@@ -16,7 +16,7 @@
 use crate::bitset::BitSet;
 use crate::cursor::{Cursor, FuncCursor};
 use crate::flowgraph::ControlFlowGraph;
-use crate::ir::types::I32;
+use crate::ir::types::{I32, I64};
 use crate::ir::{self, InstBuilder, MemFlags};
 use crate::isa::TargetIsa;
 use crate::predicates;
@@ -632,4 +632,36 @@ fn narrow_store(
         offset.try_add_i64(8).expect("store offset overflow"),
     );
     pos.remove_inst();
+}
+
+/// Expands an illegal iconst value by splitting it into two.
+fn narrow_iconst(
+    inst: ir::Inst,
+    func: &mut ir::Function,
+    _cfg: &mut ControlFlowGraph,
+    isa: &dyn TargetIsa,
+) {
+    let imm: i64 = if let ir::InstructionData::UnaryImm {
+        opcode: ir::Opcode::Iconst,
+        imm,
+    } = &func.dfg[inst]
+    {
+        (*imm).into()
+    } else {
+        panic!("unexpected instruction in narrow_iconst");
+    };
+
+    let mut pos = FuncCursor::new(func).at_inst(inst);
+    pos.use_srcloc(inst);
+
+    let ty = pos.func.dfg.ctrl_typevar(inst);
+    if isa.pointer_bits() == 32 && ty == I64 {
+        let low = pos.ins().iconst(I32, imm & 0xffffffff);
+        let high = pos.ins().iconst(I32, imm >> 32);
+        // The instruction has as many results as iconcat, so no need to replace them.
+        pos.func.dfg.replace(inst).iconcat(low, high);
+        return;
+    }
+
+    unimplemented!("missing encoding or legalization for iconst.{:?}", ty);
 }
--- a/filetests/simple_preopt/simplify32.clif
+++ b/filetests/simple_preopt/simplify32.clif
@@ -1,0 +1,61 @@
+test simple_preopt
+target i686
+
+;; 32-bits platforms.
+
+function %iadd_imm(i32) -> i32 {
+ebb0(v0: i32):
+    v1 = iconst.i32 2
+    v2 = iadd v0, v1
+    return v2
+}
+; sameln: function %iadd_imm
+; nextln: ebb0(v0: i32):
+; nextln:     v1 = iconst.i32 2
+; nextln:     v2 = iadd_imm v0, 2
+; nextln:     return v2
+; nextln: }
+
+function %isub_imm(i32) -> i32 {
+ebb0(v0: i32):
+    v1 = iconst.i32 2
+    v2 = isub v0, v1
+    return v2
+}
+; sameln: function %isub_imm
+; nextln: ebb0(v0: i32):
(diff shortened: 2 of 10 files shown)


cranelift-codegen/src/legalizer/mod.rs
```diff
@@ -498,3 +498,33 @@ fn expand_stack_store(
...
+    {
+        (*imm).into()
+    } else {
+        panic!("unexpected instruction in expand_iconst");
+    };
+
+    let mut pos = FuncCursor::new(func).at_inst(inst);
+    pos.use_srcloc(inst);
+
+    if pos.func.dfg.ctrl_typevar(inst) == I64 {
```
bjorn3: Can you please remove this if, use `pos.func.dfg.ctrl_typevar(inst).half_width().unwrap()` instead of `I32` below and adapt the splitting? That would automatically make `iconst.i128 some_imm_fitting_in_64bits` possible on my `i128` branch. Removing the if would not cause `iconst.i32` to get split unless there is no encoding for it.

cranelift-codegen/src/legalizer/mod.rs
```diff
@@ -498,3 +498,33 @@ fn expand_stack_store(
     mflags.set_aligned();
     pos.func.dfg.replace(inst).store(mflags, val, addr, 0);
 }
+
+fn expand_iconst(
```
bjorn3:
```suggestion
fn narrow_iconst(
```

bjorn3:
> Or implement real constant pools for f32/f64, which would be even better.

LLVM does that for both f32 and f64.

```rust
black_box(0.1f64);
```

```asm
[...]
movsd	.LCPI10_0(%rip), %xmm0
callq	core::hint::black_box
[...]
```

cranelift-codegen/src/legalizer/mod.rs
bnjbvr: Nice idea. It would require additional work to compute the right constants for the shift and mask according to the ctrl type, but that's easy.

bnjbvr: Thinking a bit more about this, while the idea of generalizing this code is tempting, there are 2 things that come to mind:

- the WIP PR #795 doesn't introduce 128-bits immediates (or I missed something), so the instruction format will be different and a different legalization will have to be used.
- having something very general wouldn't bring much, since as far as I know there are no plans to support 8 or 16-bits platforms. So adding one else if when we have i128 support should show simple enough for our needs.

bnjbvr:
Remaining to do: don't simplify operations into immediate operations if the ctrl value type is bigger than the native pointer type.

cranelift-codegen/src/legalizer/mod.rs
bjorn3: > the WIP PR #795 doesn't introduce 128-bits immediates (or I missed something)

correct

> having something very general wouldn't bring much, since as far as I know there are no plans to support 8 or 16-bits platforms. So adding one else if when we have i128 support should show simple enough for our needs.

+1

cranelift-codegen/src/legalizer/mod.rs
```diff
@@ -633,3 +633,34 @@ fn narrow_store(
...
+    _isa: &dyn TargetIsa,
+) {
+    let imm: i64 = if let ir::InstructionData::UnaryImm {
+        opcode: ir::Opcode::Iconst,
+        imm,
+    } = &func.dfg[inst]
+    {
+        (*imm).into()
+    } else {
+        panic!("unexpected instruction in expand_iconst");
```
abrown: `unexpected instruction in narrow_iconst`?

cranelift-codegen/src/legalizer/mod.rs
```diff
@@ -633,3 +633,34 @@ fn narrow_store(
...
+    if pos.func.dfg.ctrl_typevar(inst) == I64 {
+        let low = pos.ins().iconst(I32, imm & 0xffffffff);
+        let high = pos.ins().iconst(I32, imm >> 32);
+        // The instruction has as many results as iconcat, so no need to replace them.
+        pos.func.dfg.replace(inst).iconcat(low, high);
+        return;
+    }
+
+    unimplemented!();
+}
```
abrown: Maybe I'm missing something but I don't see where `narrow_iconst` has a predicate limiting it to 32-bit architectures; on 64-bit architectures I would expect this legalization not to be run because we can effectively load 64-bit integers directly with https://github.com/CraneStation/cranelift/blob/0fd81b590af06835c6faa67a212f4c84b21028d8/cranelift-codegen/meta/src/isa/x86/encodings.rs#L729.

filetests/simple_preopt/simplify32.clif
```diff
@@ -0,0 +1,61 @@
...
+;; support.
+function %iadd_imm(i64) -> i64 {
+ebb0(v0: i64):
+    v1 = iconst.i64 2
+    v2 = iadd v0, v1
+    return v2
+}
+; sameln: function %iadd_imm
+; nextln: ebb0(v0: i64):
+; nextln:     v1 = iconst.i64 2
```
abrown: Am I correct in thinking that this hasn't been legalized into two `iconst.i32` because legalization has not happened yet?

bnjbvr: Yes, you're right; simple_preopt doesn't run legalization.

cranelift-codegen/src/legalizer/mod.rs
bnjbvr: Good point; this doesn't run on x86 64-bits, but it shouldn't probably either run for future 64-bits architectures like aarch64. Will add an ISA check in this function.

PR #960, opened 2019-08-30
Allow pinning a certain register in Cranelift, and use it as the heap base

bnjbvr:
Based on #959. So this is very experimental, probably a bit sketchy. This allows pinning one register (r15 on x86 64 bits) so it's entirely in the control of the embedder. In particular, it tries to exclude it from register allocation. Assertions in this code plus all my tests passing make me confident it isn't horribly wrong, but it doesn't prove this is incredibly good either. (I'm a bit worried about local diversions in particular.) Also, this is not enough tested at the moment, I'll try to add more test cases and fuzz this a bit (?).

Second commit adds a second setting that enables use of the pinned register as the heap base when legalizing heap_addr instructions. If the embedder does the Right Thing, and correctly sets the heap base with set_pinned_reg, this is enough to avoid reloading the heap base before every single memory access, and is a nice win in Spidermonkey (up to 10% in some of our benchmarks).

### DIFF
--- a/cranelift-codegen/meta/src/cdsl/regs.rs
+++ b/cranelift-codegen/meta/src/cdsl/regs.rs
@@ -11,6 +11,7 @@ pub struct RegBank {
     pub names: Vec<&'static str>,
     pub prefix: &'static str,
     pub pressure_tracking: bool,
+    pub pinned_reg: Option<u16>,
     pub toprcs: Vec<RegClassIndex>,
     pub classes: Vec<RegClassIndex>,
 }
@@ -23,6 +24,7 @@ impl RegBank {
         names: Vec<&'static str>,
         prefix: &'static str,
         pressure_tracking: bool,
+        pinned_reg: Option<u16>,
     ) -> Self {
         RegBank {
             name,
@@ -31,6 +33,7 @@ impl RegBank {
             names,
             prefix,
             pressure_tracking,
+            pinned_reg,
             toprcs: Vec::new(),
             classes: Vec::new(),
         }
@@ -183,6 +186,7 @@ pub struct RegBankBuilder {
     pub names: Vec<&'static str>,
     pub prefix: &'static str,
     pub pressure_tracking: Option<bool>,
+    pub pinned_reg: Option<u16>,
 }
 
 impl RegBankBuilder {
@@ -193,6 +197,7 @@ impl RegBankBuilder {
             names: vec![],
             prefix,
             pressure_tracking: None,
+            pinned_reg: None,
         }
     }
     pub fn units(mut self, units: u8) -> Self {
@@ -205,6 +210,11 @@ impl RegBankBuilder {
     }
     pub fn track_pressure(mut self, track: bool) -> Self {
         self.pressure_tracking = Some(track);
+        self
+    }
+    pub fn pinned_reg(mut self, unit: u16) -> Self {
+        assert!(unit < (self.units as u16));
+        self.pinned_reg = Some(unit);
         self
     }
 }
@@ -246,6 +256,7 @@ impl IsaRegsBuilder {
             builder
                 .pressure_tracking
                 .expect("Pressure tracking must be explicitly set"),
+            builder.pinned_reg,
         ))
     }
 
--- a/cranelift-codegen/meta/src/shared/instructions.rs
+++ b/cranelift-codegen/meta/src/shared/instructions.rs
@@ -964,6 +964,34 @@ pub(crate) fn define(
         .operands_out(vec![addr]),
     );
 
+    // Note this instruction is marked as having other side-effects, so GVN won't try to hoist it,
+    // which would result in it being subject to spilling. While not hoisting would generally hurt
+    // performance, since a computed value used many times may need to be regenerated before each
+    // use, it is not the case here: this instruction doesn't generate any code.  That's because,
+    // by definition the pinned register is never used by the register allocator, but is written to
+    // and read explicitly and exclusively by set_pinned_reg and get_pinned_reg.
+    ig.push(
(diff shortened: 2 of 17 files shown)


cranelift-codegen/meta/src/cdsl/regs.rs
```diff
@@ -11,6 +11,7 @@ pub struct RegBank {
     pub names: Vec<&'static str>,
     pub prefix: &'static str,
     pub pressure_tracking: bool,
+    pub pinned_reg: Option<u8>,
```
iximeow: small quibble, shouldn't this be `Option<RegUnit>`? (which is u16, rather than u8)

it currently writes a literal out without a trailing `u8`, which is then inferred as `u16` when compiling the generated code, but sure would be an odd bug to track down :D

iximeow:
🎉🎉🎉 i was hoping to make time next week to do just this with the Spidermonkey changes you mentioned recently! instead i'll test run Lucet's tests + wasm spec tests with these changes instead :)

are you aware of use cases for pinning n>1 registers? there's a few we've talked about with Lucet, but we haven't measured what improvement we'd get yet.

cranelift-codegen/meta/src/cdsl/regs.rs
bnjbvr: Good catch, thanks.

bnjbvr:
> tadatadatada i was hoping to make time next week to do just this with the Spidermonkey changes you mentioned recently! instead i'll test run Lucet's tests + wasm spec tests with these changes instead :)

Note you still need to use the new instruction to set the heap base, before entering code compiled with Cranelift, and in certain precise points (when translating a call to an import, or an indirect call, or after wasm's mem.grow, and maybe others I'm not thinking of right now).

> are you aware of use cases for pinning n>1 registers? there's a few we've talked about with Lucet, but we haven't measured what improvement we'd get yet.

Well, ideally this would be its own mechanism in register allocation, maybe as a hint that some value is very important and should be preserved in its own register for the entire function's lifetime. So you could extend this mechanism to multiple heaps (when that comes in wasm), or you could imagine a very important value to live in its own register (emulated stack pointer, e.g.), or on some RISC platforms it could contain an interesting constant value (zero register, as on MIPS).

Spidermonkey also pins a local instance's storage (called thread local storage), that's the "VmCtx" parameter that's passed in the Baldrdash signatures. It was a bit more complicated to emulate what's done for VmCtx, especially to tell Cranelift how to recompute the heap base (this is done in the wasm->clif translation that's done on Spidermonkey's side).

Did you have other uses in mind?

cranelift-codegen/meta/src/cdsl/regs.rs
bnjbvr: Actually, there's no concept of RegUnit in the meta code, only in the non-meta code. I'll make it a u16 though, just for future proofing...

bnjbvr:
This is in better shape, with even more restrictions compared to previous iterations:

- the pinned register may not be the target of local moves generated by the solver (regmove/regfill).
- the pinned register may only be a definition of the current instruction if it's get_pinned_reg.

Any other case will trigger an assertion. Thanks to Julian's recent patches, the pinned reg (r15) was much likely to be taken as a fill's target, so I've got more confidence in this new version, which passes all of Cranelift, Embenchen and Spidermonkey testing. This is ready for review.

iximeow:
poppin' back in to say that Lucet's tests all pass, wasm spectests pass (well, as much as they did before), and most benchmarks look ~10-20% better! We have a few that measure slower, and I'm going to investigate why. I suspect it's unrelated to this change.

> Well, ideally this would be its own mechanism in register allocation, maybe as a hint that some value is very important and should be preserved in its own register for the entire function's lifetime.

This is about what I expected, too, yeah. I've been wondering if it makes sense to think about a register-pinned value as a particular flavor of GlobalValue, but that's not right since a global value would have a (much) wider lifetime than a "keep this value in this register" guidance that pinning a register really is.

> Did you have other uses in mind?

We have a bit of code in Lucet to optionally count the number of wasm operations executed, which currently works by adding to a counter tacked on to our VmCtx via wasm instrumented as we hand it to `cranelift-wasm`. It'd be interesting to measure the impact of keeping that counter in a pinned register as well. If memory serves, we go through ~4 memory accesses per counter update (so, control flow changes), which could be 2 if VmCtx was less spill-y, and probably a lower but non-zero average if we were more careful about not immediately writing the counter back to memory. Right now the performance is not great, especially in tight loops :(

cranelift-codegen/meta/src/shared/instructions.rs
```diff
@@ -993,6 +993,32 @@ pub fn define(
         .operands_out(vec![addr]),
     );
 
+    // Note this instruction is marked as having other side-effects, so GVN won't try to hoist it,
+    // which would result in it being subject to spilling. It's free to do so, because this
```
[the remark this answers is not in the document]
bnjbvr: Good call. Tweaked it to "While not hoisting would generally hurt performance, since a computed value used many times may need to be regenerated before each use, it is not the case here: this instruction doesn't generate any code. That's because, by definition the pinned register is entirely user controlled, so there's nothing to do to reset it."

cranelift-codegen/meta/src/shared/instructions.rs
```diff
@@ -993,6 +993,32 @@ pub fn define(
...
+    // instruction mustn't generate any code; by definition the pinned register is entirely user
+    // controlled, so there's nothing to do to reset it.
```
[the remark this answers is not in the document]
bnjbvr: ... actually tweaked the above comment to include this, thanks!

Issue #765, opened 2019-05-16
verifier error: Store must have an encoding for i686 targets when compiling wasm


htfy96:
## Sources
[sources.zip](https://github.com/CraneStation/cranelift/files/3189242/sources.zip)

cb.wasm was compiled with Emscripten 1.18.31 + stock LLVM with:
```
EMCC_WASM_BACKEND=1 emcc -s WASM=1 -s TOTAL_MEMORY=2147418112 -s ERROR_ON_UNDEFINED_SYMBOLS=0 -s EXPORTED_FUNCTIONS=['_double_plus_cnt_plus_pid','_square','_large_struct_op'] -fPIC -DPIC -g3 -O2 -I test/lib ../test/lib/libcb.c -o cb.wasm
```
## Behavior
Compiling this wasm with `clif-util wasm --target i686 ~/code/git/wasm-sandboxing/build/cb.wasm` with 42bd100d gives out `verifier error: Store must have an encoding`. Log is attached [here](https://github.com/CraneStation/cranelift/files/3189249/log.log)

sunfishcode:
Support for i686 (32-bit x86) is currently incomplete. A lot of the major pieces are in place, since we tend to implement the 32-bit x86 parts along with the 64-bit x86 parts, however it hasn't been tested yet, and there are a few more 32-bit-specific pieces needed.

PR #961, opened 2019-08-31
[codegen] add encodings for iadd carry variants

ryzokuken:
Add encodings for iadd carry variants (iadd_cout, iadd_cin, iadd_carry)
for x86_32, enabling the legalization for iadd.i64 to work.

Bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1576675
Bug: https://github.com/CraneStation/cranelift/issues/765

/cc @bnjbvr @wingo @caitp

### DIFF
--- a/cranelift-codegen/meta/src/isa/x86/recipes.rs
+++ b/cranelift-codegen/meta/src/isa/x86/recipes.rs
@@ -2528,6 +2528,61 @@ pub fn define<'shared>(
             ),
     );
 
+    // Adding with carry
+
+    // XX /r, MR form. Add two GPR registers and set carry flag.
+    recipes.add_template_recipe(
+        EncodingRecipeBuilder::new("rcout", f_binary, 1)
+            .operands_in(vec![gpr, gpr])
+            .operands_out(vec![0])
+            .clobbers_flags(true)
+            .emit(
+                r#"
+                    {{PUT_OP}}(bits, rex2(in_reg0, in_reg1), sink);
+                    modrm_rr(in_reg0, in_reg1, sink);
+                "#,
+            ),
+    );
+
+    // XX /r, MR form. Add two GPR registers and get carry flag.
+    recipes.add_template_recipe(
+        EncodingRecipeBuilder::new("rcin", f_ternary, 1)
+            .operands_in(vec![
+                OperandConstraint::RegClass(gpr),
+                OperandConstraint::RegClass(gpr),
+                OperandConstraint::FixedReg(reg_rflags),
+            ])
+            .operands_out(vec![0])
+            .clobbers_flags(true)
+            .emit(
+                r#"
+                    {{PUT_OP}}(bits, rex2(in_reg0, in_reg1), sink);
+                    modrm_rr(in_reg0, in_reg1, sink);
+                "#,
+            ),
+    );
+
+    // XX /r, MR form. Add two GPR registers with carry flag.
+    recipes.add_template_recipe(
+        EncodingRecipeBuilder::new("rcarry", f_ternary, 1)
+            .operands_in(vec![
+                OperandConstraint::RegClass(gpr),
+                OperandConstraint::RegClass(gpr),
+                OperandConstraint::FixedReg(reg_rflags),
+            ])
+            .operands_out(vec![
+                OperandConstraint::TiedInput(0),
+                OperandConstraint::FixedReg(reg_rflags),
+            ])
+            .clobbers_flags(true)
+            .emit(
+                r#"
+                    {{PUT_OP}}(bits, rex2(in_reg0, in_reg1), sink);
+                    modrm_rr(in_reg0, in_reg1, sink);
+                "#,
+            ),
+    );
+
     // Compare and set flags.
 
     // XX /r, MR form. Compare two GPR registers and set flags.
--- a/cranelift-codegen/meta/src/isa/x86/recipes.rs
+++ b/cranelift-codegen/meta/src/isa/x86/recipes.rs
@@ -2530,20 +2530,6 @@ pub fn define<'shared>(
 
     // Adding with carry
 
-    // XX /r, MR form. Add two GPR registers and set carry flag.
-    recipes.add_template_recipe(
-        EncodingRecipeBuilder::new("rcout", f_binary, 1)
-            .operands_in(vec![gpr, gpr])
-            .operands_out(vec![0])
-            .clobbers_flags(true)
-            .emit(
-                r#"
-                    {{PUT_OP}}(bits, rex2(in_reg0, in_reg1), sink);
-                    modrm_rr(in_reg0, in_reg1, sink);
-                "#,
-            ),
-    );
-
     // XX /r, MR form. Add two GPR registers and get carry flag.
     recipes.add_template_recipe(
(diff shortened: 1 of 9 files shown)


cranelift-codegen/meta/src/isa/x86/recipes.rs
```diff
@@ -2528,6 +2528,61 @@ pub fn define<'shared>(
...
+            ),
+    );
+
+    // XX /r, MR form. Add two GPR registers and get carry flag.
+    recipes.add_template_recipe(
+        EncodingRecipeBuilder::new("rcin", f_ternary, 1)
+            .operands_in(vec![
+                OperandConstraint::RegClass(gpr),
+                OperandConstraint::RegClass(gpr),
+                OperandConstraint::FixedReg(reg_rflags),
```
bjorn3: The carry flags are bools, which cant be stored in rflags if I understand correctly.

bjorn3:
Could you please add tests?

cranelift-codegen/meta/src/isa/x86/recipes.rs
ryzokuken: `cmov` uses `rflags` as the input condition, so I assumed it could be done rather similarly.

ryzokuken:
@bjorn3 I'm running tests locally, but I'm afraid I'll need https://github.com/CraneStation/cranelift/pull/957 to merge into master before adding tests in this since I'd need `iconst.i64` for that.

ryzokuken:
There *is* an issue though:

```
function u0:1(i32 [0], i32 [4]) -> i32 [%rax], i32 [%rdx] system_v {
    ss0 = incoming_arg 4, offset 0
    ss1 = incoming_arg 4, offset 4

                                ebb0(v4: i32 [ss0], v5: i32 [ss1]):
[-,-]                               v1 = iconcat v4, v5
[-,-]                               v3 = iadd_imm v1, 42
                                    v8, v9 = isplit v3
[Op1ret#c3]                         return v8, v9
;~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
; error: inst2: v8 is a ghost value used by a real [Op1ret#c3] instruction

}

; 1 verifier error detected (see above). Compilation aborted.
```

cranelift-codegen/meta/src/isa/x86/recipes.rs
bjorn3: That one has the clif name of `selectif` which takes an integer flag, not a bool.

bjorn3:
> I'll need #957 to merge into master before adding tests in this since I'd need iconst.i64 for that.

You could take 64bit as function arguments. You could also explicitly test `iadd_c*.i32` instead of `iadd.i64`.

cranelift-codegen/meta/src/isa/x86/recipes.rs
bnjbvr: Right. Looking at this more in detail, it appears that iadd_cin/cout/carry aren't encoded on any platforms, and the instruction definitions (in shared/instruction.rs) tell that they take an in/out b1 operand. In fact, I think this should be an `iflags` operand in this case, because what we really want is to use the flags register as you've done here. Materializing the flag value in an output register would defeat the purpose of this CLIF instruction, at least on x86.

Can you change the 3 instructions, so they take an iflags input/output instead of a b1 value? It might require more changes, like changing the formats, maybe updating tests too. With this change done, the recipes specified here look good.

bnjbvr:
> There is an issue though:

Yeah, that's simple preopt trying to optimize `iadd(iconst, iconst)` into `iadd_imm(iconst)`. We should disable this when the result type is wider than the native pointer type, and we should maybe split the expand legalization group into a group of default legalizations (iadd_imm -> iconst + iadd), and things that are really expansions.

In the meanwhile, @bjorn3 's suggestion of using parameters sounds great for testing.

cranelift-codegen/meta/src/shared/legalize.rs
```diff
@@ -494,16 +479,6 @@ pub fn define(insts: &InstructionGroup, immediates: &OperandKinds) -> TransformG
...
     );
 
-    expand.legalize(
-        def!((a, c) = iadd_carry(x, y, c_in)),
-        vec![
-            def!((a1, c1) = iadd_cout(x, y)),
-            def!(c_int = bint(c_in)),
-            def!((a, c2) = iadd_cout(a1, c_int)),
-            def!(c = bor(c1, c2)),
-        ],
```
bjorn3: Those legaizations are useful for platforms which dont yet support them natively. It wont hurt for x86 to keep them, as legalization is only performed when there is no encoding for an instruction.

bnjbvr: You're right. But this also implies that we can't just use the iflags as the type variable for the carry flags, which makes it impossible to properly implement add-with-carry on i686. @ryzokuken discussed this with Dan, who suggested removing the legalizations, since they were useful only for RISCV, which isn't quite properly implemented anyway. We'll try to find a better solution, like introducing new instructions maybe (iadd_ifcout/iadd_ifcin/iadd_ifcarry), that'll be involved on x86 legalizations for iadd_cout/cin/carry.

ryzokuken: +1 @bnjbvr. See https://github.com/CraneStation/cranelift/pull/961/files#diff-8d295395253ef7b2fd5b41fdd6386e73R1 and the commit messages for more.

filetests/isa/x86/legalize-i64.clif
```diff
@@ -1,10 +1,16 @@
...
 target i686 haswell
 
-function %iadd(i64, i64) {
+; regex: V=v\d+
+
+function %iadd(i64, i64) -> i64 {
 ebb0(v1: i64, v2: i64):
     v10 = iadd v1, v2
-    return
+    ; check: v1 = iconcat $(p11=$V), $(p12=$V)
```
bjorn3: Maybe use names like `v1_lsb` and `v1_msb` instead of `p11`.

filetests/isa/x86/legalize-i64.clif
```diff
@@ -0,0 +1,16 @@
+; Test the legalization of i64 instructions on x86_32.
+test legalizer
+target i686 haswell
+
+; regex: V=v\d+
```
bjorn3: Maybe this regex should be backed into filetest itself, as it is used often.

ryzokuken: @bnjbvr

bnjbvr: I wonder if this wouldn't be too magical, I'm fine with the current situation.

bjorn3:
@ryzokuken I think you forgot to stage a file in the last commit. The operand type for iadd carry variants hasn't been changed.

ryzokuken:
@bjorn3 the last commit is a fixup? I changed the type in the original commit (https://github.com/CraneStation/cranelift/pull/961/commits/4d9f584b0c8a0988c6534a94136d00937ed9345b)

bjorn3:
Didn't see that commit :)

bnjbvr:
(I squashed all the commits together because Github doesn't do autosquash on merges; next time feel free to autosquash locally before pushing!)

PR #966, opened 2019-09-03
Don't incorrectly omit a REX prefix for some encodings of `copy_to_ss…

julian-seward1:
…a`.  Mozilla bug #1576969.

Also, as a ridealong fix, removes R32 encodings for x86_64 in `enc_r32_r64`,
since the type `rXX` by definition only exists for targets with word size `XX`
bits.

### DIFF
--- a/cranelift-codegen/meta/src/isa/x86/encodings.rs
+++ b/cranelift-codegen/meta/src/isa/x86/encodings.rs
@@ -151,6 +151,16 @@ impl PerCpuModeEncodings {
     }
 
     /// Add encodings for `inst.i32` to X86_32.
+    /// Add encodings for `inst.i32` to X86_64 with a REX prefix.
+    /// Add encodings for `inst.i64` to X86_64 with a REX.W prefix.
+    fn enc_i32_i64_rex_only(&mut self, inst: impl Into<InstSpec>, template: Template) {
+        let inst: InstSpec = inst.into();
+        self.enc32(inst.bind(I32), template.nonrex());
+        self.enc64(inst.bind(I32), template.rex());
+        self.enc64(inst.bind(I64), template.rex().w());
+    }
+
+    /// Add encodings for `inst.i32` to X86_32.
     /// Add encodings for `inst.i32` to X86_64 with and without REX.
     /// Add encodings for `inst.i64` to X86_64 with a REX.W prefix.
     fn enc_i32_i64_instp(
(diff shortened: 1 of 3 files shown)


cranelift-codegen/meta/src/isa/x86/encodings.rs
```diff
@@ -964,20 +984,22 @@ pub fn define(
...
+    e.enc_i32_i64_rex_only(copy_to_ssa, rec_umr_reg_to_ssa.opcodes(vec![0x89]));
+    e.enc_r32_r64_rex_only(copy_to_ssa, rec_umr_reg_to_ssa.opcodes(vec![0x89]));
+    e.enc_both_rex_only(copy_to_ssa.bind(B1), rec_umr_reg_to_ssa.opcodes(vec![0x89]));
+    e.enc_both_rex_only(copy_to_ssa.bind(I8), rec_umr_reg_to_ssa.opcodes(vec![0x89]));
+    e.enc_both_rex_only(
         copy_to_ssa.bind(I16),
         rec_umr_reg_to_ssa.opcodes(vec![0x89]),
     );
-    e.enc_both(
+    e.enc_both_rex_only(
```
bnjbvr: Is the REX prefix meaningful for FP instructions? I wonder how this is dealt with...

cranelift-codegen/meta/src/isa/x86/encodings.rs
```diff
@@ -177,16 +187,18 @@ impl PerCpuModeEncodings {
...
-        // reg-alloc would never use r8 and up.
-        self.enc64(inst.bind_ref(R32), template.rex());
-        self.enc64(inst.bind_ref(R32), template.nonrex());
+    /// Add encodings for `inst.r32` to X86_32.
+    /// Add encodings for `inst.r64` to X86_64 with a REX.W prefix.
+    fn enc_r32_r64_rex_only(&mut self, inst: impl Into<InstSpec>, template: Template) {
+        let inst: InstSpec = inst.into();
+        self.enc32(inst.bind_ref(R32), template.nonrex());
         self.enc64(inst.bind_ref(R64), template.rex().w());
     }
```
bjorn3: No difference with `enc_r32_r64`?

cranelift-codegen/meta/src/isa/x86/encodings.rs
julian-seward1: Yes, it's meaningful for FP instructions.  Although REX.W (the 64-vs-32-bit flag) is meaningless for FP instructions that only deal with FP data, the other three REX bits are used to provide bit 3 of up to 3 register numbers, exactly as in the integer-instruction case.

cranelift-codegen/meta/src/isa/x86/encodings.rs
julian-seward1: Ah, well spotted.  I removed `enc_r32_r64` and replaced all uses of it with `enc_r32_r64_rex_only`.  The alternative would have been to remove `enc_r32_r64_rex_only` and use `enc_r32_r64` instead, but that would mean that the REX behaviour of `enc_r32_r64` was inconsistent with that of the otherwise very similar `enc_i32_i64`.  So I didn't do that.

PR #968, opened 2019-09-04
[meta] Lazy_static for immediates and entity refs

bnjbvr:
Let's try to make this incrementally. Formats will require more discussion and work (FormatRegistry would be mutable and shared, which is pretty unsafe, so we need to figure this out first), so deferring. I'd suggest looking at this PR commit by commit, but let me know if you'd prefer smaller PRs. @abrown can you take a look, please? Thanks!

### DIFF
--- a/cranelift-codegen/meta/Cargo.toml
+++ b/cranelift-codegen/meta/Cargo.toml
@@ -9,6 +9,7 @@ readme = "README.md"
 edition = "2018"
 
 [dependencies]
+lazy_static = "1.4.0"
 cranelift-entity = { path = "../../cranelift-entity", version = "0.41.0", default-features = false }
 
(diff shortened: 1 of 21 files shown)


bjorn3:
In case you missed https://github.com/bnjbvr/cranelift/commit/512b93a2793798b56f49023635e057688fedf366#r34944606:

You could try using the fragile crate instead of `RefCell` inside `TypeVar`. That would make `TypeVar` Sync (after you replace Rc with Arc) while not requiring any lock.

bnjbvr:
Removed review request; I've been suggested an even simpler way that I'd like to try first.

bnjbvr:
Closing, the other approach is simpler and agreed upon.

PR #969, opened 2019-09-04
[meta] Some random refactorings

bnjbvr:
Split off from #968.

### DIFF
--- a/cranelift-codegen/meta/src/cdsl/instructions.rs
+++ b/cranelift-codegen/meta/src/cdsl/instructions.rs
@@ -113,9 +113,12 @@ pub struct InstructionContent {
     /// polymorphic, set otherwise.
     pub polymorphic_info: Option<PolymorphicInfo>,
 
+    /// Indices in operands_in of input operands that are values.
     pub value_opnums: Vec<usize>,
-    pub value_results: Vec<usize>,
+    /// Indices in operands_in of input operands that are immediates or entities.
     pub imm_opnums: Vec<usize>,
+    /// Indices in operands_out of output operands that are values.
+    pub value_results: Vec<usize>,
 
     /// True for instructions that terminate the EBB.
     pub is_terminator: bool,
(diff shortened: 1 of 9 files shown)


cranelift-codegen/meta/src/cdsl/instructions.rs
```diff
@@ -527,57 +525,51 @@ fn verify_polymorphic(
 ///
 /// All polymorphic results must be derived from `ctrl_typevar`.
 ///
-/// Return a vector of other type variables used, or panics.
-fn verify_ctrl_typevar(
+/// Return a vector of other type variables used, or a string explaining what went wrong.
+fn is_ctrl_typevar_candidate(
```
abrown: We might consider tests for `is_ctrl_typevar_candidate` and `verify_polymorphic`; not because they are so very complex but a couple of illustrative tests might explain how these functions work and ensure that when I touch this file I don't mess up the intended behavior 😄.

cranelift-codegen/src/ir/constant.rs
```diff
@@ -44,21 +47,23 @@ impl ConstantPoolEntry {
...
 
-/// Maintains the mapping between a constant handle (i.e.
-/// [`Constant`](crate::ir::Constant)) and its constant data (i.e.
-/// [`ConstantData`](crate::ir::ConstantData)).
+/// Maintains the mapping between a constant handle (i.e.  [`Constant`](crate::ir::Constant)) and
+/// its constant data (i.e.  [`ConstantData`](crate::ir::ConstantData)).
 #[derive(Clone)]
 pub struct ConstantPool {
-    /// This mapping maintains the insertion order as long as Constants are created with sequentially increasing integers.
+    /// This mapping maintains the insertion order as long as Constants are created with
```
abrown: So should I be trying to keep comments under 100 characters?

bnjbvr: Yes please! :) It can't be enforced with the rustfmt right now because this is a Nightly feature (turns out, formatting comments containing code examples is complicated).

abrown: Sounds good, I'll find my IDE setting for that.

PR #970, opened 2019-09-04
[meta] Use giant structures for Immediates/EntityRefs

bnjbvr:
... instead of static instances.

This is nice too: it gets the explicit passing to functions that need it (instead of global variables with hacks to make them static), erroneous lookup is impossible, jump-to-definition is better (not as good as static variables: you jump to the definition, and have to look at the constructor to see how it was formed), it doesn't require a new dependency just for the build step.

Would people like this, it'd be much easier to keep on doing this for formats and other things.

cc @abrown @sunfishcode

### DIFF
--- a/cranelift-codegen/meta/src/cdsl/ast.rs
+++ b/cranelift-codegen/meta/src/cdsl/ast.rs
@@ -11,7 +11,6 @@ use std::fmt;
 pub enum Expr {
     Var(VarIndex),
     Literal(Literal),
-    Apply(Apply),
 }
 
 impl Expr {
@@ -39,7 +38,6 @@ impl Expr {
         match self {
             Expr::Var(var_index) => var_pool.get(*var_index).to_rust_code(),
             Expr::Literal(literal) => literal.to_rust_code(),
-            Expr::Apply(a) => a.to_rust_code(var_pool),
         }
     }
 }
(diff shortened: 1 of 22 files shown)


cranelift-codegen/meta/src/shared/instructions.rs
```diff
@@ -24,19 +25,19 @@ pub fn define(
...
+    let uimm8 = &imm.uimm8;
+    let uimm32 = &imm.uimm32;
+    let imm64 = &imm.imm64;
+    let uimm128 = &imm.uimm128;
+    let offset32 = &imm.offset32;
+    let memflags = &imm.memflags;
+    let ieee32 = &imm.ieee32;
+    let ieee64 = &imm.ieee64;
+    let boolean = &imm.boolean;
+    let regunit = &imm.regunit;
```
abrown: These re-assignment lists are just another level of indirection when trying to look up the code; can we just do `&imm.ieee32`, e.g., where it is used?

cranelift-codegen/meta/src/cdsl/ast.rs
```diff
@@ -11,7 +11,6 @@ use std::fmt;
 pub enum Expr {
     Var(VarIndex),
     Literal(Literal),
-    Apply(Apply),
```
abrown: I don't really understand how this can go away?

bnjbvr: This was Rustc complaining that it was unused after I changed some pub to pub(crate). In fact, Apply expressions appear only in DummyExpr, which gets manipulated by xform.rs, but never rewritten to an Expr. Instead, it's embedded as part of a Def.

cranelift-codegen/meta/src/shared/instructions.rs
bnjbvr: Yes, good idea!

PR #972, opened 2019-09-04
Limit redundant jump folding to only fold parameterless target blocks

sstangl:
This is a better solution than https://github.com/CraneStation/cranelift/pull/950.

It's obviously possible to do better than this, for example by folding edges that cannot possible be implicit phi nodes. The purpose of this phase is mostly to undo garbage blocks created by register allocation, so this simpler version serves that goal just fine.

### DIFF
--- a/cranelift-codegen/src/binemit/relaxation.rs
+++ b/cranelift-codegen/src/binemit/relaxation.rs
@@ -169,6 +169,14 @@ fn try_fold_redundant_jump(
         }
     };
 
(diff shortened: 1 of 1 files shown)


bnjbvr:
There's no policies, so let's try to reproduce Spidermonkey's: a TODO = an issue open to discuss it and not forget about it? (There are many TODOs without matching issues at the moment, so it's fine to do best-effort here.)

sstangl:
There's already a Bugzilla issue still open for fixing this thoroughly. That's still in my dashboard and I won't forget about it.

PR #988, opened 2019-09-06
Simplify isa_builder macro

nbp:
This changes replace the inner function creation by a simple if using the `cfg!` macro.

### DIFF
--- a/cranelift-codegen/src/isa/mod.rs
+++ b/cranelift-codegen/src/isa/mod.rs
@@ -88,18 +88,17 @@ pub mod registers;
 mod stack;
 
 /// Returns a builder that can create a corresponding `TargetIsa`
-/// or `Err(LookupError::Unsupported)` if not enabled.
+/// or `Err(LookupError::SupportDisabled)` if not enabled.
 macro_rules! isa_builder {
-    ($name:ident, $feature:tt) => {{
+    ($name: ident, $feature: tt, $triple: ident) => {{
         #[cfg(feature = $feature)]
-        fn $name(triple: Triple) -> Result<Builder, LookupError> {
-            Ok($name::isa_builder(triple))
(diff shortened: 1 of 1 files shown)


cranelift-codegen/src/isa/mod.rs
```diff
@@ -90,29 +90,25 @@ mod stack;
...
 macro_rules! isa_builder {
-    ($name:ident, $feature:tt) => {{
-        #[cfg(feature = $feature)]
-        fn $name(triple: Triple) -> Result<Builder, LookupError> {
-            Ok($name::isa_builder(triple))
-        };
-        #[cfg(not(feature = $feature))]
-        fn $name(_triple: Triple) -> Result<Builder, LookupError> {
-            Err(LookupError::Unsupported)
+    ($name: ident, $feature: tt, $triple: ident) => {{
```
bjorn3: The name is now unuses.

bnjbvr: It's used line 95.

bjorn3: Oops.

bnjbvr: There's probably a macro way to reuse $name for $feature, using the `stringify!` macro, though.

nbp: Doing so would imply that we always keep the same for both, and modules names are more restricted than feature names.

nbp:
Apparently we cannot refer to a conditional module name based only on the branch condition. The code has to be valid whether or not the feature is present or not.

I will rewrite it to

```rust
#[cfg(feature = $feature)]
{
   ...
}
#[cfg(feature != $feature)]
{
   ...
}
```

Which avoids the function call.

PR #991, opened 2019-09-06
Add equivalent safepoint test cases for basic blocks.

nbp:


nbp:
> Why is call-bb.clif necessary? Comparing it with the original call.clif, it looks to me that the original doesn't have any real Ebbs anyway.

`call-bb.clif` happens to be working well with and without the feature flag, I will just replace the `call.clif` content to replace the ebb then.

PR #992, opened 2019-09-06
Support running spec tests in wasmtime

abrown:
To support running spec tests in wasmtime, I had to modify cranelift with the following commits. This is a non-exhaustive list since some spec tests are currently failing and will likely require further changes.

### DIFF
--- a/cranelift-codegen/meta/src/isa/x86/encodings.rs
+++ b/cranelift-codegen/meta/src/isa/x86/encodings.rs
@@ -293,6 +293,12 @@ impl PerCpuModeEncodings {
         }
     }
 
+    /// Add the same encoding/template pairing to both X86_32 and X86_64
+    fn enc_32_64(&mut self, inst: impl Clone + Into<InstSpec>, template: Template) {
+        self.enc32(inst.clone(), template.clone());
+        self.enc64(inst, template);
+    }
+
     /// Add the same encoding/recipe pairing to both X86_32 and X86_64
     fn enc_32_64_rec(
         &mut self,
@@ -1957,6 +1963,40 @@ pub(crate) fn define(
         let instruction = bor.bind_vector_from_lane(ty, sse_vector_size);
         let template = rec_fa.nonrex().opcodes(vec![0x0f, 0x56]);
         e.enc_32_64_maybe_isap(instruction, template, None); // from SSE
+    }
+
+    // SIMD register movement: store, load, spill, fill, regmove. All of these use encodings of
+    // MOVUPS and MOVAPS from SSE (TODO ideally all of these would either use MOVAPS when we have
+    // alignment or type-specific encodings, see https://github.com/CraneStation/cranelift/issues/1039).
+    for ty in ValueType::all_lane_types().filter(allowed_simd_type) {
+        // Store
+        let bound_store = store.bind_vector_from_lane(ty, sse_vector_size).bind_any();
+        e.enc_32_64(bound_store.clone(), rec_fst.opcodes(vec![0x0f, 0x11]));
+        e.enc_32_64(bound_store.clone(), rec_fstDisp8.opcodes(vec![0x0f, 0x11]));
+        e.enc_32_64(bound_store, rec_fstDisp32.opcodes(vec![0x0f, 0x11]));
+
+        // Load
+        let bound_load = load.bind_vector_from_lane(ty, sse_vector_size).bind_any();
+        e.enc_32_64(bound_load.clone(), rec_fld.opcodes(vec![0x0f, 0x10]));
+        e.enc_32_64(bound_load.clone(), rec_fldDisp8.opcodes(vec![0x0f, 0x10]));
+        e.enc_32_64(bound_load, rec_fldDisp32.opcodes(vec![0x0f, 0x10]));
+
+        // Spill
+        let bound_spill = spill.bind_vector_from_lane(ty, sse_vector_size);
+        e.enc_32_64(bound_spill, rec_fspillSib32.opcodes(vec![0x0f, 0x11]));
+        let bound_regspill = regspill.bind_vector_from_lane(ty, sse_vector_size);
+        e.enc_32_64(bound_regspill, rec_fregspill32.opcodes(vec![0x0f, 0x11]));
+
+        // Fill
+        let bound_fill = fill.bind_vector_from_lane(ty, sse_vector_size);
+        e.enc_32_64(bound_fill, rec_fillSib32.opcodes(vec![0x0f, 0x10]));
+        let bound_regfill = regfill.bind_vector_from_lane(ty, sse_vector_size);
+        e.enc_32_64(bound_regfill, rec_fregfill32.opcodes(vec![0x0f, 0x10]));
+
+        // Regmove
+        let bound_regmove = regmove.bind_vector_from_lane(ty, sse_vector_size);
+        e.enc_32_64(bound_regmove.clone(), rec_rmov.opcodes(vec![0x0f, 0x28]));
+        e.enc_32_64(bound_regmove, rec_frmov.opcodes(vec![0x0f, 0x28]));
     }
 
     // Reference type instructions
--- a/cranelift-codegen/meta/src/isa/x86/encodings.rs
+++ b/cranelift-codegen/meta/src/isa/x86/encodings.rs
@@ -1992,11 +1992,19 @@ pub(crate) fn define(
         e.enc_32_64(bound_fill, rec_fillSib32.opcodes(vec![0x0f, 0x10]));
         let bound_regfill = regfill.bind_vector_from_lane(ty, sse_vector_size);
         e.enc_32_64(bound_regfill, rec_fregfill32.opcodes(vec![0x0f, 0x10]));
+        let bound_fill_nop = fill_nop.bind_vector_from_lane(ty, sse_vector_size);
+        e.enc_32_64_rec(bound_fill_nop, rec_ffillnull, 0);
 
         // Regmove
         let bound_regmove = regmove.bind_vector_from_lane(ty, sse_vector_size);
         e.enc_32_64(bound_regmove.clone(), rec_rmov.opcodes(vec![0x0f, 0x28]));
         e.enc_32_64(bound_regmove, rec_frmov.opcodes(vec![0x0f, 0x28]));
+
+        // Copy
+        let bound_copy = copy.bind_vector_from_lane(ty, sse_vector_size);
+        e.enc_32_64(bound_copy, rec_furm.opcodes(vec![0x0f, 0x28])); // MOVAPS from SSE
+        let bound_copy_nop = copy_nop.bind_vector_from_lane(ty, sse_vector_size);
+        e.enc_32_64_rec(bound_copy_nop, rec_stacknull, 0);
     }
 
     // SIMD integer addition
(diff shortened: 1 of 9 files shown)


cranelift-wasm/src/translation_utils.rs
```diff
@@ -76,6 +77,8 @@ pub enum GlobalInit {
     F32Const(u32),
     /// An `f64.const`.
     F64Const(u64),
+    /// A `vconst`.
+    V128Const(Uimm128),
```
abrown: @sunfishcode, would you prefer this was `V128Const([u8; 16])` to avoid exposing our Uimm128 type?

sunfishcode: Ah. I commented about this in the Wasmtime PR. I think we can expose it, because it is just a raw value with no particular implementation details baked in. But we should rename it to `V128Imm` so that it's more descriptive of what  it represents.

cranelift-wasm/src/translation_utils.rs
```diff
@@ -65,7 +66,7 @@ pub struct Global {
     pub initializer: GlobalInit,
 }
 
-/// Globals are initialized via the four `const` operators or by referring to another import.
+/// Globals are initialized via the five `const` operators or by referring to another import.
```
bnjbvr: Maybe we could just get rid of the manual count in the comment, I don't think it gives us much.

cranelift-codegen/meta/src/isa/x86/encodings.rs
```diff
@@ -1839,6 +1845,39 @@ pub(crate) fn define(
...
+        e.enc_32_64(bound_load.clone(), rec_fldDisp8.opcodes(vec![0x0f, 0x11]));
+        e.enc_32_64(bound_load, rec_fldDisp32.opcodes(vec![0x0f, 0x11]));
+
+        // Spill
+        let bound_spill = spill.bind_vector_from_lane(ty, sse_vector_size);
+        e.enc_32_64(bound_spill, rec_fspillSib32.opcodes(vec![0x0f, 0x11]));
+        let bound_regspill = regspill.bind_vector_from_lane(ty, sse_vector_size);
+        e.enc_32_64(bound_regspill, rec_fregspill32.opcodes(vec![0x0f, 0x11]));
+
+        // Fill
```
bnjbvr: You might need fill_nop and the copy_nop encodings too, since they can be inserted by regalloc to replace actuall moves and fills by typed noops.

(Also if copy isn't present already, it will be needed for regalloc.)

abrown:
With these changes and #993 to avoid a test failure in `spec_testsuite/address.wast`, I can now run the `simd_const.wast` test in https://github.com/CraneStation/wasmtime/pull/330.

cranelift-codegen/meta/src/isa/x86/encodings.rs
```diff
@@ -1899,6 +1905,39 @@ pub(crate) fn define(
         e.enc_32_64_maybe_isap(instruction, template, None); // from SSE
     }
 
+    // SIMD register movement: store, load, spill, fill, regmove. All of these use variants of
+    // MOVUPS from SSE (TODO ideally this would use the aligned MOVAPS).
+    for ty in ValueType::all_lane_types().filter(allowed_simd_type) {
+        // Store
+        let bound_store = store.bind_vector_from_lane(ty, sse_vector_size).bind_any();
+        e.enc_32_64(bound_store.clone(), rec_fst.opcodes(vec![0x0f, 0x11]));
```
bnjbvr: For all of these, shouldn't we use enc_both instead, so this adds the rex-variant on x64, allowing uses of more xmm registers? (makes me wonder if enc_32_64_maybe_isap should do this too)

cranelift-codegen/meta/src/isa/x86/encodings.rs
```diff
@@ -1899,6 +1905,39 @@ pub(crate) fn define(
         e.enc_32_64_maybe_isap(instruction, template, None); // from SSE
     }
 
+    // SIMD register movement: store, load, spill, fill, regmove. All of these use variants of
+    // MOVUPS from SSE (TODO ideally this would use the aligned MOVAPS).
```
bnjbvr: Does a perf penalty occur if we use MOVUPS instead of e.g. MOVUPD for float64x2, or MOVDQU for integer SIMD types?

cranelift-codegen/meta/src/isa/x86/encodings.rs
```diff
@@ -1938,6 +1938,17 @@ pub(crate) fn define(
         e.enc_32_64(bound_regmove, rec_frmov.opcodes(vec![0x0f, 0x11]));
     }
 
+    // SIMD integer addition
+    let mut x86_padd_mapping: HashMap<LaneType, Vec<u8>> = HashMap::new();
+    x86_padd_mapping.insert(I8.into(), vec![0x66, 0x0f, 0xfc]); // PADDB from SSE2
+    x86_padd_mapping.insert(I16.into(), vec![0x66, 0x0f, 0xfd]); // PADDW from SSE2
+    x86_padd_mapping.insert(I32.into(), vec![0x66, 0x0f, 0xfe]); // PADDD from SSE2
+    x86_padd_mapping.insert(I64.into(), vec![0x66, 0x0f, 0xd4]); // PADDQ from SSE2
+    for (ty, opcodes) in x86_padd_mapping {
```
bnjbvr: (First time I realize this, this might apply also in other places). Instead of creating a hashmap, can we do:

```
for &(ty, opcodes) in &[
  (I8, [0x66, 0x0f, 0xfc]),
  (I16, [0x66, 0x0f, 0xfd]) ...
] {
```

cranelift-codegen/meta/src/isa/x86/encodings.rs
```diff
@@ -1906,7 +1906,7 @@ pub(crate) fn define(
     }
 
     // SIMD register movement: store, load, spill, fill, regmove. All of these use variants of
-    // MOVUPS from SSE (TODO ideally this would use the aligned MOVAPS).
+    // MOVUPS from SSE (TODO ideally this would use the aligned MOVAPS) except for copy.
```
bnjbvr: Regmove is register to register too, so it could use movaps as well.

cranelift-codegen/meta/src/isa/x86/encodings.rs
abrown: Just a cursory glance seems to indicate that, yes, there seem to be faster encodings for different types. For this PR I was still just trying to get the wasmtime spec tests to work so I was not thinking too much about performance yet. Let me open an issue for this one because this is exactly the kind of thing I will be spending more time on after this is all working: https://github.com/CraneStation/cranelift/issues/1039.

cranelift-codegen/meta/src/isa/x86/encodings.rs
abrown: Let me do some more research on this one; I think you're right from what I figured out in https://github.com/CraneStation/cranelift/pull/919 but I just want to make sure.

abrown: I've convinced myself that this is safe to do. I am still planning to refactor `PerCpuModeEncodings` as a part of #955 so I can do it there (and remove all the unnecessary `enc_*` methods).

bnjbvr: Sounds good to me, thanks.

cranelift-codegen/meta/src/isa/x86/encodings.rs
```diff
@@ -1820,23 +1826,26 @@ pub(crate) fn define(
...
-    let mut x86_pinsr_mapping: HashMap<u64, (Vec<u8>, Option<SettingPredicateNumber>)> =
-        HashMap::new();
-    x86_pinsr_mapping.insert(8, (vec![0x66, 0x0f, 0x3a, 0x20], Some(use_sse41_simd))); // PINSRB
-    x86_pinsr_mapping.insert(16, (vec![0x66, 0x0f, 0xc4], None)); // PINSRW from SSE2
-    x86_pinsr_mapping.insert(32, (vec![0x66, 0x0f, 0x3a, 0x22], Some(use_sse41_simd))); // PINSRD
-    x86_pinsr_mapping.insert(64, (vec![0x66, 0x0f, 0x3a, 0x22], Some(use_sse41_simd))); // PINSRQ, only x86_64
-
-    for ty in ValueType::all_lane_types().filter(allowed_simd_type) {
-        if let Some((opcode, isap)) = x86_pinsr_mapping.get(&ty.lane_bits()) {
+    for (lane_bits, opcode, isap) in [
```
bnjbvr: Sorry, this was a bad call from me for insertlane and extractlane; this way of doing it implies to clone the content that's iterated over for no good reason (it's consumed once and for all), plus we now iterate on the all_lane_types() many more times. For these two, let's keep the previous way.

cranelift-codegen/meta/src/isa/x86/encodings.rs
```diff
@@ -1978,12 +1985,15 @@ pub(crate) fn define(
...
 
     // SIMD integer addition
-    let mut x86_padd_mapping: HashMap<LaneType, Vec<u8>> = HashMap::new();
-    x86_padd_mapping.insert(I8.into(), vec![0x66, 0x0f, 0xfc]); // PADDB from SSE2
-    x86_padd_mapping.insert(I16.into(), vec![0x66, 0x0f, 0xfd]); // PADDW from SSE2
-    x86_padd_mapping.insert(I32.into(), vec![0x66, 0x0f, 0xfe]); // PADDD from SSE2
-    x86_padd_mapping.insert(I64.into(), vec![0x66, 0x0f, 0xd4]); // PADDQ from SSE2
-    for (ty, opcodes) in x86_padd_mapping {
+    for (ty, opcodes) in [
+        (I8, vec![0x66, 0x0f, 0xfc]),  // PADDB from SSE2
```
bnjbvr: ... but for this one, you can do:

```
    for &(ty, opcodes) in &[
        (I8, &[0x66, 0x0f, 0xfc]),  // PADDB from SSE2
        (I16, &[0x66, 0x0f, 0xfd]), // PADDW from SSE2
        (I32, &[0x66, 0x0f, 0xfe]), // PADDD from SSE2
        (I64, &[0x66, 0x0f, 0xd4]), // PADDQ from SSE2
    ]
    {
        let iadd = iadd.bind_vector_from_lane(ty, sse_vector_size);
        e.enc_32_64(iadd, rec_fa.opcodes(Vec::from(opcodes.as_ref())));
    }
```

(no more clon ing! To wit: it's possible to do so here because all the opcodes have the same length.)

abrown: I still had to `ty.clone()` but I ended up using `opcodes.to_vec()` which is a bit shorter:
```
    for (ty, opcodes) in &[
        (I8, &[0x66, 0x0f, 0xfc]),  // PADDB from SSE2
        (I16, &[0x66, 0x0f, 0xfd]), // PADDW from SSE2
        (I32, &[0x66, 0x0f, 0xfe]), // PADDD from SSE2
        (I64, &[0x66, 0x0f, 0xd4]), // PADDQ from SSE2
    ] {
        let iadd = iadd.bind_vector_from_lane(ty.clone(), sse_vector_size);
        e.enc_32_64(iadd, rec_fa.opcodes(opcodes.to_vec()));
    }
```

Issue #980, opened 2019-09-05
Bad use of iadd_cout when legalizing dynamic heaps


bnjbvr:
The following test fails on master:

```
target x86_64

function u0:0(i32, i64 vmctx) baldrdash_system_v {
    gv0 = vmctx
    gv1 = load.i64 notrap aligned readonly gv0
    gv2 = load.i32 notrap aligned gv0+8
    heap0 = dynamic gv1, min 0x0001_0000, bound gv2, offset_guard 0xfff8, index_type i32

                                ebb0(v0: i32, v1: i64):
@003d                               v2 = heap_addr.i64 heap0, v0, 0x7fff_ffe1
@003d                               v3 = iadd_imm v2, 0x8000_0000
@003d                               v4 = sload8.i32 v3
@0045                               jump ebb1

                                ebb1:
@0045                               fallthrough_return
}
```

This is because heap_addr gets legalized into a sequence that involves an iadd_cout which carry flag is fed into a trapnz, but it's still assuming the carry result will be a b1, not an iflags. So the trapnz that gets generated needs to be replaced by a trapif with an Overflow condition code. Overflow isn't yet handled in Cranelift, though.

@ryzokuken Do you want to have a look?

ryzokuken:
I'll fix this.

PR #993, opened 2019-09-06
[codegen] add intcc conditions for reading carry flag

ryzokuken:
Add conditions to IntCC for checking the carry flag (Carry, NotCarry).

Fixes: https://github.com/CraneStation/cranelift/issues/980

/cc @bnjbvr @wingo @caitp

### DIFF
--- a/cranelift-codegen/src/isa/mod.rs
+++ b/cranelift-codegen/src/isa/mod.rs
@@ -371,4 +371,10 @@ pub trait TargetIsa: fmt::Display + Sync {
 
     /// Emit a whole function into memory.
     fn emit_function_to_memory(&self, func: &ir::Function, sink: &mut binemit::MemoryCodeSink);
+
+    /// IntCC condition for Unsigned Addition Overflow (Carry).
+    fn unsigned_add_overflow_condition(&self) -> ir::condcodes::IntCC;
+
+    /// IntCC condition for Unsigned Subtraction Overflow (Borrow/Carry).
+    fn unsigned_sub_overflow_condition(&self) -> ir::condcodes::IntCC;
 }
--- a/cranelift-codegen/meta/src/shared/instructions.rs
+++ b/cranelift-codegen/meta/src/shared/instructions.rs
@@ -1586,6 +1586,7 @@ pub(crate) fn define(
     let x = &operand("x", Int);
     let y = &operand("y", Int);
 
+    // TODO(ryzokuken): Add documentation for unsigned overflow.
     ig.push(
         Inst::new(
             "icmp",
--- a/cranelift-codegen/src/isa/arm32/mod.rs
+++ b/cranelift-codegen/src/isa/arm32/mod.rs
@@ -126,6 +126,14 @@ impl TargetIsa for Isa {
     fn emit_function_to_memory(&self, func: &ir::Function, sink: &mut MemoryCodeSink) {
         emit_function(func, binemit::emit_inst, sink, self)
     }
+
+    fn unsigned_add_overflow_condition(&self) -> ir::condcodes::IntCC {
+        ir::condcodes::IntCC::UnsignedLessThan
+    }
+
+    fn unsigned_sub_overflow_condition(&self) -> ir::condcodes::IntCC {
+        ir::condcodes::IntCC::UnsignedGreaterThanOrEqual
+    }
 }
 
 impl fmt::Display for Isa {
--- a/cranelift-codegen/src/isa/arm64/mod.rs
+++ b/cranelift-codegen/src/isa/arm64/mod.rs
@@ -114,6 +114,14 @@ impl TargetIsa for Isa {
     fn emit_function_to_memory(&self, func: &ir::Function, sink: &mut MemoryCodeSink) {
         emit_function(func, binemit::emit_inst, sink, self)
     }
+
+    fn unsigned_add_overflow_condition(&self) -> ir::condcodes::IntCC {
+        ir::condcodes::IntCC::UnsignedLessThan
+    }
+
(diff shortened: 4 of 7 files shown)


cranelift-codegen/src/ir/condcodes.rs
```diff
@@ -101,6 +109,8 @@ impl Display for IntCC {
             UnsignedGreaterThanOrEqual => "uge",
             UnsignedLessThan => "ult",
             UnsignedLessThanOrEqual => "ule",
+            Carry => "cr",
+            NotCarry => "nc",
```
bjorn3:
```suggestion
            NotCarry => "ncr",
```

ryzokuken: lol you're quick. I'll do a repo-wide search and replace.

ryzokuken: Done!

ryzokuken:
@bjorn3 that said, it's failing right now with:

```
FAIL filetests/isa/x86/binary32.clif: binemit(%cpu_flags):
Bad machine code for inst45: trapif cr v11, user0
Want: 72 02 user0 0f 0b
Got:  73 02 user0 0f 0b
FAIL filetests/isa/x86/binary64.clif: binemit(%cpu_flags_I64):
Bad machine code for inst45: trapif cr v11, user0
Want: 72 02 user0 0f 0b
Got:  73 02 user0 0f 0b
```

I've tried, but couldn't really find out why this could be happening. The opcodes are supposed to be in place. Please help? I'll keep looking in the meantime.

ryzokuken:
I think I have a clue.

ryzokuken:
Yay! I was expecting the wrong value. (I mean, I could see that, but I misinterpreted the pattern and realized what the actual pattern was and it was inline with what was being emitted).

@bnjbvr's test file still fails with `thread 'main' panicked at 'Expected register: Unassigned', cranelift-codegen/src/ir/valueloc.rs:44:18` though. Will try to debug over the weekend.

ryzokuken:
@sunfishcode Yes, I was planning to do it in a separate PR.

ryzokuken:
Rebased on #1014, please don't merge this without merging #1014 on master.

Okay, update, #1014 fixes the register, but now throws: `thread 'main' panicked at 'baldrdash does not expect cranelift to emit stack probes', cranelift-codegen/src/isa/x86/abi.rs:342:5`

ryzokuken:
@sunfishcode @bnjbvr I started hacking into your suggestion and towards the end reached the conclusion that the current code ins't add bad at it might seem. While the carry flag has been added to the `IntCC` enum in the common code (in the IR itself at https://github.com/CraneStation/cranelift/pull/993/files#diff-405e56b457fbe4a025f4ca292265df6f), the actual hexadecimal value for calculating the opcode for the corresponding jump instruction are all done in code exclusive to x86 (which I guess implies that the value for other ISAs is unimplemented). Also, there's no compatibility issues I could think of, since other ISAs don't seem to have any recipes that can do `IntCC` related work anyway.

bnjbvr:
So imagine a situation where you're emitting IR, emit a sub with a borrow; how do you know which condition code to use to check in the flags if there's a borrow? If you don't know ahead of time what the target ISA will be, or you want your code to work on all platforms, then you need some target ISA specific information telling you which of Carry or NotCarry is to be tested for the borrow. You could imagine a function on the TargetISA answering which is the right one, and let Carry/NotCarry in the IntCC enum; but then, there's still opportunity for misuse by embedders not knowing about this TargetISA function. Thus, not having Carry/NotCarry in the IntCC enum makes it easier to not shoot yourself in the foot in any way. Does it make sense?

Alternatively, we could imagine instructions like `check_borrow(iflags) -> b1`, but apart from working around the condition code selection issue here, I'm not sure they'd bring new optimization opportunities.

sunfishcode:
Friendly ping :-). This bug affects the wasm spec testsuite, so if you need more time, just say so, and we'll figure out a plan.

ryzokuken:
@sunfishcode I'm sorry, I was at a conference. Omw home now, will work this out within the week.

ryzokuken:
@sunfishcode regarding the code comments, I guess you mean https://github.com/CraneStation/cranelift/blob/cdf29c71ac8f314c4736031d92637a7e91bac9d6/cranelift-codegen/meta/src/shared/instructions.rs#L1593? The link you sent was relative to the file and I took a while so it moved. LMK what you feel about the code and I'll document this promptly.

sunfishcode:
I mean the documentation in the `"icmp"` instruction, which as of this writing is now [here](https://github.com/CraneStation/cranelift/blob/master/cranelift-codegen/meta/src/shared/instructions.rs#L1593).

cranelift-codegen/src/isa/mod.rs
```diff
@@ -371,4 +371,10 @@ pub trait TargetIsa: fmt::Display + Sync {
 
     /// Emit a whole function into memory.
     fn emit_function_to_memory(&self, func: &ir::Function, sink: &mut binemit::MemoryCodeSink);
+
+    /// Get the value for the condition CF=1.
```
sunfishcode: This comment is in target-independent code, so we should say something like "the condition for unsigned overflow" instead of referencing "CF". In fact, we should probably say `unsigned` somewhere in the function name too, like `unsigned_add_overflow_condition`, just to be unambiguous.

ryzokuken:
@sunfishcode fixed and rebased. Instead of adding documentation for `icmp`, I've added a TODO and will be doing the same for #995 (since this part for the two is linked together and I'd rather do it in one go). Could you please merge both if the look good apart from documentation and I'll get rid of both TODOs in one go?

PR #997, opened 2019-09-07
Correctly zero extend operand of fcvt_from_uint for 8ints and 16bit ints

### LINKED ISSUE #996: fcvt_from_uint for 8bit and 16bit uints is incorrect
```
v0 = iconst.i8 0xff
v1 = fcvt_from_uint.f32 v0
```

Gives `254.0` instead of `255.0`.

bjorn3:
Fixes #996

### DIFF
--- a/cranelift-codegen/meta/src/shared/legalize.rs
+++ b/cranelift-codegen/meta/src/shared/legalize.rs
@@ -64,7 +64,6 @@ pub(crate) fn define(insts: &InstructionGroup, imm: &Immediates) -> TransformGro
     let f64const = insts.by_name("f64const");
     let fcopysign = insts.by_name("fcopysign");
     let fcvt_from_sint = insts.by_name("fcvt_from_sint");
-    let fcvt_from_uint = insts.by_name("fcvt_from_uint");
     let fneg = insts.by_name("fneg");
     let iadd = insts.by_name("iadd");
     let iadd_cin = insts.by_name("iadd_cin");
@@ -542,21 +541,12 @@ pub(crate) fn define(insts: &InstructionGroup, imm: &Immediates) -> TransformGro
         ],
     );
 
-    // Expansions for fcvt_from_{u,s}int for smaller integer types.
-    // These use expand and not widen because the controlling type variable for
-    // these instructions are f32/f64, which are legalized as part of the expand
+    // Expansion for fcvt_from_sint for smaller integer types.
+    // This use expand and not widen because the controlling type variable for
+    // this instruction is f32/f64, which is legalized as part of the expand
     // group.
     for &dest_ty in &[F32, F64] {
         for &src_ty in &[I8, I16] {
-            let bound_inst = fcvt_from_uint.bind(dest_ty).bind(src_ty);
-            expand.legalize(
-                def!(a = bound_inst(b)),
-                vec![
-                    def!(x = uextend.I32(b)),
-                    def!(a = fcvt_from_uint.dest_ty(x)),
-                ],
-            );
-
             let bound_inst = fcvt_from_sint.bind(dest_ty).bind(src_ty);
             expand.legalize(
                 def!(a = bound_inst(b)),
--- a/cranelift-codegen/meta/src/shared/legalize.rs
+++ b/cranelift-codegen/meta/src/shared/legalize.rs
@@ -542,7 +542,7 @@ pub(crate) fn define(insts: &InstructionGroup, imm: &Immediates) -> TransformGro
     );
 
     // Expansion for fcvt_from_sint for smaller integer types.
-    // This use expand and not widen because the controlling type variable for
+    // This uses expand and not widen because the controlling type variable for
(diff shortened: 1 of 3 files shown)


cranelift-codegen/meta/src/shared/legalize.rs
```diff
@@ -449,21 +448,12 @@ pub(crate) fn define(insts: &InstructionGroup, imm: &Immediates) -> TransformGro
         }
     }
 
-    // Expansions for fcvt_from_{u,s}int for smaller integer types.
-    // These use expand and not widen because the controlling type variable for
-    // these instructions are f32/f64, which are legalized as part of the expand
+    // Expansion for fcvt_from_sint for smaller integer types.
+    // This use expand and not widen because the controlling type variable for
```
sunfishcode: With the subject of this sentence singular, "use" should change to "uses".

cranelift-codegen/meta/src/shared/legalize.rs
```diff
@@ -449,21 +448,12 @@ pub(crate) fn define(insts: &InstructionGroup, imm: &Immediates) -> TransformGro
...
     for &dest_ty in &[F32, F64] {
         for &src_ty in &[I8, I16] {
-            let bound_inst = fcvt_from_uint.bind(dest_ty).bind(src_ty);
-            expand.legalize(
-                def!(a = bound_inst(b)),
-                vec![
-                    def!(x = uextend.I32(b)),
-                    def!(a = fcvt_from_uint.dest_ty(x)),
-                ],
-            );
```
sunfishcode: To make sure I understand, this code isn't actually wrong; it's just not being used right now, right?

cranelift-codegen/src/isa/x86/enc_tables.rs
```diff
@@ -405,12 +405,21 @@ fn expand_fcvt_from_uint(
...
-        return;
+    // Conversion from an unsigned int smaller than 64bit is easy on x86-64.
+    match xty {
+        ir::types::I8 | ir::types::I16 => {
+            let wide = pos.ins().uextend(ir::types::I64, x);
+            pos.func.dfg.replace(inst).fcvt_from_sint(ty, wide);
+            return;
+        }
+        ir::types::I32 => {
+            // TODO: This should be guarded by an ISA check.
```
sunfishcode: This comment applies to the `I8` and `I16` cases too, and then those and the `I32` cases are the same code, so we can merge them into one match arm.

cranelift-codegen/meta/src/shared/legalize.rs
bjorn3: That's right. (at least for x86, maybe other archs do use it) I can put it back if you want.

cranelift-codegen/src/isa/x86/enc_tables.rs
bjorn3: I dont think that comment applies to `I8` and I16`, as those have to be `uextend`ed on 32bit x86 too. Though in that case they can be extended to `I32`. I wanted to merge both arms, but wasnt sure how to change the comment to make clear it only applies to `I32`.

cranelift-codegen/meta/src/shared/legalize.rs
sunfishcode: Ok, cool. I don't think it's necessary to keep it right now. I expect we'll eventually want to refactor the x86 code to make even more of what `expand_fcvt_from_uint` does available to other architectures, but that's a bigger change and we can figure that out later.

cranelift-codegen/src/isa/x86/enc_tables.rs
sunfishcode: We'll still want an ISA check for `I8` and `I16`, because we can't expand them to  `I64` operations on x86-32.

bjorn3:
Ready for review

cranelift-codegen/meta/src/shared/legalize.rs
```diff
@@ -542,21 +541,12 @@ pub(crate) fn define(insts: &InstructionGroup, imm: &Immediates) -> TransformGro
...
-    // These use expand and not widen because the controlling type variable for
-    // these instructions are f32/f64, which are legalized as part of the expand
+    // Expansion for fcvt_from_sint for smaller integer types.
+    // This uses expand and not widen because the controlling type variable for
+    // this instruction is f32/f64, which is legalized as part of the expand
     // group.
     for &dest_ty in &[F32, F64] {
         for &src_ty in &[I8, I16] {
-            let bound_inst = fcvt_from_uint.bind(dest_ty).bind(src_ty);
-            expand.legalize(
```
bnjbvr: Out of curiosity, is it worth removing the legalization? The fact it doesn't apply on x86 doesn't make it less valuable for other ISAs, does it?

bjorn3: https://github.com/CraneStation/cranelift/pull/997#discussion_r321977898

PR #999, opened 2019-09-07
legalizer/split.rs: simplify_branch_arguments: use SmallVec instead o…

julian-seward1:
…f Vec

This function is responsible for 8.5% of all heap allocation (calls) in CL.
This change avoids almost all of them by using a SmallVec::<[Value; 32]>
instead.  Dynamic instruction count falls by 0.25%.  The fixed size of 32 was
arrived at after profiling with fixed sizes of 1, 2, 4, 8, 16, 32, 64 and 128.
32 is as high as I can push it without the instruction count starting to creep
up again, and gets almost all the block-reduction win of 64 and 128.

### DIFF
--- a/cranelift-codegen/Cargo.toml
+++ b/cranelift-codegen/Cargo.toml
@@ -21,6 +21,7 @@ hashmap_core = { version = "0.1.9", optional = true }
 target-lexicon = "0.8.1"
 log = { version = "0.4.6", default-features = false }
 serde = { version = "1.0.94", features = ["derive"], optional = true }
+smallvec = { version = "0.6.10" }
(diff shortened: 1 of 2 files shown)


lars-t-hansen:
I would expect that for large functions, esp as results from aggressive inlining, 32 is not uncommon.  And with the new splitter, the number of ebb parameters will increase.  My experience with split-across-calls was actually that the increase was significant.  If we create profiling code for this it would be nice either to check it in or to stash it someplace where it can be found.  Indeed, we keep talking about adding telemetry to check our assumptions in the various compilers.  We have to be careful and summarize/bucket/anonymize any data we gather or we're probably doing something that looks like a fingerprint, but it can be done.

PR #1000, opened 2019-09-07
EbbHeaderBlockData::predecessors: use SmallVec instead of Vec

julian-seward1:
Allocations associated with pushes to EbbHeaderBlockData::predecessors account
for 4.9% of all heap allocation (calls) in CL.  This change avoids almost all
of them by changing it to be a SmallVec<[PredBlock; 4]>.  Dynamic instruction
count falls by 0.15%.


bnjbvr:
(Yay, PR #1000 ! :tada:)

PR #962, opened 2019-08-31
[codegen] add encodings for isub borrow variants

ryzokuken:
Add encodings for isub borrow variants (isub_bout, isub_bin, isub_borrow) for x86_32, enabling the legalization for isub.i64 to work.

Bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1576675
Bug: https://github.com/CraneStation/cranelift/issues/765

/cc @bnjbvr @wingo @caitp

PR #1005, opened 2019-09-09
[codegen] reintroduce support for carry and borrow instructions in RI…

ryzokuken:
…SC ISAs

Reintroduce support for iadd carry variants and isub borrow variants for
RISC ISAs which had been removed in
https://github.com/CraneStation/cranelift/pull/961 and
https://github.com/CraneStation/cranelift/pull/962 because of the lack
of a proper flags register in RISC architectures.

/cc @wingo @caitp

@sunfishcode @bnjbvr there's a slight problem. x86 and RISC seem to be reusing both `narrow` and `expand` so I cannot choose between the two different versions of the instructions (eg: since both x86 and RISC use narrow to expand `iadd.i64`, one would need to choose between `iadd_c*` and `iadd_ifc*`). I believe creating a specialized legalization file for `riscv` in `meta` and adding the  legalization specifically should help.

### DIFF
--- a/cranelift-codegen/meta/src/shared/instructions.rs
+++ b/cranelift-codegen/meta/src/shared/instructions.rs
@@ -1879,10 +1879,16 @@ pub(crate) fn define(
     let a = &operand("a", iB);
     let x = &operand("x", iB);
     let y = &operand("y", iB);
-    let c_in = &operand_doc("c_in", iflags, "Input carry flag");
-    let c_out = &operand_doc("c_out", iflags, "Output carry flag");
-    let b_in = &operand_doc("b_in", iflags, "Input borrow flag");
-    let b_out = &operand_doc("b_out", iflags, "Output borrow flag");
+
+    let c_in = &operand_doc("c_in", b1, "Input carry flag");
+    let c_out = &operand_doc("c_out", b1, "Output carry flag");
+    let b_in = &operand_doc("b_in", b1, "Input borrow flag");
+    let b_out = &operand_doc("b_out", b1, "Output borrow flag");
+
+    let c_if_in = &operand("c_in", iflags);
+    let c_if_out = &operand("c_out", iflags);
+    let b_if_in = &operand("b_in", iflags);
+    let b_if_out = &operand("b_out", iflags);
 
     ig.push(
         Inst::new(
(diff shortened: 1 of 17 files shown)


bnjbvr:
Thanks for looking into this! Our legalization groups have historically grown in a very ad-hoc way, and this is proof they don't quite make sense. We could split them into smaller ones, and chain them together to keep the things working almost the same way. It would be interesting to make a legalization group that's not specific to RISC but to platforms which don't have flags, for one thing.

cranelift-codegen/meta/src/shared/instructions.rs
```diff
@@ -1904,6 +1910,26 @@ pub(crate) fn define(
         .operands_out(vec![a]),
     );
 
+    ig.push(
+        Inst::new(
+            "iadd_ifcin",
+            r#"
+        Add integers with carry in.
+
+        Same as `iadd` with an additional carry input. Computes:
```
bnjbvr: nit: for all the comments of these instructions, can you add "flags" before "input" or "output", as needed, as you did for subtracts, please?

cranelift-codegen/meta/src/shared/instructions.rs
```diff
@@ -1925,6 +1951,27 @@ pub(crate) fn define(
         .operands_out(vec![a, c_out]),
     );
 
+    ig.push(
+        Inst::new(
+            "iadd_ifcout",
+            r#"
+        Add integers with carry out.
+
+        Same as `iadd` with an additional carry output.
```
bnjbvr: ditto

cranelift-codegen/meta/src/shared/instructions.rs
```diff
@@ -1946,6 +1993,27 @@ pub(crate) fn define(
         .operands_out(vec![a, c_out]),
     );
 
+    ig.push(
+        Inst::new(
+            "iadd_ifcarry",
+            r#"
+        Add integers with carry in and out.
+
+        Same as `iadd` with an additional carry input and output.
```
bnjbvr: ditto

bnjbvr:
(Actually added those for you! If the build passes, I'll squash everything.)

PR #1008, opened 2019-09-09
Fix #796: Enable basic-blocks by default.

### LINKED ISSUE #796: EBBs aren't working out all that well

nbp:
This patch enables by default the basic-blocks feature set.

It is enabled in the top-level directory, in order to test locally, as well as in sub-projects, such that any project which depends on Cranelift components will also benefit from it.

This PR depends on #991 and #1007 to be merged to get a green CI.
(cc @sstangl)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -46,7 +46,7 @@ indicatif = "0.11.0"
 walkdir = "2.2"
 
 [features]
-default = ["disas", "wasm", "cranelift-codegen/all-arch"]
+default = ["disas", "wasm", "cranelift-codegen/all-arch", "basic-blocks"]
 disas = ["capstone"]
 wasm = ["wabt", "cranelift-wasm"]
 basic-blocks = ["cranelift-codegen/basic-blocks", "cranelift-frontend/basic-blocks",
--- a/cranelift-codegen/Cargo.toml
+++ b/cranelift-codegen/Cargo.toml
@@ -32,7 +32,7 @@ smallvec = { version = "0.6.10" }
 cranelift-codegen-meta = { path = "meta", version = "0.44.0", default-features = false }
 
 [features]
-default = ["std"]
+default = ["std", "basic-blocks"]
 
 # The "std" feature enables use of libstd. The "core" feature enables use
 # of some minimal std-like replacement libraries. At least one of these two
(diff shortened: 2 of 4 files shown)


nbp:
Ok, all is green for now. I will try to re-run the benchmarks on Monday, and report the results here.
If the benchmark results are still in favor of these changes, I will merge them.

nbp:
Running the benchmarks does not seems in favor of landing this patch yet:

(before, after):
```
# mode=cranelift+cranelift, runs=21, problem size=default
# Lower score is better
box2d           964     1047    0.921
bullet          1475    1366    1.08 
conditionals    392     415     0.945
copy            745     760     0.98 
corrections     1042    1097    0.95 
fannkuch        2719    2887    0.942
fasta           889     878     1.013
fib             640     754     0.849
ifs             210     200     1.05 
binarytrees     4517    4584    0.985
memops          1615    2084    0.775
primes          1244    1168    1.065
raybench        2219    2308    0.961
rust-fannkuch   2772    2679    1.035
skinning        828     876     0.945
zlib            1505    1472    1.022
```

PR #1010, opened 2019-09-09
Fix #975 (Structure of cranelift-wasm/src/translation_utils.rs cause…

### LINKED ISSUE #975: Structure of cranelift-wasm/src/translation_utils.rs causes many pointless heap allocations
@lars-t-hansen, @bnjbvr, @sunfishcode 
Using CL to compile wasm from Spidermonkey causes translation_utils.rs to allocate heap blocks which are never read from.  For compiling wasm_lua_binarytrees.js, it causes at least 3034 blocks of average size 44 bytes to be allocated, initialised, and thrown away.

The culprit is `fn blocktype_to_type` and possibly its friends.  It looks like this:
```
pub fn blocktype_to_type(ty: wasmparser::TypeOrFuncType) -> WasmResult<ir::Type> {
    match ty {
        wasmparser::TypeOrFuncType::Type(ty) => type_to_type(ty),
        wasmparser::TypeOrFuncType::FuncType(_) => {
            wasm_unsupported!("multi-value block signature {:?}", ty);
        }
    }
}
```
`WasmResult` is a parameterisation of plain `Result`.  In the `FuncType` case, `wasm_unsupported!` formats the specified string, allocating the result on the heap, and wraps it up (indirectly) in `Err(..)`.  That all sounds reasonable, except that the use point in SM looks like this:
```
if let Ok(ty_cre) = blocktype_to_type(*ty) {
    builder.append_ebb_param(if_not, ty_cre);
}
```
so the string allocated for the `Err` return is simply thrown away.

The obvious fix is to change the return type of `blocktype_to_type` (and friends) to be `Option`-like rather than `Err`-like.  But then we lose the ability to provide a string describing a failure to callers that actually want that.  So I'm not sure what to do.

julian-seward1:
…s many pointless heap allocations)

This patch restricts the `Err(..)` return from `blocktype_to_type` to be
`Err(..)` only in the case where it really is an error to continue.  The three
use points of `blocktype_to_type` are changed actually check for an `Err(..)`
rather than silently ignoring it.  There are also cosmetic changes to
`type_to_type` and `tabletype_to_type`.

When compiling wasm_lua_binarytrees, this reduces the number of blocks
allocated by CL by 1.9%.  Instruction count falls by 0.1%.

Details:

* `type_to_type` and `tabletype_to_type`:

   - Added the function name in the failure message

   - No functional change for non-error cases

   - Push the `Ok(..)` to expression leaves, where it really applies.  This
     corrects the misleading impression that, in the case of an unsupported
     type, the function returns `Ok` wrapped around whatever
     `wasm_unsupported` returns.  It doesn't do that, but it certainly reads
     like that.  This assumes that the LLVM backend will do tail merging, so
     the generated code will be unchanged.

* `blocktype_to_type`:

  - Change return type from `WasmResult<ir::Type>` to `WasmResult<Option<ir::Type>>`

  - Manually inline the call to `type_to_type`, to make this function easier
    to read.

  - For the non-error case: map `TypeOrFuncType::Type(Type::EmptyBlockType)`
    to `Ok(None)` rather than `Err(..)`, since that's what all the call sites
    expect

  - For the error cases, add the function name in the failure messages

* cranelift-wasm/src/code_translator.rs

  - For the three uses of `blocktype_to_type`, use `?` to detect failures and
    drop out immediately, meaning that the code will no longer silently ignore
    errors.

### DIFF
--- a/cranelift-wasm/src/translation_utils.rs
+++ b/cranelift-wasm/src/translation_utils.rs
@@ -115,35 +115,43 @@ pub struct Memory {
 
 /// Helper function translating wasmparser types to Cranelift types when possible.
 pub fn type_to_type(ty: wasmparser::Type) -> WasmResult<ir::Type> {
-    Ok(match ty {
-        wasmparser::Type::I32 => ir::types::I32,
-        wasmparser::Type::I64 => ir::types::I64,
-        wasmparser::Type::F32 => ir::types::F32,
-        wasmparser::Type::F64 => ir::types::F64,
-        ty => wasm_unsupported!("unsupported wasm type {:?}", ty),
-    })
+    match ty {
+        wasmparser::Type::I32 => Ok(ir::types::I32),
+        wasmparser::Type::I64 => Ok(ir::types::I64),
+        wasmparser::Type::F32 => Ok(ir::types::F32),
+        wasmparser::Type::F64 => Ok(ir::types::F64),
+        ty => wasm_unsupported!("type_to_type: wasm type {:?}", ty),
+    }
 }
 
 /// Helper function translating wasmparser possible table types to Cranelift types when possible,
 /// or None for Func tables.
 pub fn tabletype_to_type(ty: wasmparser::Type) -> WasmResult<Option<ir::Type>> {
-    Ok(match ty {
-        wasmparser::Type::I32 => Some(ir::types::I32),
-        wasmparser::Type::I64 => Some(ir::types::I64),
-        wasmparser::Type::F32 => Some(ir::types::F32),
-        wasmparser::Type::F64 => Some(ir::types::F64),
-        wasmparser::Type::AnyFunc => None,
-        ty => wasm_unsupported!("unsupported table wasm type {:?}", ty),
-    })
+    match ty {
+        wasmparser::Type::I32 => Ok(Some(ir::types::I32)),
+        wasmparser::Type::I64 => Ok(Some(ir::types::I64)),
+        wasmparser::Type::F32 => Ok(Some(ir::types::F32)),
+        wasmparser::Type::F64 => Ok(Some(ir::types::F64)),
+        wasmparser::Type::AnyFunc => Ok(None),
+        ty => wasm_unsupported!("tabletype_to_type: table wasm type {:?}", ty),
+    }
 }
 
 /// Helper function translating wasmparser block signatures to Cranelift types when possible.
-pub fn blocktype_to_type(ty: wasmparser::TypeOrFuncType) -> WasmResult<ir::Type> {
-    match ty {
-        wasmparser::TypeOrFuncType::Type(ty) => type_to_type(ty),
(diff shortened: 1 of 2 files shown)


cranelift-wasm/src/translation_utils.rs
```diff
@@ -115,35 +115,43 @@ pub struct Memory {
...
+        wasmparser::TypeOrFuncType::Type(ty) => match ty {
+            wasmparser::Type::I32 => Ok(Some(ir::types::I32)),
+            wasmparser::Type::I64 => Ok(Some(ir::types::I64)),
+            wasmparser::Type::F32 => Ok(Some(ir::types::F32)),
+            wasmparser::Type::F64 => Ok(Some(ir::types::F64)),
+            wasmparser::Type::EmptyBlockType => Ok(None),
+            ty => wasm_unsupported!("blocktype_to_type: unsupported type {:?}", ty),
+        },
+        wasmparser::TypeOrFuncType::FuncType(_) => wasm_unsupported!(
+            "blocktype_to_type: unsupported multi-value block signature {:?}",
```
bnjbvr: light nit: feel free to remove the "unsupported" word in all the error messages: it's redundant with the one that's added in front of errors of the Unsupported variant.

nox:
It is *extremely* confusing that `wasm_unsupported` return an `Err(_)`, I would move the `Err(…)` part out of it and properly write it down whenever I use `wasm_unsupported`. Note that there is no precedent at all for doing that in any Rust project I know of.

julian-seward1:
@nox: that might be confusing, but at least this patch actually makes it less confusing than it was.  Before, the code was returning `Ok( wasm_unsupported!(..) )`, which is even more bizarre.

nox:
Yes, I understand that.

I'm saying that `Err(wasm_unsupported!(…))` requires way less knowledge to understand the code. I was pinged in the original issue about the `blocktype_to_type` function so I looked at the fix out of curiosity, and I was that close to filing an issue about how `blocktype_to_type` doesn't even return `Err(_)` anymore.

nox:
It can also hurt expressiveness, for example if you have a function which returns `Result<T, Whatever>` or `Option<T>` and you want a `WasmResult` where the error is `WasmError::Unsupported(_)`:

```rust
let thing = returns_option().ok_or_else(|| wasm_unsupported!("whatever"))?;
let thing = returns_result().map_err(|e| wasm_unsupported!("whatever: {}", e))?;
```

julian-seward1:
@nox, I agree, it would be cleaner if `wasm_unimplemented!` only produced a value `Err(..)`, rather than expanding to `return Err(..)` as at present.  Then we wouldn't have to reason about control flow effects at its use points, and (IIUC) it would then work right for your `let thing = ` examples above.

nox:
Just to be clear, I’m saying the macro should produce only the WasmError value, it shouldn’t wrap it at all not early return it.

PR #1013, opened 2019-09-09
Add x86 `icmp eq` for SIMD

abrown:
Beyond adding encodings for the very specific `eq` immediate of `icmp`, this also optimizes `vconst` when the immediate value is all ones or all zeroes. These changes may seem unrelated but serve to highlight the difficulty using predicates to make very specific encodings--see https://github.com/CraneStation/cranelift/pull/1012.


cranelift-codegen/meta/src/isa/x86/recipes.rs
```diff
@@ -2878,6 +2890,20 @@ pub(crate) fn define<'shared>(
             ),
     );
 
+    recipes.add_template_recipe(
+        EncodingRecipeBuilder::new("icscc_fpr", f_int_compare, 1)
+            .operands_in(vec![fpr, fpr])
+            .operands_out(vec![0])
+            // .inst_predicate(has_eq_condition_code)
```
abrown: It doesn't seem to matter if I place the instruction predicate here on on the `EncodingBuilder` in `encodings.rs`...

bnjbvr:
Selecting an alternative reviewer for this; Sean, feel free to bounce to somebody else if you're not comfortable with reviewing this.

cranelift-codegen/meta/src/isa/x86/recipes.rs
sstangl: What does this comment mean -- is the `EncodingRecipe::inst_predicate()` being ignored somewhere?

cranelift-codegen/meta/src/cdsl/instructions.rs
```diff
@@ -622,6 +628,9 @@ pub enum FormatPredicateKind {
 
     /// Is the referenced data object colocated?
     IsColocatedData,
+
+    /// Does the operation have a specific condition code?
+    HasConditionCode(&'static str),
```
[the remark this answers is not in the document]
abrown: Yes, but I would have to pull those enums out of `codegen` and put them in a crate accessible to both `codegen` and `codegen-meta`. Which I actually think is probably the right thing to do here but I was being lazy. What do you think--a new `codegen-shared` crate?

cranelift-codegen/meta/src/isa/x86/recipes.rs
abrown: I guess at the time I didn't understand (and wanted feedback on) why sometimes the predicates are attached where the instruction is encoded and sometimes on the recipe. I now realize that they just need to get to the `EncodingBuilder` in some way and I think the better place is in `encodings.rs`. I will remove this.

cranelift-codegen/meta/src/isa/x86/encodings.rs
```diff
@@ -1917,6 +1933,26 @@ pub(crate) fn define(
...
+        let instruction = vconst.bind_vector_from_lane(ty, sse_vector_size);
+
+        let is_zero_128bit = InstructionPredicate::new_is_all_zeroes_128bit(f_unary_uimm128, "imm");
+        let template = rec_vconst_opt.nonrex().opcodes(vec![0x66, 0x0f, 0xef]); // PXOR from SSE2
+        e.enc_32_64_func(instruction.clone(), template, |builder| {
+            builder.inst_predicate(is_zero_128bit)
+        });
+
+        let is_ones_128bit = InstructionPredicate::new_is_all_ones_128bit(f_unary_uimm128, "imm");
+        let template = rec_vconst_opt.nonrex().opcodes(vec![0x66, 0x0f, 0x74]); // PCMPEQB from SSE2
```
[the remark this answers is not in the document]
abrown: I agree. Opened https://github.com/CraneStation/cranelift/issues/1051.

cranelift-codegen/meta/src/isa/x86/legalize.rs
```diff
@@ -328,6 +328,7 @@ pub(crate) fn define(shared: &mut SharedDefinitions, x86_instructions: &Instruct
             def!(y = splat_any8x16(x)),
             vec![
                 def!(a = scalar_to_vector(x)), // move into the lowest 8 bits of an XMM register
+                // TODO replace the following two instructions with `vconst(0)` when this is possible; currently vconst receives a Constant, not a Uimm128
```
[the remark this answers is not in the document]
abrown: Opened https://github.com/CraneStation/cranelift/issues/1052 and added in with a rebased fixup.

cranelift-codegen/meta/src/isa/x86/recipes.rs
```diff
@@ -2449,6 +2449,18 @@ pub(crate) fn define<'shared>(
             ),
     );
 
+    recipes.add_template_recipe(
+        EncodingRecipeBuilder::new("vconst_opt", f_unary_const, 1)
```
bnjbvr: May I suggest to use a different name for this, please? "opt" is ambiguous in that it could mean either "optional" or "optimized". (And both would be a valid interpretation here. :))

cranelift-codegen/shared/Cargo.toml
```diff
@@ -0,0 +1,11 @@
+[package]
+authors = ["The Cranelift Project Developers"]
+name = "cranelift-codegen-shared"
```
bnjbvr: It's a thing \o/

We might need to update the top-level `./publish-all.sh` script around lines 18 and 41.

Issue #941, opened 2019-08-26
[Filetests] Use `make_exec` instead of region::protect for writes


bnjbvr:
Please use `let code_page = code_page.make_exec().map_err(|e| e.to_string())?;` instead. Maybe switch [`cranelift-simplejit`](https://github.com/CraneStation/cranelift/blob/master/cranelift-simplejit/src/memory.rs) to that method too.

_Originally posted by @bjorn3 in https://github.com/CraneStation/cranelift/pull/940_

bnjbvr:
Fixed by https://github.com/CraneStation/cranelift/pull/1017, thanks a lot @walac !

bjorn3:
@walac said he would change simplejit in a follow-up, so it is not yet completely fixed @bnjbvr.

walac:
A few questions about the implementation in `simple-jit`:
* `memmap` crate is used only for POSIX systems (except SELinux). Is there a technical reason not to use `memmap` in Windows and POSIX? If yes, I understand that we need to keep using `code::protect` in those systems.
* As `MmapMut::make_exec` and `MmapMut::make_read_only` methods return errors, we need either  to change the public interface (and the all the calls) of `Memory::set_readonly` and `Memory::set_readable_and_executable` or *panic* in case of errors. Which one would be preferable?

walac:
^ @bnjbvr

bnjbvr:
- memmap seems to be only used for SELinux, right? I don't know if we want to use the memmap crate in general; maybe we wouldn't need to rely on `region` directly, then... Let's ask @sunfishcode; I'm not very clear on this code.
- as a general principle, yielding errors is more user-friendly, since the user can decide what to do with this error in return (panic, or try something different, etc.).

PR #1017, opened 2019-09-10
Replace region::protect by make_exec in cranelift-filetests

walac:
Ref: issue #941 

I will work on simple-jit in a follow-up PR.


abrown:
If you run `rustfmt` on this the checks will likely succeed. Otherwise, looks good to me!

walac:
> If you run `rustfmt` on this the checks will likely succeed. Otherwise, looks good to me!

Done :)

abrown:
Thanks!

PR #1020, opened 2019-09-11
Minor speedup tuning for SecondaryMap

julian-seward1:
The `SecondaryMap` abstraction -- basically, resize-on-demand arrays with a
default value -- is very hot in Cranelift.  This small patch is the result of
many profiling runs.  It makes two changes:

* `fn index_mut` is changed to be `#[inline(always)]`, based on profile data.

* `fn index` and `fn index_mut` call `self.elems.resize()` directly, rather
  than via `self.resize()`.  The point of this is not to improve performance.
  Rather, it ensures that the public functions for `SecondaryMap` do not call
  each other.  When public interface functions call each other, it becomes
  difficult to interpret profiling results, because it's harder to see what
  fraction of costs for `SecondaryMap` as a whole come from outside the
  module, and what fraction is the result of "internal" calls to the external
  interface.

The overall result, for wasm_lua_binarytrees, is a 1.4% reduction in
instruction count for the compiler, and a 2.2% reduction in loads/stores.

### DIFF
--- a/cranelift-entity/src/map.rs
+++ b/cranelift-entity/src/map.rs
@@ -64,16 +64,19 @@ where
     }
 
     /// Get the element at `k` if it exists.
+    #[inline(always)]
     pub fn get(&self, k: K) -> Option<&V> {
         self.elems.get(k.index())
     }
 
     /// Is this map completely empty?
+    #[inline(always)]
     pub fn is_empty(&self) -> bool {
         self.elems.is_empty()
     }
 
     /// Remove all entries from this map.
+    #[inline(always)]
     pub fn clear(&mut self) {
         self.elems.clear()
     }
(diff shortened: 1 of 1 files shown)


sstangl:
Another option that to my understanding effectively adds `#[inline]` to *every* function is to tell Cargo to perform link-time optimization for release builds. LLVM will then greedily inline absolutely anything that it can -- and it tends to do a very good job at that. On personal projects this is a large performance improvement.

Here's an example of LTO in a `Cargo.toml`:

```toml
[profile.release]
opt-level = 3
lto = 'fat'
codegen-units = 1
```

julian-seward1:
@bjorn3 Plain `#[inline]` is only a suggestion to the compiler.  It may be ignored.  In this case I know from the structure of `index_mut` and the context in which it is typically used, that inlining will almost certainly be a win.  Hence `#[inline(always)]` so as to consistently get that win.

PR #1025, opened 2019-09-13
legalizer/split.rs: split_ebb_params: use SmallVec instead of Vec

julian-seward1:
This function is responsible for 2.2% of all heap allocation (calls) in CL.
This change avoids most of them (90%, maybe) by using a SmallVec::<[Value;
16]> instead.  The fixed size of 16 was arrived at after profiling with fixed
sizes of 1, 2, 4, 8, 16, 32 and 64.

### DIFF
--- a/cranelift-codegen/src/legalizer/split.rs
+++ b/cranelift-codegen/src/legalizer/split.rs
@@ -131,19 +131,27 @@ fn split_any(
 }
 
 pub fn split_ebb_params(func: &mut ir::Function, cfg: &ControlFlowGraph, ebb: Ebb) {
-    let mut repairs = Vec::new();
     let pos = &mut FuncCursor::new(func).at_top(ebb);
+    let ebb_params = pos.func.dfg.ebb_params(ebb);
 
-    for (num, ebb_param) in pos
-        .func
-        .dfg
-        .ebb_params(ebb)
-        .to_vec()
-        .into_iter()
-        .enumerate()
+    // Add further splittable types here.
+    fn type_requires_splitting(ty: Type) -> bool {
+        ty == ir::types::I128
+    }
+
+    // A shortcut.  If none of the param types require splitting, exit now.  This helps because
+    // the loop below necessarily has to copy the ebb params into a new vector, so it's better to
+    // avoid doing so when possible.
+    if !ebb_params
+        .iter()
+        .any(|ebb_param| type_requires_splitting(pos.func.dfg.value_type(*ebb_param)))
     {
-        let ty = pos.func.dfg.value_type(ebb_param);
-        if ty != ir::types::I128 {
+        return;
+    }
+
+    let mut repairs = Vec::new();
+    for (num, ebb_param) in ebb_params.to_vec().into_iter().enumerate() {
+        if !type_requires_splitting(pos.func.dfg.value_type(ebb_param)) {
             continue;
         }
 


cranelift-codegen/src/legalizer/split.rs
```diff
@@ -134,11 +134,7 @@ pub fn split_ebb_params(func: &mut ir::Function, cfg: &ControlFlowGraph, ebb: Eb
     let mut repairs = Vec::new();
     let pos = &mut FuncCursor::new(func).at_top(ebb);
 
-    for (num, ebb_param) in pos
-        .func
-        .dfg
-        .ebb_params(ebb)
-        .to_vec()
+    for (num, ebb_param) in SmallVec::<[Value; 16]>::from_slice(pos.func.dfg.ebb_params(ebb))
```
bnjbvr: I've tried to see if we could just not materialize the vector here, by iterating over the ebb_params, but this causes ownerships issues with "pos" which aren't trivial to fix. So we need to materialize it unfortunately.

I've looked at this code a bit more for context, and this should only serve to break i128 into smaller types, as the "continue" statement a few lines below suggests. It's unfortunate that we create this vector even if there's no i128 types in the graph, as should be the case in our wasm test suite. A couple alternatives are available here:

## Filtering out

- take an iter out of the ebb_params
- filter_map the ebb_param only when their type is i128
- then collect into a SmallVec

Then, in the general wasm case, the filtered iterator will be empty, and *hopefully* rustc is good enough to not even allocate the SmallVec in this case.

## Skipping if there's no i128 type among ebb params

So adding a check at the top of this function by iterating over ebb_params and if !any of them has the i128 type, early return. This has better memory usage certainty to it, but is less convenient to maintain (e.g. if we ever add support for other types -- would i64 need to be handled here for x86 i686? we'll see...).

julian-seward1: I redid it using the "skipping" scheme.  It works out pretty nicely, avoiding all allocation in the non-splitting case, and with a reduced instruction count too, in my test case.

cranelift-codegen/src/legalizer/split.rs
```diff
@@ -131,19 +131,27 @@ fn split_any(
...
+        .iter()
+        .any(|ebb_param| type_requires_splitting(pos.func.dfg.value_type(*ebb_param)))
     {
-        let ty = pos.func.dfg.value_type(ebb_param);
-        if ty != ir::types::I128 {
+        return;
+    }
+
+    let mut repairs = Vec::new();
+    for (num, ebb_param) in ebb_params.to_vec().into_iter().enumerate() {
```
bjorn3: Maybe use

```rust
let split_ebb_params = ebb_params.iter().enumerate().filter_map(|(i, ebb_param)| {
    if type_requires_splitting(pos.func.dfg.value_type(ebb_param)) {
        Some(i, ebb_param)
    } else {
        None
    }
}).collect::<Vec<(usize, Ebb)>>();
for (num, ebb_param) in split_ebb_params.into_iter() {
    split_ebb_param(pos, ebb, num, ebb_param, Opcode::Iconcat, &mut repairs);
}
```

instead. That would make the allocated amount proportional to the amount of ebb params which require splitting, instead of proportional to the total amount of ebb params. You should then be able to remove the `if` above.

bnjbvr: This would still allocate two Vec plus call the tail function with an empty vector, right? This PR optimizes for minimizing work, which seems even nicer...

julian-seward1: @bnjbvr, yes, I think that's correct.

bjorn3: It seems collect for `Vec` doesnt actually allocate anything until the first item. `Vec::new` doesnt actually allocate anything.

bnjbvr: The filter_map way would work too, but it's slightly less explicit in what it does. In particular, there's still need for stack allocations for the two Vecs, and the spurious call at the function tail. I don't think these particularly matter much, but the initial early exit is nice in that it does minimal work in every case (esp. when there's no ebb params on the current block). No strong opinions here, so I suggest to let it as is.

PR #1028, opened 2019-09-13
Add verifier check to ensure each SIMD lane index is within bounds; fixes #1016

### LINKED ISSUE #1016: Verifier should throw for invalid SIMD lanes in CLIF
As discussed in https://github.com/CraneStation/cranelift/pull/943#discussion_r322835465, even though cranelift uses wasmparser to verify that the SIMD lanes passed to improvelane/extractlane are valid, it would be good to add this logic to cranelift's verifier to catch this error in cranelift producers other than cranelift-wasm (the module using wasmparser).

An example of an invalid SIMD lane would be `5` in `v1 = extractlane.i32x4 v0, 5`.

abrown:
@bnjbvr, I thought I would resolve this issue (and I thought I had) but the filetests are giving me weird errors:
```
FAIL filetests/isa/x86/insertlane-run.clif: run(%test_insertlane_):
Verifier errors:
- inst2: The lane 16 does not index into the type f32x4

FAIL filetests/isa/x86/extractlane-run.clif: run(%test_extractlane):
Verifier errors:
- inst3: The lane 32 does not index into the type f32x4

212 tests
```
I do not see invalid lane indexes in either https://github.com/CraneStation/cranelift/blob/master/filetests/isa/x86/extractlane-run.clif or https://github.com/CraneStation/cranelift/blob/master/filetests/isa/x86/insertlane-run.clif, though.

### DIFF
--- a/cranelift-codegen/src/verifier/mod.rs
+++ b/cranelift-codegen/src/verifier/mod.rs
@@ -1746,21 +1746,47 @@ impl<'a> Verifier<'a> {
     ) -> VerifierStepResult<()> {
         let inst_data = &self.func.dfg[inst];
 
-        // If this is some sort of a store instruction, get the memflags, else, just return.
-        let memflags = match *inst_data {
+        match *inst_data {
             ir::InstructionData::Store { flags, .. }
-            | ir::InstructionData::StoreComplex { flags, .. } => flags,
-            _ => return Ok(()),
-        };
-
-        if memflags.readonly() {
-            fatal!(
-                errors,
-                inst,
-                "A store instruction cannot have the `readonly` MemFlag"
-            )
-        } else {
-            Ok(())
+            | ir::InstructionData::StoreComplex { flags, .. } => {
+                if flags.readonly() {
+                    fatal!(
+                        errors,
+                        inst,
+                        "A store instruction cannot have the `readonly` MemFlag"
+                    )
+                } else {
+                    Ok(())
+                }
+            }
+            ir::InstructionData::ExtractLane {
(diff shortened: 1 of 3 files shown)


cranelift-codegen/src/verifier/mod.rs
```diff
@@ -1873,4 +1892,49 @@ mod tests {
 
         assert_err_with_msg!(errors, "instruction format");
     }
+
+    #[test]
+    fn bad_lane_index() {
```
bnjbvr: (light suggestion) I think a clif test would be nice and simple to detect this. For what it's worth, you can ask for expected failures in verifier tests.

abrown:
@bnjbvr I figured out what was wrong: the `ExtractLane` and `InsertLane` formats are used for other instructions (e.g. `x86_insertps`) which may do strange things to the lane field. I made the `match` specific to the `Extractlane` and `Insertlane` opcodes and moved the testing to a CLIF file.

cranelift-codegen/src/verifier/mod.rs
```diff
@@ -1814,7 +1840,8 @@ mod tests {
     use super::{Verifier, VerifierError, VerifierErrors};
     use crate::entity::EntityList;
     use crate::ir::instructions::{InstructionData, Opcode};
-    use crate::ir::Function;
+    use crate::ir::types::*;
+    use crate::ir::{Constant, Function};
```
bnjbvr: nit: probably some of these are unused now and would generate warnings when compiling tests

abrown: Ok, let me check before I merge.

PR #1029, opened 2019-09-14
fix reflink in CONTRIBUTING.md

vitiral:


bnjbvr:
Good catch, thank you!

Issue #1021, opened 2019-09-11
cranelift-wasm::FuncTranslator::{translate, translate_from_reader} signature change


pchickey:
In #671, the signature of `FuncTranslator::translate` was changed from taking a `&mut ir::Function` and returning `WasmResult<()>` to taking a `ir::Function` and returning `WasmResult<ir::Function>`:

https://github.com/CraneStation/cranelift/commit/4cd2c76c65cf71990d4ec9b9679c4755fafa75ed#diff-3a8aa8abcffafb563225730ebca50b04L60

The Function given as an argument is required to be empty, and only provides its `name` and `signature` to the builder inside `FuncTranslator`:

https://github.com/CraneStation/cranelift/commit/4cd2c76c65cf71990d4ec9b9679c4755fafa75ed#diff-3a8aa8abcffafb563225730ebca50b04R78

This was an understandable restriction before #671, but now that the function is no longer passed mutably, it makes more sense to just pass in the `name: ExternalName` and `signature: ir::Signature` as arguments to `translate` rather than passing an entire `ir::Function`.

The change itself is trivial, but it is a breaking API change, so I'd like to seek feedback before making a PR.

abrown:
Seems like this depends on the outcome of #1018? If the API is changed back to the way it was then this might not be applicable.

bnjbvr:
Closing since #671 has been reverted. Thanks for opening an issue though! It would have been a welcome improvement.

PR #1030, opened 2019-09-14
Revert #671

### LINKED ISSUE #1018: Removing FunctionBuilderContext from the API diminishes memory reusability
Sorry I'm quite late in the party, but I noticed a bad side-effect caused by the ergonomic improvement done in https://github.com/CraneStation/cranelift/pull/671.

It is now impossible to use the same FunctionBuilder with the same Function; one has to give ownership on an entire Function to the FunctionBuilder, causing more memmoves, and more memory churn for allocating new Functions.

In particular, in Spidermonkey, we reused the storage of the codegen Context's func as the Function mutable reference given to FunctionBuilder. Now it is impossible to do so, and requires creating a new ir::Function before calling FunctionBuilder. This causes new allocations that could be avoided if we could reuse the Context's func Function storage, and there can be thousands of functions in a single wasm module, so this is likely to show up in profiles.

Making the func in FunctionBuilder a mutable reference implies that the whole struct is bound to the Function's lifetime, which defeats the purpose of reusing the FunctionBuilderContext. Using an Option is not an option here (unintended pun) because it means each use will need an unwrap() call and runtime cost.

So while the previous state was not ideal in terms of exposing many public APIs, it allowed memory's reuse in the most optimal way. Alternatives I've considered would probably imply memmoves too, e.g. have two states for the FunctionBuilder, one with a function and one without it (using the type state pattern); switching from one state to the other would cause memory churn.

Any ideas here?

cc @sunfishcode @julian-seward1

XAMPPRocky:
As shown in #1018 , #671 caused significant regressions. This PR reverts that change (39e638af99dbe6537bc935bfb1a74669b62877b3).

resolves #1018 #1021

### DIFF
--- a/cranelift-frontend/src/frontend.rs
+++ b/cranelift-frontend/src/frontend.rs
@@ -20,22 +20,22 @@ use std::vec::Vec;
 /// In order to reduce memory reallocations when compiling multiple functions,
 /// `FunctionBuilderContext` holds various data structures which are cleared between
 /// functions, rather than dropped, preserving the underlying allocations.
-struct FunctionBuilderContext {
+pub struct FunctionBuilderContext {
     ssa: SSABuilder,
     ebbs: SecondaryMap<Ebb, EbbData>,
     types: SecondaryMap<Variable, Type>,
 }
 
 /// Temporary object used to build a single Cranelift IR `Function`.
-pub struct FunctionBuilder {
+pub struct FunctionBuilder<'a> {
     /// The function currently being built.
     /// This field is public so the function can be re-borrowed.
-    pub func: Function,
+    pub func: &'a mut Function,
 
     /// Source location to assign to all new instructions.
     srcloc: ir::SourceLoc,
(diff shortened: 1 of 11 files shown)


sunfishcode:
I don't object here. I believe there's more we can do with the `FunctionBuilderContext` API, but I'm ok reverting this change for now to recover the regression.

XAMPPRocky:
@sunfishcode I agree, I think that a optimal solution will take enough time to merit reverting for now. I attempted a refactor last weekend but couldn't find a satisfactory implementation.

bnjbvr:
Thanks @XAMPPRocky, and sorry we didn't take a more closed look at your PR before merging this! Your work was still valuable and appreciated here.

For what it's worth, I've also tried a couple things to make the API nice and clean, the final state I ended up with resembled a revert as you've done here. There might be other ways to keep this tidy without causing the memory usage regression, though.

PR #1031, opened 2019-09-14
Replaced instances of SparseSet with EntitySet

### LINKED ISSUE #858: SparseSet is worse than EntitySet
Reading the code of `cranelift-entity`, I noticed that `SperseSet<T>` is implemented using a `SparseMap<T, T>`.  Such `SparseMap<T, T>` stores a `SecondaryMap<T, u32>` which is a dense map of u32 (32 bits per key), where as an `EntitySet<T>` would be a dense map of bits (1 bit per key)

We should replace `SparseSet` by `EntitySet` everywhere, or change the alias such that we default to the `EntitySet` implementation instead of the `SparseMap`.

XAMPPRocky:
resolves #858

### DIFF
--- a/cranelift-bforest/src/pool.rs
+++ b/cranelift-bforest/src/pool.rs
@@ -83,7 +83,7 @@ impl<F: Forest> NodePool<F> {
         NodeData<F>: fmt::Display,
         F::Key: fmt::Display,
     {
-        use crate::entity::SparseSet;
+        use crate::entity::EntitySet;
         use core::borrow::Borrow;
         use core::cmp::Ordering;
         use std::vec::Vec;
@@ -94,7 +94,13 @@ impl<F: Forest> NodePool<F> {
             assert!(size > 0, "Root must have more than one sub-tree");
         }
 
-        let mut done = SparseSet::new();
+        let mut done = match self[node] {
+            NodeData::Inner { size, .. } | NodeData::Leaf { size, .. } => {
+                EntitySet::with_capacity(size.into())
+            }
+            _ => EntitySet::new(),
+        };
+
         let mut todo = Vec::new();
 
         // Todo-list entries are:
@@ -104,11 +110,7 @@ impl<F: Forest> NodePool<F> {
         todo.push((None, node, None));
 
         while let Some((lkey, node, rkey)) = todo.pop() {
-            assert_eq!(
-                done.insert(node),
-                None,
-                "Node appears more than once in tree"
-            );
+            assert!(done.insert(node), "Node appears more than once in tree");
             let mut lower = lkey;
 
             match self[node] {
--- a/cranelift-codegen/src/topo_order.rs
+++ b/cranelift-codegen/src/topo_order.rs
@@ -1,7 +1,7 @@
 //! Topological order of EBBs, according to the dominator tree.
 
 use crate::dominator_tree::DominatorTree;
-use crate::entity::SparseSet;
+use crate::entity::EntitySet;
 use crate::ir::{Ebb, Layout};
 use std::vec::Vec;
 
@@ -19,7 +19,7 @@ pub struct TopoOrder {
     next: usize,
 
     /// Set of visited EBBs.
-    visited: SparseSet<Ebb>,
+    visited: EntitySet<Ebb>,
 
     /// Stack of EBBs to be visited next, already in `visited`.
     stack: Vec<Ebb>,
@@ -31,7 +31,7 @@ impl TopoOrder {
         Self {
             preferred: Vec::new(),
             next: 0,
-            visited: SparseSet::new(),
+            visited: EntitySet::new(),
             stack: Vec::new(),
         }
     }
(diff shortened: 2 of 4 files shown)


cranelift-codegen/src/verifier/flags.rs
```diff
@@ -50,12 +50,12 @@ impl<'a> FlagsVerifier<'a> {
...
         // List of EBBs that need to be processed. EBBs may be re-added to this list when we detect
         // that one of their successor blocks needs a live-in flags value.
-        let mut worklist = SparseSet::new();
+        let mut worklist = EntitySet::new();
         for ebb in self.func.layout.ebbs() {
             worklist.insert(ebb);
         }
 
-        while let Some(ebb) = worklist.pop() {
+        for ebb in worklist.keys() {
```
bjorn3: This changes behavior. Before only the ebbs which were in func.layout or were inserted into the worklist at https://github.com/XAMPPRocky/cranelift/blob/f3cb557807e2fa931c235172bba17dd8e21fa613/cranelift-codegen/src/verifier/flags.rs#L66 were visited. Now all ebbs in the range ebb0..the ebb in func.layout with the highest ebb index is visited. This causes extra ebbs to be visited, and it misses the ebbs inserted at https://github.com/XAMPPRocky/cranelift/blob/f3cb557807e2fa931c235172bba17dd8e21fa613/cranelift-codegen/src/verifier/flags.rs#L66 with index higher than the highest in func.layout to be skipped.

cranelift-bforest/src/pool.rs
```diff
@@ -94,7 +94,7 @@ impl<F: Forest> NodePool<F> {
             assert!(size > 0, "Root must have more than one sub-tree");
         }
 
-        let mut done = SparseSet::new();
+        let mut done = EntitySet::new();
```
bjorn3: We already know the amount of nodes at this point, so maybe `.resize` it before doing anything with it, to prevent having to copy all data over when it is resized inside the `.insert()` call.

cranelift-codegen/src/verifier/flags.rs
```diff
@@ -50,12 +50,12 @@ impl<'a> FlagsVerifier<'a> {
     fn check(&mut self, errors: &mut VerifierErrors) -> VerifierStepResult<()> {
         // List of EBBs that need to be processed. EBBs may be re-added to this list when we detect
         // that one of their successor blocks needs a live-in flags value.
-        let mut worklist = SparseSet::new();
+        let mut worklist = EntitySet::new();
```
bjorn3: Same here.

cranelift-codegen/src/topo_order.rs
```diff
@@ -73,7 +73,7 @@ impl TopoOrder {
                     // We have the next EBB in the preferred order.
                     self.next += 1;
                     // Push it along with any non-visited dominators.
-                    while self.visited.insert(ebb).is_none() {
+                    while self.visited.insert(ebb) {
```
bjorn3: Maybe resize at the start of this function.

cranelift-entity/src/set.rs
```diff
@@ -93,6 +103,17 @@ where
         self.elems[index / 8] |= 1 << (index % 8);
         result
     }
+
+    /// Removes and returns the last `K` from the set if any are present.
+    pub fn pop(&mut self) -> Option<K> {
+        if let Some(k) = self.keys().next_back() {
```
bjorn3: Tgis doesnt give the last set key, but the key at `self.len * 8`, which may have already been removed.

XAMPPRocky: How would you suggest to get the last key?

bjorn3: Search last item of `self.elems` for last set bit. If the resulting u8 is zero, pop it. Not sure if it is correct. In any case you will need to decrement `self.len` afterwards.

XAMPPRocky: @bjorn3 Okay, I've reimplemented and added a test to make sure it works.

XAMPPRocky:
Also cc @nbp

cranelift-entity/src/set.rs
```diff
@@ -93,6 +103,21 @@ where
         self.elems[index / 8] |= 1 << (index % 8);
         result
     }
+
+    /// Removes and returns the entity from the set if it exists.
+    pub fn pop(&mut self) -> bool {
```
bjorn3: Suppose entries 5 and 7 exist. The len will be 8 at first. When calling pop, entry 7 is removed and the new len is 7. When calling pop again, entry 6 is examined. Because it is not existing pop returns false, even though there is an existing entry.

XAMPPRocky:
@bjorn3 Okay I've updated the implementation to handle and test the scenario you brought up, this should be ready for review again.

cranelift-entity/src/set.rs
```diff
@@ -33,6 +33,14 @@ where
         }
     }
 
+    /// Creates a new empty set with the specified capacity.
+    pub fn with_capacity(capacity: usize) -> Self {
+        Self {
+            elems: Vec::with_capacity(capacity),
```
nbp: `EntitySet` stores bits, thus the capacity should be `(capacity + 7) / 8`, to match the number of elements. (see `EntitySet::resize`)

PR #1032, opened 2019-09-14
Simplify and rename `Stackmap::from_vec`

AnthonyMikh:
Functionality is preserved, all tests pass.

It is unclear if this method really should take `&Vec<bool>` rather than `&[bool]`. Should I make this change as well?

### DIFF
--- a/cranelift-codegen/src/binemit/stackmap.rs
+++ b/cranelift-codegen/src/binemit/stackmap.rs
@@ -56,7 +56,7 @@ impl Stackmap {
     }
 
     /// Create a vec of Bitsets from a vec of bools.
-    pub fn from_vec(vec: &Vec<bool>) -> Self {
+    pub fn from_vec(vec: &[bool]) -> Self {
         let len = vec.len();
         let num_word = len / 32 + (len % 32 != 0) as usize;
         let mut bitmap = Vec::with_capacity(num_word);
--- a/cranelift-codegen/src/binemit/stackmap.rs
+++ b/cranelift-codegen/src/binemit/stackmap.rs
@@ -52,11 +52,11 @@ impl Stackmap {
             }
         }
 
-        Stackmap::from_vec(&vec)
+        Stackmap::from_slice(&vec)
     }
 
     /// Create a vec of Bitsets from a vec of bools.
-    pub fn from_vec(vec: &[bool]) -> Self {
+    pub fn from_slice(vec: &[bool]) -> Self {
         let len = vec.len();
         let num_word = len / 32 + (len % 32 != 0) as usize;
         let mut bitmap = Vec::with_capacity(num_word);
@@ -89,7 +89,7 @@ mod tests {
     #[test]
     fn stackmaps() {
         let vec: Vec<bool> = Vec::new();
-        assert!(Stackmap::from_vec(&vec).bitmap.is_empty());
+        assert!(Stackmap::from_slice(&vec).bitmap.is_empty());
 
         let mut vec: [bool; 32] = Default::default();
         let set_true_idx = [5, 7, 24, 31];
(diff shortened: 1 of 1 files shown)


cranelift-codegen/src/binemit/stackmap.rs
```diff
@@ -57,20 +57,18 @@ impl Stackmap {
...
         let mut bitmap = Vec::with_capacity(num_word);
 
-        for i in 0..num_word {
+        for segment in vec.chunks(32) {
             let mut curr_word = 0;
-            let count = if rem > 32 { 32 } else { rem };
-            for j in 0..count {
-                if vec[i * 32 + j] {
-                    curr_word |= 1 << j;
+            for (set, i) in segment.iter().zip(0..) {
```
bjorn3:
```suggestion
            for (i, set) in segment.iter().enumerate() {
```

bjorn3:
> Should I make this change as well?

I think so.

vitiral:
This is really good code! I like the change from float -> int operations and the use of chunks for the vector. I bet this is more performant too, since it avoids bounds checks.

The only thing is... the method name `from_vec` is wrong now that it's from a slice instead of a vec :laughing:

AnthonyMikh:
@vitiral what name do you propose then? I think about `from_bools`/`from_flags`. It shouldn't be a hard change since `from_vec` is used in this single file only.

bjorn3:
`from_slice`?

AnthonyMikh:
@bjorn3 IMHO, `from_slice` is not descriptive enough. The fact that argument is slice is already known from types. I suppose that the name should carry a bit more information about the meaning of argument.

vitiral:
it is a constructor. If there were overloading you would just use `from` for everything and then overloading would take care of the type. `_slice` should _precisely_ communicate the type, because it is basically just a stand-in for overloading!

cranelift-codegen/src/binemit/stackmap.rs
```diff
@@ -52,25 +52,23 @@ impl Stackmap {
             }
         }
 
-        Stackmap::from_vec(&vec)
+        Stackmap::from_slice(&vec)
     }
 
     /// Create a vec of Bitsets from a vec of bools.
```
bnjbvr: nit: this comment now needs an update; it's not a vec anymore.

cranelift-codegen/src/binemit/stackmap.rs
```diff
@@ -52,25 +52,23 @@ impl Stackmap {
...
-    pub fn from_vec(vec: &Vec<bool>) -> Self {
-        let mut rem = vec.len();
-        let num_word = ((rem as f32) / 32.0).ceil() as usize;
+    pub fn from_slice(vec: &[bool]) -> Self {
+        let len = vec.len();
+        let num_word = len / 32 + (len % 32 != 0) as usize;
```
bnjbvr: light suggestion: can we replace all the occurrences of 32 (here and below line 64) by a const, and this const could use the size_of intrinsics on the type that's stored in the Stackmap bitset? (If required, we could alias the bitset's type with some local type, and reference it here in the size_of call)

cranelift-codegen/src/binemit/stackmap.rs
AnthonyMikh: 🤦‍♂️ Indeed

cranelift-codegen/src/binemit/stackmap.rs
AnthonyMikh: Done

bnjbvr:
(I'll squash before merging; in general we try to keep commits "atomic" (= do one logical thing, passes test in isolation) so as to keep git bisection trivial. No worries about not doing so this time, just wanted to let you know for next time :))

PR #1034, opened 2019-09-16
 Fix #1023: Update brnz.i128 legalization to use non-extended basic blocks.

### LINKED ISSUE #1023: br.i128 split in 2 branches, which is not compliant with BB.
```rust
    narrow.legalize(
        def!(brnz.I128(x, ebb, vararg)),
        vec![
            def!((xl, xh) = isplit(x)),
            def!(brnz(xl, ebb, vararg)),
            def!(brnz(xh, ebb, vararg)),
        ],
    );
```

When replacing one `brnz` instruction by 2 others, this causes issues with basic block assumption that we can only have a single conditional branch in a basic block.

This is blocking #1008 work.
(cc @bnjbvr , @bjorn3 )

nbp:
In order to enable basic block by default (#796) , we have to reduce the legalization of `brnz.i128` to produce a temporary basic block in which the conditional jump on the second part can be produced.

This pull request changes the previous meta code to:
```rust
    narrow.legalize(
        def!(brnz.I128(x, ebb1, vararg)),
        vec![
            def!((xl, xh) = isplit(x)),
            def!(brnz(xl, ebb1, vararg)),
            def!(jump(ebb2, Literal::empty_vararg(&imm.vararg))),
            ebb!(ebb2),
            def!(brnz(xh, ebb1, vararg)),
        ],
    );
```

Which produces the following code:
```rust
            ir::Opcode::Brnz => {
                // Unwrap () << brnz.i128(x, ebb1, vararg)
                let (x, ebb1, vararg, predicate) = if let crate::ir::InstructionData::Branch {
                    destination,
                    ref args,
                    ..
                } = pos.func.dfg[inst] {
                    let func = &pos.func;
                    let args = args.as_slice(&func.dfg.value_lists);
                    (
                        func.dfg.resolve_aliases(args[0]),
                        destination,
                        args.iter().skip(2).map(|&arg| func.dfg.resolve_aliases(arg)).collect::<Vec<_>>(),
                        func.dfg.value_type(args[0]) == ir::types::I128
                    )
                } else {
                    unreachable!("bad instruction format")
                };
                let vararg = &vararg;
                if predicate {
                    let orig_ebb = pos.current_ebb().unwrap();
                    pos.func.dfg.clear_results(inst);
                    let ebb2 = pos.func.dfg.make_ebb();
                    let curpos = pos.position();
                    let srcloc = pos.srcloc();
                    let (xl, xh) = split::isplit(pos.func, cfg, curpos, srcloc, x);
                    pos.ins().brnz(xl, ebb1, vararg);
                    pos.ins().jump(ebb2, &[]);
                    pos.insert_ebb(ebb2);
                    pos.ins().brnz(xh, ebb1, vararg);
                    let removed = pos.remove_inst();
                    debug_assert_eq!(removed, inst);
                    cfg.recompute_ebb(pos.func, orig_ebb);
                    cfg.recompute_ebb(pos.func, ebb2);
                    return true;
                }
            }
```

### DIFF
--- a/cranelift-codegen/meta/src/cdsl/ast.rs
+++ b/cranelift-codegen/meta/src/cdsl/ast.rs
@@ -111,6 +111,9 @@ pub(crate) enum Literal {
 
     /// A value of an integer immediate operand.
     Int(i64),
+
+    /// A empty list of variable set of arguments.
+    EmptyVarArgs,
 }
 
 impl Literal {
@@ -147,11 +150,16 @@ impl Literal {
         Literal::Int(value)
     }
 
+    pub fn empty_vararg() -> Self {
+        Literal::EmptyVarArgs
+    }
+
     pub fn to_rust_code(&self) -> String {
         match self {
             Literal::Enumerator { rust_type, value } => format!("{}::{}", rust_type, value),
             Literal::Bits { rust_type, value } => format!("{}::with_bits({:#x})", rust_type, value),
             Literal::Int(val) => val.to_string(),
+            Literal::EmptyVarArgs => "&[]".into(),
         }
     }
 }
@@ -388,7 +396,7 @@ impl Apply {
                              obtained enum value '{}'",
                             op.kind.name, value
                         ),
-                        Literal::Bits { .. } | Literal::Int(_) => {}
+                        Literal::Bits { .. } | Literal::Int(_) | Literal::EmptyVarArgs => {}
                     },
                     _ => {
                         panic!(
--- a/cranelift-codegen/meta/src/cdsl/ast.rs
+++ b/cranelift-codegen/meta/src/cdsl/ast.rs
@@ -4,9 +4,10 @@ use crate::cdsl::operands::{OperandKind, OperandKindFields};
 use crate::cdsl::types::ValueType;
 use crate::cdsl::typevar::{TypeSetBuilder, TypeVar};
 
-use cranelift_entity::{entity_impl, PrimaryMap};
+use cranelift_entity::{entity_impl, PrimaryMap, SparseMap, SparseMapValue};
 
 use std::fmt;
+use std::iter::IntoIterator;
 
 pub(crate) enum Expr {
     Var(VarIndex),
@@ -82,7 +83,7 @@ impl DefPool {
     pub fn next_index(&self) -> DefIndex {
         self.pool.next_key()
     }
-    pub fn create(&mut self, apply: Apply, defined_vars: Vec<VarIndex>) -> DefIndex {
+    pub fn create_inst(&mut self, apply: Apply, defined_vars: Vec<VarIndex>) -> DefIndex {
         self.pool.push(Def {
             apply,
             defined_vars,
@@ -93,6 +94,55 @@ impl DefPool {
 #[derive(Copy, Clone, PartialEq, Eq, Hash, PartialOrd, Ord)]
 pub(crate) struct DefIndex(u32);
 entity_impl!(DefIndex);
+
+/// A definition which would lead to generate a block creation.
+#[derive(Clone)]
+pub(crate) struct Block {
+    /// Instruction index after which the block entry is set.
+    pub location: DefIndex,
+    /// Variable holding the new created block.
+    pub name: VarIndex,
+}
+
+pub(crate) struct BlockPool {
+    pool: SparseMap<DefIndex, Block>,
+}
+
+impl SparseMapValue<DefIndex> for Block {
+    fn key(&self) -> DefIndex {
+        self.location
+    }
+}
+
+impl BlockPool {
+    pub fn new() -> Self {
+        Self {
+            pool: SparseMap::new(),
+        }
+    }
+    pub fn get(&self, index: DefIndex) -> Option<&Block> {
+        self.pool.get(index)
+    }
+    pub fn create_block(&mut self, name: VarIndex, location: DefIndex) {
+        if self.pool.contains_key(location) {
+            panic!("Attempt to insert 2 blocks after the same instruction")
+        }
+        self.pool.insert(Block { location, name });
+    }
+    pub fn is_empty(&self) -> bool {
+        self.pool.is_empty()
+    }
+}
+
+// Implement IntoIterator such that we can iterate over blocks which are in the block pool.
+impl<'a> IntoIterator for &'a BlockPool {
+    type Item = <&'a SparseMap<DefIndex, Block> as IntoIterator>::Item;
+    type IntoIter = <&'a SparseMap<DefIndex, Block> as IntoIterator>::IntoIter;
+
+    fn into_iter(self) -> Self::IntoIter {
+        self.pool.into_iter()
+    }
+}
 
 #[derive(Clone, Debug)]
 pub(crate) enum Literal {
@@ -518,6 +568,7 @@ pub(crate) enum DummyExpr {
     Var(DummyVar),
     Literal(Literal),
     Apply(InstSpec, Vec<DummyExpr>),
+    Block(DummyVar),
 }
 
 #[derive(Clone)]
@@ -561,6 +612,11 @@ impl ExprBuilder {
             defined_vars,
         }
     }
+
+    pub fn block(name: DummyVar) -> Self {
+        let expr = DummyExpr::Block(name);
+        Self { expr }
+    }
 }
 
 macro_rules! def_rhs {
@@ -591,4 +647,12 @@ macro_rules! def {
     ($($tt:tt)*) => {
         def_rhs!($($tt)*).assign_to(Vec::new())
     }
+}
+
+// Helper macro to define legalization recipes.
+macro_rules! ebb {
+    // An basic block definition, splitting the current block in 2.
+    ($block: ident) => {
+        ExprBuilder::block($block).assign_to(Vec::new())
+    };
 }
--- a/cranelift-codegen/meta/src/cdsl/xform.rs
+++ b/cranelift-codegen/meta/src/cdsl/xform.rs
@@ -1,5 +1,6 @@
 use crate::cdsl::ast::{
-    Apply, DefIndex, DefPool, DummyDef, DummyExpr, Expr, PatternPosition, VarIndex, VarPool,
+    Apply, BlockPool, DefIndex, DefPool, DummyDef, DummyExpr, Expr, PatternPosition, VarIndex,
+    VarPool,
 };
 use crate::cdsl::instructions::Instruction;
 use crate::cdsl::type_inference::{infer_transform, TypeEnvironment};
@@ -22,6 +23,7 @@ pub(crate) struct Transform {
     pub dst: Vec<DefIndex>,
     pub var_pool: VarPool,
     pub def_pool: DefPool,
+    pub block_pool: BlockPool,
     pub type_env: TypeEnvironment,
 }
 
@@ -31,6 +33,7 @@ impl Transform {
     fn new(src: DummyDef, dst: Vec<DummyDef>) -> Self {
         let mut var_pool = VarPool::new();
         let mut def_pool = DefPool::new();
+        let mut block_pool = BlockPool::new();
 
         let mut input_vars: Vec<VarIndex> = Vec::new();
         let mut defined_vars: Vec<VarIndex> = Vec::new();
@@ -47,6 +50,7 @@ impl Transform {
             &mut defined_vars,
             &mut var_pool,
             &mut def_pool,
+            &mut block_pool,
         )[0];
 
         let num_src_inputs = input_vars.len();
@@ -59,6 +63,7 @@ impl Transform {
             &mut defined_vars,
             &mut var_pool,
             &mut def_pool,
+            &mut block_pool,
         );
 
         // Sanity checks.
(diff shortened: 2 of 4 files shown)


cranelift-codegen/meta/src/gen_legalizer.rs
```diff
@@ -401,14 +401,30 @@ fn gen_transform<'a>(
     // Guard the actual expansion by `predicate`.
     fmt.line("if predicate {");
     fmt.indent(|fmt| {
+        // If we are adding some blocks, we need to recall the original block, such that we can
+        // recompute it.
+        if !transform.block_pool.is_empty() {
+            fmt.line("let orig_ebb = pos.current_ebb().unwrap();");
```
bjorn3: Maybe emit this unconditionally, and then always use the else case of `if transform.block_pool.is_empty() {` for `recompute_ebb`?

cranelift-codegen/meta/src/shared/legalize.rs
```diff
@@ -260,11 +263,13 @@ pub(crate) fn define(insts: &InstructionGroup, imm: &Immediates) -> TransformGro
...
     narrow.legalize(
-        def!(brnz.I128(x, ebb, vararg)),
+        def!(brnz.I128(x, ebb1, vararg)),
         vec![
             def!((xl, xh) = isplit(x)),
-            def!(brnz(xl, ebb, vararg)),
-            def!(brnz(xh, ebb, vararg)),
+            def!(brnz(xl, ebb1, vararg)),
+            def!(jump(ebb2, Literal::empty_vararg(&imm.vararg))),
+            ebb!(ebb2),
```
bjorn3:
```suggestion

            ebb!(ebb2),
```

nbp: Not sure if `cargo fmt` will like this change.

bnjbvr: I think cargo fmt doesn't run in macro's code (probably because the formatting could be part of the macro pattern matching), but not sure.

cranelift-codegen/meta/src/cdsl/ast.rs
```diff
@@ -168,6 +171,13 @@ impl Literal {
         }
     }
 
+    pub fn empty_vararg(kind: &OperandKind) -> Self {
+        Self {
+            kind: kind.clone(),
```
bnjbvr: Hmm, it seems it doesn't even need a kind, since it always generates the same literal (making the kind optional would be unsatisfactory, though)? Or we could reuse the "constant" kind for the immediate VarArgs?

cranelift-codegen/meta/src/cdsl/xform.rs
```diff
@@ -241,8 +250,17 @@ fn rewrite_def_list(
     defined_vars: &mut Vec<VarIndex>,
     var_pool: &mut VarPool,
     def_pool: &mut DefPool,
+    block_pool: &mut BlockPool,
 ) -> Vec<DefIndex> {
     let mut new_defs = Vec::new();
+    // Register variable names of new blocks first as a block name can be used to jump forward. Thus
+    // the name has to be registered first to avoid miss-interpreting it as an input-var.
```
bnjbvr: nit: misinterpreting

cranelift-codegen/meta/src/cdsl/xform.rs
```diff
@@ -254,19 +272,34 @@ fn rewrite_def_list(
...
             var_pool,
         );
-        let new_apply = rewrite_expr(position, dummy_def.expr, symbol_table, input_vars, var_pool);
+        if let DummyExpr::Block(var) = dummy_def.expr {
+            let mut temp_vars = vec![];
+            let var_index = var_index(var.name, symbol_table, &mut temp_vars, var_pool);
+            var_pool.get_mut(var_index).set_def(position, def_index);
+            block_pool.create_block(var_index, def_index);
+            assert!(
+                temp_vars.len() == 0,
```
bnjbvr: nit: is_empty()

cranelift-codegen/meta/src/gen_legalizer.rs
bnjbvr: Won't this cause recompute_ebb for each ebb containing a legalized instruction?

cranelift-codegen/meta/src/shared/legalize.rs
nbp: After testing, `cargo fmt` does re-indent and remove the empty line.

bjorn3: > probably because the formatting could be part of the macro pattern matching

No it can't. I think it doesn't fmt macro invocations because rustfmt doesn't know if a certain part of a macro invocation is an expr, item, pattern, ... However it seems that blank lines are removed.

cranelift-codegen/meta/src/gen_legalizer.rs
```diff
@@ -401,14 +401,30 @@ fn gen_transform<'a>(
...
+        for block in &transform.block_pool {
+            let var = transform.var_pool.get(block.name);
+            fmtln!(fmt, "let {} = pos.func.dfg.make_ebb();", var.name);
+        }
+
         // Emit the destination pattern.
         for &def_index in &transform.dst {
+            if let Some(block) = transform.block_pool.get(def_index) {
+                let var = transform.var_pool.get(block.name);
+                fmtln!(fmt, "pos.insert_ebb({});", var.name);
```
bnjbvr: I need to re-read this, but these two lines make me wonder if we need to tie the implementation of Block with Var, or just separate the two concepts: it seems the Block could hold its name directly and not need a Var for this.

cranelift-codegen/meta/src/gen_legalizer.rs
nbp: I am not sure, as this implies keeping a variable with the value before other transformations as the one which is currently present would just pick the value without the need for extra stack space.

I do not think the rust compiler (LLVM) is capable of moving this code around and remove this extra variable.

cranelift-codegen/meta/src/cdsl/ast.rs
nbp: An alternative would be to make the `kind` field optional, or make it a field of each of the `LiteralValue` enumerated values.

Another option would be to create the kind field within this function instead of reusing the immediate value.

cranelift-codegen/meta/src/gen_legalizer.rs
nbp: Remember that instructions have to use the block name, so this should not be a different symbol table.  Using the variable makes sense to me when thought in terms of generated code.

cranelift-codegen/meta/src/shared/immediates.rs
```diff
@@ -170,6 +173,11 @@ impl Immediates {
                     .rust_type("ir::TrapCode")
                     .build()
             },
+
+            vararg: Builder::new_imm("vararg")
+                .doc("A list of block arguments")
+                .rust_type("&[_]")
```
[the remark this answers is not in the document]
nbp: Testing the following code in the Rust playground seem to work:

```rust
fn print_chars(chars: &[u8]) {
    println!("{:?}", chars);  // [97, 97, 97]
}

fn main() {
    let aaa: &[_] = &[];
    print_chars(aaa);
}
```

However, I have not yet seen it be generated in the code. I added this line to have a complete definition.

cranelift-codegen/meta/src/cdsl/ast.rs
```diff
@@ -95,6 +96,55 @@ impl DefPool {
...
+        }
+        self.pool.insert(Block { location, name });
+    }
+    pub fn is_empty(&self) -> bool {
+        self.pool.is_empty()
+    }
+}
+
+// Implement IntoIterator such that we can iterate over blocks which are in the block pool.
+impl<'a> IntoIterator for &'a BlockPool {
```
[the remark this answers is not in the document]
nbp: `Iterator` requires the `next` function and store an index of the element being visited. Having `impl Iterator for BlockPool` would mean that BlockPool is an `Iterator` and not a collection over which we iterate over.

`IntoIterator` is used by Rust for-loops which are de-sugared using `<_ as IntoIterator>::into_iter()` function call, if I understand correctly. By implementing `impl<'a> IntoIterator for &'a BlockPool`, we had the ability to use the following syntax: `for block of &block_pool`, which would create an immutable borrow of the `block_pool` elements.

Therefore, I am not sure I understand your question.

cranelift-codegen/meta/src/cdsl/ast.rs
nbp: ok, I will do that.

cranelift-codegen/meta/src/shared/immediates.rs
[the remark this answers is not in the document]
bnjbvr: Ok fine; I'd rather not add stuff that's unused (because it's untested), but it's not a big deal in this case, so no strong opinions.

cranelift-codegen/meta/src/cdsl/ast.rs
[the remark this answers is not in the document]
bnjbvr: You're right, we'd need another data structure to be the Iterator, using IntoIterator is simpler, thanks.

cranelift-codegen/meta/src/cdsl/xform.rs
```diff
@@ -197,23 +210,14 @@ fn rewrite_expr(
...
-                    }
-                };
+                if let Some(&own_var) = symbol_table.get(var.name) {
+                    let var = var_pool.get(own_var);
+                    assert!(
+                        var.is_input() || var.get_def(position).is_some(),
+                        format!("{:?} used as both input and def", var)
+                    );
+                }
+                let own_var = var_index(var.name, symbol_table, input_vars, var_pool);
```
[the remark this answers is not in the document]
nbp: It appears that newly inserted variable should all pass the assertion. The reason being that `is_input` is true unless `set_def` was called to specify where the variable is defined. When a new variable is added with `var_index`, no `set_def` calls are made, and as such it is considered as an input. (and also goes in the `input_vars` vector)

Thus I think it is safe to move this assert out of this if-branch, and thus not have any need for an extra flag reporting whether the variable was present or not in the symbol table.

I will replace this code by:
```rust
            DummyExpr::Var(var) => {
                let own_var = var_index(var.name, symbol_table, input_vars, var_pool);
                let var = var_pool.get(own_var);
                assert!(
                    var.is_input() || var.get_def(position).is_some(),
                    format!("{:?} used as both input and def", var)
                );
                args.push(Expr::Var(own_var));
            }
```

cranelift-codegen/meta/src/cdsl/xform.rs
```diff
@@ -138,6 +138,28 @@ impl Transform {
     }
 }
 
+/// Inserts, if not present, a name in the `symbol_table`. Then returns its index in the variable
+/// pool `var_pool`. If the variable was not present in the symbol table, then add it to the list of
+/// `defined_vars`.
+fn var_index(
+    name: &'static str,
+    symbol_table: &mut SymbolTable,
+    defined_vars: &mut Vec<VarIndex>,
```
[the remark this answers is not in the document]
nbp: I think the current scheme is the simplest. Returning a boolean to then branch on it would be harder to read from my point of view.

Note, I could replace the code from the next commit to be less alien by replacing:
```rust
            let mut temp_vars = vec![];
            let var_index = var_index(var.name, symbol_table, &mut temp_vars, var_pool);
            var_pool.get_mut(var_index).set_def(position, def_index);
            block_pool.create_block(var_index, def_index);
            assert!(
                temp_vars.is_empty(),
                format!(
                    "Block {} was not registered during the first visit",
                    var.name
                )
            );
```

by

```rust
            let nb_defined_vars = defined_vars.len();
            let var_index = var_index(var.name, symbol_table, defined_vars, var_pool);
            var_pool.get_mut(var_index).set_def(position, def_index);
            block_pool.create_block(var_index, def_index);
            assert!(
                defined_vars.len() == nb_defined_vars,
                format!(
                    "Block {} was not registered during the first visit",
                    var.name
                )
            );
```

We could also completely replace the `var_index` call by:

```
            let var_index = symbol_table.get(var.name).expect(
                format!(
                    "Block {} was not registered during the first visit",
                    var.name
                )
            );
            var_pool.get_mut(var_index).set_def(position, def_index);
            block_pool.create_block(var_index, def_index);
```

Since the first loop should have registered all new block names.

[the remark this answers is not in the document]
bnjbvr: > We could also completely replace the var_index call by:

Yes, that looks more straight to the point, thanks.

cranelift-codegen/meta/src/cdsl/ast.rs
```diff
@@ -95,6 +96,55 @@ impl DefPool {
 pub struct DefIndex(u32);
 entity_impl!(DefIndex);
 
+/// A definition which would lead to generate a block creation.
+#[derive(Clone)]
+pub struct Block {
```
bnjbvr: Can you mark this struct and the one below pub(crate) instead, please? (If that means transitively marking more stuff pub(crate), please do so.)

PR #1035, opened 2019-09-17
[codegen] Check for downcasting in bitcast instruction

### LINKED ISSUE #854: Verifier should check bitcast validity
Currently, the verifier accepts this code:

```clif
function %test(i64) -> i32 {
ebb0(v0: i64):
    v1 = bitcast.i32 v0
    return v1
}
```

though it can't be compiled, because `bitcast` requires the input be the same size as the output. We should write a verifier check which checks for this.

There are two approaches, one would be to just add a custom check for bitcast instructions in the verifier, and the other would be to add a type constraint in the bitcast instruction. My rough sense is that it's better to just handle this in the verifier, as there won't be many instructions which have this constraint, and it's simpler that way, though I wouldn't be opposed to a type constraint either.

walac:
Bitcasting to a smaller size is invalid.

closes #854

### DIFF
--- a/cranelift-codegen/src/verifier/mod.rs
+++ b/cranelift-codegen/src/verifier/mod.rs
@@ -696,6 +696,13 @@ impl<'a> Verifier<'a> {
                 }
             }
 
+            Unary {
+                opcode: Opcode::Bitcast,
+                arg,
+            } => {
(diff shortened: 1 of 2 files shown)


walac:
Done!

Issue #1038, opened 2019-09-17
Document entity types better


jyn514:
Consider a frontend that wants to emit a [`call`](https://docs.rs/cranelift-codegen/0.42.0/cranelift_codegen/ir/trait.InstBuilder.html#method.call) instruction. The documentation states that you need to pass

- a [`FuncRef`](https://docs.rs/cranelift-codegen/0.42.0/cranelift_codegen/ir/entities/struct.FuncRef.html)
- a slice of [`Value`](https://docs.rs/cranelift-codegen/0.42.0/cranelift_codegen/ir/entities/struct.Value.html)s

However, neither of the documentations for those types show how you can get an instance of them! For Value you can look around on the rest of the page for `InstBuilder` and see that most instructions return a `Value`, but for `FuncRef` it's non-obvious that you need to either call [`import_function`](https://docs.rs/cranelift-codegen/0.42.0/cranelift_codegen/ir/function/struct.Function.html#method.import_function) or [`declare_func_in_func`](https://docs.rs/cranelift-module/0.42.0/cranelift_module/struct.Module.html#method.declare_func_in_func). It would be great to link to these pages in the documentation for `FuncRef`.

Similar problems exist for 

- [`GlobalValue`](https://docs.rs/cranelift-codegen/0.42.0/cranelift_codegen/ir/entities/struct.GlobalValue.html)
- [`Ebb`](https://docs.rs/cranelift-codegen/0.42.0/cranelift_codegen/ir/entities/struct.Ebb.html)
- [`Inst`](https://docs.rs/cranelift-codegen/0.42.0/cranelift_codegen/ir/entities/struct.Inst.html)
- [`JumpTable`](https://docs.rs/cranelift-codegen/0.42.0/cranelift_codegen/ir/entities/struct.JumpTable.html)
- [`Heap`](https://docs.rs/cranelift-codegen/0.42.0/cranelift_codegen/ir/entities/struct.Heap.html)
- [`Table`](https://docs.rs/cranelift-codegen/0.42.0/cranelift_codegen/ir/entities/struct.Table.html)
- [`SigRef`](https://docs.rs/cranelift-codegen/0.42.0/cranelift_codegen/ir/entities/struct.SigRef.html)
- [`StackSlot`](https://docs.rs/cranelift-codegen/0.42.0/cranelift_codegen/ir/entities/struct.StackSlot.html)
[...]

PR #1041, opened 2019-09-17
[WIP] Add more documentation for entities

jyn514:
Addresses https://github.com/CraneStation/cranelift/issues/1038

Document stackslots, funcrefs, heaps, sigrefs, and jumptables.
Also adds more inter-crate links.

## Undocumented

- [x] `Constant` - https://github.com/CraneStation/cranelift/pull/1041#issuecomment-532348093
- [x] `GlobalValue` - https://github.com/CraneStation/cranelift/pull/1041#issuecomment-532575257
- [x] `Table` - https://github.com/CraneStation/cranelift/pull/1041#issuecomment-532575257
- [x] `Inst` - https://github.com/CraneStation/cranelift/pull/1041#issuecomment-532575257

## Issues

I can't find a way to link to a type/function in a different crate if that crate isn't currently in scope for the code. For example, [the documentation for JumpTable](https://github.com/CraneStation/cranelift/compare/master...jyn514:doc-entity?expand=1#diff-dd78b3709a42d640a60ea2733dcdbe53R187) links to `FunctionBuilder::create_jump_table`, but the link is broken (it points to `cranelift_codegen/ir/entities/cranelift_frontend::FunctionBuilder::create_jump_table`, which is clearly wrong).

There are two ways I can think to resolve this:

1. Link directly to docs.rs like this: `https://docs.rs/cranelift-module/*/cranelift_module/struct.Module.html`. This has the downside that it will always point to the latest version (and won't work locally or offline).
2. Put the name of the type in prose and have users search for it. This is not ideal but will work the same everywhere.

### DIFF
--- a/cranelift-codegen/src/ir/entities.rs
+++ b/cranelift-codegen/src/ir/entities.rs
@@ -29,7 +29,8 @@ use serde::{Deserialize, Serialize};
 /// block](https://en.wikipedia.org/wiki/Extended_basic_block) in a
 /// [`Function`](super::function::Function).
 ///
-/// You can get an `Ebb` using `cranelift_frontend::FunctionBuilder::create_ebb`.
+/// You can get an `Ebb` using
+/// [`FunctionBuilder::create_ebb`](https://docs.rs/cranelift-frontend/*/cranelift_frontend/struct.FunctionBuilder.html#method.create_ebb)
 #[derive(Copy, Clone, PartialEq, Eq, Hash, PartialOrd, Ord)]
 pub struct Ebb(u32);
 entity_impl!(Ebb, "ebb");
@@ -100,7 +101,7 @@ entity_impl!(Inst, "inst");
 /// [call stack](https://en.wikipedia.org/wiki/Call_stack).
 ///
 /// `StackSlot`s can be created with
-/// `cranelift_frontend::FunctionBuilder::create_stackslot`.
+/// [`FunctionBuilder::create_stackslot`](https://docs.rs/cranelift-frontend/*/cranelift_frontend/struct.FunctionBuilder.html#method.create_stack_slot).
 ///
 /// `StackSlot`s are most often used with
 /// [`stack_addr`](super::InstBuilder::stack_addr),
@@ -133,11 +134,11 @@ impl StackSlot {
 ///
 /// - When compiling to WASM, you can use it to load values from a
 /// [`VmContext`](super::GlobalValueData::VMContext) using
-/// [`make_global`](cranelift_wasm::FuncEnvironment::make_global).
+/// [`FuncEnvironment::make_global`](https://docs.rs/cranelift-wasm/*/cranelift_wasm/trait.FuncEnvironment.html#tymethod.make_global).
 /// - When compiling to native code, you can use it for objects in static memory with
-/// [`declare_data_in_func`](cranelift_module::Module::declare_data_in_func).
+/// [`Module::declare_data_in_func`](https://docs.rs/cranelift-module/*/cranelift_module/struct.Module.html#method.declare_data_in_func).
 /// - For any compilation target, it can be registered with
-/// `create_global_value`](cranelift_frontend::FunctionBuilder::create_global_value).
+/// [`FunctionBuilder::create_global_value`](https://docs.rs/cranelift-frontend/*/cranelift_frontend/struct.FunctionBuilder.html#method.create_global_value).
 ///
 /// `GlobalValue`s can be retrieved with
 /// [`InstBuilder:global_value`](super::InstBuilder::global_value).
@@ -184,12 +185,12 @@ impl Constant {
 ///
 /// `JumpTable`s are used for indirect branching and are specialized for dense,
 /// 0-based jump offsets. If you want a jump table which doesn't start at 0,
-/// or is not contiguous, consider using a [`Switch`](cranelift_frontend::Switch) instead.
+/// or is not contiguous, consider using a [`Switch`](https://docs.rs/cranelift-frontend/*/cranelift_frontend/struct.Switch.html) instead.
 ///
 /// `JumpTable` are used with [`br_table`](super::InstBuilder::br_table).
 ///
 /// `JumpTable`s can be created with
-/// [`create_jump_table`](cranelift_frontend::FunctionBuilder::create_jump_table).
+/// [`create_jump_table`](https://docs.rs/cranelift-frontend/*/cranelift_frontend/struct.FunctionBuilder.html#method.create_jump_table).
 #[derive(Copy, Clone, PartialEq, Eq, Hash)]
 #[cfg_attr(feature = "enable-serde", derive(Serialize, Deserialize))]
 pub struct JumpTable(u32);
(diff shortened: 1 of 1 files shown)


abrown:
@jyn514 I can explain Constant: it is an integer used to index into the ConstantPool to identify constant values held there (see https://github.com/CraneStation/cranelift/blob/dcaa7f5ba618d72c9dd24a29c9a485dc24ab0d07/cranelift-codegen/src/ir/constant.rs#L97).

jyn514:
I see, and it's public because `cranelift-frontend` needs to access it?

abrown:
> I can't find a way to link to a type/function in a different crate if that crate isn't currently in scope for the code

Also, I believe I had similar issues trying to fix some documentation warnings; in the end I went with something like `crate::ir::...` and that seemed to work: https://github.com/CraneStation/cranelift/pull/952. The trick is to use correct Rust paths (see links in https://stackoverflow.com/a/53504254 and https://github.com/rust-lang/rust/issues/43466). I think that getting the Rust paths to work is better long-term than direct links or prose.

cranelift-codegen/src/ir/entities.rs
```diff
@@ -44,6 +48,18 @@ impl Ebb {
...
+///
+/// You can get a constant `Value` from the following
+/// [`InstBuilder`](super::InstBuilder) instructions:
+///
+/// - [`iconst`](super::InstBuilder::iconst) for integer constants
+/// - [`f32const`](super::InstBuilder::f32const) for 32-bit float constants
+/// - [`f64const`](super::InstBuilder::f64const) for 64-bit float constants
+/// - [`bconst`](super::InstBuilder::bconst) for boolean constants
+/// - [`vconst`](super::InstBuilder::vconst) for vector constants
+/// - [`null`](super::InstBuilder::null) for null pointer constants
```
bjorn3: *null reference

cranelift-codegen/src/ir/entities.rs
```diff
@@ -68,6 +84,16 @@ pub struct Inst(u32);
...
+///
+/// Stack slots represent an address on the
+/// [call stack](https://en.wikipedia.org/wiki/Call_stack).
+///
+/// `StackSlot`s can be created with
+/// `cranelift_frontend::FunctionBuilder::create_stackslot`.
+///
+/// `StackSlot`s are most often used with
+/// [`stack_addr`](super::InstBuilder::stack_addr) and
+/// [`stack_load`](super::InstBuilder::stack_load).
```
bjorn3: missing stack_store

jyn514:
> in the end I went with something like crate::ir::...

That only works if the crate is imported into the current scope, though. I can't put `extern crate cranelift_module;` in the code because that breaks the code of anyone who uses cranelift for WASM.

cranelift-codegen/src/ir/entities.rs
```diff
@@ -122,7 +148,13 @@ impl Constant {
...
 }
 
-/// An opaque reference to a jump table.
+/// An opaque reference to a [jump table](https://en.wikipedia.org/wiki/Branch_table).
+///
+/// `JumpTable`s are used for indirect branching.
+/// See [`br_table`](super::InstBuilder::br_table) for details.
+///
+/// `JumpTable`s can be created with
+/// [`create_jump_table`](cranelift_frontend::FunctionBuilder::create_jump_table).
```
bjorn3: If you want a jump table which doesnt start at 0, or is not contiguous, you should use cranelift_frontend::Switch instead.

abrown:
> I see, and it's public because cranelift-frontend needs to access it?

Now that I look at its usages, I don't see it anywhere except in codegen so it might be able to be downgraded to `pub(crate)`. The idea, though, is that it is part of a public API for interacting with the constant pool.

cranelift-codegen/src/ir/entities.rs
```diff
@@ -159,7 +206,17 @@ impl FuncRef {
...
+///
+/// `SigRef`s are used to declare a function with
+/// [`import_function`](cranelift_frontend::FunctionBuilder::import_function)
+/// as well as to make an [indirect function call](super::InstBuilder::call_indirect).
+///
+/// `SigRef`s can be created with
+/// [`import_signature`](cranelift_frontend::FunctionBuilder::import_signature).
+///
+/// You can retrieve the [`Signature`](super::Signature) that was used to create a `SigRef` with
+/// [`signature`](cranelift_frontend::FunctionBuilder::signature).
```
bjorn3: func.dfg.signatures

abrown:
Maybe https://github.com/rust-lang/rust/issues/43466#issuecomment-500418346 is fixed somewhere? Not exactly sure...

abrown:
> Maybe rust-lang/rust#43466 (comment) is fixed somewhere? Not exactly sure...

But if there is absolutely no way to link to external crates then I guess my vote would be for the direct link.

jyn514:
> The idea, though, is that it is part of a public API for interacting with the constant pool.

I don't see a way to get a `Constant` without inserting things into the pool yourself, though ... I guess it could be useful as a data structure so you aren't caching things twice (once in cranelift and once in the frontend).

abrown:
@jyn514, for `Inst` you might want to look at how it is used in `DataFlowGraph`.

bnjbvr:
Thanks for taking a stab at this!

Note that all of these implement EntityRef, so they're all handles to some data living somewhere else. For instance Constant has ConstantData, etc. I remember proposing to constantly suffix them with "Ref" or something, but this was a long time ago and I'm unsure what the outcome of the discussion was.

- GlobalValue can be mapped to a wasm global value, but doesn't need to; it is "just" a value that may be live across the entire function lifetime, in my understanding, and that can be preloaded from other global values. (Useful when you have a VmContext, and want to load constant things from it, for instance)
- Table is used for wasm function tables at the moment. I don't know if it's planned to use it for other table kinds in the future, could be. Useful for implementing indirect calls to known functions.
- Inst is the handle for an instruction. I imagine most usage is internal, and I wonder if this could be made not public.

bjorn3:
> Inst is the handle for an instruction. I imagine most usage is internal, and I wonder if this could be made not public.

I do use it to attach a comment to a specific instruction in the clif ir written by cg_clif.

jyn514:
Getting slightly off-topic now - is there any way that [`pretty_clif.rs`](https://github.com/bjorn3/rustc_codegen_cranelift/blob/master/src/pretty_clif.rs) could be made part of the main Cranelift API? I would love to reuse that code without having to copy/paste most of the implementation.

jyn514:
> Useful when you have a VmContext, and want to load constant things from it

Is this a WASM thing? It mentions a sandbox in [the documentation](https://docs.rs/cranelift-codegen/0.42.0/cranelift_codegen/ir/enum.ArgumentPurpose.html#variant.VMContext), and I don't see a way to construct it from `cranelift_frontend` or `cranelift_codegen`

bjorn3:
> is there any way that pretty_clif.rs could be made part of the main Cranelift API?

https://github.com/CraneStation/cranelift/pull/379#issuecomment-401552747

> I would love to reuse that code without having to copy/paste most of the implementation.

Please note that currently comments get inserted **after** the specified inst, as there is no way to know what the next inst id will be.

jyn514:
This is ready for review as soon as I fix the broken links. @abrown recommended permalinks to docs.rs, should I do that?

cranelift-codegen/src/ir/entities.rs
```diff
@@ -78,7 +78,18 @@ impl Value {
...
+/// Most usage of `Inst` is internal. `Inst`ructions are returned by
+/// [`InstBuilder`](super::InstBuilder) instructions that do not return a
+/// [`Value`], such as control flow and trap instructions.
+///
+/// If you look around the API, you can find many inventive uses for `Inst`,
+/// such as [annotating specific instructions with a comment][inst_comment]
+/// or [performing reflection at compile time](super::DataFlowGraph::analyze_branch)
+/// on the type of instruction.
+///
+/// [inst_comment]: https://github.com/CraneStation/cranelift/pull/1041#issuecomment-532575989
```
bjorn3: Maybe link https://github.com/bjorn3/rustc_codegen_cranelift/blob/0f8814fd6da3d436a90549d4bb19b94034f2b19c/src/pretty_clif.rs instead.

abrown:
@jyn514 I'm fine merging this but we need to fix the build issues: https://dev.azure.com/CraneStation/Cranelift/_build/results?buildId=897&view=logs.

jyn514:
Ok, I'll change it to link directly to docs.rs, then.

jyn514:
This is ready for review.

abrown:
@jyn514 thanks for the fix!

PR #1042, opened 2019-09-18
Continuous fuzzing

bookmoons:
Integrates with [Fuzzit](https://fuzzit.dev/) for automated fuzzing.

Uses the existing fuzz targets:
* translate-module - Fuzz valid WebAssembly modules.
* reader-parse - Fuzz IR text format parsing.

Replaces #944.


tschneidereit:
And it's working!

bookmoons:
Nice! :sunglasses:

I think you can remove me from the org in settings, then it's all yours. You should be able to add @pchickey there too.

Issue #732, opened 2019-04-08
assertion failed: !self.opcode().is_branch() during licm


bjorn3:
Compile

```rust
// run-pass

// aux-build:xcrate-reachable.rs

#![feature(generator_trait)]

extern crate xcrate_reachable as foo;

use std::ops::Generator;
use std::pin::Pin;

fn main() {
    Pin::new(&mut foo::foo()).resume();
}
```

using rustc_codegen_cranelift with `opt_level=best` and you get:

```
---- [run-pass] run-pass/generator/xcrate-reachable.rs stdout ----

error: test compilation failed although it shouldn't!
status: exit code: 101
command: "/home/bjorn/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/bin/rustc" "/home/bjorn/Documenten/rustc_codegen_cranelift/rust/src/test/run-pass/generator/xcrate-reachable.rs" "-Zthreads=1" "--target=x86_64-unknown-linux-gnu" "--error-format" "json" "-Zui-testing" "-o" "/home/bjorn/Documenten/rustc_codegen_cranelift/rust/build/x86_64-unknown-linux-gnu/test/run-pass/generator/xcrate-reachable/a" "-Crpath" "-O" "-Zunstable-options" "-Zcodegen-backend=/home/bjorn/Documenten/rustc_codegen_cranelift/rust/../target/debug/librustc_codegen_cranelift.so" "--sysroot" "/home/bjorn/Documenten/rustc_codegen_cranelift/rust/../build_sysroot/sysroot" "-Cpanic=abort" "-Lnative=/home/bjorn/Documenten/rustc_codegen_cranelift/rust/build/x86_64-unknown-linux-gnu/native/rust-test-helpers" "-L" "/home/bjorn/Documenten/rustc_codegen_cranelift/rust/build/x86_64-unknown-linux-gnu/test/run-pass/generator/xcrate-reachable/auxiliary"
stdout:
------------------------------------------
Instance { def: Item(DefId(9/1:10 ~ xcrate_reachable[8787]::foo[0]::{{closure}}[0])), substs: [(), u32, {()}] } _ZN16xcrate_reachable3foo28_$u7b$$u7b$closure$u7d$$u7d$17he526fe6105ca50ffE

------------------------------------------
stderr:
------------------------------------------
[...]

PR #1044, opened 2019-09-18
Create a new optimisation level, "nearly_best", and use it to disable…

julian-seward1:
… the insn shrink pass

The optimisation level called "best" runs at least one pass which, according
to measurements on a large wasm input, slows the compiler down whilst having
very little benefit for the final generated code.

This patch creates a new optimisation level, "nearly_best", which is intended
to be the same as "best", except with such low-win transformations removed.
The "instruction shrinking" pass is then omitted at the "nearly_best" level.
Behaviour of the compiler is unchanged for the three existing settings
"default", "fastest" and "best".  So all existing uses of Cranelift will not
see any change in behaviour from this patch.

### DIFF
--- a/cranelift-codegen/src/context.rs
+++ b/cranelift-codegen/src/context.rs
@@ -132,18 +132,18 @@ impl Context {
         self.verify_if(isa)?;
         debug!("Compiling:\n{}", self.func.display(isa));
 
+        let opt_level = isa.flags().opt_level();
+
         self.compute_cfg();
-        if isa.flags().opt_level() != OptLevel::Fastest {
+        if opt_level != OptLevel::None {
             self.preopt(isa)?;
         }
         if isa.flags().enable_nan_canonicalization() {
             self.canonicalize_nans(isa)?;
         }
         self.legalize(isa)?;
-        if isa.flags().opt_level() != OptLevel::Fastest {
+        if opt_level != OptLevel::None {
             self.postopt(isa)?;
-        }
-        if isa.flags().opt_level() == OptLevel::Best {
             self.compute_domtree();
             self.compute_loop_analysis();
             self.licm(isa)?;
@@ -151,13 +151,15 @@ impl Context {
         }
         self.compute_domtree();
         self.eliminate_unreachable_code(isa)?;
-        if isa.flags().opt_level() != OptLevel::Fastest {
+        if opt_level != OptLevel::None {
             self.dce(isa)?;
         }
         self.regalloc(isa)?;
         self.prologue_epilogue(isa)?;
-        if isa.flags().opt_level() == OptLevel::Best {
+        if opt_level == OptLevel::Speed || opt_level == OptLevel::SpeedAndSize {
             self.redundant_reload_remover(isa)?;
+        }
+        if opt_level == OptLevel::SpeedAndSize {
             self.shrink_instructions(isa)?;
         }
         let result = self.relax_branches(isa);
--- a/cranelift-codegen/meta/src/shared/settings.rs
+++ b/cranelift-codegen/meta/src/shared/settings.rs
@@ -8,11 +8,12 @@ pub fn define() -> SettingGroup {
         r#"
         Optimization level:
 
-        - default: Very profitable optimizations enabled, none slow.
-        - best: Enable all optimizations
-        - fastest: Optimize for compile time by disabling most optimizations.
+        - none: Minimise compile time by disabling most optimizations.
+        - speed: Generate the fastest possible code
+        - speed_and_size: like "speed", but also perform transformations
+          aimed at reducing code size.
         "#,
-        vec!["default", "best", "fastest"],
+        vec!["none", "speed", "speed_and_size"],
     );
 
     settings.add_bool(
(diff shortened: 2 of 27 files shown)


cranelift-codegen/src/context.rs
```diff
@@ -132,33 +132,37 @@ impl Context {
...
         self.legalize(isa)?;
-        if isa.flags().opt_level() != OptLevel::Fastest {
+        if opt_level != OptLevel::Fastest {
             self.postopt(isa)?;
         }
-        if isa.flags().opt_level() == OptLevel::Best {
+        if opt_level == OptLevel::NearlyBest || opt_level == OptLevel::Best {
             self.compute_domtree();
             self.compute_loop_analysis();
             self.licm(isa)?;
```
bjorn3: I believe I read somewhere LICM caused a wasm bench with clif to be slower. Also it being broken (#732) is the reason I don't enable opt_level=best in cg_clif.

cranelift-codegen/src/context.rs
```diff
@@ -132,33 +132,37 @@ impl Context {
...
+        if opt_level != OptLevel::Fastest {
             self.dce(isa)?;
         }
         self.regalloc(isa)?;
         self.prologue_epilogue(isa)?;
-        if isa.flags().opt_level() == OptLevel::Best {
+        if opt_level == OptLevel::NearlyBest || opt_level == OptLevel::Best {
             self.redundant_reload_remover(isa)?;
-            self.shrink_instructions(isa)?;
+            if opt_level == OptLevel::Best {
```
bjorn3: Maybe move this if out of the if above. Also performing this is necessary for gdb to show correct backtraces. (https://github.com/bjorn3/rustc_codegen_cranelift/issues/146#issuecomment-449474527)

sunfishcode:
I can also add, there hasn't been a lot of thought behind what "default" does, so one option would just be to just define "default" to be what this "nearly_best" option does.

julian-seward1:
How about this, then:
```
 OLD         NEW
 fastest     minimum    Behaviour unchanged
 default     default    Make this do what the proposed "nearly_best" does
 best        maximum    Behaviour unchanged
```
From the grammatical nitpicking aspect .. both `fastest` and `best` are ambiguous; they could describe the compiler's performance or the performance of generated code.  To me, changing to `OptLevel::{minimum, default, maximum}` feels generally cleaner and clearer.

jedisct1:
The names of the current optimization levels are indeed a little bit confusing.

For what it's worth, in Lucet, we don't use them. The API and CLI hide them behind more traditional optimization levels: `0`, `1`, `2`, `3` and `fast` (which is currently an alias for `3`).

lars-t-hansen:
Instead of picking names that require interpretation, if we're going to have three levels it would perhaps be appropriate to do "none", "speed", and "speed_and_size" (or none, fast, fast_and_small).  I can see the point of 0-3 but we're trying to avoid running the slow-and-ineffective shrinking pass and a linear scale does not capture that.  Really the minimal..maximal scale has the same problem IMO, as does the worst..best or anything else we're likely to think of.

julian-seward1:
Here's a new patch that implements @lars-t-hansen's suggestion.  In short:
```
 OLD         NEW
 fastest     none             Behaviour unchanged
 default     (deleted)        It never had any definition .. was meaningless
 best        speed_and_size   Behaviour unchanged
 (none)      speed            As 'speed_and_size', but omitting transformations
                              aimed only at reducing code size
```

bjorn3:
default used to perform preopt and postopt, which improve output quality without costing much in terms of compilation time.

julian-seward1:
How about doing preopt and postopt for `none`, then?

bjorn3:
For cold functions/a repl you might want the old fastest.

julian-seward1:
Alternatively, and I think I prefer this .. let's just leave the patch as it is.  That is, simply get rid of `default`.  If anybody later complains we can easily enough bring it back as `minimal`, which would be the same as `none` except that it would to pre-opt and post-opt.

cranelift-codegen/src/context.rs
```diff
@@ -132,33 +132,37 @@ impl Context {
...
         self.compute_domtree();
         self.eliminate_unreachable_code(isa)?;
-        if isa.flags().opt_level() != OptLevel::Fastest {
+        if opt_level != OptLevel::None {
             self.dce(isa)?;
         }
         self.regalloc(isa)?;
         self.prologue_epilogue(isa)?;
-        if isa.flags().opt_level() == OptLevel::Best {
+        if opt_level == OptLevel::Speed || opt_level == OptLevel::SpeedAndSize {
```
bnjbvr: Could this instead be `if opt_level != OptLevel::None`?

cranelift-codegen/src/context.rs
```diff
@@ -132,33 +132,37 @@ impl Context {
...
             self.redundant_reload_remover(isa)?;
-            self.shrink_instructions(isa)?;
+            if opt_level == OptLevel::SpeedAndSize {
```
bnjbvr: And could this be moved out of the `if` body?

cranelift-codegen/src/context.rs
bnjbvr: Yeah, it's time to ping Dan...

PR #1045, opened 2019-09-18
Some LiveRange improvements

bnjbvr:
This is a set of improvements for LiveRanges, updating comments, trying to make things simpler to understand, etc. Each commit does only one thing, so you can look at it commit per commit. No change in functionality here.

### DIFF
--- a/cranelift-codegen/src/regalloc/liverange.rs
+++ b/cranelift-codegen/src/regalloc/liverange.rs
@@ -58,8 +58,16 @@
 //!
 //! # Implementation notes
 //!
-//! A few notes about the implementation of this data structure. This should not concern someone
-//! only looking to use the public interface.
+//! A few notes about the implementation of the live intervals field `liveins`. This should not
+//! concern someone only looking to use the public interface.
+//!
+//! ## Current representation
+//!
+//! Our current implementation uses a B-tree map with the necessary interface for an efficient
+//! implementation of coalescing, implemented as a generic data-structure bforest::Map.
+//!
+//! A `BTreeMap<Ebb, Inst>` could have been used for the live-in intervals, but it doesn't provide
+//! the necessary API to make coalescing easy, nor does it optimize for our types' sizes.
 //!
 //! ## EBB ordering
 //!
@@ -99,13 +107,6 @@
 //! It is more complicated to work with, though, so it is probably not worth it. The performance
 //! benefits of switching to a numerical EBB order only appears if the binary search is doing
 //! EBB-EBB comparisons.
-//!
-//! ## B-tree representation
-//!
-//! A `BTreeMap<Ebb, Inst>` could also be used for the live-in intervals. It looks like the
-//! standard library B-tree doesn't provide the necessary interface for an efficient implementation
-//! of coalescing, so we would need to roll our own.
-//!
 
 use crate::bforest;
 use crate::entity::SparseMapValue;
--- a/cranelift-codegen/src/regalloc/liverange.rs
+++ b/cranelift-codegen/src/regalloc/liverange.rs
@@ -179,32 +179,6 @@ pub struct GenericLiveRange<PO: ProgramOrder> {
     po: PhantomData<*const PO>,
 }
 
-/// Context information needed to query a `LiveRange`.
-pub struct LiveRangeContext<'a, PO: 'a + ProgramOrder> {
-    /// Ordering of EBBs.
-    pub order: &'a PO,
-    /// Memory pool.
-    pub forest: &'a bforest::MapForest<Ebb, Inst>,
-}
-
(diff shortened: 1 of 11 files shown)


cranelift-codegen/src/verifier/liveness.rs
```diff
@@ -106,7 +106,9 @@ impl<'a> LivenessVerifier<'a> {
                         Some(lr) => lr,
                         None => return fatal!(errors, inst, "{} has no live range", val),
                     };
-                    if !self.live_at_use(lr, inst) {
+
+                    let ebb = lr_ctx.order.inst_ebb(inst).unwrap();
```
abrown: Should this be `let ebb = lr_ctx.order.inst_ebb(inst).ok_or(...)?;`?

cranelift-codegen/src/regalloc/liverange.rs
```diff
@@ -190,6 +190,22 @@ impl<'a, PO: ProgramOrder> bforest::Comparator<Ebb> for Cmp<'a, PO> {
...
+    ($order:ident, $a:ident >= $b:expr) => {
+        $order.cmp($a, $b) != Ordering::Less
+    };
+    ($order:ident, $a:ident < $b:expr) => {
+        $order.cmp($a, $b) == Ordering::Less
+    };
+    ($order:ident, $a:ident <= $b:expr) => {
+        $order.cmp($a, $b) != Ordering::Greater
+    };
+}
```
abrown: What about just implementing `Ord` and `Eq` on `ProgramOrder` or `ExpandedProgramPoint` or wherever it is needed so that the `<`, `<=`, `>`, etc. are overloaded and we can just use the operators directly instead of a macro or `order.cmp...`? It may require adding a method to `ProgramOrder` but the following seems more clear to me: `order.of(to) >= order.of(self.def_begin)`.

cranelift-codegen/src/verifier/liveness.rs
lars-t-hansen: If I'm reading this right, either the new `ebb` should be the same value as the previous binding (in which case why compute it) or we should assert that they are the same value?

bnjbvr:
Sorry I should have been more explicit about the intents of this PR: it was just me trying to make sense of this code, and trying to make it simpler to read and to understand in particular.

- Renaming GenLiveRange to GenericLiveRange: I think "Gen" leaves a doubt about whether we're talking about something Generic or a Generator, so making it explicit seemed worthwhile. That's probably a bias from my JS developer persona, although I heard Rust has generators too...
- Use LiveRange::reaches_use: Just trying to avoid code duplication there.
- Regalloc: remove the transient LiveRangeContext data structure: this was just used to pass parameters around, hiding its true purpose. In some cases, the structure was even constructed many times in a loop, with no good reasons. I thought it'd be simpler to just pass parameters: reducing the number of data structures that have no clear utility seems like a general valuable goal, to help understanding and reduce cognitive load. @bjorn3 you're right we could hide this behind a getter, thanks for the idea.
- Next commit updates a very outdated comment. Hopefully we want to keep documentation up to date :)
- Liverange: use a macro to make Order comparisons simpler to read: the inversion in semantics of a statement like "order.cmp(a, b) != Ordering::Less" slows comprehension down in my opinion. Even after seeing a lot of these, I wasn't used to it. Another personal bias probably. I think @abrown suggestion applies, though: we can't directly implement Equal/Compare on ProgramPoint since the order == layout uses extended information in its data structure to compare the values, but we can have the order return the relative number, and then compare this. In this case, we can even remove the cmp! macro entirely. Thanks for the suggestion again!

@lars-t-hansen does it all make sense, or would you like me to revert some of these commits? I don't have strong opinions here, these were just small refactorings that felt useful to comprehend the live range code, so I don't mind reverting any of them.

lars-t-hansen:
I'm fine with the commits.  I just think we should start enforcing meta-issues such as the justification for a patch.  My naive sense of the matter is that github discourages good hygiene with the ability to chain commits like you did here w/o individual commit messages, so it'll take vigilance and effort to improve the situation.

bjorn3:
> That's probably a bias from my JS developer persona, although I heard Rust has generators too...

Rust indeed has generators. They are the base of async/await.

cranelift-codegen/src/verifier/liveness.rs
bnjbvr: Indeed, we can assert it was the same. Previous checks verify that this instruction has been inserted, so it's fine to unwrap here.

cranelift-codegen/src/regalloc/liverange.rs
bnjbvr: Great suggestion; I'll open an issue to discuss this, since I'd like to make sure this is something we want first.

Issue #493, opened 2018-09-03
verifier: Diagnose incorrect uses of INVALID


sunfishcode:
Cranelift [has a VOID type](https://github.com/CraneStation/cranelift/blob/289035f7a136659ad7264cb16c7e0ebedc2b6f9c/lib/codegen/src/ir/types.rs#L23) but it's only used in a few contexts, because in general Cranelift prefers to use an array or vector of types which can be empty. As an example, it's particularly confusing that one can create a function with a return type of VOID, which turns out to be impossible to actually use. We should add verifier checks for this and other possible misuses of VOID.

sunfishcode:
Cranelift's VOID isn't actually used to indicate any actual typing concept, it's just a placeholder for IR data structures that have a type field but don't always need it. To hopefully clear up some of the understandable confusion, I've now renamed VOID to INVALID and updated the comments accordingly.

This issue can remain open, as we should still implement more verifier checks to ensure that INVALID doesn't appear in the IR, such as in function signatures.

ash-tamraz:
Where should the verifier checks to ensure that INVALID doesn't appear in the IR be implemented? I mean in terms of the files. Is it all in `cranelift-codegen/src/verifier`? I'm interested in working on this, where would be a good place to start?

sunfishcode:
Yes, this can be done entirely in `cranelift-codegen/src/verifier`. I think the best way to do this is to add new checks at the top of `Verifier::run` similar to `verify_global_values` etc.

PR #1046, opened 2019-09-18
Check for types::INVALID in the function signature

walac:
ref #493

### DIFF
--- a/cranelift-codegen/src/verifier/mod.rs
+++ b/cranelift-codegen/src/verifier/mod.rs
@@ -1837,12 +1837,58 @@ impl<'a> Verifier<'a> {
         Ok(())
     }
 
+    fn typecheck_function_signature(&self, errors: &mut VerifierErrors) -> VerifierStepResult<()> {
+        let pos_as_str = |i| match i {
+            0 => String::from("1st"),
+            1 => String::from("2nd"),
+            2 => String::from("3rd"),
+            v => format!("{}th", v + 1),
+        };
+
+        self.func
+            .signature
+            .params
+            .iter()
+            .enumerate()
+            .filter(|(_, &param)| param.value_type == types::INVALID)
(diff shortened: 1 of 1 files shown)


cranelift-codegen/src/verifier/mod.rs
```diff
@@ -1837,12 +1837,41 @@ impl<'a> Verifier<'a> {
         Ok(())
     }
 
+    fn verify_function_invalid_types(&self, errors: &mut VerifierErrors) -> VerifierStepResult<()> {
```
abrown: Maybe call this `typecheck_function_signature` to match `typecheck_entry_block_params`?

cranelift-codegen/src/verifier/mod.rs
```diff
@@ -1837,12 +1837,41 @@ impl<'a> Verifier<'a> {
...
+        self.func
+            .signature
+            .params
+            .iter()
+            .filter(|&param| param.value_type == types::INVALID)
+            .for_each(|_| report!(errors, AnyEntity::Function, "Parameter has an invalid type"));
```
abrown: It might be helpful to include the parameter that is invalid, e.g. `"Parameter {} has..."`

cranelift-codegen/src/verifier/mod.rs
walac: Done

cranelift-codegen/src/verifier/mod.rs
walac: Done. As the parameters have no names, I report the position of the parameter.

abrown:
@walac added a commit to tweak the message (very nitpicky... but see my commit message); if you're ok with that I will squash and merge.

walac:
lgtm!

abrown:
Thanks!

PR #1047, opened 2019-09-19
tidy: Remove extra semicolons

fitzgen:
These were causing compilation warnings.


fitzgen:
Prettyyyy sure the CI failure is not because of this PR :-p

abrown:
Yeah, don't think this broke fuzzing 😄. And it turns out I can re-run those checks. Let's see what happens this time.

fitzgen:
> And it turns out I can re-run those checks. Let's see what happens this time.

Nice! Looks like it's green now :)

PR #1048, opened 2019-09-19
Remove redundant semi-colons

abrown:
These were showing warnings when running `./test-all.sh`.


abrown:
Oops, guess I didn't see #1047.

PR #1049, opened 2019-09-19
Translate Wasm multi-value into cranelift IR

fitzgen:
f? @sunfishcode 

This is a work-in-progress.

Right now, it can translate the test WAT files into cranelift IR but the IR doesn't pass the verifier. I'd like to debug this further, but I'm in a little bit of a rut due to my unfamiliarity with this code base. As suggested on IRC, I'm trying to view the full IR, rather than just the snippet that gets printed with the verifier error, via `clif-util -p wasm`. Unfortunately, it looks like that command is still running the IR verifier, and is printing the same verifier error as `cargo test -p cranelift-wasm` does. I looked into `clif-util`'s flags and didn't see anything about disabling the verifier. Any tips here?

Additionally, if you have any suggestions for the code itself, they would be most welcome. I have some "TODO FITZGEN" comments in there that pose some design/architecture questions for things I wasn't super sure about as well.


fitzgen:
> Unfortunately, it looks like that command is still running the IR verifier, and is printing the same verifier error as `cargo test -p cranelift-wasm` does.

For example:

<details>

```
$ cargo run --bin clif-util -- wasm -p wasmtests/multi-1.wat --target riscv32-unknown-linux-gnu
   Compiling cranelift-tools v0.42.0 (/home/fitzgen/cranelift)
    Finished dev [unoptimized + debuginfo] target(s) in 6.14s
     Running `target/debug/clif-util wasm -p wasmtests/multi-1.wat --target riscv32-unknown-linux-gnu`
function u0:0(i64, i64, i32 vmctx) -> i64, i64, i64 system_v {
                                ebb0(v0: i64, v1: i64, v2: i32):
@0032                               v8 = iconst.i64 1234
@0035                               jump ebb2(v1, v0, v8)
; error: inst1: mismatched argument count for `jump ebb2(v1, v0, v8)`: got 3, expected 2


                                ebb2(v6: i64, v7: i64):
@0036                               jump ebb1(v0, v6, v7)

                                ebb1(v3: i64, v4: i64, v5: i64):
@0036                               return v3, v4, v5
}

; 1 verifier error detected (see above). Compilation aborted.
```

</details>

fitzgen:
Oh wait hm -- I think that is the full IR for this test case.

sunfishcode:
You can disable the verifier with `--set enable_verifier=false`, though yes, that is the complete IR for the function in question.

cranelift-wasm/src/code_translator.rs
```diff
@@ -133,19 +133,23 @@ pub fn translate_operator<FE: FuncEnvironment + ?Sized>(
          ***********************************************************************************/
         Operator::Block { ty } => {
             let next = builder.create_ebb();
-            if let Some(ty_cre) = blocktype_to_type(*ty)? {
+            for ty_cre in blocktype_to_types(environ, builder.func, *ty)? {
```
sunfishcode: This is subtle, but the `next` block here is the basic block we'll jump to for branches out of this wasm `block`. `blocktype_to_types` returns the params of the function type, but what we need here are the returns, because we're creating the block parameters for the block that we'll be exiting to.

cranelift-wasm/src/environ/dummy.rs
```diff
@@ -52,6 +52,17 @@ pub struct DummyModuleInfo {
...
     config: TargetFrontendConfig,
 
+    /// The raw Wasm type signatures for the module being translated. This is
+    /// necessary to keep around because multi-value blocks outline their params
+    /// and returns to signatures in the type section.
+    ///
+    /// TODO FITZGEN: should this be part of `signatures` or a secondary map or
+    /// something? `signatures` is a subset of `wasm_signatures`, where we won't
+    /// have any `signatures` entry for a wasm type signature that is only used
+    /// for multi-value blocks.
```
sunfishcode: One thing that occurs to me is that we can move `wasm_signatures` out of the trait implementation, and into cranelift-wasm proper, because all users of cranelift-wasm will need the same thing.

We could actually then use that to construct `signatures` entries lazily, so that it indeed does only hold those signatures used by functions.

cranelift-wasm/src/translation_utils.rs
```diff
@@ -138,20 +139,38 @@ pub fn tabletype_to_type(ty: wasmparser::Type) -> WasmResult<Option<ir::Type>> {
 }
 
 /// Helper function translating wasmparser block signatures to Cranelift types when possible.
-pub fn blocktype_to_type(ty_or_ft: wasmparser::TypeOrFuncType) -> WasmResult<Option<ir::Type>> {
+pub fn blocktype_to_types<FE: ?Sized + FuncEnvironment>(
+    environ: &mut FE,
+    func: &mut ir::Function,
+    ty_or_ft: wasmparser::TypeOrFuncType,
+) -> WasmResult<Vec<ir::Type>> {
```
bjorn3: Maybe use `Cow<'static, [ir::Type]>` to prevent allocation when it is not a multi-value block.

cranelift-wasm/src/environ/dummy.rs
fitzgen: Do you have an opinion about where the right place to put this is? The `TranslationState`?

cranelift-wasm/src/translation_utils.rs
fitzgen: Good idea :+1:

cranelift-wasm/src/code_translator.rs
```diff
@@ -1118,15 +1111,17 @@ fn translate_unreachable_operator(
...
             // so we don't have any branches anywhere.
-            state.push_if(ir::Inst::reserved_value(), ir::Ebb::reserved_value(), 0);
+            state.push_if(ir::Inst::reserved_value(), ir::Ebb::reserved_value(), 0, 0);
         }
         Operator::Loop { ty: _ } | Operator::Block { ty: _ } => {
-            state.push_block(ir::Ebb::reserved_value(), 0);
+            state.push_block(ir::Ebb::reserved_value(), 0, 0);
         }
         Operator::Else => {
+            println!("FITZGEN: other Else");
```
bjorn3: Just a reminder

fitzgen:
Ok, I think this is ready for a fresh round of review!

cranelift-wasm/src/environ/dummy.rs
sunfishcode: Yes, `TranslationState` sounds right.

cranelift-wasm/src/code_translator.rs
```diff
@@ -1154,6 +1150,8 @@ fn translate_unreachable_operator(
                     builder.change_jump_destination(branch_inst, else_ebb);
                     builder.seal_block(else_ebb);
                     builder.switch_to_block(else_ebb);
+                    let params = params.clone();
+                    state.pushn(&params);
```
sunfishcode: This is only called for unreachable code, so I'm not concerned about performance here, but it looks odd; why is a clone needed if you're just doing a borrow?

cranelift-wasm/src/state.rs
```diff
@@ -24,27 +24,41 @@ pub enum ControlStackFrame {
     If {
         destination: Ebb,
         branch_inst: Inst,
+        params: Vec<Value>,
```
sunfishcode: I'm concerned about the overhead of doing a dynamic allocation for each `If` and `Else` opcode that has parameters. Do you think it would be awkward to do something like this? On entry to an `If` with parameters, push a second copy of the parameter values onto the stack. Then, the `If` can use the second copy, and if there's an `Else`, it can use the first copy, and if not, we can just pop and discard the second copy.

cranelift-wasm/src/code_translator.rs
fitzgen: It's because `params` is borrowed immutably from within `state`, but calling `state.pushn` requires a `&mut self` which isn't possible while `params` is borrowed. Internally, it would be safe, since they operate on disjoint fields of `TranslationState`, but since the borrow checker can't see through method calls, it doesn't realize this.

There have been proposals for ways to work around this failing of the borrow checker, but nothing accepted and implemented yet. The only current work arounds are to either inline all the methods by hand so that the borrow checker sees everything, or make these things free functions that take the fields they operate on as parameters, rather than borrowing every field `&mut self`.

cranelift-wasm/src/state.rs
fitzgen: I can look into this. I'm a little concerned that it might be a little bit subtle / hard to follow for readers -- is translation usually CPU bottlenecked or IO/network bottlenecked?

fitzgen: Ok this was actually pretty straightforward! Done.

lars-t-hansen: FWIW, cranelift performance in Firefox is currently limited in part by the absolutely massive volume of small-vector allocations in cranelift, and it would be nice not to make this worse.  Partly all the malloc/free behavior is overhead in itself, especially with poisioning-on-free; partly jemalloc makes it much worse by limiting parallel speedup due to its architecture.  cc @julian-seward1 who's been looking into this.  Also see [this bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1578352).

julian-seward1: I understand @sunfishcode's concern here, but I'd suggest to leave this as-is for the time being.  We can profile it later with dhat and if it does significantly contribute to malloc traffic then we can maybe change it to a SmallVec.  That has been effective in the past, per the bug that @lars-t-hansen references, but using SmallVec isn't free, in the sense it's necessary to establish an optimal-ish inline size, and also it adds a conditional branch to all accesses.  So let's wait for the data.

fitzgen: Well, the double push of params to avoid the extra vec is already implemented, and I think it ended up being simpler than I had originally feared, so best of both worlds?

fitzgen:
Note: this isn't quite ready to land yet, I've found some bugs while running the spec test suite in wasmtime

fitzgen:
Update: I have most of the multi-value spec tests passing in wasmtime. The failing tests fall into these buckets:

* There is the expected failure for functions with many return values, where we only currently support returning as many values as can fit in registers. So we either need a way to disable certain functions within a spec test file (more of an issue for wasmtime) or we need to implement returning `>N` values via stack pointer (not sure yet if this is more of an embedder/environ responsibility or if cranelift itself needs to get involved). Opinions on which approach to take, @sunfishcode?

* There is something funky going on with the factorial in multi-value test. It seems to infinite loop and I'm not 100% sure why yet. I'm trying to read the cranelift IR that we translate the wasm into, but I'm running into a couple snags. I can't seem to figure out how to look at the IR exactly as it comes out of `cranelift-wasm`, before any target-specific legalization passes or optimizations or anything of that sort. The x86-64-specific IR has a lot of noise in it, which is making it pretty hard to read. Also the local, direct function calls are getting turned into `call_indirect`s when I use `clif-util`, which is pretty confusing, and I'm not sure whether it is also happening with wasmtime and running the spec tests or not.

  @sunfishcode any way to dump the pre-legalization/etc IR with `clif-util`?

fitzgen:
>     * There is something funky going on with the factorial in multi-value test. It seems to infinite loop and I'm not 100% sure why yet.

FWIW, I think I am not properly popping the initial loop params and then pushing the ebb's params. Will look into it tomorrow. The question about dumping the IR still stands though :)

bnjbvr:
`clif-util compile -dtp ./file.wasm` only does the translation (-t), prints resulting IR (-p) in stdout (-d). Is this sufficient for your needs?

fitzgen:
> FWIW, I think I am not properly popping the initial loop params and then pushing the ebb's params.

Yep, this was it. This PR passes all the multi-value spec tests other than the ones that test more return values than we can fit in registers.

Issue #1015, opened 2019-09-10
Add SIMD encodings for older SSE releases


abrown:
As @bnjbvr mentioned in https://github.com/CraneStation/cranelift/pull/943#discussion_r322831206, some cranelift users may not have the most recent SSE versions (e.g. SSE4.1). In these cases, we may need to provide alternate encodings or legalizations in order to implement certain SIMD instructions.

abrown:
For reference: https://data.firefox.com/dashboard/hardware

Issue #896, opened 2019-08-12
icmp_imm implemented incorrectly for i8 + eq/ne


bjorn3:
https://github.com/CraneStation/cranelift/blob/57fa45c8520110ee7e9b216538f1990376aed0e9/cranelift-codegen/meta/src/shared/legalize.rs#L434-L435

For example

```
v0 = iconst.i8 -128
v1 = iconst.i8 -128
v2 = icmp ne v0, v1
```

gets turned into:

```
v7 = iconst.i32 -128
v0 = ireduce.i8 v7
v9 = uextend.i32 v0
v2 = icmp_imm ne v9, -128
```

Which is equal to:

```
v9 = iconst.i32 128
v2 = icmp_imm ne v9, -128
```

That is obviously false.

bjorn3:
This is similar to #815, though a fix for it may not fix this, as the problem here is using `uextend` instead of `sextend`.

Issue #793, opened 2019-06-11
ineg


calebmer:
Currently there is no `ineg` instruction but there is an `fneg` instruction. Are there plans to add an `ineg` instruction or should one use `imul` or one of the bitwise functions?

sunfishcode:
`fneg` is special, since it's defined in IEEE 754 as being a pure bitwise operation (invert the sign bit), which makes it diffferent from `0 - x` or `-1 * x` since those forms interpret NaNs differently. For integers, NaNs aren't an issue, so you can just use a subtract and do `0 - x`.

That said, x86 and aarch64 do have a `neg` instruction, and we will eventually want to pattern-match `0 - x` and other patterns to use it, so I think we do eventually want to add an `ineg` to cranelift IR to represent that.

bnjbvr:
@abrown said in #1053:

> On x86, the new `ineg` instruction added for SIMD support (see #1050) can also be encoded for scalar integers using the `NEG` instruction. To do so, however, may require some refactoring of `PerCpuModeEncodings` to more easily pass the bits needed for the reg field of the ModR/M byte (see comment on [#1050 (comment)](https://github.com/CraneStation/cranelift/pull/1050#discussion_r326724852)). Once there is a cleaner way to add those bits, the `ineg` instruction can be encoded for x86.

PR #1050, opened 2019-09-20
Add SIMD arithmetic instructions on x86

abrown:
Includes:
 - new `ineg` instruction
 - `isub` encodings
 - `imul` encodings, but only for 16-bit and 32-bit
 - saturating addition and subtraction with new instructions `sadd_sat`, `uadd_sat`, `ssub_sat`, `usub_sat`
 - a fix for an incorrect recipe in `extractlane.t16x8`

### DIFF
--- a/cranelift-codegen/meta/src/isa/x86/encodings.rs
+++ b/cranelift-codegen/meta/src/isa/x86/encodings.rs
@@ -480,6 +480,7 @@ pub(crate) fn define(
     let sqrt = shared.by_name("sqrt");
     let sshr = shared.by_name("sshr");
     let sshr_imm = shared.by_name("sshr_imm");
+    let ssub_sat = shared.by_name("ssub_sat");
     let stack_addr = shared.by_name("stack_addr");
     let store = shared.by_name("store");
     let store_complex = shared.by_name("store_complex");
@@ -501,6 +502,7 @@ pub(crate) fn define(
     let uload8_complex = shared.by_name("uload8_complex");
     let ushr = shared.by_name("ushr");
     let ushr_imm = shared.by_name("ushr_imm");
+    let usub_sat = shared.by_name("usub_sat");
     let vconst = shared.by_name("vconst");
     let x86_bsf = x86.by_name("x86_bsf");
     let x86_bsr = x86.by_name("x86_bsr");
@@ -1964,6 +1966,24 @@ pub(crate) fn define(
         let isub = isub.bind_vector_from_lane(ty.clone(), sse_vector_size);
         e.enc_32_64(isub, rec_fa.opcodes(*opcodes));
     }
+
+    // SIMD integer saturating subtraction
+    e.enc_32_64(
+        ssub_sat.bind_vector_from_lane(I8, sse_vector_size),
+        rec_fa.opcodes(&PSUBSB),
+    );
+    e.enc_32_64(
+        ssub_sat.bind_vector_from_lane(I16, sse_vector_size),
+        rec_fa.opcodes(&PSUBSW),
+    );
+    e.enc_32_64(
+        usub_sat.bind_vector_from_lane(I8, sse_vector_size),
+        rec_fa.opcodes(&PSUBUSB),
+    );
+    e.enc_32_64(
+        usub_sat.bind_vector_from_lane(I16, sse_vector_size),
+        rec_fa.opcodes(&PSUBUSW),
+    );
 
     // SIMD integer multiplication: the x86 ISA does not have instructions for multiplying I8x16
     // and I64x2 and these are (at the time of writing) not necessary for WASM SIMD.
--- a/cranelift-codegen/meta/src/isa/x86/encodings.rs
+++ b/cranelift-codegen/meta/src/isa/x86/encodings.rs
@@ -1945,6 +1945,16 @@ pub(crate) fn define(
         e.enc_32_64(isub, rec_fa.opcodes(*opcodes));
     }
 
+    // SIMD integer multiplication: the x86 ISA does not have instructions for multiplying I8x16
+    // and I64x2 and these are (at the time of writing) not necessary for WASM SIMD.
+    for (ty, opcodes, isap) in &[
+        (I16, &PMULLW[..], None),
+        (I32, &PMULLD[..], Some(use_sse41_simd)),
+    ] {
+        let imul = imul.bind_vector_from_lane(ty.clone(), sse_vector_size);
+        e.enc_32_64_maybe_isap(imul, rec_fa.opcodes(opcodes), *isap);
+    }
+
     // SIMD icmp using PCMPEQ*
     let mut pcmpeq_mapping: HashMap<u64, (&[u8], Option<SettingPredicateNumber>)> = HashMap::new();
     pcmpeq_mapping.insert(8, (&PCMPEQB, None));
--- a/cranelift-codegen/meta/src/isa/x86/encodings.rs
+++ b/cranelift-codegen/meta/src/isa/x86/encodings.rs
@@ -464,6 +464,7 @@ pub(crate) fn define(
     let rotl_imm = shared.by_name("rotl_imm");
     let rotr = shared.by_name("rotr");
     let rotr_imm = shared.by_name("rotr_imm");
+    let sadd_sat = shared.by_name("sadd_sat");
     let safepoint = shared.by_name("safepoint");
     let scalar_to_vector = shared.by_name("scalar_to_vector");
     let selectif = shared.by_name("selectif");
@@ -490,6 +491,7 @@ pub(crate) fn define(
     let trueff = shared.by_name("trueff");
     let trueif = shared.by_name("trueif");
     let trunc = shared.by_name("trunc");
+    let uadd_sat = shared.by_name("uadd_sat");
     let uextend = shared.by_name("uextend");
     let uload16 = shared.by_name("uload16");
     let uload16_complex = shared.by_name("uload16_complex");
@@ -1938,6 +1940,24 @@ pub(crate) fn define(
         let iadd = iadd.bind_vector_from_lane(ty.clone(), sse_vector_size);
         e.enc_32_64(iadd, rec_fa.opcodes(*opcodes));
     }
+
+    // SIMD integer saturating addition
+    e.enc_32_64(
+        sadd_sat.bind_vector_from_lane(I8, sse_vector_size),
+        rec_fa.opcodes(&PADDSB),
+    );
+    e.enc_32_64(
+        sadd_sat.bind_vector_from_lane(I16, sse_vector_size),
+        rec_fa.opcodes(&PADDSW),
+    );
+    e.enc_32_64(
+        uadd_sat.bind_vector_from_lane(I8, sse_vector_size),
+        rec_fa.opcodes(&PADDUSB),
+    );
+    e.enc_32_64(
+        uadd_sat.bind_vector_from_lane(I16, sse_vector_size),
+        rec_fa.opcodes(&PADDUSW),
+    );
 
     // SIMD integer subtraction
     for (ty, opcodes) in &[(I8, &PSUBB), (I16, &PSUBW), (I32, &PSUBD), (I64, &PSUBQ)] {
--- a/cranelift-codegen/meta/src/isa/x86/encodings.rs
+++ b/cranelift-codegen/meta/src/isa/x86/encodings.rs
@@ -1940,6 +1940,12 @@ pub(crate) fn define(
         e.enc_32_64(iadd, rec_fa.opcodes(*opcodes));
     }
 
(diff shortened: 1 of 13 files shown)


cranelift-codegen/meta/src/shared/instructions.rs
```diff
@@ -1685,6 +1685,38 @@ pub(crate) fn define(
         .operands_out(vec![a]),
     );
 
+    ig.push(
+        Inst::new(
+            "uadd_sat",
```
bjorn3: Can these new instructions (`*_sat`, `ineg`) have non simd encodings too? That would make cg_clif easier and codegen better.

cranelift-codegen/src/isa/x86/enc_tables.rs
```diff
@@ -1090,3 +1091,28 @@ fn convert_insertlane(
         }
     }
 }
+
+/// For SIMD negation, convert an `irsub_imm` to a `vconst + isub`
```
bjorn3:
```suggestion
/// For SIMD negation, convert an `ineg` to a `vconst + isub`
```

cranelift-codegen/src/isa/x86/enc_tables.rs
```diff
@@ -1090,3 +1091,28 @@ fn convert_insertlane(
...
+
+    if let ir::InstructionData::Unary {
+        opcode: ir::Opcode::Ineg,
+        arg,
+    } = pos.func.dfg[inst]
+    {
+        let value_type = pos.func.dfg.value_type(arg);
+        if value_type.is_vector() && value_type.lane_type().is_int() {
+            let constant = pos.func.dfg.constants.insert(V128Imm::from(0).to_vec());
+            let vconst = pos.ins().vconst(value_type, constant); // if imm == 0 this should be
```
bjorn3:
```suggestion
            let zero_val = pos.ins().vconst(value_type, zero_imm); // this should be legalized to a PXOR
```

cranelift-codegen/src/isa/x86/enc_tables.rs
```diff
@@ -1090,3 +1091,28 @@ fn convert_insertlane(
...
+    pos.use_srcloc(inst);
+
+    if let ir::InstructionData::Unary {
+        opcode: ir::Opcode::Ineg,
+        arg,
+    } = pos.func.dfg[inst]
+    {
+        let value_type = pos.func.dfg.value_type(arg);
+        if value_type.is_vector() && value_type.lane_type().is_int() {
+            let constant = pos.func.dfg.constants.insert(V128Imm::from(0).to_vec());
```
bjorn3:
```suggestion
            let zero_imm = pos.func.dfg.constants.insert(V128Imm::from(0).to_vec());
```

cranelift-codegen/meta/src/shared/instructions.rs
abrown: `ineg` could use x86 `NEG` but I think saturating add/sub would have to be legalized (e.g. https://stackoverflow.com/a/166393). This could be done in a separate PR, though.

cranelift-codegen/src/isa/x86/enc_tables.rs
abrown: I used fixup commits on the original commit for a cleaner history but thanks for these changes (this one and the ones below). Wish I could apply these as fixup commits and have GitHub autosquash them in...

cranelift-codegen/meta/src/shared/instructions.rs
abrown: You think I should add the `NEG` encoding to this one?

bjorn3: Yes please.

abrown: He, easier said than done apparently. The `NEG` instructions wants to insert `3` in the reg field of the ModR/M byte. Not hard, right? But the recipe/template distinction and `PerCpuModeEncodings` make it hard to put this `3` in `EncodingBuilder::new(..., ..., bits)`. Do you see a good way to do this? I've been thinking about refactor `PerCpuModeEncodings`...

bjorn3: I am not familiar with `PerCpuEncodings`. It's fine to leave it out of this PR if it is hard to add. I am currently using `v1 = iconst 0` + `isub v1, v0` in cg_clif.

abrown: Ok, let me add it as an issue: https://github.com/CraneStation/cranelift/issues/1053.

bnjbvr:
(Taking the review, unless somebody beats me to it.)

filetests/isa/x86/simd-arithmetic.clif
```diff
@@ -0,0 +1,220 @@
...
+    v3 = extractlane v2, 0
+    v4 = icmp_imm eq v3, 65535
+
+    return v4
+}
+; run
+
+function %sub_sat_i8x16() -> b1 {
+ebb0:
+[-, %xmm2]    v0 = vconst.i8x16 [128 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0] ; 120 == 0x80 == -128
```
bjorn3:
```suggestion
[-, %xmm2]    v0 = vconst.i8x16 [128 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0] ; 128 == 0x80 == -128
```

cranelift-codegen/meta/src/isa/x86/encodings.rs
```diff
@@ -1798,23 +1798,22 @@ pub(crate) fn define(
...
-    x86_pextr_mapping.insert(64, (&PEXTR, Some(use_sse41_simd)));
+    let mut x86_pextr_mapping: HashMap<u64, &'static [u8]> = HashMap::new();
+    x86_pextr_mapping.insert(8, &PEXTRB);
+    x86_pextr_mapping.insert(16, &PEXTRW);
+    x86_pextr_mapping.insert(32, &PEXTR);
+    x86_pextr_mapping.insert(64, &PEXTR);
 
     for ty in ValueType::all_lane_types().filter(allowed_simd_type) {
-        if let Some((opcode, isap)) = x86_pextr_mapping.get(&ty.lane_bits()) {
+        if let Some(opcode) = x86_pextr_mapping.get(&ty.lane_bits()) {
```
bnjbvr: (unrelated, feel free to dismiss and we can do it in a follow-up, but) I realized we can avoid the HashMap with code doing:

```
let encoding = match ty.lane_bits() {
  8 => &PEXTRB,
  16 => &PEXTRW,
  ...
  _ => { continue; } // or a panic if that makes sense, but i don't think so
};
```

cranelift-codegen/meta/src/isa/x86/encodings.rs
```diff
@@ -1798,23 +1798,22 @@ pub(crate) fn define(
...
     // SIMD extractlane
-    let mut x86_pextr_mapping: HashMap<u64, (&'static [u8], Option<SettingPredicateNumber>)> =
-        HashMap::new();
-    x86_pextr_mapping.insert(8, (&PEXTRB, Some(use_sse41_simd)));
-    x86_pextr_mapping.insert(16, (&PEXTRW_SSE2, None));
-    x86_pextr_mapping.insert(32, (&PEXTR, Some(use_sse41_simd)));
-    x86_pextr_mapping.insert(64, (&PEXTR, Some(use_sse41_simd)));
+    let mut x86_pextr_mapping: HashMap<u64, &'static [u8]> = HashMap::new();
+    x86_pextr_mapping.insert(8, &PEXTRB);
+    x86_pextr_mapping.insert(16, &PEXTRW);
```
bnjbvr: (meta comment about this commit's message: using `#1` in the message will mention unrelated issues. Not a big deal of course.)

Also, would we want to still have the shorter encoding in the future, when we want to optimize for size (and don't have access to SSE4.1)?

cranelift-codegen/meta/src/shared/instructions.rs
```diff
@@ -1733,6 +1733,36 @@ pub(crate) fn define(
...
+        .operands_out(vec![a]),
+    );
+
+    ig.push(
+        Inst::new(
+            "ssub_sat",
+            r#"
+        Subtract with signed saturation.
+
+        This is similar to `iadd` but the operands are interpreted as signed integers and their
```
bnjbvr: nit: "iadd" => "isub"

filetests/isa/x86/simd-arithmetic.clif
```diff
@@ -190,3 +190,31 @@ ebb0:
...
 ; run
+
+function %sub_sat_i8x16() -> b1 {
+ebb0:
+[-, %xmm2]    v0 = vconst.i8x16 [128 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0] ; 120 == 0x80 == -128
+[-, %xmm3]    v1 = vconst.i8x16 [1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1]
+
+[-, %xmm2]    v2 = ssub_sat v0, v1 ; bin: 66 0f e8 d3
+    v3 = extractlane v2, 0
+    v4 = icmp_imm eq v3, 0x80 ; still -128, TODO it's unclear why I can't use -128 here
```
bnjbvr: Is it the same sign-extending issue that showed up in the other test?

cranelift-codegen/meta/src/isa/x86/encodings.rs
abrown: Yes, we will need the shorter encoding to finish #1015 but for now I thought it best to get rid of it; it just caused confusion when investigating this bug because it is a valid opcode but requires a completely different recipe. It will be easy to re-add in #1015 when we get to that.

cranelift-codegen/src/isa/x86/enc_tables.rs
```diff
@@ -1090,3 +1091,27 @@ fn convert_insertlane(
...
+        opcode: ir::Opcode::Ineg,
+        arg,
+    } = pos.func.dfg[inst]
+    {
+        let value_type = pos.func.dfg.value_type(arg);
+        if value_type.is_vector() && value_type.lane_type().is_int() {
+            let zero_immediate = pos.func.dfg.constants.insert(V128Imm::from(0).to_vec());
+            let zero_value = pos.ins().vconst(value_type, zero_immediate); // this should be legalized to a PXOR
+            pos.func.dfg.replace(inst).isub(zero_value, arg);
+        }
```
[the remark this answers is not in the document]
abrown: There is a `NEG` in x86 so I think we will just encode `ineg` to that for scalar integers; that work has #793. On @bjorn3's request I tried to add it as a part of this PR but ran into issues (https://github.com/CraneStation/cranelift/pull/1050#discussion_r326724852)--maybe you see an easy way to do that without re-factoring a lot?

filetests/isa/x86/simd-arithmetic.clif
abrown: Probably; I looked at the CLIF parsing (I thought something in `ir::immediates::parse_i64` was preventing me from using negative integers) but it's not clear to me why this doesn't work.

filetests/isa/x86/simd-arithmetic.clif
```diff
@@ -1,5 +1,6 @@
 test run
 test binemit
+test legalizer
```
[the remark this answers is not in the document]
abrown: Good call!

cranelift-codegen/meta/src/isa/x86/encodings.rs
abrown: Sounds good to me: https://github.com/CraneStation/cranelift/issues/1091.

filetests/isa/x86/simd-arithmetic.clif
abrown: Thanks; rebased the change in.

filetests/isa/x86/simd-arithmetic.clif
bjorn3: I believe `v3` is an `i8`, so `icmp_imm eq v3, -128` would get zero extended to `v99 = uextend.i32 v3; icmp_imm eq v99, -128`, which is not true, as `v99` would be `0x80`, not `-128`.

See https://github.com/CraneStation/cranelift/issues/896

abrown: Ah, thanks! Let me change the comment from a TODO to a note to clarify this.

cranelift-codegen/src/isa/x86/enc_tables.rs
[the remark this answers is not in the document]
bnjbvr: You're right; in the meanwhile, even materializing 0 and the isub would be nice, just so people trying to use `ineg` on scalar values don't run into ghost instructions issues. I'll look at #793 though.

Issue #1057, opened 2019-09-22
Allow switching between EBBs before they are filled


jyn514:
Copied from IRC:

Is there a way to switch to another ebb and then switch back to the current one? I'm getting 'you have to fill your block before switching' when I try. My use case is I want to generate a switch statement without knowing all the jump EBBs ahead of time. I'm generating each ebb, which calls `switch.set_entry` as it gets translated, and then at the end switching back to the original block and calling `switch.emit`. Looking at https://docs.rs/cranelift-frontend/0.43.1/src/cranelift_frontend/frontend.rs.html#256-274 there doesn't seem to be any intrinsic reason to not switch blocks, it's just a debug assertion.

I think this hasn't popped up before because Rust's `match` statements have very clear entry points and exit points, unlike C (how many EBBs are in [this code](https://godbolt.org/z/5jBPZ7)?) and `cranelift-wasm` [uses `br_table` directly](https://github.com/CraneStation/cranelift/blob/master/cranelift-wasm/src/code_translator.rs#L267) instead of a `Switch`.

jyn514:
I have a prototype using this locally (I used `--release` to disable debug assertions) and there doesn't seem to be an issue with the verifier or anything like that.

jyn514:
If this was intended as a debugging aid, maybe there could be a setting to enable/disable it in `settings::builder`? That way you get the checks by default, but can still turn them off if you know what you're doing.

nbp:
For your information, we are going to drop EBB's support soon.

jyn514:
@nbp would that affect whether users of `FunctionBuilder` can switch between blocks without filling the Basic Block?

nbp:
I presume yes, as we have restriction on the terminator of the basic blocks. On the other hand, you can always end you basic block with a jump to a block which is allocated right after it, and jump to a different block to fill it.  The problem then would be to reorder the block later on to ensure that we can use `fallthrough` instructions instead of `jump`.

jyn514:
I don't think this is worth the effort to change, I worked around it in some of my own code.

PR #1062, opened 2019-09-22
Don't require block to be filled before switching

jyn514:
Closes https://github.com/CraneStation/cranelift/issues/1057

### DIFF
--- a/cranelift-frontend/src/frontend.rs
+++ b/cranelift-frontend/src/frontend.rs
@@ -254,14 +254,6 @@ impl<'a> FunctionBuilder<'a> {
     /// successor), the block will be declared filled and it will not be possible to append
     /// instructions to it.
     pub fn switch_to_block(&mut self, ebb: Ebb) {
-        // First we check that the previous block has been filled.
-        debug_assert!(
-            self.position.is_default()
-                || self.is_unreachable()
-                || self.is_pristine()
-                || self.is_filled(),
-            "you have to fill your block before switching"
-        );
         // We cannot switch to a filled block
         debug_assert!(
             !self.func_ctx.ebbs[ebb].filled,
--- a/cranelift-frontend/src/frontend.rs
+++ b/cranelift-frontend/src/frontend.rs
@@ -68,10 +68,6 @@ impl Position {
             basic_block: PackedOption::from(basic_block),
         }
     }
-
-    fn is_default(&self) -> bool {
-        self.ebb.is_none() && self.basic_block.is_none()
-    }
 }
 
 impl FunctionBuilderContext {
--- a/cranelift-codegen/src/verifier/mod.rs
+++ b/cranelift-codegen/src/verifier/mod.rs
@@ -1885,6 +1885,9 @@ impl<'a> Verifier<'a> {
         self.typecheck_function_signature(errors)?;
 
         for ebb in self.func.layout.ebbs() {
+            if self.func.layout.first_inst(ebb).is_none() {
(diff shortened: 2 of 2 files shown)


cranelift-frontend/src/frontend.rs
```diff
@@ -254,14 +250,6 @@ impl<'a> FunctionBuilder<'a> {
...
     /// instructions to it.
     pub fn switch_to_block(&mut self, ebb: Ebb) {
-        // First we check that the previous block has been filled.
-        debug_assert!(
-            self.position.is_default()
-                || self.is_unreachable()
-                || self.is_pristine()
-                || self.is_filled(),
-            "you have to fill your block before switching"
-        );
```
bjorn3: Without this, there is no check that all switched to ebbs are filled. Also would it be possible to instead of switch to an unfilled block, create a new ebb, add a jump to it and then switch to the target ebb and finally continue with the new ebb if necessary?

```
ebb0:
; some code
; need to switch to ebb1
; create ebb2
jump ebb2

ebb1:
; some other code

ebb2:
; switched to ebb2 instead of ebb0
```

jyn514: > Without this, there is no check that all switched to ebbs are filled. 

Shouldn't that sort of check live in the verifier? I can write a pass as part of this PR if it doesn't exist already.

> instead of switch to an unfilled block, create a new ebb

Would that dead code be eliminated by the optimizer?

bjorn3: > Shouldn't that sort of check live in the verifier? I can write a pass as part of this PR if it doesn't exist already.

After adding one instruction, it does. When there is no instruction added yet, it doesnt however.

> Would that dead code be eliminated by the optimizer?

Not in all cases.

blittable:
Unsubscribe

cranelift-frontend/src/frontend.rs
jyn514: I added a verifier pass for empty EBBs.

jyn514:
This is ready for review.

sunfishcode:
The tricky case here is when you switch blocks, when the block you're switching to has an unfilled block as a predecessor. If the SSA algorithm needs to look at the predecessors, it needs them to be filled. Do you know if this is covered by an assert anywhere?

Also, we should update [this comment](https://github.com/CraneStation/cranelift/blob/master/cranelift-frontend/src/frontend.rs#L216) in frontend.rs.

jyn514:
Ah, and that runs before the verifier. Yeah that is tricky, and since it can be run in multiple places can't be fixed with a single assert ... Could we make a wrapper around `Vec<Inst>` that `panic!`s if the invariants aren't upheld?

jyn514:
It looks like this is asserted in at least one place: https://github.com/CraneStation/cranelift/blob/master/cranelift-codegen/src/dominator_tree.rs#L441

jyn514:
Maybe it would be easier to ensure that blocks that contain only a single jump are optimized out? Then I could use use a temporary EBB as a workaround [as mentioned by @bjorn3](https://github.com/CraneStation/cranelift/pull/1062#discussion_r326960931).

sunfishcode:
What if we added a function to `FunctionBuilder` named `declare_ebb_predecessor` which wraps `self.func_ctx.ssa.declare_ebb_predecessor` and asserts that the predecessor is filled, and then made all calls to `self.func_ctx.ssa.declare_ebb_predecessor` call that wrapper instead?

jyn514:
I don't think this is worth the effort to change, I worked around it in some of my own code.

PR #1065, opened 2019-09-23
Re-export ReplaceBuilder

XAMPPRocky:
ReplaceBuilder is available in the public API through [`DataFlowGraph::replace`](https://docs.rs/cranelift-codegen/0.43.1/cranelift_codegen/ir/dfg/struct.DataFlowGraph.html#method.replace), however it's documentation is not available through rustdoc as the type isn't publicly importable.


pchickey:
Thanks for the patch! This just needs formatting fixes - please run `./format-all.sh` in the root of the repo.

XAMPPRocky:
@pchickey Done!

PR #1073, opened 2019-09-24
Add wasm reference/pointers translation.

yurydelendik:
cranelift-wasm is missing conversion of wasm reftypes to cranelift IR. The PR adds capability to translate ref params and locals to IR.

Example: https://gist.github.com/yurydelendik/47b51002f998bccbcad86ace5b62d3d4

/cc @fitzgen

### DIFF
--- a/cranelift-wasm/src/environ/spec.rs
+++ b/cranelift-wasm/src/environ/spec.rs
@@ -103,12 +103,8 @@ pub enum ReturnMode {
     FallthroughReturn,
 }
 
-/// Environment affecting the translation of a single WebAssembly function.
-///
-/// A `FuncEnvironment` trait object is required to translate a WebAssembly function to Cranelift
-/// IR. The function environment provides information about the WebAssembly module as well as the
-/// runtime environment.
-pub trait FuncEnvironment {
+/// Environment affecting the translation of a WebAssembly.
+pub trait TargetEnvironment {
     /// Get the information needed to produce Cranelift IR for the given target.
     fn target_config(&self) -> TargetFrontendConfig;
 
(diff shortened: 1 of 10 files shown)


cranelift-wasm/src/environ/spec.rs
```diff
@@ -141,6 +129,23 @@ pub trait FuncEnvironment {
...
+}
+
+/// Environment affecting the translation of a single WebAssembly function.
+///
+/// A `FuncEnvironment` trait object is required to translate a WebAssembly function to Cranelift
+/// IR. The function environment provides information about the WebAssembly module as well as the
+/// runtime environment.
+pub trait FuncEnvironment: PointerEnvironment {
+    /// Get the information needed to produce Cranelift IR for the given target.
+    fn target_config(&self) -> TargetFrontendConfig;
```
fitzgen: Since `PointerEnvironment` has this method, it shouldn't be necessary for `FuncEnvironment` to have it anymroe. I kind of wanted to factor this method out into its own trait anyways, since `ModuleEnvironment` requires it as well.

cranelift-wasm/src/environ/spec.rs
```diff
@@ -301,7 +306,7 @@ pub trait FuncEnvironment {
 /// An object satisfying the `ModuleEnvironment` trait can be passed as argument to the
 /// [`translate_module`](fn.translate_module.html) function. These methods should not be called
 /// by the user, they are only for `cranelift-wasm` internal use.
-pub trait ModuleEnvironment<'data> {
+pub trait ModuleEnvironment<'data>: PointerEnvironment {
     /// Get the information needed to produce Cranelift IR for the current target.
     fn target_config(&self) -> TargetFrontendConfig;
```
fitzgen: Ditto.

bnjbvr:
Thanks for working on this. Just to make sure to understand how this moves the needle: this only translates from wasm to clif (obviously), but since enable_safepoints is still a thing, embedders need to explicitly enable it, which means implementing more on their side. Is that right? (I am just checking that this won't break e.g. all of Spidermonkey testing.0)

yurydelendik:
> since enable_safepoints is still a thing, embedders need to explicitly enable it, which means implementing more on their side

Correct, this patch mostly prepares cranelift API surface to add wasmtime anyref tests and, for me, to continue wasmtime-api work. You need to specify `--set enable_safepoints=true` in the clif-util.

bnjbvr:
Hey @yurydelendik, no pressure here, just wanted to check in what the status of this PR is, and if there's a way we can help?`

yurydelendik:
>  just wanted to check in what the status of this PR is, and if there's a way we can help?

I'm seeking the initial feedback, and if I'm close to the "target" I'll rebase that in timely manner. Either way #1176 is the related issue and might influence this PR.

PR #1077, opened 2019-09-25
Convert SIMD load and store to their respective CLIF instructions

abrown:
For x86 at least, the encodings for SIMD `load` and `store` already exist (added as a part of #992). This adds the WASM-to-CLIF conversion for these instructions.

### DIFF
--- a/cranelift-wasm/src/code_translator.rs
+++ b/cranelift-wasm/src/code_translator.rs
@@ -500,6 +500,11 @@ pub fn translate_operator<FE: FuncEnvironment + ?Sized>(
         } => {
             translate_load(*offset, ir::Opcode::Load, F64, builder, state, environ)?;
         }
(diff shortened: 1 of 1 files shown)


cranelift-wasm/src/code_translator.rs
```diff
@@ -928,6 +928,16 @@ pub fn translate_operator<FE: FuncEnvironment + ?Sized>(
             // the v128.const is typed in CLIF as a I8x16 but raw_bitcast to a different type before use
             state.push1(value)
         }
+        Operator::V128Load {
+            memarg: MemoryImmediate { offset, .. },
```
bnjbvr: For consistency, can you please put this one next to the other load operators (resp. store operators), and make it more apparent that the `flags` immediate isn't used right now? (as is done for other memory operators)

PR #1078, opened 2019-09-25
[codegen] add documentation for overflow

ryzokuken:
Add documentation to the icmp instruction text for both signed and
unsigned overflow, making it very clear why unsigned overflow is
complicated and where to find it.

/cc @wingo @caitp

@sunfishcode @bnjbvr PTAL.

### DIFF
--- a/cranelift-codegen/meta/src/shared/instructions.rs
+++ b/cranelift-codegen/meta/src/shared/instructions.rs
@@ -1586,8 +1586,6 @@ pub(crate) fn define(
     let x = &operand("x", Int);
     let y = &operand("y", Int);
 
-    // TODO(ryzokuken): Add documentation for unsigned overflow.
-    // TODO(ryzokuken): Add documentation for signed overflow.
     ig.push(
         Inst::new(
             "icmp",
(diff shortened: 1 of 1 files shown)


cranelift-codegen/meta/src/shared/instructions.rs
```diff
@@ -1606,10 +1604,15 @@ pub(crate) fn define(
...
         sle    ule      Less than or equal
+        of     *        Overflow
+        nof    *        No Overflow
         ====== ======== =========
 
         When this instruction compares integer vectors, it returns a boolean
         vector of lane-wise comparisons.
+
+        * The unsigned version of overflow conditions have ISA-specific
+        semantics and thus have been kept as methods on the TargetIsa trait.
```
bnjbvr: Since this ends up in doc comments, would it be useful to put the name and/or a link to the unsigned_add_overflow_condition function?

ryzokuken:
@bnjbvr this better?

bnjbvr:
Thanks, I think rustdoc allows putting real links to these functions, using the fully qualified function name, which would be even better. How does it sound?

PR #1079, opened 2019-09-25
Name opcodes statically in isa/x86. Closes #1051

### LINKED ISSUE #1051: Name opcodes statically
As @sstangl mentions in https://github.com/CraneStation/cranelift/pull/1013#discussion_r326379569, it would be nicer if 

```rec_vconst_opt.nonrex().opcodes(vec![0x66, 0x0f, 0x74]); // PCMPEQB from SSE2```

were actually more like

```rec_vconst_opt.nonrex().opcodes(PCMPEQB);```.

This would reduce the opportunity for programmer error and duplication and could be implemented by placing all of the opcodes in a separate `opcodes.rs` file referenced by `encodings.rs`. I do not think we need to use `lazy_static` in this case (unlike we discussed in #955) because we could write:
```
static PCMPEQB: [u8; 3] = [0x66, 0x0f, 0x74]; // or...
static PCMPEQB: &'static [u8] = &[0x66, 0x0f, 0x74];
```
And modify `Template::opcodes` to receive more than just a `Vec<u8`.

### PR DESCRIPTION
**Pull request by sstangl**
This names opcode bytes statically in meta isa/x86 encodings. The `Template::opcodes()` was changed to enforce use of such static, named bytes. As a small benefit, that also means no more owned vectors, and no more vector cloning every time you call a copy-constructor function!

I believe this caught three bugs, but I didn't want to change the bytes just in case I am mistaken. I have marked them with "FIXME" comments. In all cases, it looks like we're encoding a 32-bit form where a 64-bit form is expected (which is the 32-bit form prefixed with `0x66`).

Passes all tests locally. The tests caught one typo: I accidentally specified `ADD` as `0x09` instead of `0x01`. The test failure made that super easy to identify and fix. We have good tests :)

### DIFF
--- a/cranelift-codegen/meta/src/isa/x86/encodings.rs
+++ b/cranelift-codegen/meta/src/isa/x86/encodings.rs
@@ -17,6 +17,8 @@ use crate::shared::types::Int::{I16, I32, I64, I8};
 use crate::shared::types::Reference::{R32, R64};
 use crate::shared::Definitions as SharedDefinitions;
 
+use crate::isa::x86::opcodes::*;
+
 use super::recipes::{RecipeGroup, Template};
 
 pub(crate) struct PerCpuModeEncodings {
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Comment by sstangl** [comment]
Did a force-push: forgot to run `cargo fmt` now that lines are shorter.

**Comment by abrown** [comment]
Looks good to me! Re: your `FIXME` comments: I suspect for bitwise operations that it doesn't matter if we use, e.g., `ANDPS` versus `ANDPD` because the bitwise operation will be applied to every one of the 128 bits regardless of the size of the data in the vector. I would just leave it as is unless you want to write a `run` test that proves that it doesn't make a difference.

**Comment by abrown** [comment]
...and this sure is going to make things easier to read!

File: cranelift-codegen/meta/src/isa/x86/encodings.rs [hunk 1]
```diff
@@ -1669,156 +1575,138 @@ pub(crate) fn define(
...
-        e.enc_both(inst.bind(F64), rec_fa.opcodes(vec![0x0f, opc]));
-    }
+    e.enc_both(band.bind(F32), rec_fa.opcodes(&ANDPS));
+    e.enc_both(band.bind(F64), rec_fa.opcodes(&ANDPS)); // FIXME: ANDPD? Missing 0x66
+
+    e.enc_both(bor.bind(F32), rec_fa.opcodes(&ORPS));
+    e.enc_both(bor.bind(F64), rec_fa.opcodes(&ORPS)); // FIXME: ORPD? Missing 0x66
+
+    e.enc_both(bxor.bind(F32), rec_fa.opcodes(&XORPS));
+    e.enc_both(bxor.bind(F64), rec_fa.opcodes(&XORPS)); // FIXME: XORPD? Missing 0x66
```
sunfishcode: The 0x66 here is the difference between `andps` and `andpd`, which are semantically equivalent, so skipping the `0x66` is just saving a byte in the encoding. I think the original idea of `andpd` was to let hardware optimize for the case where the data is coming from or going to an instruction using the f64 format, but in practice, no CPUs are known to care about the difference. So I think we can just add a comment here saying something like "Use the "ps" forms of and/or/xor instead of "pd" because CPU's don't treat them differently, and the "ps" form is one byte shorter."

PR #1081, opened 2019-09-26
[codegen] legalize icmp.i64 for 32-bit ISAs

ryzokuken:
Add legalizations for icmp and icmp_imm with i64 operands for the narrow
legalization set, allowing 32-bit ISAs (like x86-32) to compare 64-bit
integers.

Fixes: https://github.com/bnjbvr/cranelift-x86/issues/2

/cc @wingo @caitp @bnjbvr @sunfishcode 

Will add other conditions next. I imagine that'll involve adding some methods to the `CondCodes` trait, but will take care of it.

Ref: https://searchfox.org/mozilla-central/source/js/src/jit/arm/MacroAssembler-arm-inl.h#1147-1150

### DIFF
--- a/cranelift-codegen/meta/src/shared/legalize.rs
+++ b/cranelift-codegen/meta/src/shared/legalize.rs
@@ -5,6 +5,7 @@ use crate::cdsl::xform::{TransformGroupBuilder, TransformGroups};
 use crate::shared::immediates::Immediates;
 use crate::shared::types::Float::{F32, F64};
 use crate::shared::types::Int::{I128, I16, I32, I64, I8};
+use cranelift_codegen_shared::condcodes::{CondCode, IntCC};
 
 pub(crate) fn define(insts: &InstructionGroup, imm: &Immediates) -> TransformGroups {
     let mut narrow = TransformGroupBuilder::new(
@@ -271,6 +272,69 @@ pub(crate) fn define(insts: &InstructionGroup, imm: &Immediates) -> TransformGro
             def!(brnz(xh, ebb1, vararg)),
         ],
     );
+
+    // TODO(ryzokuken): benchmark this and decide if branching is a faster
+    // approach than evaluating boolean expressions.
+
+    narrow.custom_legalize(icmp_imm, "narrow_icmp_imm");
+
+    let intcc_eq = Literal::enumerator_for(&imm.intcc, "eq");
+    let intcc_ne = Literal::enumerator_for(&imm.intcc, "ne");
+    for &(int_ty, int_ty_half) in &[(I64, I32), (I128, I64)] {
+        narrow.legalize(
+            def!(b = icmp.int_ty(intcc_eq, x, y)),
+            vec![
+                def!((xl, xh) = isplit(x)),
+                def!((yl, yh) = isplit(y)),
+                def!(b1 = icmp.int_ty_half(intcc_eq, xl, yl)),
+                def!(b2 = icmp.int_ty_half(intcc_eq, xh, yh)),
+                def!(b = band(b1, b2)),
+            ],
+        );
+
+        narrow.legalize(
+            def!(b = icmp.int_ty(intcc_ne, x, y)),
+            vec![
+                def!((xl, xh) = isplit(x)),
+                def!((yl, yh) = isplit(y)),
+                def!(b1 = icmp.int_ty_half(intcc_ne, xl, yl)),
+                def!(b2 = icmp.int_ty_half(intcc_ne, xh, yh)),
+                def!(b = bor(b1, b2)),
+            ],
+        );
+
+        use IntCC::*;
+        for cc in &[
+            SignedGreaterThan,
+            SignedGreaterThanOrEqual,
+            SignedLessThan,
+            SignedLessThanOrEqual,
+            UnsignedGreaterThan,
+            UnsignedGreaterThanOrEqual,
+            UnsignedLessThan,
+            UnsignedLessThanOrEqual,
+        ] {
+            let intcc_cc = Literal::enumerator_for(&imm.intcc, cc.to_static_str());
+            let cc1 = Literal::enumerator_for(&imm.intcc, cc.without_equal().to_static_str());
+            let cc2 =
+                Literal::enumerator_for(&imm.intcc, cc.inverse().without_equal().to_static_str());
+            let cc3 = Literal::enumerator_for(&imm.intcc, cc.unsigned().to_static_str());
+            narrow.legalize(
+                def!(b = icmp.int_ty(intcc_cc, x, y)),
+                vec![
+                    def!((xl, xh) = isplit(x)),
+                    def!((yl, yh) = isplit(y)),
+                    // X = cc1 || (!cc2 && cc3)
+                    def!(b1 = icmp.int_ty_half(cc1, xh, yh)),
+                    def!(b2 = icmp.int_ty_half(cc2, xh, yh)),
+                    def!(b3 = icmp.int_ty_half(cc3, xl, yl)),
+                    def!(c1 = bnot(b2)),
+                    def!(c2 = band(c1, b3)),
+                    def!(b = bor(b1, c2)),
+                ],
+            );
+        }
+    }
 
     // Widen instructions with one input operand.
     for &op in &[bnot, popcnt] {
--- a/cranelift-codegen/src/legalizer/mod.rs
+++ b/cranelift-codegen/src/legalizer/mod.rs
@@ -665,3 +665,67 @@ fn narrow_iconst(
 
     unimplemented!("missing encoding or legalization for iconst.{:?}", ty);
 }
+
+fn narrow_icmp_imm(
+    inst: ir::Inst,
+    func: &mut ir::Function,
+    _cfg: &mut ControlFlowGraph,
+    _isa: &dyn TargetIsa,
+) {
+    use crate::ir::condcodes::{CondCode, IntCC};
+
+    let (arg, cond, imm): (ir::Value, IntCC, i64) = match func.dfg[inst] {
+        ir::InstructionData::IntCompareImm {
+            opcode: ir::Opcode::IcmpImm,
+            arg,
+            cond,
+            imm,
+        } => (arg, cond, imm.into()),
+        _ => panic!("unexpected instruction in narrow_icmp_imm"),
+    };
+
+    let mut pos = FuncCursor::new(func).at_inst(inst);
+    pos.use_srcloc(inst);
+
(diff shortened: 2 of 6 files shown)


cranelift-codegen/meta/src/shared/legalize.rs
```diff
@@ -272,48 +289,92 @@ pub(crate) fn define(insts: &InstructionGroup, imm: &Immediates) -> TransformGro
...
     );
 
+    for cc in &["sgt", "sge", "slt", "sle", "ugt", "uge", "ult", "ule"] {
+        let intcc_cc = Literal::enumerator_for(&imm.intcc, cc);
+
+        let cc1 =
+            Literal::enumerator_for(&imm.intcc, Box::leak(without_equal(cc).into_boxed_str()));
+        let cc2 = Literal::enumerator_for(
+            &imm.intcc,
+            Box::leak(inverse_without_equal(cc).into_boxed_str()),
```
bjorn3: Maybe add a method to IntCC returning `&'static str` instead, and use it in the Disply impl of IntCC too.

cranelift-codegen/src/legalizer/mod.rs
```diff
@@ -665,3 +665,69 @@ fn narrow_iconst(
...
+            imm,
+        } => (arg, cond, imm.into()),
+        _ => panic!("unexpected instruction in narrow_icmp_imm"),
+    };
+
+    let mut pos = FuncCursor::new(func).at_inst(inst);
+    pos.use_srcloc(inst);
+
+    let ty = pos.func.dfg.ctrl_typevar(inst);
+    if isa.pointer_bits() == 32 && ty == I64 {
```
bjorn3: Please drop this check and use `ty.half_width().unwrap()` instead of `I32`, `ty.half_width().unwrap().bits()` instead of 32 and `imm & (1u128 << ty.half_width().unwrap().bits() - 1) as i64` instead of `imm & 0xffffffff`. Legalizations have lower precedence than encodings, so this wouldn't get called for `i64` on 64bit x86 anyway. As already noted in https://github.com/CraneStation/cranelift/pull/1081#pullrequestreview-294285205, legalizing this for `i128` on both 32bit and 64bit x86 would be very useful for me.

cranelift-codegen/meta/src/shared/legalize.rs
```diff
@@ -272,6 +289,59 @@ pub(crate) fn define(insts: &InstructionGroup, imm: &Immediates) -> TransformGro
         ],
     );
 
+    narrow.custom_legalize(icmp_imm, "narrow_icmp_imm");
+
+    let intcc_eq = Literal::enumerator_for(&imm.intcc, "eq");
```
bnjbvr: Now that there's the custom legalization, should these legalizations be dropped instead?

cranelift-codegen/meta/src/shared/legalize.rs
```diff
@@ -272,6 +289,59 @@ pub(crate) fn define(insts: &InstructionGroup, imm: &Immediates) -> TransformGro
...
+            &imm.intcc,
+            Box::leak(inverse_without_equal(cc).into_boxed_str()),
+        );
+        let cc3 = Literal::enumerator_for(&imm.intcc, Box::leak(unsigned(cc).into_boxed_str()));
+        narrow.legalize(
+            def!(b = icmp.I64(intcc_cc, x, y)),
+            vec![
+                def!((xl, xh) = isplit(x)),
+                def!((yl, yh) = isplit(y)),
+                // X = A || (!A && !B && C)
```
bnjbvr: What's the A / B / C this comment refers to?

filetests/isa/x86/legalize-i64.clif
```diff
@@ -25,3 +25,25 @@ ebb0(v1: i64, v2: i64):
...
+    v10 = icmp eq v1, v2
+    ; check: v1 = iconcat $(v1_lsb=$V), $(v1_msb=$V)
+    ; nextln: v2 = iconcat $(v2_lsb=$V), $(v2_msb=$V)
+    ; nextln: $(v10_lsb=$V) = icmp eq $v1_lsb, $v2_lsb
+    ; nextln: $(v10_msb=$V) = icmp eq $v1_msb, $v2_msb
+    ; nextln: v10 = band $v10_lsb, $v10_msb
+    return v10
+}
+
+function %icmp_ne(i64, i64) -> b1 {
```
bnjbvr: Can you also add tests for one of the other condition codes, please?

cranelift-codegen/meta/src/shared/legalize.rs
bjorn3: The custom legalization is for icmp_imm, these are for icmp.

cranelift-codegen/src/legalizer/mod.rs
ryzokuken: @bjorn3 on it :)

ryzokuken: @bjorn3 done! PTAL.

cranelift-codegen/meta/src/shared/legalize.rs
ryzokuken: @bjorn3 on spot :) I already dropped the legalizations for `icmp_imm`.

cranelift-codegen/meta/src/shared/legalize.rs
ryzokuken: The results of cc1, cc2 and cc3. This is just a helpful comment to help visualize the final boolean expression.

Depending on what you prefer, I could either remove or expand this comment.

filetests/isa/x86/legalize-i64.clif
ryzokuken: On it :)

cranelift-codegen/meta/src/shared/legalize.rs
```diff
@@ -272,6 +293,60 @@ pub(crate) fn define(insts: &InstructionGroup, imm: &Immediates) -> TransformGro
...
+            let intcc_cc = Literal::enumerator_for(&imm.intcc, cc);
+            let cc1 = Literal::enumerator_for(&imm.intcc, without_equal(cc));
+            let cc2 = Literal::enumerator_for(&imm.intcc, inverse_without_equal(cc));
+            let cc3 = Literal::enumerator_for(&imm.intcc, unsigned(cc));
+            narrow.legalize(
+                def!(b = icmp.int_ty(intcc_cc, x, y)),
+                vec![
+                    def!((xl, xh) = isplit(x)),
+                    def!((yl, yh) = isplit(y)),
+                    // X = A || (!A && !B && C)
```
bnjbvr: nit: Can you expand these names please? A, B and C don't seem to match any of the previous variable names.

Also, one comparison can be avoided because `X || (!X && Y)` is equal to `X || Y` (short-circuiting ftw).

cranelift-codegen/src/legalizer/mod.rs
```diff
@@ -665,3 +665,71 @@ fn narrow_iconst(
...
+        _ => panic!("unexpected instruction in narrow_icmp_imm"),
+    };
+
+    let mut pos = FuncCursor::new(func).at_inst(inst);
+    pos.use_srcloc(inst);
+
+    let ty = pos.func.dfg.ctrl_typevar(inst);
+    let ty_half = ty.half_width().unwrap();
+
+    if isa.pointer_bits() as u16 != ty_half.bits() {
```
bnjbvr: This check won't pass when legalizing i128 on 32 bits, right? I expect it should, since we can legalize to 64 bits first, and then the legalizer will reapply this legalization from 64 to 32 bits.

cranelift-codegen/meta/src/shared/legalize.rs
```diff
@@ -272,6 +293,60 @@ pub(crate) fn define(insts: &InstructionGroup, imm: &Immediates) -> TransformGro
...
+            vec![
+                def!((xl, xh) = isplit(x)),
+                def!((yl, yh) = isplit(y)),
+                def!(b1 = icmp.int_ty_half(intcc_ne, xl, yl)),
+                def!(b2 = icmp.int_ty_half(intcc_ne, xh, yh)),
+                def!(b = bor(b1, b2)),
+            ],
+        );
+
+        for cc in &["sgt", "sge", "slt", "sle", "ugt", "uge", "ult", "ule"] {
```
bnjbvr: Should we instead iterate on the IntCC condition codes themselves, instead of having (below) hidden to_string/from_string calls? It might make this code more straightforward and remove the need for the helper functions above.

cranelift-codegen/meta/src/shared/legalize.rs
```diff
@@ -272,6 +293,60 @@ pub(crate) fn define(insts: &InstructionGroup, imm: &Immediates) -> TransformGro
...
+                    def!((yl, yh) = isplit(y)),
+                    // X = A || (!A && !B && C)
+                    def!(b1 = icmp.int_ty_half(cc1, xh, yh)),
+                    def!(b2 = icmp.int_ty_half(cc2, xh, yh)),
+                    def!(b3 = icmp.int_ty_half(cc3, xl, yl)),
+                    def!(c1 = bnot(b1)),
+                    def!(c2 = bnot(b2)),
+                    def!(d1 = band(c1, c2)),
+                    def!(d2 = band(d1, b3)),
+                    def!(b = bor(b1, d2)),
```
bnjbvr: In this case, there are so many comparisons that I expect branching to be faster...

@sunfishcode, @abrown, you may know about this: in general, which is faster: materializing the flags register and grouping results of several comparisons with binary OR/AND, or using branches to short-circuit unnecessary comparisons?

abrown: Not sure...

bnjbvr:
> This fails with, could you take a quick look? I'll try to debug this tomorrow morning. FAIL filetests/isa/x86/icmp-i128.clif: panicked in worker #1: attempt to shift right with overflow

What's going on here is that the immediate `imm` is 64-bits, while for `ty` := i128, `ty_half.bits()` is 64 bits too. So we're doing `(imm64) >> 64` which will zero out the result. It's again because we can't store more than 64-bits in a single instruction, otherwise we need to use handles to constants stored somewhere else (like SIMD does for 128-bits immediates). I think it's fine to use the wrapping variant here "wrapping_shr".

> Also, I'd need a final work regarding #1081 (comment). Would you be okay with me leaving a TODO comment asking me to revisit this once we've benchmarked this?

Sounds good to me, thanks.

ryzokuken:
Fixed committed and squashed, ready to merge!

Issue #1084, opened 2019-09-27
Change the representation of livein intervals


bnjbvr:
All operations related to the `GenericLiveRange::liveins` field appear high in CPU profiles, and seem to be costly. There's a big comment in `cranelift-codegen/src/regalloc/liverange.rs` file which explains their meaning: they store compressed intervals in which an SSA value is live, outside the initial block in which it's been defined.

In the past, **and what I'm proposing to do in a PR**, is to represent those with a sorted array of (Ebb, Inst) intervals. Using an EBB as the left bound makes sense, because being live in an EBB means starting to be live from the beginning of this EBB. The provided Inst is the instruction at which the value isn't live anymore (in a world of BBs, it could be a basic block too). Note a single interval can span multiple EBBs, that is, Inst can be an in a different EBB than ebb: that's the compressing part. This is used to coalesce intervals when inserting them during liveness analysis: if we want to mark the current SSA value live in [ebb2, inst2] while we already have [ebb1, inst1] in the set of livein intervals, with ebb1 (resp inst1) appearing in the layout before ebb2 (resp inst2), then we can store the resulting interval as [ebb1, inst2], storing only one interval and not two.

It means maintaining the array in sorted order during insertions, using binary search. If N is the number of elements, lookup of an entry / insertion position can be done in O(N log N), inserting in O(N). Analysis has proven that most SSA values (at least >95% in my memories) have 0, 1 or 2 livein entries. Distribution seems to follow a power law, with a long tail showing rare occurrences of SSA values with several hundreds of livein entries.

A future requirement would be that the LiveRange data structure allows to be split between two ones, so this is something to keep in mind when looking at the underlying data structures representing liveins.

# Current implementation

This uses a custom BTreeMap implementation, that's the Map in `cranelift-bforest`. It's a BTreeMap mapping Ebb to Inst, and allowing coalescing by retrieving the previous (in order) entry when a looked up entry is missing. According to the [commit which introduced it for livein intervals](https://github.com/CraneStation/cranelift/commit/1ba9eca253dd5451d194cc27180d0d1fa74c2d0f), it was done so to 1. spare memory per entry, and 2. to avoid a quadratic behavior when there are many livein intervals.

However, this comes with a cost in both complexity (maintaining and making sure the custom Map work as intended), and a high CPU cost. The quadratic behavior can definitely happen (when there are hundreds of livein entries in particular), but it's not as bad as the commit seemed to mention.

# Proposed alternatives

[...]

PR #1085, opened 2019-09-27
Use a sorted array for (Ebb, Inst) interval again

bnjbvr:
Explanations are in #1084. @julian-seward1 please take a look!

### DIFF
--- a/cranelift-codegen/src/regalloc/liverange.rs
+++ b/cranelift-codegen/src/regalloc/liverange.rs
@@ -63,11 +63,11 @@
 //!
 //! ## Current representation
 //!
-//! Our current implementation uses a B-tree map with the necessary interface for an efficient
-//! implementation of coalescing, implemented as a generic data-structure bforest::Map.
-//!
-//! A `BTreeMap<Ebb, Inst>` could have been used for the live-in intervals, but it doesn't provide
-//! the necessary API to make coalescing easy, nor does it optimize for our types' sizes.
+//! Our current implementation uses a sorted array of compressed intervals, represented by their
+//! boundaries (Ebb, Inst), sorted by Ebb. This is a simple data structure, enables coalescing of
+//! intervals easily, and shows some nice performance behavior. See
+//! https://github.com/CraneStation/cranelift/issues/1084 for benchmarks against using a
+//! bforest::Map<Ebb, Inst>.
 //!
 //! ## EBB ordering
 //!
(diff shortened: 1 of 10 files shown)


cranelift-codegen/src/regalloc/liverange.rs
```diff
@@ -167,29 +179,18 @@ pub struct GenericLiveRange<PO: ProgramOrder> {
 
     /// Additional live-in intervals sorted in program order.
     ///
-    /// This map is empty for most values which are only used in one EBB.
+    /// This vector is empty for most values which are only used in one EBB.
     ///
-    /// A map entry `ebb -> inst` means that the live range is live-in to `ebb`, continuing up to
+    /// An entry `ebb -> inst` means that the live range is live-in to `ebb`, continuing up to
```
bjorn3:
```suggestion
    /// An entry `ebb -> inst` means that the live range is live at the start of `ebb`, continuing up to
```

lars-t-hansen: Actually, live-in is the term of art and is used elsewhere too, so I sort of think it would be good to revert this change.

cranelift-codegen/src/regalloc/liverange.rs
```diff
@@ -63,11 +63,11 @@
...
-//! Our current implementation uses a B-tree map with the necessary interface for an efficient
-//! implementation of coalescing, implemented as a generic data-structure bforest::Map.
-//!
-//! A `BTreeMap<Ebb, Inst>` could have been used for the live-in intervals, but it doesn't provide
-//! the necessary API to make coalescing easy, nor does it optimize for our types' sizes.
+//! Our current implementation uses a sorted array of compressed intervals, represented by their
+//! boundaries: (Ebb, Inst). This is a simple data structure, enables coalescing of intervals
+//! easily, and shows some nice performance behavior. See
+//! https://github.com/CraneStation/cranelift/issues/1084 for benchmarks against using a
+//! bforest::Map<Ebb, Inst>.
```
lars-t-hansen: Might mention what the sort key is.

bnjbvr: Good idea.

lars-t-hansen:
I see a number of references to possible quadratic behavior in this ticket and in those it references and in the original commit, but it hasn't been explained precisely where the quadratic behavior comes from, nor why we should fear it in the code that we have here.  A deeper analysis would actually be welcome.  Eyeballing it, the worst case seems to be that we're inserting a new piece of a live range that doesn't coalesce with any previous piece; in this case it's O(lg n) to find the location and then O(n) to shift elements that are to the right in the array to the right, ie, O(n + lg n) = O(n) iirc.  We do this for all the elements, so in the worst case O(n^2) is possible if the live range pieces are almost completely randomized (no useful compression).  But in reality, we traverse the CFG top-down, thus we will usually be inserting near the end of the array, and there should be fairly effective compression of any live range that does span more than one block as these ranges will tend to cover entire subgraphs of blocks; it is only when a live range ends inside the two different arms of a diamond (say) that we get partly uncompressible data.  And the compression is not sensitive to interference of live ranges, I think.

In sum, I'm not very worried :-)

bnjbvr:
> A good end to end check that the live-range semantics are unchanged, is to check that the generated machine code is identical. If it is possible to verify this quickly (compile; dump asm; diff) before landing, that would be good.

Good idea. To wit: there are also tests at the bottom of the liverange.rs file, and they test for the optimal behavior (i.e. maximal coalescing). This in addition to the verifier running extensive tests made me sure it was at least correct.

I've now run the Spidermonkey shell, and compared disassembly as suggested, for wasm_zlib: generated code doesn't change at all.

Regarding both comments about the quadratic behavior: indeed the quadratic behavior was meant to be when doing N insertions that would each cause every element to be moved by copy. Agreed with Lars' analysis here, and in case this matters, for analyze purposes (ie code not to land), I've added a field that counts the number of moves happening. That is, if you insert a position N in an array of size M, you'll get M - N moves. The overall distribution of moves is the following, for wasm_zlib:

<details>

```
  60826 0
  11696 1
   2760 2
    692 3
    346 4
    264 5
    122 6
    116 7
     70 8
     28 9
     22 10
     16 11
      2 12
     10 13
      4 14
    112 15
     34 16
     44 17
     40 18
      2 20
      4 21
      2 22
    110 35
      6 36
      2 38
```

</details>

In practice, for the 77330 live SSA values (that's all the functions in wasm_zlib), the maximum number of moves for a single SSA value is 38, and this happens twice. So this matches Lars' explanation above, and I'm not too worried about this too.

PR #1086, opened 2019-09-27
Add a "hidden" argument purpose.

sunfishcode:
Add `ArgumentPurpose::Hidden` to allow wasm VMs to pass additional
hidden arguments that don't correspond to wasm arguments.

### DIFF
--- a/cranelift-codegen/src/ir/extfunc.rs
+++ b/cranelift-codegen/src/ir/extfunc.rs
@@ -279,13 +279,14 @@ pub enum ArgumentPurpose {
 }
 
 /// Text format names of the `ArgumentPurpose` variants.
-static PURPOSE_NAMES: [&str; 8] = [
+static PURPOSE_NAMES: [&str; 9] = [
     "normal",
     "sret",
     "link",
     "fp",
     "csr",
     "vmctx",
+    "hidden",
     "sigid",
     "stack_limit",
 ];
--- a/cranelift-codegen/src/ir/extfunc.rs
+++ b/cranelift-codegen/src/ir/extfunc.rs
@@ -259,6 +259,12 @@ pub enum ArgumentPurpose {
     /// used as a base pointer for `vmctx` global values.
     VMContext,
 
+    /// A hidden parameter.
+    ///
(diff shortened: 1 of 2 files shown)


bnjbvr:
Out of curiosity, can you expand a bit on what this is solving? Is this a generalization of the VmContext argument, and if so, could we remove it? Or is it an addition, making VmContext a first "hidden argument" and this one would be the second one? If so, is it worth generalizing the notion of hidden arguments so we don't have to special case a third hidden argument when the need arises? How is this supposed to interact with register allocation: will the argument be put in a location that Cranelift can reuse (thus clobbering the hidden argument), or should Cranelift not try to reuse the physical storage at all? Is it planned that Cranelift instructions can interact with this value?

Is it worth discussing this in an issue, to understand what's the problem this is solving, and consider alternative ideas?

(For what it's worth, tests are failing right now.)

sunfishcode:
> Out of curiosity, can you expand a bit on what this is solving? Is this a generalization of the VmContext argument, and if so, could we remove it? 

This is solving the fact that `declare_wasm_parameters` in cranelift-wasm/src/func_translator.rs assumes that all `Normal` parameters correspond to wasm parameters, and that it should create locals for them. My idea was that `Hidden` wouldn't imply anything special in terms of register allocation or calling convention, just that it's something the VM needs, rather than a wasm-level argument.

bnjbvr:
Ah right, it's still usable within the function's body and then behaves as a standard SSA value. It is a bit weird that a thing that's used only in the cranelift-wasm crate leaks into the cranelift-codegen crate, though, because its only usage right now is indirectly in cranelift-wasm. I mean, it seems useful, but the fact it's just handled as a Normal parameter everywhere in cranelift-codegen suggests it's not the right place to put it.

Another way to do this would be to have declare_wasm_parameters give more control to the embedder about its desired behavior; a function argument to filter out some parameters? or it makes its way into a trait with a default impl and the embedder can override it?

sunfishcode:
I've now submitted https://github.com/bytecodealliance/cranelift/pull/1329 which takes a different approach.

PR #1093, opened 2019-10-01
Fixes #1091: Use match statements instead of HashMaps in x86 encodings

### LINKED ISSUE #1091: Replace `HashMap` with `match` in `x86/encodings.rs`
See https://github.com/CraneStation/cranelift/pull/1050#discussion_r329659688.

bnjbvr:
This avoids Hashmap allocations in the meta crate, and is also more straightforward. Note the cute `let (opcodes, isa_predicate): (&[_], _)` type annotations so that Rustc doesn't infer a fixed-length array type as the first member's type.

@abrown PTAL!


bjorn3:
Fixes https://github.com/CraneStation/cranelift/issues/1091?

PR #1094, opened 2019-10-01
Insert split-edge of conditional branches before the destination.

nbp:
This patch changes the logic for the insertion of new blocks when splitting conditional branches.

Until now we added at the end or before the last block, which tend to mislead the branch predictor and resulted in bad benchmark results.

This patch change the logic to insert such edges before the destination. This has some advantages, such as when dealing with tight loop, to reduce the number of cache lines which are loaded for executing loops, as well as playing well with branch predictors.

### DIFF
--- a/cranelift-codegen/src/regalloc/branch_splitting.rs
+++ b/cranelift-codegen/src/regalloc/branch_splitting.rs
@@ -22,7 +22,6 @@ pub fn run(
 ) {
     let mut ctx = Context {
         has_new_blocks: false,
-        has_fallthrough_return: None,
         cur: EncCursor::new(func, isa),
         domtree,
         topo,
@@ -34,12 +33,6 @@ pub fn run(
 struct Context<'a> {
     /// True if new blocks were inserted.
     has_new_blocks: bool,
-
-    /// Record whether newly inserted empty blocks should be inserted last, or before the last, to
-    /// avoid disturbing the expected control flow of `fallthroug_return` statements.
-    ///
-    /// This value is computed when needed. The Option wraps the computed value if any.
-    has_fallthrough_return: Option<bool>,
 
     /// Current instruction as well as reference to function and ISA.
     cur: EncCursor<'a>,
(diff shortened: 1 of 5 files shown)


nbp:
Embenchen results (with ebb, with bb):
(to be compared with https://github.com/CraneStation/cranelift/pull/1008#issuecomment-536563505)

```
# mode=cranelift+cranelift, runs=41, problem size=default
# Lower score is better
box2d           1014    1077    0.942
bullet          1388    1359    1.021
conditionals    391     389     1.005
copy            752     755     0.996
corrections     1025    1035    0.99
fannkuch        2724    2958    0.921
fasta           920     907     1.014
fib             622     766     0.812
ifs             214     224     0.955
binarytrees     4533    4549    0.996
memops          1629    1718    0.948
primes          1231    1224    1.006
raybench        2207    2221    0.994
rust-fannkuch   2807    2793    1.005
skinning        833     879     0.948
zlib            1503    1475    1.019
```

lars-t-hansen:
Not a big improvement, I guess.  I'm curious about what's going on here.  I should think that the machine code for eg fib is easy enough to understand that it might be possible to zero in on that progam for a start, to see why perf drops by 20%.  Have you tried to do that?

nbp:
I tried to do that on memops, but I can do it again on fib. I presume the compilation speed is part of the issue as well.

lars-t-hansen:
Normally, wasm_bench will only report runtime, compile time is not a factor.

bjorn3:
https://mozilla.logbot.info/cranelift/20191001

> 16:29 nbp    Hum … both the BB and EBB versions of cranelift generate the same code. I wonder why one would be slower than the other in optimized builds: paste.mozilla.org/kzhvfiie


There are a few new unconditional jumps with ebb params, which are not turned into fallthrough, even though they are immediately before the target.

lars-t-hansen:
I don't know what @bjorn3 is referring to, everything seems to turn into a fallthrough just fine in that diff.  I notice that branch tightening seems to leave a NOP in the instruction stream where it removes a jump (at 0x4F), but also that that NOP seems to have been there before?  There's no encoding shown so I'm not sure...  Would love to see the actual disassembly actually.

bjorn3:
At https://paste.mozilla.org/kzhvfiie#L192 there is:

```
+[Op1jmpb#eb]                        jump ebb2(v0)
+
                                 ebb2(v6: i32):
```

That jump should have been a fallthrough.

lars-t-hansen:
I think, if you read all the way to the end, it is turned into a fallthrough later.  Am I misreading something?

bjorn3:
I thought that the log included multiple completely compiled functions, but it shows one function in multiple compilation stages. My bad.

PR #1103, opened 2019-10-02
Enhance generated code for legalizations

bnjbvr:
- First commit avoids unwrapping several times the InstructionData for each legalization of a specific opcode. Instead, we unwrap it once and, all the legalizations can look at the fields shared in the local scope. This required renaming a few variables in legalizations definitions, so they match in an opcode's legalization group.
- Second commit avoids the dumb "if true" statement in legalization's generated code.
- Third commit does a small refactoring of how we handle `varargs` in legalizations: before/after:
    <details>

    ```
            ir::Opcode::Brnz => {
                // Unwrap () << brnz.i128(x, ebb1, vararg)
                let (x, ebb1, vararg, predicate) = if let crate::ir::InstructionData::Branch {
                    destination,
                    ref args,
                    ..
                } = pos.func.dfg[inst] {
                    let func = &pos.func;
                    let args = args.as_slice(&func.dfg.value_lists);
                    (
                        func.dfg.resolve_aliases(args[0]),
                        destination,
                        args.iter().skip(2).map(|&arg| func.dfg.resolve_aliases(arg)).collect::<Vec<_>>(),
                        func.dfg.value_type(args[0]) == ir::types::I128
                    )
                } else {
                    unreachable!("bad instruction format")
                };
                let vararg = &vararg;
                // etc.
            }
   ```

    </details>
    <details>

    ```
                // Unwrap fields from instruction format () := brnz.i128(x, ebb1, vararg)
                let (x, ebb1, args) = if let ir::InstructionData::Branch {
                    destination,
                    ref args,
                    ..
                } = pos.func.dfg[inst] {
                    let args = args.as_slice(&pos.func.dfg.value_lists);
                    (
                        pos.func.dfg.resolve_aliases(args[0]),
                        destination,
                        args
                    )
                } else {
                    unreachable!("bad instruction format")
                };

                let vararg = &Vec::from(&args[1..]);
    ```
    </details>

    (Note the `skip()` call was off by one.)
- last commit makes more things `pub(crate)` instead of `pub`, finding one more dead method in the meta code.

Overall, this reduces the number of LOC in generated files as follows:

- legalizer.rs: from 5639 to 3826 non-blank LOC.
- legalize-x86.rs: from 828 to 502 non-blank LOC.

This measure is a good proxy for code size, thus time it takes for Rustc to compile cranelift-codegen, etc. I haven't seen a noticeable difference in compile time with the `time` command.

### DIFF
--- a/cranelift-codegen/meta/src/gen_encodings.rs
+++ b/cranelift-codegen/meta/src/gen_encodings.rs
@@ -76,9 +76,9 @@ use crate::unique_table::UniqueSeqTable;
 /// The generated code is an `if let` pattern match that falls through if the instruction has an
 /// unexpected format. This should lead to a panic.
 fn emit_instp(instp: &InstructionPredicate, has_func: bool, fmt: &mut Formatter) {
-    if instp.is_type_predicate() {
+    if let Some(type_predicate) = instp.type_predicate("func") {
         fmt.line("let args = inst.arguments(&func.dfg.value_lists);");
-        fmt.line(instp.rust_predicate("func"));
+        fmt.line(type_predicate);
         return;
     }
 
@@ -127,7 +127,7 @@ fn emit_instp(instp: &InstructionPredicate, has_func: bool, fmt: &mut Formatter)
             // Silence dead argument.
             fmt.line("let _ = func;");
         }
-        fmtln!(fmt, "return {};", instp.rust_predicate("func"));
+        fmtln!(fmt, "return {};", instp.rust_predicate("func").unwrap());
     });
     fmtln!(fmt, "}");
 
--- a/cranelift-codegen/meta/src/gen_encodings.rs
+++ b/cranelift-codegen/meta/src/gen_encodings.rs
@@ -78,7 +78,7 @@ use crate::unique_table::UniqueSeqTable;
 fn emit_instp(instp: &InstructionPredicate, has_func: bool, fmt: &mut Formatter) {
     if instp.is_type_predicate() {
         fmt.line("let args = inst.arguments(&func.dfg.value_lists);");
-        fmt.line(instp.rust_predicate());
+        fmt.line(instp.rust_predicate("func"));
         return;
     }
 
@@ -127,7 +127,7 @@ fn emit_instp(instp: &InstructionPredicate, has_func: bool, fmt: &mut Formatter)
             // Silence dead argument.
             fmt.line("let _ = func;");
         }
-        fmtln!(fmt, "return {};", instp.rust_predicate());
+        fmtln!(fmt, "return {};", instp.rust_predicate("func"));
     });
     fmtln!(fmt, "}");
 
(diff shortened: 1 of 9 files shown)


cranelift-codegen/meta/src/gen_legalizer.rs
```diff
@@ -162,10 +159,14 @@ fn unwrap_inst(
...
-
-            // Above name is set to an `Vec` representing the varargs. However it is expected to be
-            // `&[Value]` below, so we borrow it.
-            fmtln!(fmt, "let {} = &{};", name, name);
+            let n = inst
+                .imm_opnums
+                .iter()
+                .chain(inst.value_opnums.iter())
+                .max()
+                .map(|n| *n)
```
fitzgen: FWIW:

```suggestion
                .cloned()
```

bjorn3:
```suggestion
                .copied()
```

https://doc.rust-lang.org/stable/core/iter/trait.Iterator.html#method.copied stable since 1.36.0

bnjbvr:
For a better understanding of the changes brought here, you can see the diff at https://github.com/bnjbvr/cranelift-compare-gencode/commit/2f4c6a39e29ccd24324acc9e142be3fce83a88e9 .

@sstangl, gentle review ping, please? :)

cranelift-codegen/meta/src/gen_legalizer.rs
```diff
@@ -34,27 +35,61 @@ fn unwrap_inst(
...
+    // value list as args.
+    // - otherwise, if there's the count of immediate operands added to the presence of a value list exceeds one.
+    let need_tuple = iform.num_value_operands >= 1
+        || (inst
+            .operands_in
+            .iter()
+            .filter(|op| op.is_immediate())
+            .count()
+            + if iform.has_value_list { 1 } else { 0 })
+            > 1;
```
sstangl: Personally, I try to avoid lines that get broken up like this, because I find them difficult to comfortably parse at a glance.

This is more manual, but I think it's probably easier to read and still avoids the calculation in the short-circuit case:

```rust
let need_tuple = if iform.num_value_operands >= 1 {
    true
} else {
    let mut immediates = inst.operands_in.iter().filter(|op| op.is_immediate()).count();
    if iform.has_value_list {
        immediates += 1;
    }
    immediates > 1
};
```

Something like that.

cranelift-codegen/meta/src/gen_encodings.rs
```diff
@@ -78,7 +78,7 @@ use crate::unique_table::UniqueSeqTable;
 fn emit_instp(instp: &InstructionPredicate, has_func: bool, fmt: &mut Formatter) {
     if instp.is_type_predicate() {
         fmt.line("let args = inst.arguments(&func.dfg.value_lists);");
-        fmt.line(instp.rust_predicate());
+        fmt.line(instp.rust_predicate("func").unwrap());
```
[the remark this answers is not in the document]
bnjbvr: It would mean publicizing the node, which was prevented in previous code. However, I can morph "is_type_predicate" into "type_predicate(func_str: &str) -> Option<String>", with the same effect and nice typesystem-ness.

bnjbvr:
Thanks for the review!

PR #1104, opened 2019-10-02
Add SIMD logical operators for x86; closes #1092

### LINKED ISSUE #1092: Shared and x86 legalizations interfere for `bnot`
To support Wasm SIMD's `v128.not` operation I decided to legalize `bnot` to a `vconst + bxor`:

```
ebb0(v0: b32x4):
    v1 = bnot v0
    ; check: v2 = vconst.b32x4 0xffffffffffffffffffffffffffffffff
    ; nextln: v1 = bxor v2, v0
    return v1
```

However, I legalized this in the x86 legalizations, which I believe is causing some issues with the shared legalizations:

```
STALLED for 3 seconds with 32/98 tests finished
slow: filetests/isa/x86/legalize-isplit-backwards.clif
thread 'main' panicked at 'worker threads stalled for 10 seconds.', cranelift-filetests/src/runner.rs:283:21
```

I think c66e1b6292272580628f5c2b7aaa918e5c40dd7b is the change that introduces the legalization for splitting and joining `i128s`. @bjorn3, @bnjbvr: what is the best way to add my legalization and not break the existing legalization?

abrown:
This makes changes to SIMD `bor`, `band`, `bxor` and `bnot`. `bnot` is the most interesting in that it required changes in https://github.com/CraneStation/cranelift/commit/bc6c1244bc5e6cf2eae323b98b66d341460de6e0 to support constants during legalization. As discussed in #1092, custom legalizations sort of stomps on the regular legalizations (e.g. see `legalize.rs`) and before this change custom legalizations were the only way to add constants to the pool.

### DIFF
--- a/cranelift-codegen/meta/src/cdsl/ast.rs
+++ b/cranelift-codegen/meta/src/cdsl/ast.rs
@@ -54,7 +54,7 @@ impl Def {
         let results = self
             .defined_vars
             .iter()
-            .map(|&x| var_pool.get(x).name)
+            .map(|&x| var_pool.get(x).name.as_str())
             .collect::<Vec<_>>();
 
         let results = if results.len() == 1 {
@@ -238,7 +238,7 @@ pub(crate) enum PatternPosition {
 ///
 /// Temporary values are defined only in the destination pattern.
 pub(crate) struct Var {
-    pub name: &'static str,
+    pub name: String,
 
     /// The `Def` defining this variable in a source pattern.
     pub src_def: Option<DefIndex>,
@@ -254,7 +254,7 @@ pub(crate) struct Var {
 }
 
 impl Var {
-    fn new(name: &'static str) -> Self {
+    fn new(name: String) -> Self {
         Self {
             name,
             src_def: None,
@@ -346,7 +346,7 @@ impl Var {
     }
 
     pub fn to_rust_code(&self) -> String {
-        self.name.into()
+        self.name.clone()
     }
     fn rust_type(&self) -> String {
         self.type_var.as_ref().unwrap().to_rust_code()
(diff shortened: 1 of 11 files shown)


abrown:
@bjorn3 you may want to take a look at this as well.

cranelift-codegen/meta/src/isa/x86/encodings.rs
```diff
@@ -1992,6 +1985,21 @@ pub(crate) fn define(
...
 
+    // SIMD logical operations
+    for ty in ValueType::all_lane_types().filter(allowed_simd_type) {
+        // and
+        let band = band.bind_vector_from_lane(ty, sse_vector_size);
+        e.enc_32_64(band, rec_fa.opcodes(&PAND));
+
+        // or
+        let bor = bor.bind_vector_from_lane(ty, sse_vector_size);
+        e.enc_32_64(bor, rec_fa.opcodes(&POR));
```
bjorn3: What is the difference between the old ORPS and POR?

cranelift-codegen/meta/src/isa/x86/legalize.rs
```diff
@@ -383,6 +387,16 @@ pub(crate) fn define(shared: &mut SharedDefinitions, x86_instructions: &Instruct
         );
     }
 
+    // SIMD bxor
```
bjorn3: *bnot

cranelift-codegen/meta/src/isa/x86/opcodes.rs
```diff
@@ -263,6 +263,9 @@ pub static PADDUSB: [u8; 3] = [0x66, 0x0f, 0xdc];
 /// Add packed unsigned word integers from xmm2/m128 and xmm1 saturate the results (SSE).
 pub static PADDUSW: [u8; 3] = [0x66, 0x0f, 0xdd];
 
+/// Bitwise AND of xmm2/m128 and xmm1 (SSE2).
```
bjorn3: Why do you mention specific registers instead of just the register class (xmm)?

cranelift-codegen/meta/src/isa/x86/encodings.rs
abrown: See the commit message on 62ad8de

cranelift-codegen/meta/src/isa/x86/opcodes.rs
abrown: It's not specific registers; I think in the Intel manual they choose to use this (sort of confusing) nomenclature for indicating "first XMM operand" and "second XMM operand." Following what I think @sstangl intended for this file I've been copying the text directly from the Intel manual.

bjorn3:
> any thoughts on the DummyConstant/ConstPool stuff?

Not really

> Also, BTW, I think @bnjbvr was trying to get in touch with you on IRC

Thanks for letting me know!

bnjbvr:
(Bouncing reviewer; from a quick skim it seemed pretty good.)

abrown:
@sstangl ping

bnjbvr:
Re-taking this PR's review to help move things forward.

cranelift-codegen/meta/src/cdsl/xform.rs
```diff
@@ -148,16 +153,17 @@ impl Transform {
 /// pool `var_pool`. If the variable was not present in the symbol table, then add it to the list of
 /// `defined_vars`.
 fn var_index(
-    name: &'static str,
+    name: impl Into<String>,
```
bnjbvr: Could this just be a `&str` instead? (in the spirit of reducing the number of type-parameterized functions in general)

cranelift-wasm/src/code_translator.rs
```diff
@@ -1028,6 +1028,22 @@ pub fn translate_operator<FE: FuncEnvironment + ?Sized>(
...
+        }
+        Operator::V128And => {
+            let (a, b) = state.pop2();
+            state.push1(builder.ins().band(a, b));
+        }
+        Operator::V128Or => {
+            let (a, b) = state.pop2();
+            state.push1(builder.ins().bor(a, b));
+        }
+        Operator::V128Xor => {
```
bnjbvr: Light suggestion: it seems we could commonize all this code with the non-vector branches too, which would result in less code, but would break the grouping of SIMD opcodes. I leave it to your personal taste.

cranelift-codegen/meta/src/cdsl/ast.rs
```diff
@@ -384,8 +384,51 @@ impl VarPool {
...
+}
+
+impl ConstPool {
+    /// Create an empty constant pool.
+    pub fn new() -> Self {
+        Self { pool: vec![] }
+    }
+
+    /// Create a name for a constant from its position in the pool.
+    fn stringify(position: usize) -> String {
```
bnjbvr: Forgot to comment here: could we rename this function `constant_name_for`, maybe? Stringify may be confusing since it has other meanings (the macro just transforms any token tree into a Rust string, JS has JSON.stringify which does something different, etc.).

cranelift-wasm/src/code_translator.rs
abrown: Opened https://github.com/CraneStation/cranelift/issues/1139.

PR #1105, opened 2019-10-02
clif-util: Make the `-t` flag work with the `wasm` subcommand

fitzgen:
The presence of the flag was checked in the code, so it was essentially already supported, but it was not properly configured when constructing the CLI arguments parser.

### DIFF
--- a/src/clif-util.rs
+++ b/src/clif-util.rs
@@ -119,6 +119,12 @@ fn add_enable_multi_value<'a>() -> clap::Arg<'a, 'a> {
         .help("Enable WASM's multi-value support")
 }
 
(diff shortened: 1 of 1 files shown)


fitzgen:
We might want to consider using `structopt`, so that we parse CLI args right into a struct that we pass to the subcommand function impls, and it isn't possible to have this sort of mix up anymore.

It is a derive macro, and would therefore increase build times a bit, however. Worth considering, IMO.

fitzgen:
FWIW, the `-c` flag to just check the translation was suffering this same issue. Pushed another commit to fix that one too.

PR #1109, opened 2019-10-02
[codegen] legalize imul for 64-bit and 128-bit operands

ryzokuken:
Add a legalization that legalizes imul.I64 for 32-bit ISAs.

Refs: https://github.com/bnjbvr/cranelift-x86/issues/4

/cc @wingo @caitp

@bnjbvr PTAL.

There's one small problem though, there's no existing way to handle carry output in `imul`. Good news? According to https://www.felixcloutier.com/x86/imul, `imul` does set the appropriate flags on x86-32 atleast, so I can just add an alternate `imul_ifcout` instruction, I guess? wdyt?

### DIFF
--- a/cranelift-codegen/meta/src/shared/legalize.rs
+++ b/cranelift-codegen/meta/src/shared/legalize.rs
@@ -115,6 +115,7 @@ pub(crate) fn define(insts: &InstructionGroup, imm: &Immediates) -> TransformGro
     let uextend = insts.by_name("uextend");
     let uload8 = insts.by_name("uload8");
     let uload16 = insts.by_name("uload16");
+    let umulhi = insts.by_name("umulhi");
     let ushr = insts.by_name("ushr");
     let ushr_imm = insts.by_name("ushr_imm");
     let urem = insts.by_name("urem");
@@ -334,6 +335,25 @@ pub(crate) fn define(insts: &InstructionGroup, imm: &Immediates) -> TransformGro
                 ],
             );
         }
+    }
+
+    // TODO(ryzokuken): explore the perf diff w/ x86_umulx and consider have a
+    // separate legalization for x86.
+    for &ty in &[I64, I128] {
+        narrow.legalize(
+            def!(a = imul.ty(x, y)),
+            vec![
+                def!((xl, xh) = isplit(x)),
+                def!((yl, yh) = isplit(y)),
+                def!(a1 = imul(xh, yl)),
+                def!(a2 = imul(xl, yh)),
+                def!(a3 = iadd(a1, a2)),
+                def!(a4 = umulhi(xl, yl)),
+                def!(ah = iadd(a3, a4)),
+                def!(al = imul(xl, yl)),
+                def!(a = iconcat(al, ah)),
+            ],
+        );
     }
 
     // Widen instructions with one input operand.
--- a/filetests/isa/x86/imul-i128.clif
+++ b/filetests/isa/x86/imul-i128.clif
@@ -1,0 +1,20 @@
+test run
+target x86_64 haswell
+
+function %test_imul_i128() -> b1 {
+ebb0:
+    v11 = iconst.i64 0xf2347ac4503f1e24
+    v12 = iconst.i64 0x0098fe985354ab06
+    v1 = iconcat v11, v12
+    v21 = iconst.i64 0xf606ba453589ef89
+    v22 = iconst.i64 0x042e1f3054ca7432
+    v2 = iconcat v21, v22
+    v31 = iconst.i64 0xbe2044b2742ebd44
+    v32 = iconst.i64 0xa363ce3b6849f307
+    v3 = iconcat v31, v32
+    v4 = imul v1, v2
+    v5 = icmp eq v3, v4
+    return v5
+}
+
+; run
--- a/filetests/isa/x86/legalize-i128.clif
+++ b/filetests/isa/x86/legalize-i128.clif
@@ -1,0 +1,20 @@
+; Test the legalization of i128 instructions on x86_64.
+test legalizer
+target x86_64 haswell
+
+; regex: V=v\d+
+
+function %imul(i128, i128) -> i128 {
+ebb0(v1: i128, v2: i128):
+    v10 = imul v1, v2
+    ; check: v1 = iconcat $(v1_lsb=$V), $(v1_msb=$V)
+    ; nextln: v2 = iconcat $(v2_lsb=$V), $(v2_msb=$V)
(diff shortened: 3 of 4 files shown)


bjorn3:
Have you seen https://docs.rs/cranelift-codegen/0.44.0/cranelift_codegen/ir/trait.InstBuilder.html#method.umulhi and https://docs.rs/cranelift-codegen/0.44.0/cranelift_codegen/ir/trait.InstBuilder.html#method.smulhi ?

ryzokuken:
@bjorn3 this is exactly what I needed! Thanks.

bjorn3:
On x86 you can directly use x86_u/smulhi to get both low and hi bits at once:

https://github.com/CraneStation/cranelift/blob/6ed06a1e8e428aedb7adfd825b131abb9a7a3d15/cranelift-codegen/meta/src/isa/x86/legalize.rs#L85

cranelift-codegen/meta/src/shared/legalize.rs
```diff
@@ -272,6 +272,22 @@ pub(crate) fn define(insts: &InstructionGroup, imm: &Immediates) -> TransformGro
...
+    // separate legalization for x86.
+    narrow.legalize(
+        def!(a = imul.I64(x, y)),
+        vec![
+            def!((xl, xh) = isplit(x)),
+            def!((yl, yh) = isplit(y)),
+            def!(al = imul(xl, yl)),
+            def!(a1 = imul(xh, yl)),
+            def!(a2 = imul(xl, yh)),
+            def!(a3 = umulhi(xl, yl)),
```
bjorn3: a3 isnt used.

cranelift-codegen/meta/src/shared/legalize.rs
```diff
@@ -272,6 +272,22 @@ pub(crate) fn define(insts: &InstructionGroup, imm: &Immediates) -> TransformGro
         ],
     );
 
+    // TODO(ryzokuken): explore the perf diff w/ x86_umulx and consider have a
+    // separate legalization for x86.
+    narrow.legalize(
+        def!(a = imul.I64(x, y)),
```
bjorn3: Could you please also implement this and further i64 -> i32 narrowing legalizations for i128?

cranelift-codegen/meta/src/shared/legalize.rs
ryzokuken: oops.

cranelift-codegen/meta/src/shared/legalize.rs
ryzokuken: I could legalize `imul.ty` in terms of `imul.ty_half` which is good, but I guess I'll need to legalize `umulhi` for I64 for that too?

bjorn3: For x86_64 it is already implemented: https://github.com/CraneStation/cranelift/blob/6ed06a1e8e428aedb7adfd825b131abb9a7a3d15/cranelift-codegen/meta/src/isa/x86/encodings.rs#L725

32bit support

ryzokuken: so `umulhi` is defined for x86_64. I guess it's okay to add the legalization since it's not defined for any other ISA anyway...

ryzokuken:
@bjorn3 @bnjbvr done, with tests and all.

cranelift-codegen/meta/src/shared/legalize.rs
```diff
@@ -272,6 +273,25 @@ pub(crate) fn define(insts: &InstructionGroup, imm: &Immediates) -> TransformGro
...
+            vec![
+                def!((xl, xh) = isplit.ty(x)),
+                def!((yl, yh) = isplit.ty(y)),
+                def!(al = imul.ty_half(xl, yl)),
+                def!(a1 = imul.ty_half(xh, yl)),
+                def!(a2 = imul.ty_half(xl, yh)),
+                def!(a3 = umulhi.ty_half(xl, yl)),
+                def!(a4 = iadd.ty_half(a1, a2)),
+                def!(ah = iadd.ty_half(a4, a3)),
+                def!(a = iconcat.ty_half(al, ah)),
```
bjorn3: The type params can be left out as they are inferrable, right?

bjorn3:
Could you please add a run test with very large 128bit ints, say 0x98fe985354ab06f2347ac4503f1e24 and 0x42e1f3054ca7432f606ba453589ef89. It should return 0xa363ce3b6849f307be2044b2742ebd44. They have to be very large, so that no parts are zero and 128bit, so the legalizations actually get used on 64bit archs.

Tested using:

```rust
fn main() {
    println!("{:x}", 0x98fe985354ab06f2347ac4503f1e24u128.wrapping_mul(0x42e1f3054ca7432f606ba453589ef89u128));
}
```

Note: I just typed some random numbers and letters to generate those numbers.

ryzokuken:
@bjorn3 I added a test, but it'll need both `iconst.I128` (non existent, but should be fairly straightforward after #1081) and `icmp_imm.I128` (#1081) in .order to pass...

bjorn3:
iconst.i64 + iconcat.i128 could be used instead of iconst.i128 for now.

ryzokuken:
@bjorn3 `FAIL filetests/isa/x86/imul-i128.clif: 8: i128 is not a valid typevar for iconcat` 😅

bjorn3:
It will probably work when leaving `.i128` out.

ryzokuken:
@bjorn3 the current test has issues: `FAIL filetests/isa/x86/imul-i128.clif: 13: Too many hexadecimal digits`. I guess it'll work once we have landed #1081.

bjorn3:
Immediate values are limited to 64bit as 128bit would drastically increase memory usage. You could use the non imm variant of `icmp` instead.

ryzokuken:
@bjorn3 looks like this will now work as soon as `icmp` is merged. Current error: 

```
FAIL filetests/isa/x86/imul-i128.clif: run(%test_imul_i128):
Verifier errors:
- inst11: v5 is a ghost value used by a real [Op1ret#c3] instruction
```

cranelift-codegen/meta/src/shared/legalize.rs
bnjbvr: Agreed, and it should make the legalization easier to read too.

bnjbvr: Can you also swap the order of definitions of a3 and a4, to reduce the corresponding SSA values' lifetimes, please?

ryzokuken: @bnjbvr on it.

PR #1110, opened 2019-10-02
cranelift-wasm: Jump to the destination block at end of consequent

fitzgen:
This commit fixes a bug where at the end of an `if..else..end`'s consequent block, we would sometimes erroneously jump to the `else` block instead of to the following destination block. Not good!


fitzgen:
(Note: this was introduced by my multi-value PR, and is only possible to hit this bug with multi-value enabled)

sunfishcode:
Thanks!

fitzgen:
Thank *you!*

PR #1111, opened 2019-10-03
cranelift-wasm: Create `ModuleTranslationState` and polish API a little

### LINKED ISSUE #1080: Revamping cranelift_wasm's `{Func,Module}Environment` and public-facing API
Since I'm still fairly new to the code base, and since this ended up a bit more involved than I expected it to be, I figured it would be good to bounce these ideas off y'all before whipping up an implementation and submitting a PR.

### Problem

With the introduction of [support for multi-value Wasm](https://github.com/CraneStation/cranelift/pull/1049), we need to maintain some state about the whole module's translation<sup id="back-foot-0">[0](#foot-0)</sup> that we later use during each function's translation. However, `cranelift`'s `{Module,Func}Environment` traits and `wasmtime`'s implementation of these traits doesn't make it easy to get that shared module translation state when we need it to translate a particular function.

Here is some psuedo code describing the current state of `wasmtime`'s interactions with `cranlift_wasm` (my understanding is that SpiderMonkey takes a similar approach). The nesting roughly represents the call stack.

```
let data: &'data [u8] = ...;
let module_env: impl cranelift_wasm::ModuleEnvironment<'data> = ...;
let module_translation: wasmtime::ModuleTranslation = module_env.translate();
    cranelift_wasm::translate_module()
        enqueues function bodies into module_env's inner wasmtime::ModuleTranslation
parallel for each function body in module_translation:
    let func_env: impl cranelift_wasm::FuncEnvironment = ...;
    cranelift_wasm::FuncTranslator::translate(&mut func_env)
```

The shared module translation state needs to be initialized while we are decoding the Wasm module inside `cranelift_wasm::translate_module`, but it needs to be used during `cranelift_wasm::FuncTranslator::translate`.

### Desiridata

1. We want the creation, upkeep, and lifetime of the shared module translation state to be managed by `cranelift_wasm` and not external API consumers. Since the module translation state is the same for every user of `cranelift_wasm`, we don't want to put its interface into the `*Environment` traits, and force every `cranelift_wasm` API consumer to implement the same set of methods the same way. That's just busy work.

2. We would like to use plain borrows to access the shared module translation state during function translation. The module translation state is relatively temporary: it is only required during function translation, and afterwards we can forget about it. It is additionally only needed in a read-only fashion. This suggests allocating the module translation state on the stack and borrowing it during function translation. However, this is a bit tricky because `cranelift_wasm::translate_module` is not an ancestor stack frame above `cranelift_wasm::FuncTranslate::translate`.

3. We would like it to be difficult -- impossible even -- to use the shared module state for module A while translating a function from module B. At worst, if this mixup occurred, there would be a verification error, so this point is more of a nice-to-have than a show stopper. (This is of course assuming that we always verify the clif in, for example, `wasmtime-jit`. If we ever call JIT'd functions without verifying the clif, then this point is a must-have for memory safety!)

### Proposed Solution

* Introduce `ModuleTranslationState` for the Wasm module's associated translation state that is shared across all functions' translations. This is analagous to the current `TranslationState` but associated with a module rather than a function.

* We would rename to `FuncTranslationState` to help differentiate between module's and function's associated states.

* `ModuleEnvironment<'data>` would become `ModuleEnvironment<'state>` where `'state` is a `ModuleTranslationState`'s lifetime. Similarly, `FuncEnvironment` would become `FuncEnvironment<'state>`. We would thread a `&ModuleTranslationState<'state>` parameter through all these trait methods.

  For example:

  ```rust
  trait ModuleEnvironment<'state> {
      // ...

      fn define_function_body(
          &mut self,
          state: &ModuleTranslationState<'state>,
          body_bytes: &'state [u8],
          body_offset: usize,
      ) -> WasmResult<()>;

      // ...
  }
  ```

* We would provide a "scoped" API that "brands" its state and data to a single, invariant lifetime. When combined with replacing `'data` with `'state`, this gets the borrow checker to guarantee (3) for us, because the lifetime can't be extended, contracted, or unified with any other liftime by rustc.

  ```rust
  impl ModuleTranslationState {
      pub fn scoped<F, T>(data: &[u8], f: F) -> T
      where
          F: for<'state> FnOnce(&mut ModuleTranslationState<'state>) -> T
      {
          let mut state = ModuleTranslationState::new(data);
          f(&mut state)
      }

      /// Get the underlying data with our unique, invariant lifetime.
      pub fn data(&self) -> &'state [u8] {
          self.data
      }
  }
  ```

* Because ownership of the state stays inside `scoped`, API consumers never have to manage it themselves. While references to it are passed to API consumers, they just pass it along to other `cranelift_wasm` API functions. This gets us (1).

Here is the pseudocode for what `wasmtime`'s new usage of `cranelift_wasm` would look like:

```
let data: &'data [u8] = ...;
cranelift_wasm::ModuleTranslationState::scoped(data, |state: &ModuleTranslationState<'state>| {
    let module_env: impl cranelift_wasm::ModuleEnvironment<'state> = ...;
    let module_translation: wasmtime::ModuleTranslation = module_env.translate();
        cranelift_wasm::translate_module()
            enqueues function bodies into module_env's inner wasmtime::ModuleTranslation
    parallel for each function body in module_translation:
        let func_env: impl cranelift_wasm::FuncEnvironment<'state> = ...;
        cranelift_wasm::FuncTranslator::translate(&mut func_env)
});
```

#### Further Improvements

From here, there are some orthogonal API clean ups we could make, namely converting `cranelift_wasm::translate_module` and `cranelift_wasm::FuncTranslator::translate` into methods of `ModuleTranslationState`:

```
impl<'state> ModuleTranslationState<'state> {
    /// Note: `&mut self` since this populates the shared module state.
    pub fn translate_module(
        &mut self,
        module_env: &mut dyn ModuleEnvironment<'state>,
    ) -> WasmResult<()> {
        // ...
    }

    /// Note: `&self` since this only reads the already-populated state. This
    /// allows it to be shared across threads when translating functions in
    /// parallel.
    pub fn translate_function(
        &self,
        func_env: &mut dyn FuncEnvironment<'state>,
    ) -> WasmResult<()> {
        // ...
    }
}
```

This would make it pretty hard to even get (3)-style compilation errors by accident, and would further improve the API's usage ergonomics:

```
let data: &'data [u8] = ...;
cranelift_wasm::ModuleTranslationState::scoped(data, |state: &ModuleTranslationState<'state>| {
    let module_env: impl cranelift_wasm::ModuleEnvironment<'state> = ...;
    state.translate_module(&mut env);
        enqueues function bodies into module_env's inner wasmtime::ModuleTranslation

    let module_translation: wasmtime::ModuleTranslation = module_env.result;
    parallel for each function body in module_translation:
        let func_env: impl cranelift_wasm::FuncEnvironment<'state> = ...;
        state.translate_function(&mut func_env);
});
```

### Alternatives

* If we drop the nice-to-have (3), we can avoid the word-y `scoped` API, avoid introducing closures and indentation, and avoid threading invariant lifetimes through a bunch of functions, types, and traits. Instead, we would return the `ModuleTranslationState` from `cranelift::translate_module`, save it in `wasmtime::ModuleTranslation`, and then finally pass shared references to it back into each call of `cranelift_wasm::FuncTranslator::translate`. This does, however, make the user manage the module translation state's lifetime themselves, as well.

  If we pursued this further, we could still make `cranelift_wasm::FuncTranslator::translate` into a method of `ModuleTranslationState`, which would improve the ergonomics of function translation, IMO.

  The API usage would then end up something like this:

  ```
  let data: &'data [u8] = ...;
  let module_env: impl cranelift_wasm::ModuleEnvironment<'data> = ...;
  let state: ModuleTranslationState<'data> = cranelift_wasm::translate_module(&mut env);
      enqueues function bodies into module_env's inner wasmtime::ModuleTranslation

  let module_translation: wasmtime::ModuleTranslation = module_env.result;
  parallel for each function body in module_translation:
      let func_env: impl cranelift_wasm::FuncEnvironment = ...;
      state.translate_function(&mut func_env);
  ```

  I think this is also pretty good, and ultimately the decision comes down to how much we care about statically enforcing (3).

----------

<sup id="foot-0">0</sup> To be precise, we need to keep the Wasm module's decoded type section around so we can determine the type signature for a given multi-value block later on when translating a function. [↩](#back-foot-0)

fitzgen:
* Replace `WasmTypesMap` with `ModuleTranslationState`.

  The `ModuleTranslationState` contains information decoded from the Wasm module that must be referenced during each Wasm function's translation.

  This is only for data that is maintained by `cranelift-wasm` itself, as opposed to being maintained by the embedder. Data that is maintained by the embedder is represented with `ModuleEnvironment`.

  A `ModuleTranslationState` is returned by `translate_module`, and can then be used when translating functions from that module.

* Rename `TranslationState` to `FuncTranslationState` to disambiguate a bit with the new `ModuleTranslationState`.

* Reorganize the internal `state` module into submodules. One module for the `ModuleTranslationState` and another for the `FuncTranslationState`.

* Replace `FuncTranslator` with methods on `ModuleTranslationState`.

  `FuncTranslator` was two methods that always took ownership of `self`, so it didn't really make sense as an object as opposed to two different functions, or in this case methods on the object that actually persists for a longer time (the `ModuleTranslationState`).

  I think this improves ergonomics nicely.

  Before:

  ```rust
  let module_translation = translate_module(...)?;
  for body in func_bodies {
      let mut translator = FuncTranslator::new();
      translator.translate(&module_translation, body, ...)?;
  }
  ```

  After:

  ```rust
  let module_translation = translate_module(...)?;
  for body in func_bodies {
      module_translation.translate_func(body, ...)?;
  }
  ```

  Note that this commit does not remove `FuncTranslator`. It still exists, but is just a wrapper over the `ModuleTranslationState` methods, and it is marked deprecated, so that downstream users get a heads up. This should make the transition easier.

Fixes #1080.


cranelift-wasm/src/state/module_state.rs
```diff
@@ -24,4 +31,361 @@ impl ModuleTranslationState {
...
+        info!(
+            "translate({} bytes, {}{})",
+            reader.bytes_remaining(),
+            func.name,
+            func.signature
+        );
+        debug_assert_eq!(func.dfg.num_ebbs(), 0, "Function must be empty");
+        debug_assert_eq!(func.dfg.num_insts(), 0, "Function must be empty");
+
+        let mut func_ctx = FunctionBuilderContext::new();
```
bnjbvr: This will raise the number of mallocs by a large number. This context structure only exists for the purpose of sharing memory across function compilations. Can you store it in the `ModuleTranslationState` instead, and pass it by mut ref below?

cranelift-wasm/src/state/module_state.rs
```diff
@@ -24,4 +31,361 @@ impl ModuleTranslationState {
...
+        let mut state = FuncTranslationState::new();
```
bnjbvr: Ditto, this could be stored in the module's state. (If you follow through the chain of calls of `initialize`, it does clear its data structures.)

fitzgen:
> assuming the last commit does only code motion

Correct.

> This will raise the number of mallocs by a large number. This context structure only exists for the purpose of sharing memory across function compilations. Can you store it in the ModuleTranslationState instead, and pass it by mut ref below?

This is true only if the embedder was translating functions sequentially, and reusing the `FuncTranslator`.

If the embedder is translating functions in parallel, they already have to have a different `FuncTranslator` and context per thread.

`wasmtime` is doing the latter. I thought SpiderMonkey was too, but after digging through searchfox it seems like that might not actually be the case.

Anyways, I've figured out an even better design for both parallel and sequential situations. `ModuleTranslationState::translate_func` will take an `&mut FunctionBuilderContext` and an `&mut FuncTranslationState`. This will allow sequential translation to easily manage and reuse them.

Additionally, for parallel translation, we can use `rayon::ParallelIterator::map_{init,with}` in `wasmtime` to reuse the context and state for each function that a particular thread translates.

Best of both worlds!

Will update PR soon.

fitzgen:
Also: thanks for review!

fitzgen:
Hm actually at this point it might make sense to keep `FuncTranslator` around...

fitzgen:
I reverted the last commit, which deprecated `FuncTranslator` since reusing its internal context and state is nice for minimizing allocations, as discussed above. Assuming CI comes back green and that the r+ carries over, I'll merge this.

PR #1112, opened 2019-10-03
deps: bump wasmparser dependency to 0.39.2

fitzgen:
This has [a bug fix](https://github.com/yurydelendik/wasmparser.rs/pull/135) for Wasm multi-value blocks that is necessary to getting the spec tests passing.


fitzgen:
Thanks Yury!

PR #1114, opened 2019-10-04
codegen: Assert that the calling convention supports multi return

fitzgen:
If the pre-legalized function is returning multiple logical values, assert that the calling convention supports multiple return values.

Ideally, this check would happen in the verifier, rather than as an assertion during legalization. The problem is that legalization can (correctly!) split a single logical return value into multiple physical register returns, and the verifier will only see the multiple returns and doesn't know whether they are multiple logical return values, or a single logical value that's been split.

### DIFF
--- a/cranelift-codegen/src/isa/call_conv.rs
+++ b/cranelift-codegen/src/isa/call_conv.rs
@@ -64,6 +64,27 @@ impl CallConv {
             _ => false,
         }
     }
+
+    /// Does this calling convention have first-class support for multiple
+    /// return values?
+    pub fn supports_multiple_return_values(&self) -> bool {
+        match self {
+            // With our internal calling conventions, we can do whatever we
+            // want!
+            CallConv::Fast | CallConv::Cold => true,
+            // Existing, stable calling conventions can return aggregate values
+            // as structs or arrays, but we don't have that concept in
+            // cranelift. We also can't really treat multiple return values as
+            // an aggregate because there are subtleties about layout, whether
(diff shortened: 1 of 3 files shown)


cranelift-codegen/src/isa/call_conv.rs
```diff
@@ -64,6 +64,27 @@ impl CallConv {
...
+    pub fn supports_multiple_return_values(&self) -> bool {
+        match self {
+            // With our internal calling conventions, we can do whatever we
+            // want!
+            CallConv::Fast | CallConv::Cold => true,
+            // Existing, stable calling conventions can return aggregate values
+            // as structs or arrays, but we don't have that concept in
+            // cranelift. We also can't really treat multiple return values as
+            // an aggregate because there are subtleties about layout, whether
+            // something is adderessable, and constructors/destructors that only
```
bjorn3: *addressable

fitzgen:
> I would prefer a `assert!(sig.returns.len() <= 3)` for SysV for now to prevent breaking cg_clif.

Is cg_clif explicitly using sys v? If it is using the default "fast" calling convention, then it shouldn't break.

bjorn3:
It uses the sys v abi for everything, as it is required for C ffi anyway, and using fast for the Rust abi broke some things wrt allocators. Maybe you could add a flag to opt out of the check implemented in this PR.

fitzgen:
Closing since we agreed in #1147 that it doesn't make sense to have this assert until we have a better way of explicitly talking about compound return types in calling conventions.

PR #1121, opened 2019-10-08
Make post order invariant to EBB splits

### LINKED ISSUE #145: Canonical, split-invariant CFG post-order
In e0fd5252d5c9d0fb0eaae5f33a2d55b197c33ad4 we landed the function `DominatorTree::recompute_split_ebb()` which updates the dominator tree and its cached CFG post-order after an EBB split. It looks like the way we repair the post order is not correct. Consider this example:

```
ebb0:
    brnz v1, ebb1
    brnz v2, ebb2
    brnz v3, ebb3
    brnz v4, ebb4
    jump ebb5
```

If we assume that all of ebb0's successors are not visited before ebb0 in the DFS, we get the following post order: ebb5, ebb4, ebb3, ebb2, ebb1, ebb0.

Now assume we split the EBB by inserting the ebb99 header:

```
ebb0:
    brnz v1, ebb1
    brnz v2, ebb2
    jump ebb99
ebb99
    brnz v3, ebb3
    brnz v4, ebb4
    jump ebb5
```

The post order is now: ebb5, ebb4, ebb3, **ebb99**, ebb2, ebb1, ebb0. Two claims:

1. The current implementation of `recompute_split_ebb()` is not correct because it inserts the new EBB adjacent to ebb0. As the example shows, that is not always the right insertion spot.
2. The post order currently computed by the dominator tree is *split invariant*. By this I mean that the effect on the post order of splitting an EBB is just to insert the new EBB header somewhere. The relative order of existing EBBs does not change. It would be good to prove this property more strictly.

Some implementation notes for fixing this:

- Assuming the post order really is split invariant, the strategy in `recompute_split_ebb()` of inserting the new EBB somewhere in the existing post order is correct. It is only the insertion position that needs to be fixed.
- The new EBB will always appear before the old EBB header in the post order.
- The successors to the old EBB (ebb0 in the example) do not necessarily come before the old EBB in the post order. It is possible for any of the to be visited before ebb0 along another CFG path, which would cause them to appear *after* ebb0 in the post order.
- If one of the successors after the split appear before ebb0 in the post order, the correct insertion point is after the first such successor.

KiChjang:
Resolves #145.

Sorry for letting this issue sit for 2 years; hopefully I got the semantics of everything correct right now.

### DIFF
--- a/cranelift-codegen/src/dominator_tree.rs
+++ b/cranelift-codegen/src/dominator_tree.rs
@@ -452,27 +452,68 @@ impl DominatorTree {
     /// When splitting an `Ebb` using `Layout::split_ebb`, you can use this method to update
     /// the dominator tree locally rather than recomputing it.
     ///
-    /// `old_ebb` is the `Ebb` before splitting, and `new_ebb` is the `Ebb` which now contains
-    /// the second half of `old_ebb`. `split_jump_inst` is the terminator jump instruction of
-    /// `old_ebb` that points to `new_ebb`.
-    pub fn recompute_split_ebb(&mut self, old_ebb: Ebb, new_ebb: Ebb, split_jump_inst: Inst) {
+    /// `func` is the `Function` where the splitting of the `Ebb` into `old_ebb` and `new_ebb`
+    /// happened. `old_ebb` is the `Ebb` before splitting, and `new_ebb` is the `Ebb` which now
+    /// contains the second half of `old_ebb`. `split_jump_inst` is the terminator jump instruction
+    /// of `old_ebb` that points to `new_ebb`.
+    pub fn recompute_split_ebb(&mut self, func: &Function, old_ebb: Ebb, new_ebb: Ebb, split_jump_inst: Inst) {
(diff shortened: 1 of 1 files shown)


bnjbvr:
Hi @KiChjang ! Thanks for opening a PR. Sorry to be the bearer of bad news, but a lot of things have changed over the last two years, and we've realized during a meeting today that this method wasn't even used in the first place, except during testing. So we came to the conclusion that in place of maintaining this already complicated code, we could instead remove this function and the associated tests. (Unless somebody proved that recomputing the whole post-order is a big performance issue.) That doesn't make your contribution less worthwhile, and we all appreciate you taking the time to work on this.

Would you do us the honors of removing it, instead? Otherwise, we can close this PR and the associated issue.

KiChjang:
I just dug through my notifications, and I'm sorry to say that I would not be as free as I once was to continue contributing to cranelift. I'll close this issue.

bnjbvr:
No worries, and thank you for contributing, despite the bummer this could have been! You're always welcome to come back if/when you want to :)

PR #1087, opened 2019-09-28
Make no std default

bjorn3:
This removes the `std` feature flag where possible. liballoc is supported on stable since a few releases.

PR #1126, opened 2019-10-09
cleanup: remove spurious macro_use in cranelift-bforest;

bnjbvr:
This removes a compile warning when compiling tests for cranelift-bforest. Andrew, PTAL :)

### DIFF
--- a/cranelift-bforest/src/lib.rs
+++ b/cranelift-bforest/src/lib.rs
@@ -33,7 +33,6 @@
 #![no_std]
 
 #[cfg(test)]
-#[macro_use]
 extern crate alloc;
 
 #[macro_use]


bjorn3:
Does `./test-no_std.sh` succeed afterwards? If not you can ignore the warning instead like at https://github.com/CraneStation/cranelift/pull/1087/files#diff-46ee9825d31534aa2bce44b9289075bcR45.

bnjbvr:
> Does ./test-no_std.sh succeed afterwards?

Compilation of cranelift-bforest works; but it seems no_std is busted even on master in general, resulting to a failure of this test script.

bnjbvr:
Added a commit that also fixes no-std support.

cranelift-bforest/Cargo.toml
```diff
@@ -14,6 +14,11 @@ edition = "2018"
 [dependencies]
 cranelift-entity = { path = "../cranelift-entity", version = "0.44.0", default-features = false }
 
+[features]
+default = ["std"]
```
bjorn3: :(

1. cranelift-bforest doesn't have any extra functionality in std mode, so could you instead change the test script to not pass the core feature.
2. Please use `default-features = false` when depending on this or the other crates with `no_std` support. Otherwise depending on for example cranelift-codegen would enable `std` even when in `no_std` mode. I had removed `default-features = false` in https://github.com/CraneStation/cranelift/pull/1087, as I removed the `std` feature.

bnjbvr: Ok, I'll admit I didn't quite know what i was doing. Let's move this to a separate branch/PR.

PR #1127, opened 2019-10-09
Check for existing variables in solver's can_add_var

bnjbvr:
See commit message for the explanation. Our fuzzers at Mozilla found several instances of this particular bug very easily, so I've added all the test cases there, after reduction with bugpoint.

Note the smallest commit (which adds tests) is the actual fix (first commit only renames things). There might be a better way to avoid this bad situation, for instance adding a new array of "vars added because of a fixed input conflicts" in the solver, but that also mean iteration on an array. Plus, this fix is conceptually simpler, even though it might be more costly for compile time. I'll do a quick check to see if it has a dramatically bad impact.

Not sure who'd be the best reviewer for this (it's unclear from looking at this history of this file, even across renamings). Trying a few ones for starters.

### DIFF
--- a/cranelift-codegen/src/regalloc/coloring.rs
+++ b/cranelift-codegen/src/regalloc/coloring.rs
@@ -588,8 +588,8 @@ impl<'a> Context<'a> {
         // Copy register assignments from tied inputs to tied outputs.
         if let Some(constraints) = constraints {
             if constraints.tied_ops {
-                for (op, lv) in constraints.outs.iter().zip(defs) {
-                    if let ConstraintKind::Tied(num) = op.kind {
+                for (constraint, lv) in constraints.outs.iter().zip(defs) {
+                    if let ConstraintKind::Tied(num) = constraint.kind {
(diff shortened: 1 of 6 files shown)


lars-t-hansen:
I don't know that I have the insight in the solver to review this (except to say that the fix sounds plausible), but as a general comment, test cases of this size really want some comments embedded in them explaining where the problem is so that whoever stumbles across the test can understand what it is they're seeing, and perhaps more important, can take care not to invalidate it if the test case file has to be changed for any reason.

bnjbvr:
It's a good point, but unfortunately it's really hard to tell the required conditions for this to happen. (Maybe writing a Rust test would actually be simpler.) The three tests exhibited three different assertions, so they seemed useful. Having large tests in CI is also more useful, since they can be sort of integration tests; but I do agree it's a shame these are not unit tests. I'll try to write one.

bnjbvr:
Reduced total size of the 3 test cases from ~1600 lines to ~500 lines, thanks to latest bugpoint improvements in #1138.

PR #1129, opened 2019-10-09
Share more code between the meta and codegen crates

bnjbvr:
Now that there's a shared crate for codegen and the meta crate, we can share more constants and even some code that was duplicated between the two crates:

- constant_hash was almost entirely duplicated. Sharing the probe() function wasn't trivial and would have required some kind of closure passed to the function, so in the end it was simpler to keep it in the codegen crate. Also the Trait had to stay in the codegen crate, because if it were defined in the shared crate, then there couldn't be generic implementations of this trait (since a generic impl can only relate to types defined in the current crate).
- also clarify a comment in constant_hash that I left as a TODO, after experimenting with alternative ways to do it.
- share some constants that were implicitly synchronized between the two crates: now they're enforced to use the same value, either by using constants or assertions.

@abrown, can you have a look, please? Thanks!


cranelift-codegen/src/isa/registers.rs
```diff
@@ -12,25 +12,16 @@ use core::fmt;
...
 pub type RegClassMask = u32;
 
-/// Guaranteed maximum number of top-level register classes with pressure tracking in any ISA.
+/// A bit mask indexed by register units.
 ///
-/// This can be increased, but should be coordinated with meta/src/cdsl/regs.rs.
-pub const MAX_TRACKED_TOPRCS: usize = 4;
+/// The size of this type is determined by the target ISA that has the most register units defined.
+/// Currently that is arm32 which has 64+16 units.
+pub type RegUnitMask = [RegClassMask; 3];
```
bjorn3: Should the 3 here be in constants too?

cranelift-codegen/src/isa/registers.rs
```diff
@@ -338,3 +329,21 @@ impl<'a> fmt::Display for DisplayRegUnit<'a> {
...
+fn assert_sizes() {
+    use cranelift_codegen_shared::constants;
+    use std::mem::size_of;
+
+    // In these tests, size_of returns number of bytes: we actually want the number of bits, so
+    // multiply these by 8.
+    assert!(
+        (size_of::<RegClassMask>() * 8) <= constants::MAX_NUM_REG_CLASSES,
+        "need to bump MAX_NUM_REG_CLASSES or change RegClassMask type"
+    );
```
bjorn3: Maybe use a const assertion (`const A: () = [()][(size_of::<RegClassMask>() * 8) > constants::MAX_NUM_REG_CLASSES) as usize]` When size_of > max num, then casting the bool to an usize would return 1, which is out of bounds of `[()]`.

cranelift-codegen/meta/src/cdsl/regs.rs
```diff
@@ -364,19 +365,21 @@ impl IsaRegsBuilder {
...
-        // code.
         let num_toplevel = self
             .classes
             .values()
             .filter(|x| x.toprc == x.index && self.banks.get(x.bank).unwrap().pressure_tracking)
             .count();
-        assert!(num_toplevel <= 4, "Too many top-level register classes");
+
+        assert!(
+            num_toplevel <= constants::MAX_TRACKED_TOPRCS,
```
abrown: I read "TOPRCS" as "topics"; maybe "MAX_TRACKED_TOP_REG_CLASSES"?

bjorn3: Or `MAX_TRACKED_TOP_RCS` if you want to keep it shortish.

cranelift-codegen/src/isa/registers.rs
bjorn3: I just saw this crate in this week in rust:

cranelift-codegen/src/isa/registers.rs
bnjbvr: Not sure, since it doesn't need to be synced between meta and codegen.

bjorn3: Maybe meta can emit it as constant to match the isa with the most reg units.

bnjbvr: It's a nice idea: we could add the total number of regunits belonging to register classes that track register pressure, and use the maximum number in this array. Maybe something to do later yeah.

Issue #1132, opened 2019-10-10
wasm translation regression in wasm_linpack_float


bnjbvr:
[wasm_linpack_float](https://github.com/lars-t-hansen/embenchen/blob/master/asm_v_wasm/wasm_linpack_float.c.wasm) doesn't validate anymore when compiled when running `cargo run wasm --target x86_64 -d /tmp/wasm_linpack_float.c.wasm`. This impacts more test cases, since the wasm test suite in Spidermonkey now passes fewer tests.

There seems to be a jump to an incorrect block.

Bisection shows the following:

```
ee6fe21f924179466dd10c7b5b1e3c8e2422a7fa is the first bad commit
commit ee6fe21f924179466dd10c7b5b1e3c8e2422a7fa
Author: Nick Fitzgerald </snip>
Date:   Wed Oct 2 16:28:46 2019 -0700

    cranelift-wasm: Jump to the destination block at end of consequent
```

Will try to make a reduced test case.

@fitzgen Could you have a look, please? Thanks!

fitzgen:
@bnjbvr thanks for the heads up. Did you end up creating a reduced test case? Are there instructions somewhere for how I can get those benchmarks up and running somewhere?

fitzgen:
nvm, I've got it working locally.

fitzgen:
Reduced test case:

```wat
(module
  (type (;0;) (func (param i32) (result i32)))
  (import "env" "table" (table (;0;) 12 12 funcref))
  (func (;0;) (type 0) (param i32) (result i32)
    i32.const 0
    if  ;; label = @1
      i32.const 0
      if  ;; label = @2
        nop
      else
        unreachable
      end
      unreachable
    end
    unreachable)
  (elem (;0;) (i32.const 1)))
```

PR #1134, opened 2019-10-10
[wasm] Include more large tests

bnjbvr:
It seems important to not break translation of large, existing, realistic wasm code bases when touching the translator. I propose to add some test cases in this PR, taken from [Embenchen](https://github.com/lars-t-hansen/embenchen) or [wasm-collection](https://github.com/sunfishcode/wasm-collection/tree/master/misc-valid), so they serve as large tests for translation. They all compile in less than 2.5 seconds on my machine with `cargo run wasm`, and would have helped catching stuff like #1132.

I think a next step would be to *compile* wasm test cases in automation, since we don't do this during testing. They would serve as large integration tests, again, but they're useful to have, in addition to all the unit test cases we're adding as we improve Cranelift.

Marking as WIP, because this won't pass automation right now :innocent:.

### DIFF
--- a/wasmtests/embenchen_fannkuch.wat
+++ b/wasmtests/embenchen_fannkuch.wat
@@ -1,0 +1,16725 @@
+(module
+ (type $0 (func (param i32 i32 i32) (result i32)))
+ (type $1 (func (param i32) (result i32)))
+ (type $2 (func (param i32)))
+ (type $3 (func (result i32)))
+ (type $4 (func (param i32 i32) (result i32)))
+ (type $5 (func (param i32 i32)))
+ (type $6 (func))
+ (type $7 (func (param i32 i32 i32 i32 i32) (result i32)))
+ (type $8 (func (param i32 i32 i32)))
+ (type $9 (func (param i64 i32) (result i32)))
+ (type $10 (func (param i32 i32 i32 i32 i32)))
+ (type $11 (func (param f64 i32) (result f64)))
+ (type $12 (func (param i32 i32 i32 i32) (result i32)))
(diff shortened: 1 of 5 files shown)


fitzgen:
We may also want to put them in a subdirectory with the license from where they were taken from too.

abrown:
> Do we want to include the binary .wasms and not the wat text?

I would prefer the WAT personally.

bnjbvr:
Both points are great, thanks for bringing them up!

- I don't have any of the WAT for these tests, and while it could certainly be produced with one of wabt/binaryen's tool, we'd still lose a lot of developer context.
- I don't know what the license for these tests is, and need to find more information about this.

bnjbvr:
Using binaryen, I've recreated the wat equivalent of each of these files, and have removed a few of them for the following reasons:

- addm1, break-to-return, dylib, simple, stacky, test, trivial.obj: trivial test cases
- stripped, s: "yet another fibonacci function"
- a.out: no idea where it comes from (it seems to be a pthreads program compiled with emscripten), so no idea about licensing, so removed.

Kept:

- rust-fannkuch, which is an older build of https://github.com/alexcrichton/rust-fannkuch-js-and-wasm (Apache 2 licensed)
- other wasm_ prefixed files, which comes from https://github.com/kripken/embenchen (MIT licensed, so compatible with Apache 2 used in this repository)

Will let tests re-run in automation, and merge this tomorrow, unless people have objections or new insights.

bnjbvr:
(Full wat files of all the tests initially added are all available on [this branch](https://github.com/bnjbvr/cranelift/tree/tests-add-wasm-wat).)

PR #1138, opened 2019-10-11
A few bugpoint improvements: merge blocks, make it faster, etc.

bnjbvr:
In addition to cosmetic improvements, this implements replacing an instruction with N results with N constant instructions. This helped reduced the size of my test cases in #1127 by 20% in the best case.

@nbp please take a look!


src/bugpoint.rs
```diff
@@ -67,8 +67,9 @@ pub fn run(
 }
 
 enum MutationKind {
-    /// The mutation reduced the amount of instructions or ebbs.
-    Shrinked,
+    /// The mutation raised or reduced the amount of instructions or ebbs.
+    ExpandedOrShrinked,
```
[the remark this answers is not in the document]
bnjbvr: Agreed the name I've chosen isn't the best. However, renaming this "Progress" would be confusing with the below "Changed", which ought to be renamed too.

Maybe this whole enum's name should be about the size, so something like "SizeEffect" instead of "MutationKind", "ExpandedOrShrinked" => "Changed", "Changed" => "None".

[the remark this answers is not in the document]
bnjbvr: (I'll defer this for the moment.)

bnjbvr:
Re-requesting review: I've added support for merging consecutive blocks, and implemented a few perf enhancements (skip mutations that didn't cause a change). Let's see if the CI issue was a fluke or not...

src/bugpoint.rs
```diff
@@ -547,6 +552,97 @@ impl Mutator for RemoveUnusedEntities {
...
+        }
+
+        // Remove the terminator branch to the current EBB.
+        func.layout.remove_inst(pred.inst);
+
+        // Move all the instructions to the predecessor.
+        while let Some(inst) = func.layout.first_inst(ebb) {
+            func.layout.remove_inst(inst);
+            func.layout.append_inst(inst, pred.ebb);
+        }
```
nbp: Can you test this with basic-blocks flag, we have debug assertions which are preventing us from generating blocks which have a conditional branch followed by anything else than a jump.

If this fails with basic-blocks, please add a check that the previous instruction is not a branch, and if so skip this merge.

src/bugpoint.rs
```diff
@@ -660,24 +605,66 @@ fn reduce(
...
+                                should_keep_reducing = true;
+                                "shrink"
+                            }
+                            MutationKind::Changed => "changed",
+                        };
+                        if verbose {
+                            progress_bar.println(format!("{}: {}", msg, verb));
+                        }
+                    }
+                }
```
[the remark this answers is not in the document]
bjorn3: If you want to check for differential behavior, you could change check_for_crash to report a crash on behavior diff.

src/bugpoint.rs
bnjbvr: I'm pretty sure there's no risk here: basic blocks structure is stable across the mutation. If blocks were respecting the basic blocks structure before, they'll still be afterwards, the branch instruction position being conserved when it's moved from the ebb to its predecessor.

bjorn3: What about:

```
    brz
    jump ebb1

ebb1:
    iconst
    trap
```

That would result in:

```
    brz
    iconst
    trap
```

which us not allowed with basic blocks.

bnjbvr: Duh, I was very wrong :). Thanks.

So my understanding is that it becomes a verifier error and not a crash anymore if such a mutation is performed. In which case, it's just an invalid mutation and it'll be discarded. Is there anything else that needs to be done here?

bjorn3: I think you are right. You may want to check for that case in the mutator, because otherwise there would be time wasted verifying the whole function.

src/bugpoint.rs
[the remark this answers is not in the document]
bnjbvr: It's nice that the entire validation is linear in a single function at the moment. When the need arises for more validation schemes would be the good time to move this into another helper function, in my opinion.

src/bugpoint.rs
bnjbvr: Yeah, it's nicer indeed. Done.

src/bugpoint.rs
```diff
@@ -601,52 +697,116 @@ fn reduce(
...
+                    Some(res) => res,
+                    None => {
+                        break;
+                    }
+                };
+
+                if let ProgressStatus::Skip = mutation_kind {
+                    // The mutator didn't change anything, but we want to try more mutator
+                    // iterations.
+                    continue;
```
bjorn3: Did you read my reaction at https://github.com/CraneStation/cranelift/commit/e110a3027d4cde53593129cedf938f9dcf8fcc17#r35503802?

bnjbvr: Yep. To be fair, this is a very minor point: in debug (non-optimized) builds, skipping states are non-visible, and in optimized + debug-assertions mode, everything is so fast that i couldn't see a single progress message.

At some point, I imagined that we could also have a way to disable progress messages at all, since they were likely to make reduction slower, but the optimized build convinced me it's not even useful to implement, since things are so fast right now.

bnjbvr: Are you fine with me merging like this? If you have a very strong opinion about this, I can move the code around.

bjorn3: > Are you fine with me merging like this?

Yes

> At some point, I imagined that we could also have a way to disable progress messages at all

Makes sense, but I would like to retain the ability to show them. If it takes a lot of time to reduce, then it is useful to estimate the amount of time it will require. And it makes it possible to see when it is hanging. (for example infinite loop during legalization)

PR #1142, opened 2019-10-14
Increase legibility of the SSABuilder

sstangl:
I read through the SSABuilder to learn how it works. While doing so I updated comments to make things more dead-obvious.

The major point of contention is just that `temp_arg_val` is not some arbitrary temporary value, but a local Ebb parameter inserted for purposes of cycle detection.

There's no corresponding issue: this is just a legibility patch.

### DIFF
--- a/cranelift-frontend/src/ssa.rs
+++ b/cranelift-frontend/src/ssa.rs
@@ -4,6 +4,8 @@
 //! Zwinkau A. (2013) Simple and Efficient Construction of Static Single Assignment Form.
 //! In: Jhala R., De Bosschere K. (eds) Compiler Construction. CC 2013.
 //! Lecture Notes in Computer Science, vol 7791. Springer, Berlin, Heidelberg
+//!
+//! https://link.springer.com/content/pdf/10.1007/978-3-642-37051-9_6.pdf
 
 use crate::Variable;
 use alloc::vec::Vec;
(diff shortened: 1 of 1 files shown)


cranelift-frontend/src/ssa.rs
```diff
@@ -527,31 +565,39 @@ impl SSABuilder {
...
-        for _ in 0..self.predecessors(dest_ebb).len() {
-            // For each predecessor, we query what is the local SSA value corresponding
-            // to var and we put it as an argument of the branch instruction.
-            let pred_val = self.results.pop().unwrap();
+        let num_results = self.results.len();
+        let num_predecessors = self.predecessors(dest_ebb).len();
+
+        // Determine how many predecessors are yielding unique, non-temporary Values.
+        for i in 0..num_predecessors {
+            let pred_val = self.results[num_results - 1 - i];
```
abrown: This might be a bit hard to read... `self.results.iter().rev()...`?

cranelift-frontend/src/ssa.rs
```diff
@@ -527,31 +565,39 @@ impl SSABuilder {
...
                 ZeroOneOrMore::One(old_val) => {
-                    if pred_val != temp_arg_val && pred_val != old_val {
+                    if pred_val != sentinel && pred_val != old_val {
                         pred_values = ZeroOneOrMore::More;
+                        break;
                     }
                 }
-                ZeroOneOrMore::More => {}
+                ZeroOneOrMore::More => {
+                    break;
```
abrown: If we're trying to keep the logic the same as before, don't these `break`s change it?

sstangl: The logic should be the same, unless I have misread something -- the multiple `pop()` calls were moved to a single `truncate()` below.

cranelift-frontend/src/ssa.rs
sstangl: Ah yeah, that would definitely be better.

cranelift-frontend/src/ssa.rs
abrown: Yeah, what I mean is that before the loop would keep iterating and `pred_values` could have been set more than once--which now that I look at it seems like a bug since it is a `sentinel` that guards this. Perhaps this fixes an unlikely bug?

sstangl:
It looks like it's failing on all recent pull requests.

sstangl:
Oh, Nicolas turned on the "basic-blocks" feature by default, and some tests weren't updated. This patch is fine, though.

PR #1144, opened 2019-10-15
[ci] Don't have cargo nightly fuzz cause failures of the whole pipeline

bnjbvr:
It has started breaking in ways unrelated to Cranelift, making it hard
to spot real CI failures in Cranelift. We should re-enable it at some
point, but disable it in the meanwhile.

### DIFF
--- a/.azure-pipelines.yml
+++ b/.azure-pipelines.yml
@@ -88,6 +88,7 @@ jobs:
     - bash: cargo install cargo-fuzz
       displayName: Install cargo-fuzz
       condition: and(succeeded(), eq(variables['toolchain'], 'nightly'))
+
     - bash: |
         fuzz_module="ffaefab69523eb11935a9b420d58826c8ea65c4c"
         cargo fuzz run fuzz_translate_module \
@@ -96,6 +97,7 @@ jobs:
       env:
         RUST_BACKTRACE: 1
       condition: and(succeeded(), eq(variables['toolchain'], 'nightly'))
+      continueOnError: true
 
 - job: Fuzz_regression
   displayName: Fuzz regression


bjorn3:
Maybe add `exit 0` instead?

bnjbvr:
> Maybe add exit 0 instead?

Not sure it's very useful: it'd mean we'd never notice when it's broken, unless we look at the logs, and I don't expect people to read at the logs when CI passes are successful :)

If there was a way to make a CI pass optional, that is, simply report its completion status but not require it to merge things, that would be better than disabling, though.

tschneidereit:
> If there was a way to make a CI pass optional, that is, simply report its completion status but not require it to merge things, that would be better than disabling, though.

There is! If you add `continueOnError: true` to the task, it'll show up as a warning, but not make the overall job fail.

bjorn3:
There is `continueOnError`, but I dont know if it would still mark the whole build as failed.

bnjbvr:
Interesting, it seems the CI didn't fail on #1145, let's see if this happens again or if it was fixed in rustc nightly.

bjorn3:
Or you could add `exit 1` as test.

bnjbvr:
Ok, it's not failing anymore, but I've updated this PR to just use continueOnError for this step, in case this happens again in the future.

bnjbvr:
After many trial and errors, I've just used "true" as the condition for `continueOnError`: it seems I can't put other complex conditions at this level.

@tschneidereit This is ready if you want to give it a small look! It's now just a one-liner change. Thanks!

bnjbvr:
Thanks! For what it's worth, the job is already conditionally triggered, so the "true" condition really reads "if not nightly etc".

tschneidereit:
> For what it's worth, the job is already conditionally triggered, so the "true" condition really reads "if not nightly etc".

Right, I thought you had more specific conditions in mind for when this is allowed to fail, but as I said, I don't think these are necessary.

PR #1145, opened 2019-10-15
[wasm] Make the ModuleTranslationState ctor public

bnjbvr:
It's useful for consumers which don't want to translate a whole module,
but just need translation of functions, like Spidermonkey.

@fitzgen PTAL.


fitzgen:
Yep, just understand that when you want multi-value support, you'll need a properly initialized `ModuleTranslationState`.

bnjbvr:
You're right. There are many ABI changes to implement in Spidermonkey before being able to support multi-value yet. I'll keep this in mind though, thanks for the heads-up.

PR #1148, opened 2019-10-16
i128-isplit-forward-jump.clif:  brz followed by trap

nbp:
This fix the last test case which prevent merging #1008.


bjorn3:
Why is `brz` + `trap` not allowed? It is a conditional jump followed by a terminator.

nbp:
I will merge it as-is to merge #1008 , and we can later discuss whether to allow `trap` instructions.

Issue #1101, opened 2019-10-02
Use variable-length encodings for REX instructions


bnjbvr:
Variable-length encodings (introduced to support the proper machine code for loads/stores which memory operand is in a given register on x86) could allow removing the need to have both the REX vs non-REX encodings.

This would have a few benefits:

- reduce the number of encodings in the legalizer interpreter, making it faster
- allow some encodings to accept REX registers, when they couldn't before (at least regmove is affected by this issue)
- reduce the number of methods in PerCpuModeEncodings

bjorn3:
It would also make the instruction shrinking pass unnecessary when you want to avoid unnecessary REX prefixes. That would for example avoid https://github.com/bjorn3/rustc_codegen_cranelift/issues/146#issuecomment-449474527 when using `opt_level=none`.

PR #1149, opened 2019-10-16
Encode x86 SIMD instructions with REX prefix; fixes #1090

### LINKED ISSUE #1090: Encode x86 SIMD instructions with REX prefix to use more XMM registers
See discussion https://github.com/CraneStation/cranelift/pull/992#discussion_r325063181

abrown:
As a preliminary step to this, I re-factored `PerCpuModeEncodings` to reduce some of the issues observed in #955. Although this was not strictly necessary, I felt that changing many of the SIMD encodings was a good time to take a look at `PerCpuModeEncodings`. Stepping through the commits one-by-one should show the re-factoring progression. The final step, adding the REX prefix, should be more clear in the code after the re-factoring.

### DIFF
--- a/cranelift-codegen/meta/src/isa/x86/encodings.rs
+++ b/cranelift-codegen/meta/src/isa/x86/encodings.rs
@@ -1,14 +1,13 @@
 #![allow(non_snake_case)]
 
 use cranelift_codegen_shared::condcodes::IntCC;
-use std::collections::HashMap;
 
 use crate::cdsl::encodings::{Encoding, EncodingBuilder};
 use crate::cdsl::instructions::{
     vector, Bindable, InstSpec, Instruction, InstructionGroup, InstructionPredicate,
     InstructionPredicateNode, InstructionPredicateRegistry,
 };
-use crate::cdsl::recipes::{EncodingRecipe, EncodingRecipeNumber, Recipes};
+use crate::cdsl::recipes::{EncodingRecipe, Recipes};
 use crate::cdsl::settings::{SettingGroup, SettingPredicateNumber};
 use crate::cdsl::types::{LaneType, ValueType};
 use crate::shared::types::Bool::{B1, B16, B32, B64, B8};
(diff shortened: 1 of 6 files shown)


cranelift-codegen/meta/src/isa/x86/encodings.rs
```diff
@@ -22,6 +22,10 @@ use super::recipes::{RecipeGroup, Template};
 use crate::cdsl::formats::FormatRegistry;
 use crate::cdsl::instructions::BindParameter::Any;
 
+enum EncodingPattern {
```
bnjbvr: How do you feel about naming this CpuMode? (even if there's already something named this way somewhere else, I don't think the two namings would interact or conflict in any way)

cranelift-codegen/meta/src/isa/x86/encodings.rs
```diff
@@ -41,6 +45,32 @@ impl<'defs> PerCpuModeEncodings<'defs> {
...
+        &self,
+        inst: impl Into<InstSpec>,
+        recipe: EncodingRecipe,
+        bits: u16,
+    ) -> EncodingBuilder {
+        EncodingBuilder::new(inst.into(), recipe.clone(), bits, self.formats)
+    }
+
+    /// Create an encoding builder from a template.
+    fn enc(&self, inst: impl Into<InstSpec>, template: Template) -> EncodingBuilder {
```
bnjbvr: Is the idea to remove the other helper functions, and just use a combination of these new ones instead?

abrown: Yes! For me it was initially very confusing to figure out what was going on in this file, if/should I create new helper methods, figuring out if a helper method I needed already existed, etc. I think having fewer methods (`add` and `enc`) would have helped me figure out more clearly what was going on.

sstangl:
Just as a note, for #1101 , I'm going to try having a single Encoding for REX/non-REX. Hopefully that can be applied to SIMD instructions too.

I think the vague team consensus is to move away from Recipes as much as possible.

PR #1153, opened 2019-10-17
Fix build by marking tests as incompatible with basic-blocks. Closes #1152

### LINKED ISSUE #1152: Build fails with verifier issues
On master, build is failing during `cargo test` with:
```
AIL filetests\regalloc\solver-fixedconflict-var-3.clif: function u0:0(i32, i32, i32, i64 vmctx) -> i64 uext system_v {
ebb0(v0: i32, v1: i32, v2: i32, v3: i64):
    v5 = iconst.i32 -8
    v114 = iconst.i32 0
    v16 = iconst.i32 -8
    v17 = popcnt v16
    v192 = ifcmp_imm v17, -1
    brif eq v192, ebb12
    trap user0
;   ^~~~~~~~~~
; error: inst6: post-branch instruction not fallthrough or jump
...
```
Looks like it starts after merging #1127

sstangl:
There are two options for fixing the build:

1. Rewrite the tests so that they pass the basic-blocks verifier.
2. Make the test feature-gated on the absence of the `basic-blocks` feature.

Many tests are already disabled by `basic-blocks` (and therefore are now off by default! We need to fix that soon). I added these to the list because I think we should probably fix these tests by just allowing `trap` to be treated as, effectively, a valid jump instruction for the purposes of a terminal `BranchGroup`.

### DIFF
--- a/filetests/regalloc/solver-fixedconflict-var-3.clif
+++ b/filetests/regalloc/solver-fixedconflict-var-3.clif
@@ -2,6 +2,7 @@ test compile
 set opt_level=speed
 set enable_pinned_reg=true
 target x86_64 haswell
+feature !"basic-blocks"
 
 function u0:0(i32, i32, i32, i64 vmctx) -> i64 uext system_v {
 ebb0(v0: i32, v1: i32, v2: i32, v3: i64):
--- a/filetests/regalloc/solver-fixedconflict-var.clif
+++ b/filetests/regalloc/solver-fixedconflict-var.clif
@@ -2,6 +2,7 @@ test compile
 set opt_level=speed
 set enable_pinned_reg=true
 target x86_64 haswell
+feature !"basic-blocks"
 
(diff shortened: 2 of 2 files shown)


sstangl:
Requesting from abrown since this is addressing a failure on `master` and Europe is off-duty :)

sstangl:
I think we should solve this by accepting `trap` instructions in basic block terminal groups, and that the tests ought to be valid as written.

bnjbvr:
First instance (I know of!) of merge skew! Thanks for fixing it.

Agreed that we should enable (non-resumable) traps after conditional jumps in the future. Changing the test may tweak its behavior with respect to register allocation, so I'd prefer not to do this.

nbp:
> Many tests are already disabled by basic-blocks (and therefore are now off by default! We need to fix that soon). 

@sstangl , any of the test case which are feature gated should have a clone of them which is not. Unless someone else added such a test case in the mean time.

PR #1154, opened 2019-10-17
Fix test merge conflict from #1127

### LINKED ISSUE #1152: Build fails with verifier issues

iximeow:
Filetests now have to be valid as basic blocks, but these tests were written against a revision before that was default. Fixes #1152.

### DIFF
--- a/filetests/regalloc/solver-fixedconflict-var-3.clif
+++ b/filetests/regalloc/solver-fixedconflict-var-3.clif
@@ -11,6 +11,9 @@ ebb0(v0: i32, v1: i32, v2: i32, v3: i64):
     v17 = popcnt v16
     v192 = ifcmp_imm v17, -1
     brif eq v192, ebb12
+    jump ebb1
+
+ebb1:
     trap user0
 
 ebb12:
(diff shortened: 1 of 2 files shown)


sstangl:
I fixed this a different way in https://github.com/CraneStation/cranelift/pull/1153 -- I don't think it should be invalid to end a basic block with a `trap` instruction, so I think the long-term fix is to just make these tests valid.

sstangl:
Sorry for the dup -- #1153 was merged in.

PR #1155, opened 2019-10-17
Implement emitting Windows unwind information for fastcall functions.

peterhuene:
This PR implements emitting Windows unwind information for x64 fastcall
calling convention functions.

The unwind information can be used to construct a Windows function table at
runtime for JIT'd code, enabling stack walking and unwinding by the operating
system.


peterhuene:
I've marked this as a draft so I can get early feedback of the changes.

I'm still working on consuming this change from Wasmtime so that Windows can walk/unwind across JIT frames.  I'll make this PR ready when those changes have been properly tested.

cranelift-codegen/src/isa/x86/unwind.rs
```diff
@@ -0,0 +1,525 @@
...
+        self.write(v);
+    }
+
+    pub fn write_u32(&mut self, v: u32) {
+        self.write(v);
+    }
+
+    fn write<T>(&mut self, x: T) {
+        unsafe {
+            #[cfg_attr(feature = "cargo-clippy", allow(clippy::cast_ptr_alignment))]
```
peterhuene: This should probably always write little endian so that the host endianness doesn't affect it.

cranelift-codegen/src/isa/x86/unwind.rs
```diff
@@ -0,0 +1,525 @@
...
+use alloc::vec::Vec;
+
+// Maximum (inclusive) size of a "small" stack allocation
+const SMALL_ALLOC_MAX_SIZE: u32 = 128;
+// Maximum (inclusive) size of a "large" stack allocation that can represented in 16-bits
+const LARGE_ALLOC_16BIT_MAX_SIZE: u32 = 524280;
+
+struct MemoryWriter {
+    ptr: *mut u8,
+    offset: usize,
```
bjorn3: Is there any reason you cant use a plain `Vec<u8>`? That would be a lot safer.

cranelift-codegen/src/isa/x86/unwind.rs
```diff
@@ -0,0 +1,525 @@
...
+
+    pub fn write_u16(&mut self, v: u16) {
+        self.write(v);
+    }
+
+    pub fn write_u32(&mut self, v: u32) {
+        self.write(v);
+    }
+
+    fn write<T>(&mut self, x: T) {
```
bjorn3: This is unsafe, as it allows writing out of bounds of the backing storage.

cranelift-codegen/src/isa/x86/unwind.rs
```diff
@@ -0,0 +1,525 @@
+//! Unwind information for x64 Windows.
```
bjorn3: Please add a blank line after this doc comment.

iximeow:
Woah! This is really cool! I think some of the stack layout information you're collecting in `UnwindInfo::try_from_func` is also collected in https://github.com/CraneStation/cranelift/pull/679 (which my ELF unwind information PR builds on: https://github.com/CraneStation/cranelift/pull/902). #679 should be close to landing, it's just waiting on someone to have review bandwidth.

How does Windows unwind information handle functions with multiple epilogues, if at all? The `Preserve` and `Restore` frame layout changes map directly to equivalent DWARF unwind information, but I dunno if Windows has similar operators.

Upside, if you can phrase this as information built on top of `FrameLayoutChange`, you can use whatever frame register the layout happens to have encoded. It should correctly describe `rsp` for functions that elide frame pointer as it is today, even though Cranelift doesn't do that yet.

cranelift-codegen/src/isa/x86/unwind.rs
```diff
@@ -0,0 +1,525 @@
...
+use super::registers::RU;
+use crate::ir::{Function, InstructionData, Opcode};
+use crate::isa::{CallConv, RegUnit, TargetIsa};
+use alloc::vec::Vec;
+
+// Maximum (inclusive) size of a "small" stack allocation
+const SMALL_ALLOC_MAX_SIZE: u32 = 128;
+// Maximum (inclusive) size of a "large" stack allocation that can represented in 16-bits
```
bjorn3: Please use doc comments here and above.

cranelift-filetests/src/test_unwind.rs
```diff
@@ -0,0 +1,233 @@
...
+
+        match code.additional_nodes() {
+            0 => {
+                writeln!(text).unwrap();
+                i += 1;
+            }
+            1 => {
+                let offset = offset + core::mem::size_of::<UnwindCode>() as isize;
+                writeln!(text, ", value = {}", unsafe {
+                    core::ptr::read_unaligned(mem.as_ptr().offset(offset) as *const u16)
```
bjorn3: Please use `LittleEndian::read_u16(&mem[offset..offset+2])` instead. (LittleEndian comes from the byteorder crate)

cranelift-filetests/src/test_unwind.rs
```diff
@@ -0,0 +1,233 @@
...
+struct UnwindInfo {
+    version_and_flags: u8,
+    prologue_size: u8,
+    unwind_code_count: u8,
+    frame_register: u8,
+}
+
+impl UnwindInfo {
+    unsafe fn from_ptr(ptr: *const u8) -> Self {
+        core::ptr::read_unaligned(ptr as *const Self)
```
bjorn3: Please take a slice and read the individual fields instead.

cranelift-filetests/src/test_unwind.rs
```diff
@@ -0,0 +1,233 @@
...
+            7 => Self::SaveXmm128Far,
+            8 => Self::PushMachineFrame,
+            _ => panic!("unsupported unwind operation"),
+        }
+    }
+}
+
+impl UnwindCode {
+    unsafe fn from_ptr(ptr: *const u8) -> Self {
+        core::ptr::read_unaligned(ptr as *const Self)
```
bjorn3: Please take a slice and read the individual fields instead.

peterhuene:
Epilogues on Windows x64 ABI are strictly defined, so the OS can tell if it's in an epilogue by disassembling forward a few bytes from the frame IP.  Functions are allowed to have multiple epilogues.  For unwind, if it detects that the frame IP is in the epilogue, it will simulate the remaining execution of the epilogue to unwind the frame; thus the OS needs no description of epilogues in the unwind information.

I did look at #679 and I think there's work there I can piggy back off of, although it does seem to be DWARF-tailored.  For Windows x64 (we'll need an Windows ARM implementation eventually, but unwind information on ARM is quite different), we just need information about the prologues for unwind; at least until such time where a Cranelift function can have an exception handler.  Still, I would like #679 to go in first and then we can figure out what makes sense to merge this in with the `FrameLayoutChange` information, I think.

cranelift-codegen/src/isa/x86/unwind.rs
peterhuene: Originally I was passing in a `MemoryCodeSink`, which was a little too "heavy" of an abstraction for what this code was doing, so I ripped off it's memory writing functionality.

I agree that this could just simply write to the `Vec` directly which would be safer.  I'll remove this in favor of that.

cranelift-codegen/src/isa/x86/unwind.rs
peterhuene: I'll remove this in favor of writing to the Vec.

cranelift-codegen/src/isa/x86/unwind.rs
peterhuene: I generally don't use doc comment style for privates (I like it to warn/error if suddenly these get exposed as public since the comment might not be "up to par" for public documentation), but if that's the convention for Cranelift, I will fix.

cranelift-filetests/src/test_unwind.rs
peterhuene: I didn't see a dependency on `byteorder` yet, so I was reluctant to add it, but that's what I will likely use.

cranelift-filetests/src/test_unwind.rs
peterhuene: Agreed, that would be better.  I'll fix.

cranelift-filetests/src/test_unwind.rs
peterhuene: Agreed, that would be better.  I'll fix.

peterhuene:
I've tested these changes with corresponding changes to Wasmtime and was able to get successful stack walks (by Visual Studio's debugger) and unwinds through JIT frames.

I think we may want to focus on getting #679 in first and then I can rebase this on top of those changes and see what makes sense to merge in with those changes.

cranelift-codegen/src/isa/x86/unwind.rs
```diff
@@ -0,0 +1,525 @@
...
+        func: &Function,
+        isa: &dyn TargetIsa,
+        frame_register: Option<RegUnit>,
+    ) -> Option<Self> {
+        // Only Windows fastcall is supported for unwind information
+        if func.signature.call_conv != CallConv::WindowsFastcall || func.prologue_end.is_none() {
+            return None;
+        }
+
+        let prologue_end = func.prologue_end.unwrap();
```
bjorn3: What if the function never returns?

cranelift-codegen/src/isa/x86/unwind.rs
bjorn3: I dont now if it is convention, but it is useful when using the rustdoc option to document private items.

cranelift-codegen/Cargo.toml
```diff
@@ -23,6 +23,7 @@ target-lexicon = "0.8.1"
 log = { version = "0.4.6", default-features = false }
 serde = { version = "1.0.94", features = ["derive"], optional = true }
 smallvec = { version = "0.6.10" }
+byteorder = "1.3.2"
```
sunfishcode: Can you add `default-features = false` here, to enable support for `no_std` builds? (as described in [the README](https://crates.io/crates/byteorder))

`no_std` isn't a high priority right now, but it may be in the future, so it's useful to keep an eye out for it when adding new dependencies.

cranelift-filetests/src/test_unwind.rs
```diff
@@ -0,0 +1,197 @@
...
+        match value {
+            0 => Self::PushNonvolatileRegister,
+            1 => Self::LargeStackAlloc,
+            2 => Self::SmallStackAlloc,
+            3 => Self::SetFramePointer,
+            4 => Self::SaveNonvolatileRegister,
+            5 => Self::SaveNonvolatileRegisterFar,
+            6 => Self::SaveXmm128,
+            7 => Self::SaveXmm128Far,
+            8 => Self::PushMachineFrame,
```
sunfishcode: Are these numbers defined by the Windows ABI? Please add a comment here, and ideally a link to the documentation that these come from.

cranelift-codegen/src/isa/x86/unwind.rs
```diff
@@ -0,0 +1,508 @@
...
+}
+
+impl UnwindInfo {
+    pub fn try_from_func(
+        func: &Function,
+        isa: &dyn TargetIsa,
+        frame_register: Option<RegUnit>,
+    ) -> Option<Self> {
+        // Only Windows fastcall is supported for unwind information
+        if func.signature.call_conv != CallConv::WindowsFastcall || func.prologue_end.is_none() {
```
sunfishcode: In functions that don't require their own stack space, we don't currently emit a prologue, so `prologue_end` will be `None` here. Is it intended to emit no unwind info for such functions?

cranelift-codegen/src/isa/x86/unwind.rs
```diff
@@ -0,0 +1,508 @@
...
+        // Stores the stack size when SP is not adjusted via an immediate value
+        let mut stack_size = None;
+        let mut prologue_size = 0;
+        let mut unwind_codes = Vec::new();
+        let mut found_end = false;
+
+        for (offset, inst, size) in func.inst_offsets(entry_block, &isa.encoding_info()) {
+            // x64 ABI prologues cannot exceed 256 bytes in length
+            if (offset + size) > 256 {
+                return None;
```
sunfishcode: If we ever do emit a prologue that big, what are the consequences if we try to unwind through such a function, without unwind information?

In any case, this seems like a good place for a `debug!` message.

cranelift-codegen/Cargo.toml
peterhuene: I'll add.

cranelift-filetests/src/test_unwind.rs
peterhuene: Yes, they are defined in the ABI.  I'll add a link to the document (it's the same as linked in other places in the PR).

cranelift-codegen/src/isa/x86/unwind.rs
peterhuene: It would likely result in the operating system raising an unhandled invalid stack exception whenever it attempts to unwind through such a frame.  This is what's currently happening for most traps in Wasmtime on Windows.

It shouldn't be possible for us to emit a prologue that big for the Windows ABI, so I think we could perhaps panic/ICE if we hit this condition.

cranelift-codegen/src/isa/x86/unwind.rs
peterhuene: In Windows x64 ABI, the only functions that don't need unwind information are "leaf functions", which have specific requirements:

* cannot modify non-volatile registers
* cannot call other functions
* does not need to re-align the stack (i.e. don't need a prologue).
* can be unwound via simulation of a return instruction

As such, they do not require unwind information because the OS can simply "simulate a return" for any function that doesn't have unwind information.

However, frame functions (i.e. non-leaf functions) must have prologues because they are required to realign the stack even if they have no need for their own stack space, provided the frame function calls another function.  If the frame function does not call another function, then it is not required to realign the stack, but then there's probably another reason why the function isn't a "leaf" and would require a prologue anyway (stack allocation, saving of non-volatile registers, etc).

cranelift-codegen/src/isa/x86/unwind.rs
sunfishcode: Makes sense. A panic sounds reasonable here.

cranelift-filetests/src/test_unwind.rs
sunfishcode: Ah, ok. A link isn't necessary then, just a comment mentioning that these are Windows ABI numbers would do it.

peterhuene: I'll add.

cranelift-codegen/src/isa/x86/unwind.rs
peterhuene: I'll add.

cranelift-codegen/src/isa/x86/unwind.rs
peterhuene: Looking over the code, it does seem like we currently always generate a function prologue for fastcall functions that, at least, allocates homing space for callees (even if the function doesn't call), so we could instead panic if the prologue end marker isn't set and unwind information is requested, rather than silently omit unwind information.

In the future, the code generation should probably omit frame pointers and detect "leaf" functions that don't need prologues.  At that point we could stop panicking and instead not emit unwind information.

peterhuene:
Rather than waiting on #679 to merge first (which hasn't had much progress in the past week), I think we may want to get this merged (pending approval) to unblock getting the Wasm test suites passing on Windows for Wasmtime.  The Windows unwind information is fairly isolated and I can easily refactor this, as needed, once the tracking frame layout changes PR is merged.

cranelift-codegen/src/isa/x86/unwind.rs
peterhuene: FYI: I've filed https://github.com/CraneStation/cranelift/issues/1201 to track potential improvements for fastcall code generation.

PR #1160, opened 2019-10-18
Remove dynamic name indirections for formats

bnjbvr:
This does what's been discussed in #955 regarding formats. Now they're static attributes of a big structure containing all of them, so there's no risk of using an incorrect format by mistake. Sorry for the last commit which is kind of big; there's some fundamental changes in there, but most of it is pretty mechanical (pass explicit format to instructions).

@abrown, are you interested in reviewing this?

### DIFF
--- a/cranelift-codegen/meta/src/cdsl/formats.rs
+++ b/cranelift-codegen/meta/src/cdsl/formats.rs
@@ -2,9 +2,8 @@ use crate::cdsl::operands::{Operand, OperandKind};
 
 use std::collections::{HashMap, HashSet};
 use std::fmt;
+use std::rc::Rc;
 use std::slice;
-
-use cranelift_entity::{entity_impl, PrimaryMap};
 
 /// An immediate field in an instruction format.
 ///
@@ -152,15 +151,10 @@ impl InstructionFormatBuilder {
     }
 }
 
-#[derive(Copy, Clone, Debug, PartialEq, Eq, Hash, PartialOrd, Ord)]
-pub struct InstructionFormatIndex(u32);
-entity_impl!(InstructionFormatIndex);
-
 pub struct FormatRegistry {
-    /// Map (immediate kinds names, number of values, has varargs) to an instruction format index
-    /// in the actual map.
-    sig_to_index: HashMap<(Vec<String>, usize, bool), InstructionFormatIndex>,
-    map: PrimaryMap<InstructionFormatIndex, InstructionFormat>,
+    /// Map (immediate kinds names, number of values, has varargs) to an instruction format.
+    sig_to_index: HashMap<(Vec<String>, usize, bool), usize>,
+    formats: Vec<Rc<InstructionFormat>>,
     name_set: HashSet<&'static str>,
 }
 
@@ -168,14 +162,14 @@ impl FormatRegistry {
     pub fn new() -> Self {
         Self {
             sig_to_index: HashMap::new(),
-            map: PrimaryMap::new(),
+            formats: Vec::new(),
             name_set: HashSet::new(),
         }
     }
 
     /// Find an existing instruction format that matches the given lists of instruction inputs and
     /// outputs.
-    pub fn lookup(&self, operands_in: &Vec<Operand>) -> InstructionFormatIndex {
+    pub fn lookup(&self, operands_in: &Vec<Operand>) -> &Rc<InstructionFormat> {
         let mut imm_keys = Vec::new();
         let mut num_values = 0;
         let mut has_varargs = false;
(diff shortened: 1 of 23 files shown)


cranelift-codegen/meta/src/shared/instructions.rs
```diff
@@ -906,6 +948,7 @@ pub(crate) fn define(
         refer to a byte inside the stack slot:
         `0 <= Offset < sizeof(SS)`.
         "#,
+            &formats.stack_load,
```
bjorn3: I think the stack_load format should be renamed, as it is used for stack_addr here too.

bjorn3:
Re 0d5c3e9:

It would be nice to actually use the doc. It would be possible to split `InstBuilder` into multiple traits (one for each inst group) and then make `InstBuilder` depend on them all. The traits can then have doc comments. It would also be useful to split an internal instruction group from the shared one. That way when viewing the shared trait, only the insts supposed to be used by the user are visible.

cranelift-codegen/meta/src/gen_inst.rs
```diff
@@ -1060,36 +1062,34 @@ fn gen_builder(instructions: &AllInstructions, formats: &FormatRegistry, fmt: &m
...
+        for format in formats {
             gen_format_constructor(format, fmt);
         }
     });
     fmt.line("}");
 }
 
 pub(crate) fn generate(
-    shared_defs: &SharedDefinitions,
+    formats: &Vec<&InstructionFormat>,
```
abrown: I would expect `Vec<InstructionFormat>` here?

cranelift-codegen/meta/src/shared/mod.rs
```diff
@@ -30,21 +35,64 @@ pub(crate) fn define() -> Definitions {
...
+impl Definitions {
+    /// Verifies certain properties of formats.
+    ///
+    /// - Formats must be uniquely named: if two formats have the same name, they must refer to the
+    /// same data. Otherwise, two format variants in the codegen crate would have the same name.
+    /// - Formats must be structurally different from each other. Otherwise, this would lead to
+    /// code duplicate in the codegen crate.
+    ///
+    /// Returns a list of all the instruction formats effectively used.
+    pub fn verify_instruction_formats(&self) -> Vec<&InstructionFormat> {
```
abrown: Is this new or does it replace something that I missed?

cranelift-codegen/meta/src/cdsl/formats.rs
```diff
@@ -75,6 +72,18 @@ impl InstructionFormat {
                 )
             })
     }
+
+    /// Returns a tuple that uniquely identifies the structure.
+    pub fn structure(&self) -> (usize, bool, Vec<&'static str>) {
```
abrown: Ok, I see now how this is used; for clarity could this be a struct instead of an anonymous tuple? (if it introduces weird import issues then never mind)

cranelift-codegen/meta/src/cdsl/instructions.rs
```diff
@@ -533,6 +533,57 @@ impl Bindable for BoundInstruction {
     }
 }
 
+/// Checks that the input operands actually match the given format.
+fn verify_format(inst_name: &str, operands_in: &Vec<Operand>, format: &InstructionFormat) {
```
abrown: Is this new code or does it replace something else?

cranelift-codegen/meta/src/gen_inst.rs
abrown: (Ah, I see that we are referring to the `Rc<InstructionFormat>` as a `&InstructionFormat` but what about the `Vec`?)

bnjbvr:
> It would also be useful to split an internal instruction group from the shared one. That way when viewing the shared trait, only the insts supposed to be used by the user are visible.

I don't see a good reason from disallowing group (=ISA) specific instructions, as long as the producer is aware of their behavior and limitations.

> It would be nice to actually use the doc. It would be possible to split InstBuilder into multiple traits (one for each inst group) and then make InstBuilder depend on them all. The traits can then have doc comments.

Apart from documentation, what value would be brought by splitting InstBuilder into Instruction-group specific traits? So far, all the instructions' names in the x86 group are prefixed with `x86_`, which looks quite self-documenting?

cranelift-codegen/meta/src/gen_inst.rs
bnjbvr: Indeed, it could consume the vector instead of letting it live after the call. Since there's no subsequent uses of this vector, I'll do this.

cranelift-codegen/meta/src/shared/mod.rs
bnjbvr: It is replacing code in `lookup`, that made sure to find the right format. Since we're now specifying it manually, a check is required to ensure it matches what the instruction actually needs.

cranelift-codegen/meta/src/cdsl/formats.rs
bnjbvr: Sure!

cranelift-codegen/meta/src/cdsl/instructions.rs
bnjbvr: It is replacing code in FormatRegistry::lookup, that made sure to find the right format. Since we're now specifying it manually, a check is required to ensure it matches what the instruction actually needs.

bjorn3:
> I don't see a good reason from disallowing group (=ISA) specific instructions, as long as the producer is aware of their behavior and limitations.

I actually meant that when showing the docs of the shared trait, you dont see isa specific and internal insts. The other traits will still be public.

> Apart from documentation, what value would be brought by splitting InstBuilder into Instruction-group specific traits? So far, all the instructions' names in the x86 group are prefixed with x86_, which looks quite self-documenting?

Not much, except maybe preventing accidential use of private insts during pre-opt. It does make the docs easier to rwad though, as the pages are shorter, and you know which ones are internal.

bnjbvr:
> Not much, except maybe preventing accidential use of private insts during pre-opt. It does make the docs easier to rwad though, as the pages are shorter, and you know which ones are internal.

I think there's a misunderstanding here. The x86 instruction group doesn't contain internal instructions, but IR instructions only mapping 1:1 to x86 machine instructions. There's no reason these can't be used when producing CLIF code, as stated below. There are already internal instructions in the "shared" instruction group (e.g. iconcat/isplit). Maybe you were suggesting to split the existing "shared" group into a public vs a private one (e.g containing iconcat/isplit)? If so, it does definitely make sense, and is definitely outside the scope for this PR :)

bjorn3:
> Maybe you were suggesting to split the existing "shared" group into a public vs a private one (e.g containing iconcat/isplit)?

Yes

> If so, it does definitely make sense, and is definitely outside the scope for this PR :)

+1

PR #1162, opened 2019-10-18
Add `all_true` and `any_true` reductions

abrown:
These are mainly legalization changes though I had to add a new `x86_ptest` instruction as well as the  `vany_true` and `vall_true` instructions to legalize from (see details in the commit messages). One typing question: I chose to implement `vany_true` and `vall_true` for any `TxN` type even though the spec only calls for `IxN` because there was no overhead besides generating an extra `raw_bitcast` (a noop) in `vall_true`; however, do we want to restrict these to only `IxN` types?

### DIFF
--- a/cranelift-codegen/meta/src/isa/x86/encodings.rs
+++ b/cranelift-codegen/meta/src/isa/x86/encodings.rs
@@ -521,6 +521,7 @@ pub(crate) fn define(
     let x86_psll = x86.by_name("x86_psll");
     let x86_psra = x86.by_name("x86_psra");
     let x86_psrl = x86.by_name("x86_psrl");
+    let x86_ptest = x86.by_name("x86_ptest");
     let x86_push = x86.by_name("x86_push");
     let x86_sdivmodx = x86.by_name("x86_sdivmodx");
     let x86_smulx = x86.by_name("x86_smulx");
@@ -1988,6 +1989,10 @@ pub(crate) fn define(
         // xor
         let bxor = bxor.bind(vector(ty, sse_vector_size));
         e.enc_32_64(bxor, rec_fa.opcodes(&PXOR));
+
+        // ptest
+        let x86_ptest = x86_ptest.bind(vector(ty, sse_vector_size));
+        e.enc_32_64_maybe_isap(x86_ptest, rec_fcmp.opcodes(&PTEST), Some(use_sse41_simd));
     }
 
     // SIMD bitcast from I32/I64 to the low bits of a vector (e.g. I64x2); this register movement
--- a/cranelift-codegen/meta/src/isa/x86/legalize.rs
+++ b/cranelift-codegen/meta/src/isa/x86/legalize.rs
@@ -55,12 +55,14 @@ pub(crate) fn define(shared: &mut SharedDefinitions, x86_instructions: &Instruct
     let shuffle = insts.by_name("shuffle");
     let srem = insts.by_name("srem");
     let sshr = insts.by_name("sshr");
+    let trueif = insts.by_name("trueif");
     let udiv = insts.by_name("udiv");
     let umulhi = insts.by_name("umulhi");
     let ushr_imm = insts.by_name("ushr_imm");
     let urem = insts.by_name("urem");
     let ushr = insts.by_name("ushr");
     let vconst = insts.by_name("vconst");
+    let vany_true = insts.by_name("vany_true");
 
     let x86_bsf = x86_instructions.by_name("x86_bsf");
     let x86_bsr = x86_instructions.by_name("x86_bsr");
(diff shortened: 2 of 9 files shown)


cranelift-wasm/src/code_translator.rs
```diff
@@ -1151,8 +1151,15 @@ pub fn translate_operator<FE: FuncEnvironment + ?Sized>(
         | Operator::I16x8AnyTrue
         | Operator::I32x4AnyTrue
         | Operator::I64x2AnyTrue => {
-            let bool = builder.ins().vany_true(state.pop1());
-            state.push1(builder.ins().bint(I32, bool))
+            let bool_result = builder.ins().vany_true(state.pop1());
+            state.push1(builder.ins().bint(I32, bool_result))
```
bnjbvr: Here we go, this belongs in the previous wasm translation commit ;)

cranelift-codegen/meta/src/isa/x86/legalize.rs
```diff
@@ -446,6 +451,39 @@ pub(crate) fn define(shared: &mut SharedDefinitions, x86_instructions: &Instruct
...
+    let eq = Literal::enumerator_for(&imm.intcc, "eq");
+    for ty in ValueType::all_lane_types().filter(allowed_simd_type) {
+        let vall_true = vall_true.bind(vector(ty, sse_vector_size));
+        let lane_type_as_int = LaneType::int_from_bits(ty.lane_bits() as u16);
+        let raw_bitcast_to_int = raw_bitcast.bind(vector(lane_type_as_int, sse_vector_size));
+        narrow.legalize(
+            def!(y = vall_true(x)),
+            vec![
+                def!(a = vconst(zeroes)),
+                def!(b = raw_bitcast_to_int(x)),
```
bnjbvr: I think it would be fine to branch on x's type to avoid this raw_bitcast_to_int. Trying to minimize the number of IR instructions seems like a valuable goal in general, because it reduces the amount of allocated memory in the whole pipeline, makes each optimization pass run faster etc.

cranelift-codegen/meta/src/isa/x86/encodings.rs
```diff
@@ -1812,9 +1813,7 @@ pub(crate) fn define<'defs>(
 
     // SIMD bitcast all 128-bit vectors to each other (for legalizing splat.x16x8).
     for from_type in ValueType::all_lane_types().filter(allowed_simd_type) {
-        for to_type in
-            ValueType::all_lane_types().filter(|t| allowed_simd_type(t) && *t != from_type)
```
bnjbvr: I *think* we should keep this, see also other comment below.

cranelift-codegen/meta/src/isa/x86/legalize.rs
```diff
@@ -446,6 +449,20 @@ pub(crate) fn define(shared: &mut SharedDefinitions, x86_instructions: &Instruct
         );
     }
 
+    // SIMD vany_true; `PTEST x, x` will set the ZF flag to 0 if any bits in x contain a 1;
+    // we then set a 1 in a GPR using `SETNZ` (in CLIF, `trueif ne ...`). Remember that the b1
+    // value saved in the GPR by `SETNZ` will not set the upper bits so `bint.i32` or `bextend`
+    // should be used to zero these upper bits. See the Intel manual (vol. 1, appdx. B) to
+    // correlate `ZF = 0` (from PTEST) to `NZ` (used in SETcc) to `ne` (used by CLIF).
+    let ne = Literal::enumerator_for(&imm.intcc, "ne");
+    for ty in ValueType::all_lane_types().filter(allowed_simd_type) {
```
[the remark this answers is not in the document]
abrown: `PMOVMSKB` uses the MSB and I think the Cranelift boolean is set in the LSB.

cranelift-codegen/meta/src/isa/x86/legalize.rs
```diff
@@ -463,6 +465,25 @@ pub(crate) fn define(shared: &mut SharedDefinitions, x86_instructions: &Instruct
...
+        let vall_true = vall_true.bind(vector(ty, sse_vector_size));
+        let lane_type_as_int = LaneType::int_from_bits(ty.lane_bits() as u16);
+        let raw_bitcast_to_int = raw_bitcast.bind(vector(lane_type_as_int, sse_vector_size));
+        narrow.legalize(
+            def!(y = vall_true(x)),
+            vec![
+                def!(a = vconst(zeroes)),
+                def!(b = raw_bitcast_to_int(x)),
+                def!(c = icmp(eq, b, a)),
+                def!(d = x86_ptest(c, c)),
```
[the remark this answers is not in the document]
abrown: Yes, and I think we can do somethin like this as a part of #1015. I didn't use `PMOVMSKB` here for the reason mentioned earlier (MSB vs LSB) but also it would involve an extra instruction or two at the end to convert the extracted `11111...` to Wasm's expected `...00001`.

PR #1167, opened 2019-10-23
Simplify binding of IntCC::Equals to SIMD `icmp`

abrown:
  here. Related to #955 and already merged #1012
  first. See title
- [x] This PR contains test cases, if meaningful. No significant change to behavior, should pass existing tests.

### DIFF
--- a/cranelift-codegen/meta/src/cdsl/instructions.rs
+++ b/cranelift-codegen/meta/src/cdsl/instructions.rs
@@ -426,17 +426,15 @@ impl From<Immediate> for BindParameter {
 }
 
 #[derive(Clone)]
-#[allow(dead_code)] // TODO(#1150): remove this once we use it in legalization patterns.
 pub enum Immediate {
-    UInt8(u8),
-    UInt128(u128),
+    // When needed, this enum should be expanded to include other immediate types (e.g. u8, u128).
+    IntCC(IntCC),
 }
 
 impl Display for Immediate {
     fn fmt(&self, f: &mut Formatter) -> Result<(), Error> {
         match self {
-            Immediate::UInt8(x) => write!(f, "{}", x),
-            Immediate::UInt128(x) => write!(f, "{}", x),
+            Immediate::IntCC(x) => write!(f, "IntCC::{:?}", x),
         }
     }
 }
@@ -773,9 +771,6 @@ pub enum FormatPredicateKind {
 
     /// Is the referenced data object colocated?
     IsColocatedData,
-
-    /// Does the operation have a specific condition code?
-    HasConditionCode(IntCC),
 }
 
 #[derive(Clone, Hash, PartialEq, Eq)]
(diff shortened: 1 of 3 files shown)


cranelift-codegen/meta/src/cdsl/operands.rs
```diff
@@ -74,6 +74,15 @@ impl Operand {
...
     }
+
+    /// Determine the name to use for this operand value in the generated Rust code.
+    pub fn rust_name(&self) -> &'static str {
+        if let Some(name) = self.kind.default_member {
+            name
+        } else {
+            self.name
+        }
+    }
```
abrown: @bnjbvr: is this correct? This change uses the `self.kind.default_member` branch of the `if-else` but I don't know if the `else` is correct (otherwise `self.kind.name`?).

cranelift-codegen/meta/src/cdsl/instructions.rs
```diff
@@ -430,13 +430,15 @@ impl From<Immediate> for BindParameter {
 pub enum Immediate {
```
bnjbvr: Can you also remove the dead_code allowance above (and mention in the commit message that this closes this issue)

cranelift-codegen/meta/src/cdsl/operands.rs
bnjbvr: Good question; at some point I'll try to clean up all the "name", because they're not very eloquent. In this case, I think we can use `kind.imm_name()` in the caller, and not add the new function; can you check this, please? If not, then since we're really asking for the operand kind *member* name, it should be `self.kind.name`.

abrown: I tried `InstructionPredicate::new_is_field_equal(..., immediate_operand.kind.imm_name().unwrap(), ..)` but this failed with:
```
thread 'main' panicked at 'unexpected immediate field named intcc in instruction format IntCompare', cranelift-codegen/meta/src/cdsl/formats.rs:77:17
stack backtrace:
```
I think the issue here is that `predicates::is_equal` needs to use the actual Rust name for the IntCC value (https://github.com/CraneStation/cranelift/blob/d2feec1bb5854ca03f9fcdda6d58209c3acbd124/cranelift-codegen/meta/src/cdsl/instructions.rs#L827). And I think this is generated from `default_member` if it is set. In the cases where `default_member` is not set, I think (but am not sure) that the generator uses `Operand::name` and not `OperandKind::name`, which would make sense because you want to use names in Rust like `y` or `x` and not `intcc` and `uint8` that could be duplicated in certain formats.

cranelift-codegen/meta/src/cdsl/instructions.rs
abrown: Ok, I'll pull out the Uint8 and Uint128 for now then since they are still unused and leave a comment.

cranelift-codegen/meta/src/cdsl/operands.rs
bnjbvr: Right (definitely needs refactoring :)). So the default member is the operand's name when it's not defined, and it's not defined only for variable arguments. We can just use `default_member` here and `expect` with an error message in the caller.

PR #1169, opened 2019-10-24
Fix clippy warnings.

peterhuene:
This PR fixes the current set of (stable) clippy warnings in the repo.


peterhuene:
I'm open to undoing any of these changes in favor of disabling specific lints, provided we can get clean runs of clippy so that contributors can lint their own changes locally without the (hundreds?) of warnings currently being generated as noise to sort through.

PR #1171, opened 2019-10-25
Add SIMD integer comparisons on x86

abrown:
  here: _this is part of an ongoing effort to add SIMD support to Cranelift_.
- [X] A short description of what this does, why it is needed; if the
  first: _see title._

### DIFF
--- a/cranelift-codegen/meta/src/isa/x86/encodings.rs
+++ b/cranelift-codegen/meta/src/isa/x86/encodings.rs
@@ -517,6 +517,10 @@ pub(crate) fn define(
     let x86_pop = x86.by_name("x86_pop");
     let x86_pextr = x86.by_name("x86_pextr");
     let x86_pinsr = x86.by_name("x86_pinsr");
+    let x86_pmaxs = x86.by_name("x86_pmaxs");
+    let x86_pmaxu = x86.by_name("x86_pmaxu");
+    let x86_pmins = x86.by_name("x86_pmins");
+    let x86_pminu = x86.by_name("x86_pminu");
     let x86_pshufd = x86.by_name("x86_pshufd");
     let x86_pshufb = x86.by_name("x86_pshufb");
     let x86_psll = x86.by_name("x86_psll");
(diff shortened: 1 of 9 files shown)


abrown:
The first commit has its own PR (#1167) so it can be discussed separately from the actual SIMD implementation.

cranelift-codegen/meta/src/isa/x86/encodings.rs
```diff
@@ -2028,21 +2033,46 @@ pub(crate) fn define(
...
+        (I64, IntCC::Equal, &PCMPEQQ[..], Some(use_sse41_simd)),
+        (I8, IntCC::SignedGreaterThan, &PCMPGTB[..], None),
+        (I16, IntCC::SignedGreaterThan, &PCMPGTW[..], None),
+        (I32, IntCC::SignedGreaterThan, &PCMPGTD[..], None),
+        (
+            I64,
+            IntCC::SignedGreaterThan,
+            &PCMPGTQ,
+            Some(use_sse42_simd),
+        ),
```
sstangl: It's a little unfortunate that the formatter is putting this one-liner over six lines.

Maybe some part of that could be factored out? `use INTCC::*`?

cranelift-reader/src/parser.rs
```diff
@@ -3255,6 +3261,20 @@ mod tests {
         );
     }
 
+    #[test]
+    fn u8_as_hex() {
+        macro_rules! parse_as_uimm8 {
+            ($text:expr) => {{
+                Parser::new($text).match_uimm8("unable to parse u8")
+            }};
+        }
```
sstangl: I understand that this macro is keeping with the style of the file, but I wonder why it's using a macro at all instead of just a normal helper function. It seems that if a normal function would suffice, that we should just use that.

Honestly this code is simple enough that even just inlining `Parser::new()` would be perfectly fine.

sstangl:
Also, sorry for the long review time! I'm still getting used to the GitHub interface, and it's surprisingly easy to overlook things.

abrown:
> Also, sorry for the long review time!

Not a big deal but how can I contact you outside of GitHub? My e-mail should be available on my profile and I try to be online at the #cranelift IRC--what do you tend to use?

cranelift-reader/src/parser.rs
abrown: Yeah, good point; I went with a nested function.

PR #1174, opened 2019-10-27
Tls support for ELF and MachO

### LINKED ISSUE #963: TLS support
- What is the feature or code improvement you would like to do in Cranelift? Supporting native TLS.
- What is the value of adding this in Cranelift? This is required for https://github.com/bjorn3/rustc_codegen_cranelift/issues/388 (support multithreading in cg_clif)
- Do you have an implementation plan, and/or ideas for data structures or
  algorithms to use? Do the same thing as LLVM
- Have you considered alternative implementations? If so, how are they better
  or worse than your proposal?
    * Use a TLS base like Spidermonkey. That would make it impossible to interface with TLS variables created by LLVM or any other compiler which uses the native TLS facilities. It also wouldn't work when linking multiple objects together, as the native linker doesn't understand anything about it. Lastly it requires reserving a register, which won't work well when interfacing with C.

Edit: filled in new issue template

bjorn3:
This adds TLS (thread local storage) support for ELF targets to Cranelift

* [x] ASM of the `tls_example.clif` file has been verifier to match the ASM emitted by LLVM at https://github.com/bjorn3/rustc_codegen_cranelift/issues/388#issuecomment-531466812
* [x] Test runtime: object::write doesn't have enough TLS support yet to test execution.
* [x] Verify encodings are correct and idiomatic.
* [x] Rustfmt
* [x] Add flag to switch between TLS styles (ELF, machO) and set it automatically from either codegen or backends.
* [x] Add tests

  `bnjbvr`. <strike>The list of suggested reviewers on the right can help you.</strike> No list of suggested reviewers shown on the right side.

Fixes #963

cc @philipc https://github.com/bjorn3/rustc_codegen_cranelift/issues/388#issuecomment-526909856
cc @abrown Because you implemented SIMD support for x86, I think you know enough of x86 encodings to verify that the encodings are correct.

### DIFF
--- a/cranelift-codegen/meta/src/isa/x86/instructions.rs
+++ b/cranelift-codegen/meta/src/isa/x86/instructions.rs
@@ -552,7 +552,6 @@ pub(crate) fn define(
 
     let GV = &Operand::new("GV", &entities.global_value);
     let addr = &Operand::new("addr", i64_t);
-    let clobber = &Operand::new("clobber", i64_t); // FIXME remove need for this
 
     ig.push(
         Inst::new(
@@ -564,7 +563,7 @@ pub(crate) fn define(
             &formats.unary_global_value,
         )
         .operands_in(vec![GV])
-        .operands_out(vec![addr, clobber]),
+        .operands_out(vec![addr]),
     );
     ig.push(
         Inst::new(
@@ -576,7 +575,7 @@ pub(crate) fn define(
             &formats.unary_global_value,
         )
         .operands_in(vec![GV])
-        .operands_out(vec![addr, clobber]),
+        .operands_out(vec![addr]),
     );
 
     ig.build()
--- a/cranelift-codegen/meta/src/isa/x86/instructions.rs
+++ b/cranelift-codegen/meta/src/isa/x86/instructions.rs
@@ -7,6 +7,7 @@ use crate::cdsl::operands::Operand;
 use crate::cdsl::types::ValueType;
 use crate::cdsl::typevar::{Interval, TypeSetBuilder, TypeVar};
 
+use crate::shared::entities::EntityRefs;
 use crate::shared::formats::Formats;
 use crate::shared::immediates::Immediates;
 use crate::shared::types;
@@ -16,6 +17,7 @@ pub(crate) fn define(
     mut all_instructions: &mut AllInstructions,
     formats: &Formats,
     immediates: &Immediates,
+    entities: &EntityRefs,
 ) -> InstructionGroup {
     let mut ig = InstructionGroupBuilder::new(&mut all_instructions);
 
@@ -540,6 +542,41 @@ pub(crate) fn define(
         )
         .operands_in(vec![x, y])
         .operands_out(vec![a]),
+    );
+
+    let i64_t = &TypeVar::new(
+        "i64_t",
+        "A scalar 64bit integer",
+        TypeSetBuilder::new().ints(64..64).build(),
+    );
+
+    let GV = &Operand::new("GV", &entities.global_value);
+    let addr = &Operand::new("addr", i64_t);
+    let clobber = &Operand::new("clobber", i64_t); // FIXME remove need for this
+
+    ig.push(
+        Inst::new(
+            "x86_elf_tls_get_addr",
+            r#"
+        Elf tls get addr -- This implements the GD TLS model for ELF. The clobber output should
+        not be used.
+            "#,
+            &formats.unary_global_value,
+        )
+        .operands_in(vec![GV])
+        .operands_out(vec![addr, clobber]),
+    );
+    ig.push(
+        Inst::new(
+            "x86_macho_tls_get_addr",
+            r#"
+        Mach-O tls get addr -- This implements TLS access for Mach-O. The clobber output should
+        not be used.
+            "#,
+            &formats.unary_global_value,
+        )
+        .operands_in(vec![GV])
+        .operands_out(vec![addr, clobber]),
     );
 
     ig.build()
--- a/cranelift-codegen/meta/src/isa/x86/recipes.rs
+++ b/cranelift-codegen/meta/src/isa/x86/recipes.rs
@@ -393,7 +393,6 @@ pub(crate) fn define<'shared>(
     let reg_rax = Register::new(gpr, regs.regunit_by_name(gpr, "rax"));
     let reg_rcx = Register::new(gpr, regs.regunit_by_name(gpr, "rcx"));
     let reg_rdx = Register::new(gpr, regs.regunit_by_name(gpr, "rdx"));
-    let reg_rdi = Register::new(gpr, regs.regunit_by_name(gpr, "rdi"));
     let reg_r15 = Register::new(gpr, regs.regunit_by_name(gpr, "r15"));
 
     // Stack operand with a 32-bit signed displacement from either RBP or RSP.
@@ -3265,10 +3264,13 @@ pub(crate) fn define<'shared>(
         ),
     );
 
+    // Both `elf_tls_get_addr` and `macho_tls_get_addr` require all caller-saved registers to be spilled.
+    // This is currently special cased in `regalloc/spilling.rs` in the `visit_inst` function.
+
     recipes.add_recipe(
         EncodingRecipeBuilder::new("elf_tls_get_addr", &formats.unary_global_value, 16)
             // FIXME Correct encoding for non rax registers
-            .operands_out(vec![reg_rax, reg_rdi])
+            .operands_out(vec![reg_rax])
             .emit(
                 r#"
                     // output %rax
@@ -3303,7 +3305,7 @@ pub(crate) fn define<'shared>(
     recipes.add_recipe(
         EncodingRecipeBuilder::new("macho_tls_get_addr", &formats.unary_global_value, 9)
             // FIXME Correct encoding for non rax registers
-            .operands_out(vec![reg_rax, reg_rdi])
+            .operands_out(vec![reg_rax])
             .emit(
                 r#"
                     // output %rax
--- a/cranelift-codegen/meta/src/isa/x86/recipes.rs
+++ b/cranelift-codegen/meta/src/isa/x86/recipes.rs
@@ -393,6 +393,7 @@ pub(crate) fn define<'shared>(
     let reg_rax = Register::new(gpr, regs.regunit_by_name(gpr, "rax"));
     let reg_rcx = Register::new(gpr, regs.regunit_by_name(gpr, "rcx"));
     let reg_rdx = Register::new(gpr, regs.regunit_by_name(gpr, "rdx"));
+    let reg_rdi = Register::new(gpr, regs.regunit_by_name(gpr, "rdi"));
     let reg_r15 = Register::new(gpr, regs.regunit_by_name(gpr, "r15"));
 
     // Stack operand with a 32-bit signed displacement from either RBP or RSP.
@@ -3259,9 +3260,69 @@ pub(crate) fn define<'shared>(
     recipes.add_recipe(
         EncodingRecipeBuilder::new("safepoint", &formats.multiary, 0).emit(
             r#"
-            sink.add_stackmap(args, func, isa);
-        "#,
+                sink.add_stackmap(args, func, isa);
+            "#,
         ),
+    );
+
+    recipes.add_recipe(
+        EncodingRecipeBuilder::new("elf_tls_get_addr", &formats.unary_global_value, 16)
+            // FIXME Correct encoding for non rax registers
+            .operands_out(vec![reg_rax, reg_rdi])
+            .emit(
+                r#"
+                    // output %rax
+                    // clobbers %rdi
+
+                    // Those data16 prefixes are necessary to pad to 16 bytes.
+
+                    // data16 lea gv@tlsgd(%rip),%rdi
+                    sink.put1(0x66); // data16
+                    sink.put1(0b01001000); // rex.w
+                    const LEA: u8 = 0x8d;
+                    sink.put1(LEA); // lea
+                    modrm_riprel(0b111/*out_reg0*/, sink); // 0x3d
+                    sink.reloc_external(Reloc::ElfX86_64TlsGd,
+                                        &func.global_values[global_value].symbol_name(),
+                                        -4);
+                    sink.put4(0);
+
+                    // data16 data16 callq __tls_get_addr-4
+                    sink.put1(0x66); // data16
+                    sink.put1(0x66); // data16
+                    sink.put1(0b01001000); // rex.w
+                    sink.put1(0xe8); // call
+                    sink.reloc_external(Reloc::X86CallPLTRel4,
+                                        &ExternalName::LibCall(LibCall::ElfTlsGetAddr),
+                                        -4);
+                    sink.put4(0);
+                "#,
+            ),
+    );
+
+    recipes.add_recipe(
+        EncodingRecipeBuilder::new("macho_tls_get_addr", &formats.unary_global_value, 9)
+            // FIXME Correct encoding for non rax registers
+            .operands_out(vec![reg_rax, reg_rdi])
+            .emit(
+                r#"
+                    // output %rax
+                    // clobbers %rdi
+
+                    // movq gv@tlv(%rip), %rdi
+                    sink.put1(0x48); // rex
+                    sink.put1(0x8b); // mov
+                    modrm_riprel(0b111/*out_reg0*/, sink); // 0x3d
+                    sink.reloc_external(Reloc::MachOX86_64Tlv,
+                                        &func.global_values[global_value].symbol_name(),
+                                        -4);
+                    sink.put4(0);
+
+                    // callq *(%rdi)
+                    sink.put1(0xff);
+                    sink.put1(0x17);
+                "#,
+            ),
     );
 
     recipes
--- a/cranelift-codegen/src/legalizer/globalvalue.rs
+++ b/cranelift-codegen/src/legalizer/globalvalue.rs
@@ -40,7 +40,7 @@ pub fn expand_global_value(
             global_type,
             readonly,
         } => load_addr(inst, func, base, offset, global_type, readonly, isa),
-        ir::GlobalValueData::Symbol { .. } => symbol(inst, func, gv, isa),
+        ir::GlobalValueData::Symbol { tls, .. } => symbol(inst, func, gv, isa, tls),
     }
 }
 
@@ -123,7 +123,18 @@ fn load_addr(
 }
 
 /// Expand a `global_value` instruction for a symbolic name global.
-fn symbol(inst: ir::Inst, func: &mut ir::Function, gv: ir::GlobalValue, isa: &dyn TargetIsa) {
+fn symbol(
+    inst: ir::Inst,
+    func: &mut ir::Function,
+    gv: ir::GlobalValue,
+    isa: &dyn TargetIsa,
+    tls: bool,
+) {
     let ptr_ty = isa.pointer_type();
-    func.dfg.replace(inst).symbol_value(ptr_ty, gv);
+
+    if tls {
+        func.dfg.replace(inst).tls_value(ptr_ty, gv);
+    } else {
+        func.dfg.replace(inst).symbol_value(ptr_ty, gv);
+    }
 }
--- a/cranelift-codegen/meta/src/cdsl/instructions.rs
+++ b/cranelift-codegen/meta/src/cdsl/instructions.rs
@@ -340,7 +340,7 @@ impl InstructionBuilder {
         let polymorphic_info =
             verify_polymorphic(&operands_in, &operands_out, &self.format, &value_opnums);
 
-        // Infer from output operands whether an instruciton clobbers CPU flags or not.
+        // Infer from output operands whether an instruction clobbers CPU flags or not.
         let writes_cpu_flags = operands_out.iter().any(|op| op.is_cpu_flags());
 
         let camel_name = camel_case(&self.name);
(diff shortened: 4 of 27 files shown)


cranelift-codegen/meta/src/isa/x86/recipes.rs
```diff
@@ -3035,10 +3036,87 @@ pub(crate) fn define<'shared>(
...
+    );*/
+
+    recipes.add_recipe(
+        EncodingRecipeBuilder::new("elf_tlsld", &formats.unary_global_value, 7)
+            // FIXME Should use gpr instead of reg_rdi, but that causes an unnecessary regmove instruction.
+            .operands_out(vec![reg_rdi])
+            .emit(
+                r#"
+                    // leaq @tlsld(%rip), %rdi
+                    sink.put1(0b0100_1000);
```
iximeow: Can this get a `const` like for `LEA` like:
```suggestion
                    const REX_W: u8 = 0x48;
                    sink.put1(REX_W);
```?

File: cranelift-codegen/meta/src/isa/x86/recipes.rs [hunk 2]
```diff
@@ -3035,10 +3036,87 @@ pub(crate) fn define<'shared>(
...
+        EncodingRecipeBuilder::new("elf_dtpoff32", &formats.tls_offset, 7)
+            // FIXME Correct encoding for non rax registers
+            .operands_in(vec![reg_rax])
+            .operands_out(vec![reg_rax])
+            .emit(
+                r#"
+                    // output %rax
+
+                    // leaq @dtpoff32(%rax), %rax
+                    sink.put1(0b0100_1000);
```
iximeow: Same as above re. naming this constant.

cranelift-codegen/meta/src/isa/x86/recipes.rs
bjorn3: I think one of those `rex` functions would be better, but I don't know which one to use.

philipc:
We probably should start with the "General Dynamic TLS Model", and only use "Local Dynamic TLS Model" if the required conditions are met (the variable is in the same object and ideally only if there are multiple TLS variable accesses that can share the `__tls_get_addr` call).

Also, if there are multiple TLS variable accesses, will cranelift be able to optimize the away the unneeded `__tls_get_addr` calls?

bjorn3:
> We probably should start with the "General Dynamic TLS Model"

Yeah, I didnt fully read the docs, but just followed LLVM's example.

> Also, if there are multiple TLS variable accesses, will cranelift be able to optimize the away the unneeded __tls_get_addr calls?

That would require a new optimization pass. It probably wouldn't help much for cg_clif, because of lack of inlining when MIR opts are not set to max.

bnjbvr:
Can you detail in the issue the problem you're trying to solve, and what possible implementations exist, please? The mentioned issue doesn't contain any text but a link to another issue in cg_clif. I'll comment in there.

cranelift-codegen/meta/src/isa/x86/recipes.rs
```diff
@@ -3035,10 +3036,87 @@ pub(crate) fn define<'shared>(
...
+
+    recipes.add_recipe(
+        EncodingRecipeBuilder::new("elf_tlsld", &formats.unary_global_value, 7)
+            // FIXME Should use gpr instead of reg_rdi, but that causes an unnecessary regmove instruction.
+            .operands_out(vec![reg_rdi])
+            .emit(
+                r#"
+                    // leaq @tlsld(%rip), %rdi
+                    sink.put1(0b0100_1000);
+                    const LEA: u8 = 0x8d;
```
abrown: This could probably be imported from `opcodes.rs`

cranelift-codegen/meta/src/isa/x86/instructions.rs
```diff
@@ -487,5 +489,29 @@ pub(crate) fn define(
...
     );
 
+    let GV = &operand("GV", &entities.global_value);
+    let addr = &operand("addr", iWord);
+    let x = &operand("x", iWord);
+
+    ig.push(
+        Inst::new(
+            "x86_elf_tlsld",
+            "FIXME",
```
abrown: Reminder to fill this in

cranelift-codegen/meta/src/isa/x86/instructions.rs
```diff
@@ -487,5 +489,29 @@ pub(crate) fn define(
...
+            &formats.unary_global_value,
+        )
+        .operands_in(vec![GV])
+        .operands_out(vec![addr]),
+    );
+
+    ig.push(
+        Inst::new(
+            "x86_elf_dtpoff32",
+            "FIXME",
```
abrown: Reminder to fill this in

cranelift-codegen/meta/src/isa/x86/instructions.rs
```diff
@@ -487,5 +489,29 @@ pub(crate) fn define(
         .operands_out(vec![f]),
     );
 
+    let GV = &operand("GV", &entities.global_value);
+    let addr = &operand("addr", iWord);
+    let x = &operand("x", iWord);
```
abrown: Looks like `iWord` is `32..64`: do you only want 64 here?

cranelift-codegen/meta/src/isa/x86/recipes.rs
bjorn3: Do you know what the most ideomatic way would be to encode this? Using a Template instead of Recipe and then putting the rex.w prefix and opcode in encodings.rs? And if so, what should be the declaration of that Template?

cranelift-codegen/meta/src/isa/x86/recipes.rs
```diff
@@ -3035,10 +3036,87 @@ pub(crate) fn define<'shared>(
...
+                    sink.put1(LEA);
+                    modrm_riprel(0b111 /*out_reg0*/, sink);
```
abrown: From looking up `0b111` it seems that this is using `%rdi` (from your comments I take it that that is a requirement to make this work?) but we might want to make that more clear here.

cranelift-codegen/meta/src/isa/x86/recipes.rs
```diff
@@ -3035,10 +3036,87 @@ pub(crate) fn define<'shared>(
...
+            .operands_out(vec![reg_rax])
+            .emit(
+                r#"
+                    // output %rax
+
+                    // leaq @dtpoff32(%rax), %rax
+                    sink.put1(0b0100_1000);
+                    const LEA: u8 = 0x8d;
+                    sink.put1(LEA);
+                    modrm_disp32(0b000/*in_reg0*/, 0b000/*out_reg0*/, sink);
```
abrown: Were you planning to actually use `in_reg0` and `out_reg0` here but temporarily hard-coded the `RegUnit`s?

cranelift-codegen/meta/src/isa/x86/recipes.rs
bjorn3: Technically it isnt, but it causes suboptimal codegen otherwise. Also I will change this to move to the GD tls model, which requires a specific sequence of prefixes and insts in one block.

cranelift-codegen/meta/src/isa/x86/instructions.rs
bjorn3: Indeed

abrown:
@bjorn3, I would like to understand this at a higher-level; what are you using as reference? Perhaps "ELF Handling for Thread-Local Storage"?

bjorn3:
I use https://akkadia.org/drepper/tls.pdf + LLVM output as reference.

cranelift-codegen/meta/src/isa/x86/recipes.rs
bnjbvr: It doesn't seem future-proof to add register allocation constraints because regalloc isn't doing the best job right now. We're working on replacing the current regalloc by something simpler in the future. Please either make it a TODO and open an issue about it (when this lands), or remove the constraint.

bjorn3:
Switched from the local only LD model to the global GD tls model.

cranelift-codegen/meta/src/isa/x86/recipes.rs
abrown: I just meant using the `LEA` already in https://github.com/CraneStation/cranelift/blob/a18dec8b22739e95a3f0395de26e5539b7e04afb/cranelift-codegen/meta/src/isa/x86/opcodes.rs#L132

But I think I see what you mean: we've been able to keep most of the other opcodes in `encodings.rs` and use templates, a la https://github.com/CraneStation/cranelift/blob/cabbacbe63477fcc7d1f1e36717d84f08f749a98/cranelift-codegen/meta/src/isa/x86/recipes.rs#L2417-L2428

I think that `{{PUT_OP}} ...` stuff should work for you here too. (But our lack of clarity on how to use this is probably a motivator for taking another look at recipes/templates, #955).

bjorn3: Now that I switched to the TLS GD model, I dont even think it would be possible to use `Template` anymore, as I need a specfic sequence of inst prefixes and two asm insts, so using two clif insts may cause problems if some pass decides to put an inst inbetween.

bjorn3:
Implemented MachO support. Requires https://github.com/gimli-rs/object/pull/142.

cranelift-codegen/meta/src/isa/x86/recipes.rs
```diff
@@ -3035,10 +3036,106 @@ pub(crate) fn define<'shared>(
...
+                                        0);
+                    sink.put4(0);
+                "#,
+            ),
+    );*/
+
+    recipes.add_recipe(
+        EncodingRecipeBuilder::new("elf_tls_get_addr", &formats.unary_global_value, 16)
+            // FIXME Correct encoding for non rax registers
+            .operands_out(vec![reg_rax, reg_rdi])
```
bjorn3: Is there any better way to tell regalloc that `%rdi` gets clobbered by this instruction, than just returning it?

bnjbvr: There's not really a way to signal something is clobbered. Passing it as an input or output operand constraint will cause the expected behavior, though (redirecting values living in `rdi` to other places).

bjorn3: Ok, it would be nice to avoid this extra return value though, so that I can encode tls_symbol directly, instead of legalizing to an instruction with an extra return value.

bjorn3:
Note to self: I will need to write `-4` to the `X86_64_RELOC_TLV` location and add `-4` to the addend for mach-O.

philipc:
I thought the Mach-O relocations were designed such that the implicit -4 meant the actual addend would be 0, so you wouldn't need to write an addend to the location, and I think the disassembly you did agrees with that:
```
       4:	48 8b 3d 00 00 00 00 	movq	(%rip), %rdi
		0000000000000007:  X86_64_RELOC_TLV	__ZN3tls1A17h13af140b93a46e97E@TLVP
```

bjorn3:
I actually meant substract -4. The encoding for x86_64_macho_tls_symbol already uses -4 as addend, so substracting -4 would indeed give 0.

bjorn3:
Can somebody review this?

abrown:
It looks like I reviewed what I understood above; perhaps @bnjbvr or @julian-seward1?

bjorn3:
Ping?

abrown:
@bjorn3 I'm fine with this once the conflicts are resolved and tests are added but that is because I'm not familiar enough with other ways of implementing TLS to have much comment. There's a Cranelift meeting on Monday mornings and there might be someone there that has more insight into the future Cranelift's threading support; I mention that because the whole threading thing might require some back-and-forth discussion.

bjorn3:
> I'm not familiar enough with other ways of implementing TLS to have much comment.

While there are other ways to implement TLS support, this matches the native TLS handling of ELF cq Mach-O. This way, it is possible to import/export tls variables from/to native objects. While this is not really necessary for cg_clif, as Rust makes a getter for the TLS var anyway and doesn't export the raw TLS var, for rcc it will likely be needed to interoperate with native code.

bjorn3:
I rebased this and added tests for legalization of `global_value` and encoding of `x86_{elf,macho}_tls_get_addr`. Should I also add a test that `cranelift-object` correctly writes the tls section?

> I mention that because the whole threading thing might require some back-and-forth discussion.

For cg_clif this is technically the only thing that prevents multithreading at the moment. I have made atomics atomic by using a global lock. While this is slow, at least it works. I would love to see atomic support in Cranelift, but for now TLS is more important for me.

bjorn3:
@abrown Has this been discussed?

abrown:
I think the cranelift meeting yesterday was cancelled due to a US holiday (at least I didn't attend) but I posted in a [zulip chat thread](https://bytecodealliance.zulipchat.com/#narrow/stream/217117-cranelift/topic/threading) and I will bring it up next Monday.

bjorn3:
Ok, thanks!

cranelift-codegen/src/legalizer/globalvalue.rs
```diff
@@ -123,7 +123,18 @@ fn load_addr(
...
+    tls: bool,
+) {
     let ptr_ty = isa.pointer_type();
-    func.dfg.replace(inst).symbol_value(ptr_ty, gv);
+
+    if tls {
+        func.dfg.replace(inst).tls_value(ptr_ty, gv);
+    } else {
+        func.dfg.replace(inst).symbol_value(ptr_ty, gv);
+    }
```
sunfishcode: Instead of having a `tls_value` instruction, would it work to just continue to use `symbol_value` here, and use the `tls` flag of `gv` to determine whether it's TLS or not? You could add a `is_tls_data` predicate function which could work the same way as `is_colocated_data`, `FormatPredicateKind::IsColocatedData`, and so on, and then `define_entity_ref` in cranelift-codegen/meta/src/isa/x86/encodings.rs could use that to decide when to apply the TLS legalizations.

cranelift-codegen/meta/src/isa/x86/instructions.rs
```diff
@@ -542,5 +544,40 @@ pub(crate) fn define(
...
+            "x86_macho_tls_get_addr",
+            r#"
+        Mach-O tls get addr -- This implements TLS access for Mach-O. The clobber output should
+        not be used.
+            "#,
+            &formats.unary_global_value,
+        )
+        .operands_in(vec![GV])
+        .operands_out(vec![addr, clobber]),
+    );
```
sunfishcode: The ELF and Mach-O instructions here have the same signature, and essentially represent the same functionality, except that they record which target to codegen for. Would it simplify the code to just have one instruction, and move the target check into the encoding phase?

sunfishcode (commented):
Overall this approach looks reasonable to me. TLS makes sense as a GlobalValue flag, and starting with global-dynamic sounds like a good way to get started. Just a few questions below:

cranelift-codegen/meta/src/isa/x86/instructions.rs
bjorn3: Both `x86_elf_tls_get_addr` and `x86_macho_tls_get_addr` happen to clobber a register. I am currently representing this as an extra output value. In the future there may be other TLS models added that don't clobber exactly one, but zero, two or more registers. In that case a separate instruction is needed. For symmetry I think it would be nice to keep separate instructions for ELF and Mach-O too.

sunfishcode: Makes sense!

cranelift-codegen/src/legalizer/globalvalue.rs
bjorn3: `symbol_value` can directly be encoded, while `tls_value` needs to be legalized to the correct instruction for the respective object format. See comment below for why. As far as I know, unlike an encoding, there is no way to predicate a legalization.

bjorn3:
I decided to look a bit more at the `__tls_get_addr` ABI and noticed that I made a wrong assumption. In one place, it is said that the full code sequence needs to be 16 bytes (used to for example relax GD to other TLS models, https://github.com/llvm/llvm-project/blob/deb5819d6249cfe110da9377910f7e5c88e8ee09/lld/ELF/Arch/X86_64.cpp#L186), suggesting a custom ABI, but the glibc implementation seems to require the C abi: https://github.com/bminor/glibc/blob/5f72f9800b250410cad3abfeeb09469ef12b2438/elf/dl-tls.c#L821. When looking at LLVM, this seems to indeed be the case: https://github.com/llvm/llvm-project/blob/78be61871704a451a5d9462d7e96ed6c982746d4/llvm/lib/Target/X86/X86InstrCompiler.td#L453-L471. The current code works for cg_clif, but that may be caused by the fact that rust creates a very small function just to get the tls addr, so none of the clobbered registers are used after the `global_value` instruction.

Clobbering all necessary registers will mean that I have to add even more clobber registers. Another option would be to use the existing infrastructure for the `call` and `call_indirect` instruction.

bjorn3:
I tried to add more output registers for the clobbering, but the maximum amount of outputs is 8, which is way less than the amount of registers than are clobbered.

I also tried to use the same spill logic as used by normal calls, but I can't find where it is determined which registers are caller-saved and should spilled to the stack.

iximeow:
> I also tried to use the same spill logic as used by normal calls, but I can't find where it is determined which registers are caller-saved and should spilled to the stack.

I think you might not have been able to find it because all register values that live through a call are spilled - callee-save/caller-save ABI guarantees are disregarded at the moment. [There's a note in `spilling.rs` for this](https://github.com/bytecodealliance/cranelift/blob/master/cranelift-codegen/src/regalloc/spilling.rs#L267-L269). This `if` could be extended to spill around clobber-happy TLS operations.

bjorn3:
Thanks @iximeow! It now correctly spills all caller-saved registers.

abrown (approved):

bjorn3:
:tada:

stefson:
@bjorn3  

I'm getting a compile error with building for aarch64-unknown-linux-gnu on current master, and found this pullrequest as the culprit when bisecting the issue. Here's the error: 

```
   Compiling cranelift-codegen-meta v0.59.0 (/tmp/wasmtime/cranelift/codegen/meta)
   Compiling cranelift-bforest v0.59.0 (/tmp/wasmtime/cranelift/bforest)
   Compiling wasi-common v0.12.0 (/tmp/wasmtime/crates/wasi-common)
   Compiling cranelift-codegen v0.59.0 (/tmp/wasmtime/cranelift/codegen)
warning: unused import: `self::call::expand_call`
  --> cranelift/codegen/src/legalizer/mod.rs:35:5
   |
35 | use self::call::expand_call;
   |     ^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: `#[warn(unused_imports)]` on by default

warning: unused import: `self::globalvalue::expand_global_value`
  --> cranelift/codegen/src/legalizer/mod.rs:36:5
   |
36 | use self::globalvalue::expand_global_value;
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

warning: unused import: `self::heap::expand_heap_addr`
  --> cranelift/codegen/src/legalizer/mod.rs:37:5
   |
37 | use self::heap::expand_heap_addr;
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^

warning: unused import: `self::table::expand_table_addr`
  --> cranelift/codegen/src/legalizer/mod.rs:39:5
   |
39 | use self::table::expand_table_addr;
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

error[E0599]: no variant or associated item named `X86ElfTlsGetAddr` found for type `ir::instructions::Opcode` in the current scope
    --> cranelift/codegen/src/regalloc/spilling.rs:272:45
     |
272  |             || opcode == crate::ir::Opcode::X86ElfTlsGetAddr
     |                                             ^^^^^^^^^^^^^^^^ variant or associated item not found in `ir::instructions::Opcode`
     | 
    ::: /tmp/wasmtime/target/aarch64-unknown-linux-gnu/debug/build/cranelift-codegen-dd375821e82eb469/out/opcodes.rs:1427:1
     |
1427 | pub enum Opcode {
     | --------------- variant or associated item `X86ElfTlsGetAddr` not found here

error[E0599]: no variant or associated item named `X86MachoTlsGetAddr` found for type `ir::instructions::Opcode` in the current scope
    --> cranelift/codegen/src/regalloc/spilling.rs:273:45
     |
273  |             || opcode == crate::ir::Opcode::X86MachoTlsGetAddr
     |                                             ^^^^^^^^^^^^^^^^^^ variant or associated item not found in `ir::instructions::Opcode`
     | 
    ::: /tmp/wasmtime/target/aarch64-unknown-linux-gnu/debug/build/cranelift-codegen-dd375821e82eb469/out/opcodes.rs:1427:1
     |
1427 | pub enum Opcode {
     | --------------- variant or associated item `X86MachoTlsGetAddr` not found here

error: aborting due to 2 previous errors

For more information about this error, try `rustc --explain E0599`.
error: could not compile `cranelift-codegen`.
```

It's propably worth noting that this only happens when executing `cargo build` from the wasmtime dir, and it doesn't when `cd wasmtime/cranelift & cargo build`. Still the error seems to be from cranelift. 

P.S: the regression is caused by https://github.com/bytecodealliance/wasmtime/commit/0a1bb3ba6ccadd4b716cd259a097bb31435c0c6d , I hope this thread is the correct one to note the author of that regression.

bjorn3:
The problem is probably that Cranelift got built without x86 support, but I added a special case to regalloc with an x86 only instruction without checking for x86 support. I think the solution is to add an instruction attribute, just like `is_call` and `is_branch`, for this special case instead.

abrown:
@bjorn3, can you create an issue in wasmtime so we can track this there?

PR #1178, opened 2019-10-28
Rename Operand::is_immediate + remove bitvector functionality

bnjbvr:
No issue:
  - trivial renaming mentioned in a TODO of the meta code.
  - finish adding "pub(crate)" declarations everywhere in the meta crate, and remove dead code.
  - remove "bit vector" functionality, unused since the transition of the meta crate to rust. Note it's not the feature for emitting SIMD constants; but a different feature to model every type variable as a bit-vector.

### DIFF
--- a/cranelift-codegen/meta/src/cdsl/encodings.rs
+++ b/cranelift-codegen/meta/src/cdsl/encodings.rs
@@ -86,11 +86,12 @@ impl EncodingBuilder {
                 }
 
                 // Add immediate value predicates
-                for (immediate_value, immediate_operand) in inst
-                    .immediate_values
-                    .iter()
-                    .zip(inst.inst.operands_in.iter().filter(|o| o.is_immediate()))
-                {
+                for (immediate_value, immediate_operand) in inst.immediate_values.iter().zip(
+                    inst.inst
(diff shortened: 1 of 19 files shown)


cranelift-codegen/meta/src/cdsl/encodings.rs
```diff
@@ -86,11 +86,12 @@ impl EncodingBuilder {
...
-                for (immediate_value, immediate_operand) in inst
-                    .immediate_values
-                    .iter()
-                    .zip(inst.inst.operands_in.iter().filter(|o| o.is_immediate()))
-                {
+                for (immediate_value, immediate_operand) in inst.immediate_values.iter().zip(
+                    inst.inst
+                        .operands_in
+                        .iter()
+                        .filter(|o| o.is_immediate_or_entityref()),
```
abrown: I probably really should have used `is_pure_immediate` here but must have missed it so I used `is_immediate` instead; I think we can probably leave this as `is_immediate`.

bnjbvr:
Github hasn't noticed the rustfmt job has successfully passed; the Azure pipelines dashboard shows this properly.

Issue #1177, opened 2019-10-28
Refactor Operand's "name" vs OperandKind's "name" / "default_member"


bnjbvr:
They're quite confusing at the moment, while serving different roles (to be confirmed):

- for an OperandKind, the default_member is the name, unless it was explicitly defined.
- an OperandKind's name is actually the type's name of the operand (duh!).
- an Operand's name is the variable name used in functions and in the AST in general.

We should at least rename these fields, at best try to remove some if they're redundant.

Issue #851, opened 2019-07-13
Document instruction input/output types in rust docs


bjorn3:

bnjbvr:
Good idea. For polymorphic instructions, we'd show type variables descriptions.

PR #1180, opened 2019-10-29
Add doc comments for operands + make handling of OperandKind simpler

bnjbvr:
This implements #851 and #1177, and removes the need for the Operand/OperandKind builders, which were hiding default decisions that were hard to find about. Now defaulting is done at runtime, or things happen very explicitly (especially for OperandKind), which I find nicer (no more magic, no need to guess how things relate to each other).

I'd recommend looking at it commit by commit, to make the review simpler. Note that some commits do some changes that are later reverted in commits, which is unfortunate because it complicates the review. But it has the benefit of keeping changes focused, and see my train of thoughts for this particular refactoring. Happy to split/squash commits if this makes things simpler.


cranelift-codegen/meta/src/gen_inst.rs
```diff
@@ -895,11 +895,17 @@ fn gen_inst_builder(inst: &Instruction, format: &InstructionFormat, fmt: &mut Fo
...
 
+    let mut args_doc = Vec::new();
+
     // The controlling type variable will be inferred from the input values if
     // possible. Otherwise, it is the first method argument.
     if let Some(poly) = &inst.polymorphic_info {
         if !poly.use_typevar_operand {
             args.push(format!("{}: crate::ir::Type", poly.ctrl_typevar.name));
+            args_doc.push(format!(
+                "- {} (control type variable): {}",
```
abrown: "controlling type variable" instead?

cranelift-codegen/meta/src/gen_inst.rs
```diff
@@ -938,6 +951,15 @@ fn gen_inst_builder(inst: &Instruction, format: &InstructionFormat, fmt: &mut Fo
     );
 
     fmt.doc_comment(&inst.doc);
+    if !args_doc.is_empty() {
+        fmt.line("///");
+        fmt.doc_comment("Inputs:");
+        fmt.line("///");
+        for doc_line in args_doc {
+            fmt.doc_comment(doc_line);
+        }
```
abrown: Why not the outputs?

cranelift-codegen/meta/src/gen_inst.rs
```diff
@@ -1058,9 +1080,13 @@ fn gen_builder(
...
         for inst in instructions.values() {
             gen_inst_builder(inst, &*inst.format, fmt);
+            fmt.empty_line();
         }
-        for format in formats {
+        for (i, format) in formats.iter().enumerate() {
             gen_format_constructor(format, fmt);
+            if i + 1 != formats.len() {
+                fmt.empty_line();
+            }
```
abrown: If we ever bring in `itertools`: https://docs.rs/itertools/0.8.1/itertools/trait.Itertools.html#method.with_position.

cranelift-codegen/meta/src/cdsl/operands.rs
```diff
@@ -19,18 +19,34 @@ use crate::cdsl::typevar::TypeVar;
...
+        }
+    }
+
+    pub fn with_doc(name: &'static str, kind: impl Into<OperandKind>, doc: &'static str) -> Self {
+        Self {
+            name,
+            doc: Some(doc),
+            kind: kind.into(),
+        }
+    }
```
abrown: I would have written this in a different API style:
```
pub fn with_doc(self, doc: &'static str) -> Self {
  Self {
    doc: Some(doc),
    ..self
  }
}
```
So that callers do `Operand::new(...).with_doc(...)`. Just a style thing so feel free to ignore, but it avoids the duplicated constructor code.

cranelift-codegen/meta/src/shared/instructions.rs
```diff
@@ -3653,9 +3653,9 @@ pub(crate) fn define(
             .build(),
     );
 
-    let lo = &operand("lo", NarrowInt);
-    let hi = &operand("hi", NarrowInt);
-    let a = &operand_doc(
+    let lo = &Operand::new("lo", NarrowInt);
+    let hi = &Operand::new("hi", NarrowInt);
+    let a = &Operand::with_doc(
```
abrown: Just looking at this makes me think: are we going to do the same thing for instructions that we did with formats and immediates?

cranelift-codegen/meta/src/cdsl/operands.rs
```diff
@@ -115,18 +115,55 @@ pub(crate) enum OperandKindFields {
...
+
+    /// Name of this OperandKind in the format's member field.
+    pub fn default_member(&self) -> Option<&'static str> {
+        if let Some(member) = &self.default_member {
+            return Some(member);
+        }
+        match &self.fields {
+            OperandKindFields::ImmEnum(_) | OperandKindFields::ImmValue => Some("imm"),
```
abrown: I was wondering about this a few days ago: if we were to have an instruction with multiple immediates then there would be a compilation error of the generated code, right?

cranelift-codegen/meta/src/cdsl/operands.rs
```diff
@@ -116,7 +116,7 @@ pub(crate) enum OperandKindFields {
 pub(crate) struct OperandKind {
     pub name: &'static str,
     doc: Option<&'static str>,
-    default_member: Option<&'static str>,
+    format_field_name: Option<&'static str>,
```
abrown: I'm not sure I understand why this is called `format_field_name` and not just `field_name`--is it because it is related to the format somehow? We might need to document why this struct field is needed and what it is used for.

cranelift-codegen/meta/src/cdsl/operands.rs
```diff
@@ -202,7 +202,7 @@ impl OperandKindBuilder {
         Self {
             name,
             doc: None,
-            default_member: None,
+            format_field_name: None,
```
abrown: Along the same lines of a previous comment: to eliminate duplication we could use the `new(...).with_?` pattern or, alternately, keep the current constructor methods but do something like:
```
Self {
  name, fields, ..Default::default()
}
```
(we would have to `#[derive(Default)]` to do this).

cranelift-codegen/meta/src/cdsl/operands.rs
abrown: Or `rust_field_name`...

cranelift-codegen/meta/src/cdsl/formats.rs
```diff
@@ -49,15 +49,15 @@ pub(crate) struct InstructionFormat {
 pub(crate) struct FormatStructure {
     pub num_value_operands: usize,
     pub has_value_list: bool,
-    pub imm_field_names: Vec<&'static str>,
+    pub imm_field_names: Vec<(&'static str, &'static str)>,
```
abrown: Might be good to make clear, through a doc string or by type-aliasing, that this tuple is `(Rust field name, Rust type)`

cranelift-codegen/meta/src/cdsl/xform.rs
```diff
@@ -214,7 +214,7 @@ fn rewrite_expr(
             .inst()
             .operands_in
             .iter()
-            .map(|operand| format!("{}: {}", operand.name, operand.kind.name))
+            .map(|operand| format!("{}: {}", operand.name, operand.kind.rust_type))
```
abrown: Yeah, this is way clearer 👍

cranelift-codegen/meta/src/shared/immediates.rs
```diff
@@ -76,45 +76,40 @@ pub(crate) struct Immediates {
...
 
-            pool_constant: Builder::new_imm("poolConstant")
+            pool_constant: Builder::new_imm("constant_handle", "ir::Constant")
                 .doc("A constant stored in the constant pool.")
-                .format_field_name("constant_handle")
-                .rust_type("ir::Constant")
                 .build(),
 
-            offset32: Builder::new_imm("offset32")
+            offset32: Builder::new_imm("offset", "ir::immediates::Offset32")
```
abrown: I see what you mean by this being very explicit but this level of explicitness would have really helped me when I was getting started.

cranelift-codegen/meta/src/cdsl/instructions.rs
```diff
@@ -1296,15 +1296,15 @@ impl Into<InstSpec> for BoundInstruction {
...
+    use crate::cdsl::operands::{OperandKind, OperandKindFields};
     use crate::cdsl::typevar::TypeSetBuilder;
     use crate::shared::types::Int::{I32, I64};
 
     fn field_to_operand(index: usize, field: OperandKindFields) -> Operand {
         // Pretend the index string is &'static.
         let name = Box::leak(index.to_string().into_boxed_str());
         // Format's name / rust_type don't matter here.
-        let kind = OperandKindBuilder::new(name, name, field).build();
+        let kind = OperandKind::new(name, name, field);
```
abrown: I don't understand the comment on 1306: why can we do `name, name` here and it doesn't matter?

cranelift-codegen/meta/src/shared/instructions.rs
bnjbvr: Probably, yeah. Instructions need to know one final index (OpcodeNumber), so it's not trivial as it was for immediates, and might require some working around like we did for formats.

cranelift-codegen/meta/src/cdsl/operands.rs
bnjbvr: If they all have the default format field name, yeah.

cranelift-codegen/meta/src/cdsl/operands.rs
bnjbvr: Yeah, the default approach doesn't work because the name and fields wouldn't be optional, so we'd need a struct just to contain all the default parameters passed to this particular function. I actually tried this and thought it was adding a lot of noise for few value. After this commit, there's only one OperandKind ctor, and the fields are transmitted through helper functions in the appropriate files.

bnjbvr:
Thanks for the review!

PR #1182, opened 2019-10-29
Add x86 SIMD float arithmetic

abrown:
This builds on work in #1181; wait to review until that is merged.

  here: _no issue, continued SIMD implementation_
  first: _see title_

### DIFF
--- a/cranelift-codegen/meta/src/isa/x86/encodings.rs
+++ b/cranelift-codegen/meta/src/isa/x86/encodings.rs
@@ -2069,8 +2069,8 @@ pub(crate) fn define(
         (I16, x86_pminu, &PMINUW[..], Some(use_sse41_simd)),
         (I32, x86_pminu, &PMINUD[..], Some(use_sse41_simd)),
     ] {
-        let inst_ = inst.bind(vector(*ty, sse_vector_size));
-        e.enc_32_64_maybe_isap(inst_, rec_fa.opcodes(opcodes), *isa_predicate);
+        let inst = inst.bind(vector(*ty, sse_vector_size));
+        e.enc_32_64_maybe_isap(inst, rec_fa.opcodes(opcodes), *isa_predicate);
     }
 
     // SIMD float comparisons
@@ -2098,12 +2098,12 @@ pub(crate) fn define(
         (F32, fmax, &MAXPS[..]),
         (F64, fmax, &MAXPD[..]),
     ] {
-        let inst_ = inst.bind(vector(*ty, sse_vector_size));
-        e.enc_both(inst_, rec_fa.opcodes(opcodes));
+        let inst = inst.bind(vector(*ty, sse_vector_size));
+        e.enc_both(inst, rec_fa.opcodes(opcodes));
     }
     for (ty, inst, opcodes) in &[(F32, sqrt, &SQRTPS[..]), (F64, sqrt, &SQRTPD[..])] {
-        let inst_ = inst.bind(vector(*ty, sse_vector_size));
-        e.enc_both(inst_, rec_furm.opcodes(opcodes));
+        let inst = inst.bind(vector(*ty, sse_vector_size));
+        e.enc_both(inst, rec_furm.opcodes(opcodes));
     }
 
     // Reference type instructions
--- a/cranelift-codegen/meta/src/isa/x86/encodings.rs
+++ b/cranelift-codegen/meta/src/isa/x86/encodings.rs
@@ -417,6 +417,8 @@ pub(crate) fn define(
     let fill = shared.by_name("fill");
     let fill_nop = shared.by_name("fill_nop");
     let floor = shared.by_name("floor");
+    let fmax = shared.by_name("fmax");
+    let fmin = shared.by_name("fmin");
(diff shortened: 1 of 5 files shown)


cranelift-codegen/meta/src/isa/x86/legalize.rs
```diff
@@ -496,6 +499,54 @@ pub(crate) fn define(shared: &mut SharedDefinitions, x86_instructions: &Instruct
...
+    let sle = Literal::enumerator_for(&imm.intcc, "sle");
+    let ule = Literal::enumerator_for(&imm.intcc, "ule");
+    for ty in &[I8, I16, I32] {
+        // greater-than
+        let icmp_ = icmp.bind(vector(*ty, sse_vector_size));
+        narrow.legalize(
+            def!(c = icmp_(ugt, a, b)),
+            vec![def!(x = x86_pmaxu(a, b)), def!(c = icmp(eq, a, x))],
+        );
+        let icmp_ = icmp.bind(vector(*ty, sse_vector_size));
```
bjorn3: Why binding this multiple times with the same type?

abrown: `def_rhs!` (used inside `def!`) uses the `Into` trait of `Instruction/BoundInstruction` to get to an `InstSpec`: https://github.com/CraneStation/cranelift/blob/cabbacbe63477fcc7d1f1e36717d84f08f749a98/cranelift-codegen/meta/src/cdsl/ast.rs#L678-L682

I tried `$inst.clone().into()` but unfortunately neither `Instruction` nor `InstSpec` implements `Clone`. I guess I could add those... other suggestions?

abrown:
@sstangl, @bnjbvr: tagging you both so you can be aware of the bitcasts I have to start adding during translation (and the bitcasts in https://github.com/CraneStation/cranelift/pull/1182/commits/a42607a0b1dcbd79fa94c10a036c636ae2556a20 aren't even totally correct, I think we also need to bitcast before adding the value back to the stack because we may be exiting a called function and the types may not match up). I discuss this a bit more in #1192.

cranelift-codegen/meta/src/isa/x86/encodings.rs
```diff
@@ -2081,6 +2083,29 @@ pub(crate) fn define(
...
+        (F32, fmul, &MULPS[..]),
+        (F64, fmul, &MULPD[..]),
+        (F32, fdiv, &DIVPS[..]),
+        (F64, fdiv, &DIVPD[..]),
+        (F32, fmin, &MINPS[..]),
+        (F64, fmin, &MINPD[..]),
+        (F32, fmax, &MAXPS[..]),
+        (F64, fmax, &MAXPD[..]),
+    ] {
+        let inst_ = inst.bind(vector(*ty, sse_vector_size));
```
bnjbvr: light suggestion: here and below, `inst_` could even be named `inst`.

filetests/isa/x86/simd-arithmetic-run.clif
```diff
@@ -153,3 +153,69 @@ ebb0:
...
+}
+; run
+
+function %sqrt_f64x2() -> b1 {
+ebb0:
+    v0 = vconst.f64x2 [0x9.0 0x1.0]
+    v1 = sqrt v0
+    v2 = extractlane v1, 0
+    v3 = f64const 0x3.0
+    v4 = fcmp eq v2, v3
```
bnjbvr: Can you also test the second lane, please?

filetests/isa/x86/simd-arithmetic-run.clif
```diff
@@ -153,3 +153,69 @@ ebb0:
...
 }
 ; run
+
+function %add_sub_f32x4() -> b1 {
+ebb0:
+    v0 = vconst.f32x4 [0x4.2 0.0 0.0 0.0]
+    v1 = vconst.f32x4 [0x1.0 0x1.0 0x1.0 0x1.0]
+
+    v2 = fadd v0, v1
+    v3 = fsub v2, v1
```
bnjbvr: nit: this could be optimized out as a noop in the future, so maybe we'd want to introduce the result vector and directly test the result of the add against it? And introduce a new test for sub?

filetests/isa/x86/simd-arithmetic-run.clif
```diff
@@ -153,3 +153,69 @@ ebb0:
...
+}
+; run
+
+function %mul_div_f32x4() -> b1 {
+ebb0:
+    v0 = vconst.f32x4 [0x4.2 -0x2.1 0x2.0 0.0]
+    v1 = vconst.f32x4 [0x3.4 0x6.7 0x8.9 0xa.b]
+
+    v2 = fmul v0, v1
+    v3 = fdiv v2, v1
```
bnjbvr: same two remarks here

bnjbvr:
> I think we also need to bitcast before adding the value back to the stack because we may be exiting a called function and the types may not match up

Could we insert an optional raw_bitcast to the input of the return, so it matches the function's signature return type instead? That would avoid adding conversions before pushing back to the stack, and could be generalized to other control flow operators, if I understand correctly.

PR #1183, opened 2019-10-29
Tweak comments in regalloc code

bnjbvr:
This tweaks comments (the removed TODO was already implemented), and slightly simplifies one simple function.

@iximeow Can you have a look, please? Feel free to bounce this review to somebody else if you'd prefer.


cranelift-codegen/src/regalloc/affinity.rs
```diff
@@ -92,10 +92,9 @@ impl Affinity {
                 // to change anything.
                 if constraint.kind != ConstraintKind::Stack && !constraint.regclass.has_subclass(rc)
                 {
-                    // If the register classes don't overlap, `intersect` returns `Unassigned`, and
-                    // we just keep our previous affinity.
+                    // If the register classes don't overlap, we just keep our previous affinity;
```
iximeow: Wording nit: how's `If the register classes overlap, try to shrink our preferred register class.` sound? "try" because there's no guarantee the chosen register class is actually smaller - it could be the same one.

cranelift-codegen/src/regalloc/liveness.rs
```diff
@@ -389,7 +389,6 @@ impl Liveness {
         // The liveness computation needs to visit all uses, but the order doesn't matter.
         // TODO: Perhaps this traversal of the function could be combined with a dead code
         // elimination pass if we visit a post-order of the dominator tree?
-        // TODO: Resolve value aliases while we're visiting instructions?
```
iximeow: I appreciate the three year old TODO that was TODONE two years ago :)

bnjbvr:
Thanks for the review! Pushed another revision.

PR #1185, opened 2019-10-30
Use slice::iter

lzutao:
This helps https://github.com/rust-lang/rust/pull/65819


sinkuu:
Nitpicking: s/array::iter/[`slice::iter`](https://doc.rust-lang.org/std/primitive.slice.html#method.iter)/

lzutao:
You're right. Sorry for this confusing.

PR #1188, opened 2019-10-30
Migrate from failure to thiserror

joshtriplett:
The failure crate invents its own traits that don't use
std::error::Error (because failure predates certain features added to
Error); this prevents using ? on an error from failure in a function
using Error. The thiserror crate integrates with the standard Error
trait instead.

### DIFF
--- a/cranelift-codegen/Cargo.toml
+++ b/cranelift-codegen/Cargo.toml
@@ -16,13 +16,12 @@ edition = "2018"
 cranelift-codegen-shared = { path = "./shared", version = "0.46.1" }
 cranelift-entity = { path = "../cranelift-entity", version = "0.46.1" }
 cranelift-bforest = { path = "../cranelift-bforest", version = "0.46.1" }
-failure = { version = "0.1.1", default-features = false, features = ["derive"] }
-failure_derive = { version = "0.1.1", default-features = false }
 hashbrown = { version = "0.6", optional = true }
 target-lexicon = "0.8.1"
 log = { version = "0.4.6", default-features = false }
 serde = { version = "1.0.94", features = ["derive"], optional = true }
 smallvec = { version = "0.6.10" }
+thiserror = "1.0.4"
 # It is a goal of the cranelift-codegen crate to have minimal external dependencies.
 # Please don't add any unless they are essential to the task of creating binary
 # machine code. Integration tests that need external dependencies can be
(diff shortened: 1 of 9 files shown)


bnjbvr:
Thanks for opening a PR! However, this is typically the kind of PRs that would benefit from filling in the PR template, and even opening an issue to discuss alternatives and consider whether this is something we'd like to do.

In particular, there are other crates that also handle errors, and it'd be nice to get some pros/cons for each of them. Also, changing dependencies in `cranelift-codegen` is a bit sensitive because it has an impact for embedders using (cargo) vendoring (I'm thinking about Firefox's code base, which already includes `failure` but not `thiserror`).

Code looks great and I'm very happy to see the manual `From` implementations disappear, though. It would be nice to compare against alternative implementations, though.

joshtriplett:
@bnjbvr I did discuss and coordinate this with @sunfishcode before submitting it, and filed an issue at https://github.com/CraneStation/wasmtime/issues/436 to discuss as well before starting this work.

A bit more on thiserror vs failure:

thiserror doesn't add any dependencies that failure doesn't already have (the standard proc macro dependencies), other than itself. thiserror uses the standard `std::error::Error` type (whereas failure invents its own Fail type), so it smoothly integrates with other users of `Error`. It also carefully avoids exposing any implementation details of itself in the public API; error types derived using thiserror have the same API that you would get if you wrote them by hand.

Thus, I'd suggest that thiserror has the minimum possible impact on users of cranelift. I hope that there's no further churn in error-handling conventions in the community, but on the off chance there is, I think that it'd be entirely transparent to the users of cranelift, and could be done without in any way breaking the cranelift ABI.

joshtriplett:
err-derive's approach to `no_std` support is to not derive Error, and only support From and Display.

thiserror has fewer dependencies than either err-derive or failure_derive.

thiserror does have plans to handle `no_std` through a common `core-error` trait; see https://github.com/dtolnay/anyhow/issues/9 for details.

joshtriplett:
Thanks for the review and the merge, @bnjbvr and @sunfishcode!

A quick note: sorry for not using the pull request template for this PR. I didn't see it, as I don't typically submit pull requests via the github web interface. I use [`git-hub`](https://github.com/sociomantic-tsunami/git-hub/issues/273), and just type `git hub pull new` to create a new pull request. I've filed https://github.com/sociomantic-tsunami/git-hub/issues/273 asking `git-hub` to respect pull request templates.

Thanks again for your patience and review.

bnjbvr:
> I use git-hub

Ah, that explains the lack of context here. Your answers definitely cleared out all my questions, so I appreciate you taking the time providing them.

PR #1190, opened 2019-10-30
[cranelift-native] add isa::Builder without feature detection

acfoltzer:
Just a small addition; doesn't change the existing API but allows making an `isa::Builder` without adding CPUID-detected features. This is useful when the target triples are the same, but the host and target might have different CPU features.

### DIFF
--- a/cranelift-native/src/lib.rs
+++ b/cranelift-native/src/lib.rs
@@ -31,20 +31,31 @@ use target_lexicon::Triple;
 #[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
 use raw_cpuid::CpuId;
 
-/// Return an `isa` builder configured for the current host
-/// machine, or `Err(())` if the host machine is not supported
-/// in the current configuration.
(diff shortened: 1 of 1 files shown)


bjorn3:
I dont think cranelift-native is supposed to be used when compiling for another computer. Quoting the crate level docs:

> Performs autodetection of the host for the purposes of running
> Cranelift to generate code to run on the same machine.

You can just copy the content of your new function to your own crate when you need to use just the host triple, and not host features.

sunfishcode:
If the use case here is to disable use of certain features, would it work to just run the full CPUID detecting version, and then call `.set` on the resulting `Builder` to set undesired features to false?

acfoltzer:
Ah, this is what I get for not paying attention to the comments :slightly_smiling_face: 

You're right that this crate is just focused on autodetection, so this code doesn't belong here. And I thought there was only `.enable`, not `.set`. Ah well, thanks for taking a look :slightly_smiling_face:

Issue #832, opened 2019-07-08
Missing Cranelift IR documentation


abrown:
@bnjbvr, after merging #799 and associated PRs, I no longer see the Cranelift IR documentation I used to see at https://cranelift.readthedocs.io/en/latest/ir.html#vector-operations. I suspect it has something to do with the `meta-python` to `meta` switch but I haven't verified this.

abrown:
When I look at https://github.com/CraneStation/cranelift/blob/master/docs/ir.rst I see an `autoinst` that may no longer be wired up correctly.

bnjbvr:
Right, the meta-python module was used to document instructions. First step would be to fix this by making sure the doc build doesn't depend on the meta-python crate anymore; then it'd be nice to have access to the instructions comments somewhere else (maybe on docs.rs).

bnjbvr:
@sunfishcode is the meta crate published on Crates.io? I'm asking because it seems the documentation that's published [there](https://docs.rs/cranelift-codegen-meta/0.34.0/cranelift_codegen_meta/) seems to be unaware of many data structures / functions. Am I doing something else wrong?

sunfishcode:
It is published on crates.io, but the documentation only includes the things which are visible outside the crate, which right now just `generate` and a few helper routines needed by cranelift-codegen's build.rs. I wonder if there's a setting we can change, or perhaps we could just make more things `pub` in lib.rs.

bjorn3:
```toml
# Cargo.toml
[package.metadata.docs.rs]
rustdoc-args = [ "--document-private-items" ]
```

See https://docs.rs/about for more info about this section.

PR #1191, opened 2019-10-31
Make cranelift-codegen/meta documentation visible

abrown:
- [x] This has been discussed in issue #832.
- [x] A short description of what this does, why it is needed; _see title_.
- [x] This PR contains test cases, if meaningful: _none needed_.

### DIFF
--- a/.azure-pipelines.yml
+++ b/.azure-pipelines.yml
@@ -33,7 +33,9 @@ jobs:
   - checkout: self
     submodules: true
   - template: ci/azure-install-rust.yml
-  - script: cargo doc
+  - script: |
+      cargo doc --all --exclude cranelift-codegen-meta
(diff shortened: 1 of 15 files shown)


abrown:
A note on this change: I tested this locally using `pushd cranelift-codegen/meta; cargo doc --document-private-items; popd`. I realize that by adding 5669f52 docs.rs will be seeing documentation that is not built in `test-all.sh` or the CI, so there is a possibility our builds could pass but docs.rs would fail. One way to fix this would be to split the `cargo doc` line in `test-all.sh` and `.azure-pipelines.yml` to be `cargo doc --exclude cranelift_codegen_meta; cargo doc --package cranelift_codegen_meta --document-private-items`; another way to fix this would be to do some magic (that I don't know how to do) in a `.cargo/config`--thoughts?

Issue #1131, opened 2019-10-10
Add run tests for i128.icmp


bnjbvr:
It'd be nice to add run tests for the following condition codes:

- [ ] eq with non-zero values
- [ ] gt with non-zero values
- [ ] ge with non-zero values

Start from https://github.com/CraneStation/cranelift/blob/master/filetests/isa/x86/icmp-i128.clif and add more tests in there!

ash-tamraz:
I'm interested in tackling this.

How should these tests be run? My initial thought would be to run this with `cargo run compile --target=x86_64 filetests/isa/x86/icmp-i128.clif`, but I'm not sure what we're checking against to determine that the test passed. I get `filetests/isa/x86/icmp-i128.clif: 8: Unknown opcode: 'asdfasdfasgd'` when I add nonsense to the test file, but I know there are more concerns than unknown opcodes.

bnjbvr:
Thanks for looking into this! So you can use clif-util's `test` command: `cargo run test ./filetests/isa/x86/icmp-i128.clif`. Because of the metadata in this file (namely, the `test run` directive at the top), the filetests framework knows it has to parse, compile and run the test. Runtime support is very limited, but it is sufficient for our needs here.

Each test in this file is composed of a function that returns a single boolean value: by convention, the test passes if and only if the function returns true at runtime. A test directive `; run` has to be put just after the function's body, as is done in this example file. So the first test consists of checking that when comparing twice the same i128 value (both low and high 64-bits parts being equal to 0), the comparison shall return true.

Does it make sense?

ash-tamraz:
> Does it make sense?
 
Yes, your explanation is very helpful, thank you very much! I'll see if I can put together these other test cases.

ash-tamraz:
I'm not sure why I'm struggling with this so much. I can't seem to get the non-zero greater than test to work. 

Here's what my file looks like:
```
test run
target x86_64 haswell

function %test_icmp_eq_i128() -> b1 {
ebb0:
    v11 = iconst.i64 0x0
    v12 = iconst.i64 0x0
    v1 = iconcat v11, v12
    v21 = iconst.i64 0x0
    v22 = iconst.i64 0x0
    v2 = iconcat v21, v22
    v10 = icmp.i128 eq v1, v2
    return v10
}

; run

function %test_icmp_imm_eq_i128() -> b1 {
ebb0:
    v11 = iconst.i64 0x0
    v12 = iconst.i64 0x0
[...]

Issue #1117, opened 2019-10-06
Tracking issue for booleans


jyn514:
See vaguely #922, #466, and https://github.com/jyn514/rcc/issues/92. An easy way to do this might be to convert from bool -> int -> bool and back as suggested in https://github.com/CraneStation/cranelift/issues/922#issuecomment-523931097.

- [ ] bnot
Only occurs if you return the value (maybe would also happen for function calls?)
```
function u0:0() -> i32 system_v {
ebb0:
    v0 = iconst.i32 1
    v1 = icmp_imm ne v0, 0
    v2 = bnot v1
    v3 = bint.i32 v2
    return v3
}

codegen error: Compilation error: Verifier errors:
- inst2: v2 is a real GPR value defined by a ghost instruction
```
- [ ] bnz
Note: this only occurs when the values are `i8`, not for `i32`, and it occurs during codegen, not verification.
```
function u0:0() system_v {
ebb0:
    v0 = iconst.i8 0
    v1 = iconst.i8 0
    v2 = icmp ne v0, v1
    brz v2, ebb1
    return

ebb1:
    return
}

codegen error: Compilation error: Verifier errors:
- inst3: Branch must have an encoding
```
- [ ] iconst
```
function u0:0() system_v {
ebb0:   
        v0 = iconst.b1 0
        return
}
-: 3: b1 is not a valid typevar for iconst
```

bjorn3:
`iconst.b1` should be `bconst.b1`.

EliSnow:
`bxor_not` seems to also not work with a boolean value

PR #1193, opened 2019-11-03
Added run tests for i128.icmp

ash-tamraz:
### Summary

This PR is for issue #1131. Writing the greater than test case required I write a legalization for `bnot.b1`, so this PR also affects issue #1117.

Added run tests for `i128.icmp` to the file `filetests/isa/x86/icmp-i128.clif`, for condition codes equal, greater than, and greater than or equal¹ using non-zero values.

Legalized `bnot.b1` as it was required for at least the `greater than` `icmp` condition code.

### Notes

I squashed the commits for both the test case additions and the legalization into one commit. Let me know if it would be better to have those as two commits, I can separate those changes again.

I'm not sure if my legalization and/or tests cases are correct. The documentation on `icmp`s condition codes doesn't mention operand order in the comparisons. I also only added test cases for unsigned values, but can easily add test cases for unsigned values.

1: https://docs.rs/cranelift-codegen/0.46.1/cranelift_codegen/ir/trait.InstBuilder.html#method.icmp

### DIFF
--- a/cranelift-codegen/meta/src/isa/x86/encodings.rs
+++ b/cranelift-codegen/meta/src/isa/x86/encodings.rs
@@ -723,6 +723,7 @@ pub(crate) fn define(
     e.enc_both(band.bind(B1), rec_rr.opcodes(&AND));
     e.enc_both(bor.bind(B1), rec_rr.opcodes(&OR));
     e.enc_both(bxor.bind(B1), rec_rr.opcodes(&XOR));
+    e.enc_both(bnot.bind(B1), rec_ur.opcodes(&NOT).rrr(2));
 
     e.enc_i32_i64(imul, rec_rrx.opcodes(&IMUL));
     e.enc_i32_i64(x86_sdivmodx, rec_div.opcodes(&IDIV).rrr(7));
--- a/filetests/isa/x86/icmp-i128.clif
+++ b/filetests/isa/x86/icmp-i128.clif
@@ -50,3 +50,59 @@ ebb0:
 }
 
 ; run
+
+function %test_icmp_nz_eq_i128() -> b1 {
+ebb0:
+    v11 = iconst.i64 0x1
+    v12 = iconst.i64 0x1
+    v1 = iconcat v11, v12
+    v21 = iconst.i64 0x1
+    v22 = iconst.i64 0x1
+    v2 = iconcat v21, v22
+    v10 = icmp.i128 eq v1, v2
+    return v10
+}
+
+; run
+
+function %test_icmp_nz_gt_i128() -> b1 {
+ebb0:
+    v11 = iconst.i64 0x1
+    v12 = iconst.i64 0x1
+    v1 = iconcat v11, v12
+    v21 = iconst.i64 0x2
+    v22 = iconst.i64 0x2
+    v2 = iconcat v21, v22
+    v10 = icmp.i128 ugt v2, v1
+    return v10
+}
+
+; run
+
+function %test_icmp_nz_ge_not_eq_i128() -> b1 {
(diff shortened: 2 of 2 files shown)


bnjbvr (approved):
That's great, thanks!

For what it's worth, `icmp ugt v2, v1` means `v2 ugt v1`, so this is correct. It looks fine to have a single commit with both changes (encoding + tests), thanks for squashing them.

Are you interested in adding tests for signed values too? We can do it as a separate PR or the same one, as you'd see fit.

cranelift-codegen/meta/src/isa/x86/encodings.rs
```diff
@@ -723,6 +723,7 @@ pub(crate) fn define(
     e.enc_both(band.bind(B1), rec_rr.opcodes(&AND));
     e.enc_both(bor.bind(B1), rec_rr.opcodes(&OR));
     e.enc_both(bxor.bind(B1), rec_rr.opcodes(&XOR));
+    e.enc_both(bnot.bind(B1), rec_ur.opcodes(&NOT).rrr(2));
```
abrown: x86 `NOT` definition: "Performs a bitwise NOT operation (each 1 is set to 0, and each 0 is set to 1) on the destination operand and stores the result in the destination operand location. The destination operand can be a register or a memory location." 

If we `NOT` something like `...000001` we will get `...111110`. @bnjbvr, we're sure Cranelift will prevent subsequent operations from accessing these newly-set bits?

bjorn3: No. `brz` requires the full byte to be zero, not just the lsb: https://github.com/ash-tamraz/cranelift/blob/5ac23d2845c971b5fef5583ad2b05d57affc9878/cranelift-codegen/meta/src/isa/x86/encodings.rs#L1346

abrown: Then we might need a different encoding; how about the following:
 - `BTC %reg, 0`: Selects the bit in a bit string (specified with the first operand, called the bit base) at the bit-position designated by the bit offset operand (second operand), stores the value of the bit in the CF flag, and complements the selected bit in the bit string.

bjorn3: > complements the selected bit in the bit string.

That would mean an extra copie is required when using a bool after a brz/brnz.

abrown: Hm, here's a less destructive version: `BT  + SETNC` (store selected bit in the CF flag + set byte if not carry (CF=0)). But it is two instructions...

bjorn3: `bt` + `jc`/`jnc` for implementing `brz` and `brnz` is enough right?

bnjbvr: Duh, right; that's probably why it wasn't implemented in the first place. We could probably add a check in the verifier that any `b1` value only contains a 0 or 1.

So just thinking a bit about this: I wonder if we could use Bool's algebra and replace `bnot.b1 x` with `bxor_imm.b1 x, 1`. This is a bit harder to implement, though...

bnjbvr (changes requested):
@ash-tamraz Ping; the error @abrown noticed means we can't just merge this as is. Are you interested in following-up on this work, following the suggestion from @abrown last comment?

tschneidereit:
We'll archive this repository in favor of the Wasmtime repository — see Alex's comment about the merge above. I'm closing this PR now in order to have an explanation in a comment before archiving the repo auto-closes all PRs.

PR #1200, opened 2019-11-05
Migrate CI to GitHub Actions

alexcrichton:
This commit migrates from Azure Pipelines to Github Actions for CI for
cranelift. The CI configuration was relatively straightforward, and the
intention here is not to change what's actually being done on CI, just
change where it's being done. The previous CI configuration had build
targets for producing releases, but these weren't actually applicable
for cranelift itself (mostly just copied from wasmtime), so they've been
folded into the main "test everything" matrix which now includes
`--release` mode items.

See cranestation/wasmtime#474 for some more context as well in terms of
benefits of Github Actions vs Azure Pipelines.


alexcrichton:
Note that for CI to kick off on this PR at least one github-actions run needs to already happen in this repository. To do that the best way I've found is to push this commit to a temporary branch of the repository, wait for actions to kick in, then delete the branch.

alexcrichton:
For posterity, [here's a full run for my local repository](https://github.com/alexcrichton/cranelift/commit/2f20c40056e6cc51becbcc02e8b3ddb3b3a803c4/checks?check_suite_id=297103741).

Additionally I should mention that this will need configuration of the `FUZZIT_API_KEY` secret in the repository configuration on GitHub as well.

alexcrichton:
To confirm, was `FUZZIT_API_KEY` set in the repo secrets? I may have misconfigured it here because I didn't confirm that this works, but [the builds on master are failing](https://github.com/CraneStation/cranelift/runs/290333220#step:5:361)

abrown:
Yeah, I pinged @sunfishcode who was tracking down @till to set that up.

tschneidereit:
I set up the key as a repo secret, and CI is now passing.

PR #1204, opened 2019-11-06
Track frame layout changes. (cont.d)

iximeow:
  here.
  first.

Continuation of #679 as Yury asked I open a new PR from my fork.

This PR teaches `cranelift-codegen` to track changes to a call frame layout, on top of which we can generate unwind and debug information (as done in #902)

@bnjbvr you were looking at this most recently in #679, so I figure you might be most interested in reviewing?

### DIFF
--- a/cranelift-codegen/src/ir/framelayout.rs
+++ b/cranelift-codegen/src/ir/framelayout.rs
@@ -51,7 +51,7 @@ pub struct FrameLayout {
     /// Initial frame layout.
     pub initial: FrameLayoutChanges,
 
-    /// Instruction frame layout (changes). The map will not be dense,
+    /// Instruction frame layout (changes). Because the map will not be dense,
     /// a HashMap is used instead of a SecondaryMap.
     pub instructions: HashMap<Inst, FrameLayoutChanges>,
 }
--- a/cranelift-codegen/src/ir/framelayout.rs
+++ b/cranelift-codegen/src/ir/framelayout.rs
@@ -5,7 +5,7 @@ use crate::isa::RegUnit;
 use std::boxed::Box;
 
 #[cfg(not(feature = "std"))]
-use hashmap_core::HashMap;
+use crate::HashMap;
 #[cfg(feature = "std")]
 use std::collections::HashMap;
 
--- a/cranelift-codegen/src/ir/framelayout.rs
+++ b/cranelift-codegen/src/ir/framelayout.rs
@@ -31,6 +31,15 @@ pub enum FrameLayoutChange {
         /// Offset in the frame (offset from CFA).
         cfa_offset: isize,
     },
+    /// The entire frame layout must be preserved somewhere to be restored at a corresponding
+    /// `Restore` change.
+    ///
+    /// This likely maps to the DWARF call frame instruction `.cfa_remember_state`
+    Preserve,
+    /// Restore the entire frame layout from a corresponding prior `Preserve` frame change.
+    ///
+    /// This likely maps to the DWARF call frame instruction `.cfa_restore_state`
+    Restore,
 }
 
 /// Set of frame layout changes.
--- a/cranelift-codegen/src/ir/framelayout.rs
+++ b/cranelift-codegen/src/ir/framelayout.rs
@@ -4,15 +4,12 @@ use crate::ir::entities::Inst;
 use crate::isa::RegUnit;
 use std::boxed::Box;
 
-#[cfg(not(feature = "std"))]
 use crate::HashMap;
-#[cfg(feature = "std")]
-use std::collections::HashMap;
 
 /// Change in the frame layout information.
 #[derive(Clone, Copy, Debug, Eq, PartialEq, Hash)]
 pub enum FrameLayoutChange {
-    /// Base CFA pointer moved to different register/offset.
+    /// Base CallFrameAddress (CFA) pointer moved to different register/offset.
     CallFrameAddressAt {
         /// CFA register.
         reg: RegUnit,
@@ -34,11 +31,11 @@ pub enum FrameLayoutChange {
     /// The entire frame layout must be preserved somewhere to be restored at a corresponding
     /// `Restore` change.
     ///
-    /// This likely maps to the DWARF call frame instruction `.cfa_remember_state`
+    /// This likely maps to the DWARF call frame instruction `.cfa_remember_state`.
     Preserve,
     /// Restore the entire frame layout from a corresponding prior `Preserve` frame change.
     ///
-    /// This likely maps to the DWARF call frame instruction `.cfa_restore_state`
+    /// This likely maps to the DWARF call frame instruction `.cfa_restore_state`.
     Restore,
 }
 
--- a/cranelift-codegen/src/ir/framelayout.rs
+++ b/cranelift-codegen/src/ir/framelayout.rs
@@ -54,7 +54,7 @@ pub struct FrameLayout {
 }
 
 impl FrameLayout {
-    /// Creates instance of FrameLayout.
+    /// Create instance of FrameLayout.
     pub fn new() -> Self {
         FrameLayout {
             initial: vec![].into_boxed_slice(),
--- a/cranelift-codegen/src/ir/framelayout.rs
+++ b/cranelift-codegen/src/ir/framelayout.rs
@@ -1,0 +1,64 @@
+//! Frame layout item changes.
+
+use crate::ir::entities::Inst;
+use crate::isa::RegUnit;
+use std::boxed::Box;
+
+#[cfg(not(feature = "std"))]
+use hashmap_core::HashMap;
+#[cfg(feature = "std")]
+use std::collections::HashMap;
+
+/// Change in the frame layout information.
+#[derive(Clone, Copy, Debug, Eq, PartialEq, Hash)]
+pub enum FrameLayoutChange {
+    /// Base CFA pointer moved to different register/offset.
+    CallFrameAddressAt {
+        /// CFA register.
+        reg: RegUnit,
+        /// CFA offset.
+        offset: isize,
+    },
+    /// Register saved at.
+    RegAt {
+        /// Saved register.
+        reg: RegUnit,
+        /// Offset in the frame (offset from CFA).
+        cfa_offset: isize,
+    },
+    /// Return address saved at.
+    ReturnAddressAt {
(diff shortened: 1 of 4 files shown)


bnjbvr:
Yes, I'll be looking at it since it's still fresh in my mind. Thanks for picking this up!

cranelift-codegen/src/ir/framelayout.rs
```diff
@@ -0,0 +1,70 @@
...
+    /// Initial frame layout.
+    pub initial: FrameLayoutChanges,
+
+    /// Instruction frame layout (changes). Because the map will not be dense,
+    /// a HashMap is used instead of a SecondaryMap.
+    pub instructions: HashMap<Inst, FrameLayoutChanges>,
+}
+
+impl FrameLayout {
+    /// Creates instance of FrameLayout.
```
bnjbvr: nit: to be consistent with the rest of the code, I think we want `Create` and not `Creates` here.

cranelift-codegen/src/isa/x86/abi.rs
```diff
@@ -689,22 +733,87 @@ fn insert_common_prologue(
...
+        Some(CFAState {
+            cf_ptr_reg: RU::rsp as RegUnit,
+            cf_ptr_offset: word_size as isize,
+            current_depth: -(word_size as isize),
+        })
+    } else {
+        None
+    };
+
+    if let Some(ref mut frame_layout) = pos.func.frame_layout {
```
bnjbvr: Could this `if` be combined with the `if` above? That would result in one fewer `expect` below, and less convoluted code.

cranelift-codegen/src/isa/x86/abi.rs
```diff
@@ -680,7 +723,8 @@ fn insert_common_prologue(
     reg_type: ir::types::Type,
     csrs: &RegisterSet,
     isa: &dyn TargetIsa,
-) {
+) -> Option<CFAState> {
+    let word_size = isa.pointer_bytes();
```
bnjbvr: Could this be converted as `isize` directly here? It seems all the users want to use it as such (or `i64` anyways).

cranelift-codegen/src/isa/x86/abi.rs
```diff
@@ -713,7 +822,22 @@ fn insert_common_prologue(
...
+
+        if let Some(ref mut frame_layout) = pos.func.frame_layout {
+            let mut cfa_state = cfa_state
+                .as_mut()
+                .expect("cfa state exists when recording frame layout");
+            cfa_state.current_depth -= word_size;
+            frame_layout.instructions.insert(
+                reg_push_inst,
+                vec![FrameLayoutChange::RegAt {
+                    reg: reg,
```
bnjbvr: nit: `reg,`

cranelift-codegen/src/isa/x86/abi.rs
```diff
@@ -791,12 +917,46 @@ fn insert_common_epilogues(
...
-                insert_common_epilogue(inst, stack_size, pos, reg_type, csrs);
+                // Figure out if we need to insert end-of-function-aware frame layout information.
+                let following_inst = pos
+                    .func
+                    .layout
+                    .next_ebb(ebb)
+                    .and_then(|next_ebb| pos.func.layout.first_inst(next_ebb));
+
+                if following_inst.is_some() {
+                    if let Some(ref mut frame_layout) = pos.func.frame_layout {
```
bnjbvr: Could you invert these two tests, that is put the test on `if let Some()` outwards, to limit the impact of fetching the next instruction when debugging isn't enabled? It would look like:

```
let following_inst = if let Some(ref mut frame_layout) = pos.func.frame_layout {
  let inst = pos
                    .func
                    .layout
                    .next_ebb(ebb)
                    .and_then(|next_ebb| pos.func.layout.first_inst(next_ebb));
  ...
  inst
} else {
  None
};
```

cranelift-codegen/src/isa/x86/abi.rs
```diff
@@ -791,12 +917,46 @@ fn insert_common_epilogues(
...
+                    stack_size,
+                    pos,
+                    reg_type,
+                    csrs,
+                    isa,
+                    cfa_state.clone(),
+                );
+
+                if let Some(following_inst) = following_inst {
+                    if let Some(ref mut frame_layout) = pos.func.frame_layout {
```
bnjbvr: The suggested change above might require to invert the two conditions here too.

cranelift-codegen/src/isa/x86/abi.rs
```diff
@@ -810,26 +970,72 @@ fn insert_common_epilogue(
...
+            offset: cfa_state.cf_ptr_offset,
+        };
+
+        frame_layout
+            .instructions
+            .entry(fp_pop_inst)
+            .and_modify(|insts| {
+                *insts = insts
+                    .into_iter()
+                    .map(|x| *x)
```
bnjbvr: Can this be replaced with a `.clone()`?

cranelift-codegen/src/isa/x86/abi.rs
```diff
@@ -810,26 +970,72 @@ fn insert_common_epilogue(
...
     // preserve the correct order.
     let fp_ret = pos.ins().x86_pop(reg_type);
+    let fp_pop_inst = pos.built_inst();
+
+    if let Some(ref mut cfa_state) = cfa_state.as_mut() {
+        // Account for CFA state in the reverse of `insert_common_prologue`.
+        cfa_state.current_depth += word_size;
+        cfa_state.cf_ptr_offset -= word_size;
+        // And now that we're going to overwrite `rbp`, `rsp` is the only way to get to the call frame.
+        cfa_state.cf_ptr_reg = RU::rsp as RegUnit;
```
bnjbvr: I feel I've asked this before, but: don't we need to "publish" the changes as FrameLayoutChange here? (and below when popping out nonvolatile regs)

cranelift-codegen/src/ir/framelayout.rs
```diff
@@ -0,0 +1,70 @@
...
+    /// This likely maps to the DWARF call frame instruction `.cfa_remember_state`.
+    Preserve,
+    /// Restore the entire frame layout from a corresponding prior `Preserve` frame change.
+    ///
+    /// This likely maps to the DWARF call frame instruction `.cfa_restore_state`.
+    Restore,
+}
+
+/// Set of frame layout changes.
+pub type FrameLayoutChanges = Box<[FrameLayoutChange]>;
```
bnjbvr: I guess we need data in the first place, but I wonder if using a `SmallVec` here wouldn't be as efficient and as simple. This looks a bit like premature optimization, since as far as I recall there's no good justification of it being a problem or causing too many allocations. We can let it as is right now, but using the simplest thing first and then revisiting it if it's a problem would be a better strategy.

cranelift-codegen/src/isa/x86/abi.rs
iximeow: done!

cranelift-codegen/src/isa/x86/abi.rs
iximeow: I moved this into the conditional above, since at that point we know:
* we want to record frame information
* where the epilogue starts
* where the epilogue ends (following instruction, if any)

and instruction numbers are stable indices, so even though we'll add instructions in the epilogue the end stays the same.

cranelift-codegen/src/isa/x86/abi.rs
iximeow: We do need to record the layout change here, [and do just that](https://github.com/bytecodealliance/cranelift/pull/1204/files#diff-08ef6b0c3d44ab300b5158dab40ea830R1024-R1027). It's definitely not immediately obvious since tracks frame layout after `pop rbp` but isn't applied until popping other registers.

I [added some comments](https://github.com/bytecodealliance/cranelift/pull/1204/commits/0895a73e9d55b048c9d0fdcf5c57abeaec4b5b63#diff-08ef6b0c3d44ab300b5158dab40ea830R987-L1002) to try clarifying this for future readers.

cranelift-codegen/src/isa/x86/abi.rs
```diff
@@ -544,6 +545,32 @@ fn baldrdash_prologue_epilogue(func: &mut ir::Function, isa: &dyn TargetIsa) ->
     Ok(())
 }
 
+/// CFAState is `cranelift`'s model of the call frame layout at any particular point in a function.
```
bjorn3:
```suggestion
/// CFAState is cranelift's model of the call frame layout at any particular point in a function.
```

iximeow:
> Looking great, mostly nits here, **so approving**. Thanks for continuing this!

Emphasis mine, on the part I missed when I saw the notification last week :) Another small round of tweaks in line with comments.

@bnjbvr if https://github.com/bytecodealliance/cranelift/pull/1204#discussion_r347505528 and the comments referenced there read well to you I'll squash, merge, and get back to https://github.com/bytecodealliance/cranelift/pull/902 :)

cranelift-codegen/src/isa/x86/abi.rs
```diff
@@ -544,6 +545,32 @@ fn baldrdash_prologue_epilogue(func: &mut ir::Function, isa: &dyn TargetIsa) ->
...
+    /// `rbp`, but at function entry and exit may be `rsp` while the call frame is being
+    /// established.
+    cf_ptr_reg: RegUnit,
+    /// Given that `cf_ptr_reg` is a register containing a pointer to some memory, `cf_ptr_offset`
+    /// is the offset from that pointer to the address of the start of this function's call frame.
+    ///
+    /// For a concrete x86_64 example, we will start this at 8 - the call frame begins immediately
+    /// before the return address. This will typically then be set to 16, after pushing `rbp` to
+    /// preserve the parent call frame. It is very unlikely the offset should be anything other
+    /// than one or two `usize`.
```
bjorn3:
```suggestion
    /// than one or two pointer widths.
```

Current statement is ambiguous for cross-compilation, as `usize` in code refers to an int with the native pointer size.

bjorn3:
@iximeow Tip: If you switch to the "files changed" tab, you can queue multiple suggestions to be committed together in a single commit.

iximeow:
Ah that would have been good, I was wondering about that.. thanks.

PR #1207, opened 2019-11-06
Don't align the stack frame for leaf functions not using the stack

bnjbvr:
Not discussed in any issue: this is a very small optimization to not force stack frame alignment for leaf functions that don't use the stack. This spurious stack alignment would cause stack allocations that weren't needed, adding some instruction traffic.

It's a regular 1% speedup in most of embenchen benchmarks (in Spidermonkey), with `primes` showing a 6% speedup (I expect micro-architectural issues there). Leaf functions are frequent, I guess.

Requesting review from @fitzgen since he's looked at this code recently, and this PR also contains a commit that fixes a small issue I've overlooked during review of the multi-value PR.

### DIFF
--- a/cranelift-codegen/src/ir/dfg.rs
+++ b/cranelift-codegen/src/ir/dfg.rs
@@ -104,6 +104,7 @@ impl DataFlowGraph {
         self.value_lists.clear();
         self.values.clear();
         self.signatures.clear();
+        self.old_signatures.clear();
         self.ext_funcs.clear();
         self.values_labels = None;
         self.constants.clear();
--- a/cranelift-codegen/src/ir/function.rs
+++ b/cranelift-codegen/src/ir/function.rs
@@ -279,6 +279,14 @@ impl Function {
 
         Ok(())
     }
+
+    /// Returns true if the function is function that doesn't call any other functions. This is not
+    /// to be confused with a "leaf function" in Windows terminology.
+    pub fn is_leaf(&self) -> bool {
+        // Conservative result: if there's at least one function signature referenced in this
+        // function, assume it may call.
+        !self.dfg.signatures.is_empty()
+    }
 }
 
 /// Additional annotations for function display.
--- a/cranelift-codegen/src/stack_layout.rs
+++ b/cranelift-codegen/src/stack_layout.rs
@@ -7,15 +7,20 @@ use core::cmp::{max, min};
 
 /// Compute the stack frame layout.
 ///
-/// Determine the total size of this stack frame and assign offsets to all `Spill` and
-/// `Explicit` stack slots.
+/// Determine the total size of this stack frame and assign offsets to all `Spill` and `Explicit`
+/// stack slots.
 ///
-/// The total frame size will be a multiple of `alignment` which must be a power of two.
+/// The total frame size will be a multiple of `alignment` which must be a power of two, unless the
+/// function doesn't perform any call.
 ///
 /// Returns the total stack frame size which is also saved in `frame.frame_size`.
 ///
 /// If the stack frame is too big, returns an `ImplLimitExceeded` error.
-pub fn layout_stack(frame: &mut StackSlots, alignment: StackSize) -> CodegenResult<StackSize> {
+pub fn layout_stack(
+    frame: &mut StackSlots,
+    is_leaf: bool,
+    alignment: StackSize,
+) -> CodegenResult<StackSize> {
     // Each object and the whole stack frame must fit in 2 GB such that any relative offset within
     // the frame fits in a `StackOffset`.
     let max_size = StackOffset::max_value() as StackSize;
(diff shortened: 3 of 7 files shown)


filetests/isa/x86/windows_fastcall_x64_unwind.clif
```diff
@@ -160,7 +160,7 @@ ebb0(v0: i64, v1: i64):
 ; nextln:         UnwindCode {
 ; nextln:             offset: 19,
 ; nextln:             op: SmallStackAlloc,
-; nextln:             info: 4,
+; nextln:             info: 3,
```
bnjbvr: Just so you don't have to read all the code related to UnwindCode as I did: this is the multiple of 8-bytes pushed by the stack allocation; the function in the test is a leaf function, so it matches the criteria to not align the stack frame, thus reducing stack frame allocation.

cranelift-codegen/src/stack_layout.rs
```diff
@@ -103,7 +114,10 @@ pub fn layout_stack(frame: &mut StackSlots, alignment: StackSize) -> CodegenResu
     offset = offset
         .checked_sub(outgoing_max)
         .ok_or(CodegenError::ImplLimitExceeded)?;
-    offset &= -(alignment as StackOffset);
+
+    if must_align {
```
bjorn3:
```suggestion
    if frame.values().next().is_some() {
```

bnjbvr: This will align the stack if there are incoming arguments only, which seems unnecessary.

bjorn3: Didn't notice that `IncomingArg` doesn't have `must_align = true`.

cranelift-codegen/src/ir/dfg.rs
```diff
@@ -104,6 +104,7 @@ impl DataFlowGraph {
         self.value_lists.clear();
         self.values.clear();
         self.signatures.clear();
+        self.old_signatures.clear();
```
fitzgen: Woops, nice catch! thanks :)

cranelift-codegen/src/ir/function.rs
```diff
@@ -279,6 +279,13 @@ impl Function {
 
         Ok(())
     }
+
+    /// Returns true if the function may perform calls to other functions.
+    pub fn may_call(&self) -> bool {
+        // Conservative result: if there's at least one function signature referenced in this
+        // function, assume it will call.
+        !self.dfg.signatures.is_empty()
```
fitzgen: Does it make sense walk the blocks and instructions in debug mode and assert that if `self.dfg.signatures` is empty, then there are no calls? That is, assert that this conservative property holds in practice.

cranelift-codegen/src/stack_layout.rs
```diff
@@ -7,15 +7,20 @@ use core::cmp::{max, min};
...
+/// The total frame size will be a multiple of `alignment` which must be a power of two, unless the
+/// function doesn't perform any call.
 ///
 /// Returns the total stack frame size which is also saved in `frame.frame_size`.
 ///
 /// If the stack frame is too big, returns an `ImplLimitExceeded` error.
-pub fn layout_stack(frame: &mut StackSlots, alignment: StackSize) -> CodegenResult<StackSize> {
+pub fn layout_stack(
+    frame: &mut StackSlots,
+    may_call: bool,
```
fitzgen: This is a total nitpick, and maybe you disagree, but I think `is_leaf` would be a better name than `may_call` for both this parameter and the method.

filetests/isa/x86/windows_fastcall_x64_unwind.clif
fitzgen: :+1:

cranelift-codegen/src/stack_layout.rs
peterhuene: Copying my comment from Slack for posterity:

> Note that "leaf" on Windows x64 ABI has additional restrictions, though, beyond just not calling.
>
> Primarily it has to be able to unwind via a ret instruction alone.
>
> I'm good with `is_leaf`, but in the future when we have code generation tailored for Windows (primarily to exclude prologue/epilogue generation and unwind information for leafs), the nomenclature might get a little muddled.

This comment really applies to `Function.may_call` if we rename it to `Function.is_leaf`.  I think `is_leaf` here in the context of `layout_stack` makes total sense.

If we make a corresponding change to have a `Function.is_leaf`, then, as Dan suggested, perhaps it would be a good idea for a comment that makes it clear that the determination is solely based off of whether or not the function makes a call so that it isn't confused with "leaf function" in the Windows terminology.

cranelift-codegen/src/ir/function.rs
bjorn3: With `self.dfg.signatures` empty, legalizing/encoding any calls would panic, as clif doesnt know what the signature of the function is, which is necessary to know which value should be in which register.

bnjbvr: Agreed with @bjorn3; the only opcodes that return true for `is_call` are call and call_indirect, which both take in a mandatory signature reference.

cranelift-codegen/src/stack_layout.rs
bnjbvr: Good idea, renamed the function and added a comment.

bnjbvr:
> Since this touches the same spots, would it also be a good idea to skip shadow space allocation for the fastcall convention when the function doesn't call as part of these changes? This would be one of the improvements I've called out for Windows code generation in #1201.

Good idea, but I'd prefer to do it in a separate PR.

PR #1209, opened 2019-11-07
Use zeroinit API for faerie and object

jyn514:
  here.
This has not been discussed yet. Discussion of why this feature was implemented is in https://github.com/m4b/faerie/issues/97.

  first.
This uses the new `zero_init` api for Faerie, which uses `.bss` sections to reduce the amount of space binaries take up in memory (while being generated) and on disk (when stored).

This PR does not contain new test cases. All existing test cases pass.

I am not sure who should review this.

### DIFF
--- a/cranelift-faerie/src/backend.rs
+++ b/cranelift-faerie/src/backend.rs
@@ -217,20 +217,6 @@ impl Backend for FaerieBackend {
             ref data_relocs,
         } = data_ctx.description();
 
-        let size = init.size();
-        let mut bytes = Vec::with_capacity(size);
-        match *init {
-            Init::Uninitialized => {
-                panic!("data is not initialized yet");
-            }
-            Init::Zeros { .. } => {
-                bytes.resize(size, 0);
-            }
-            Init::Bytes { ref contents } => {
-                bytes.extend_from_slice(contents);
-            }
-        }
-
         for &(offset, id) in function_relocs {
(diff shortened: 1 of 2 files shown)


bjorn3:
This could be done for `object` too: https://docs.rs/object/0.15.0/object/write/struct.Object.html#method.add_symbol_bss

jyn514:
Hmm, what is the role of `object`? It looks very similar to `faerie`

bjorn3:
`object` is a bit more low level and much more flexible. For example it supports TLS and a symbol scope of [linkage](https://docs.rs/object/0.15.0/object/enum.SymbolScope.html#variant.Linkage).

philipc:
Feel free to update `cranelift-object` if you want to, but I was going to handle that after #1174 is merged since I think it will conflict with that, and it needs to handle `.tbss` too.

jyn514:
@philipc I'm happy to defer to you, no sense implementing it twice if you're going to have to redo it.

jyn514:
@philipc since it's been a few months, I think I'm going to add .bss for object and you can add .tbss when #1174 is merged.

jyn514:
@bnjbvr will you have time to review this sometime this week or next? (tagging since I know this is a few months old)

bnjbvr:
@fitzgen, can you take this review, as this is your domain of expertise, please?

PR #1212, opened 2019-11-07
Use a struct interface for creating and reading encoding bits on x86.

sstangl:
Related to #1156 . The x86 `Encoding.bits` field is an open-coded `u16`. Probably because it was created before we had `cranelift-codegen-shared`.

This adds a `cranelift_codegen_shared::isa::x86::EncodingBits` and friends that understand how to encode and decode that `u16`. It can then be inspected by using named functions, instead of by open-coding bit extraction. We're going to have more dynamic dispatch based on those bits, and I didn't want to continue open-coding it.

I'm trying to land smaller patches now to avoid having one huge, interleaved thing at the end, because I suspect this way will be significantly easier to review.

### DIFF
--- a/cranelift-codegen/shared/src/isa/x86/encoding_bits.rs
+++ b/cranelift-codegen/shared/src/isa/x86/encoding_bits.rs
@@ -1,0 +1,236 @@
+//! Provides a named interface to the `u16` Encoding bits.
+
+use packed_struct::prelude::*;
+
+/// Named interface to the `u16` Encoding bits, representing an opcode.
+///
+/// Cranelift requires each recipe to have a single encoding size in bytes.
+/// X86 opcodes are variable length, so we use separate recipes for different
+/// styles of opcodes and prefixes. The opcode format is indicated by the
+/// recipe name prefix.
+///
+/// VEX/XOP and EVEX prefixes are not yet supported.
+/// Encodings using any of these prefixes are represented by separate recipes.
+///
+/// The encoding bits are:
+///
+/// 0-7:   The opcode byte <op>.
+/// 8-9:   pp, mandatory prefix:
+///        00: none (Op*)
+///        01: 66   (Mp*)
+///        10: F3   (Mp*)
+///        11: F2   (Mp*)
+/// 10-11: mm, opcode map:
+///        00: <op>        (Op1/Mp1)
+///        01: 0F <op>     (Op2/Mp2)
+///        10: 0F 38 <op>  (Op3/Mp3)
+///        11: 0F 3A <op>  (Op3/Mp3)
+/// 12-14  rrr, opcode bits for the ModR/M byte for certain opcodes.
+/// 15:    REX.W bit (or VEX.W/E)
+#[derive(Copy, Clone, PartialEq, PackedStruct)]
+#[packed_struct(size_bytes = "2", bit_numbering = "lsb0")]
+pub struct EncodingBits {
+    /// Instruction opcode byte, without the prefix.
+    #[packed_field(bits = "0:7")]
+    pub opcode_byte: u8,
+
+    /// Prefix kind for the instruction, as an enum.
+    #[packed_field(bits = "8:11", ty = "enum")]
+    pub prefix: OpcodePrefix,
+
+    /// Bits for the ModR/M byte for certain opcodes.
+    #[packed_field(bits = "12:14")]
+    pub rrr: Integer<u8, packed_bits::Bits3>,
+
+    /// REX.W bit (or VEX.W/E).
+    #[packed_field(bits = "15")]
+    pub rex_w: Integer<u8, packed_bits::Bits1>,
+}
+
+impl From<u16> for EncodingBits {
+    fn from(bits: u16) -> EncodingBits {
+        let bytes: [u8; 2] = [((bits >> 8) & 0xff) as u8, (bits & 0xff) as u8];
+        EncodingBits::unpack(&bytes).expect("failed creating EncodingBits")
+    }
+}
+
+impl EncodingBits {
+    /// Constructs a new EncodingBits from parts.
+    pub fn new(op_bytes: &[u8], rrr: u16, rex_w: u16) -> Self {
+        EncodingBits {
+            opcode_byte: op_bytes[op_bytes.len() - 1],
+            prefix: OpcodePrefix::from_opcode(op_bytes),
+            rrr: (rrr as u8).into(),
+            rex_w: (rex_w as u8).into(),
+        }
+    }
+
+    /// Returns the raw bits.
+    #[inline]
+    pub fn bits(self) -> u16 {
+        let bytes: [u8; 2] = self.pack();
+        ((bytes[0] as u16) << 8) | (bytes[1] as u16)
+    }
+
+    /// Extracts the PP bits of the OpcodePrefix.
+    #[inline]
+    pub fn pp(self) -> u8 {
+        self.prefix.to_primitive() & 0x3
+    }
+
+    /// Extracts the MM bits of the OpcodePrefix.
+    #[inline]
+    pub fn mm(self) -> u8 {
+        (self.prefix.to_primitive() >> 2) & 0x3
+    }
+}
+
+/// Opcode prefix representation.
+///
+/// The prefix type occupies four of the EncodingBits.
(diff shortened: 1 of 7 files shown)


cranelift-codegen/shared/src/isa/x86/encoding_bits.rs
```diff
@@ -0,0 +1,193 @@
...
+    /// Extracts the MM bits of the OpcodePrefix.
+    #[inline]
+    pub fn mm(self) -> u8 {
+        ((self.0 >> 10) & 0x3) as u8
+    }
+
+    /// Extracts the RRR bits.
+    #[inline]
+    pub fn rrr(self) -> u8 {
+        ((self.0 >> 12) & 0x3) as u8
```
abrown: If this is a 3-bit field, we need to `& 0x7`, right?

cranelift-codegen/shared/src/isa/x86/encoding_bits.rs
```diff
@@ -0,0 +1,193 @@
...
+        let prefix: u16 = OpcodePrefix::from_opcode(op_bytes) as u16;
+        let opcode_byte: u16 = u16::from(op_bytes[op_bytes.len() - 1]);
+
+        debug_assert!(opcode_byte <= 0b1111_1111);
+        debug_assert!(prefix <= 0b1111);
+        debug_assert!(rrr <= 0b111);
+        debug_assert!(rex_w <= 0b1);
+
+        EncodingBits {
+            0: opcode_byte | (prefix << 8) | (rrr << 12) | (rex_w << 15),
```
abrown: Where's `mm` as described in the docs? Shouldn't we have a `| (mm << 10)` here somewhere?

cranelift-codegen/shared/src/isa/x86/encoding_bits.rs
```diff
@@ -0,0 +1,193 @@
...
+impl EncodingBits {
+    /// Constructs a new EncodingBits from parts.
+    pub fn new(op_bytes: &[u8], rrr: u16, rex_w: u16) -> Self {
+        let prefix: u16 = OpcodePrefix::from_opcode(op_bytes) as u16;
+        let opcode_byte: u16 = u16::from(op_bytes[op_bytes.len() - 1]);
+
+        debug_assert!(opcode_byte <= 0b1111_1111);
+        debug_assert!(prefix <= 0b1111);
+        debug_assert!(rrr <= 0b111);
+        debug_assert!(rex_w <= 0b1);
```
abrown: I wonder if these shouldn't be just regular checks that panic even in optimized builds...

cranelift-codegen/shared/src/isa/x86/encoding_bits.rs
```diff
@@ -0,0 +1,193 @@
...
+///        10: 0F 38 <op>  (Op3/Mp3)
+///        11: 0F 3A <op>  (Op3/Mp3)
+/// 12-14  rrr, opcode bits for the ModR/M byte for certain opcodes.
+/// 15:    REX.W bit (or VEX.W/E)
+#[derive(Copy, Clone, Eq, PartialEq)]
+pub struct EncodingBits(u16);
+
+impl From<u16> for EncodingBits {
+    fn from(bits: u16) -> EncodingBits {
+        EncodingBits { 0: bits }
```
abrown: style question: this is equivalent to `EncodingBits(bits)`?

cranelift-codegen/src/isa/x86/binemit.rs
```diff
@@ -101,63 +103,61 @@ fn put_rexop2<CS: CodeSink + ?Sized>(bits: u16, rex: u8, sink: &mut CS) {
...
     sink.put1(bits as u8);
 }
 
 // Emit single-byte opcode with mandatory prefix and REX.
 fn put_rexmp1<CS: CodeSink + ?Sized>(bits: u16, rex: u8, sink: &mut CS) {
+    let enc = EncodingBits::from(bits);
     debug_assert_eq!(bits & 0x0c00, 0, "Invalid encoding bits for Mp1*");
-    let pp = (bits >> 8) & 3;
-    sink.put1(PREFIX[(pp - 1) as usize]);
+    sink.put1(PREFIX[(enc.pp() - 1) as usize]);
```
abrown: Seems like the `let enc = ...` should come after the check to see if it the bits are valid; I think even cleaner would be to somehow use the new `OpcodePrefix` code to do the check instead of a bitwise `&`.

cranelift-codegen/shared/src/isa/x86/encoding_bits.rs
```diff
@@ -0,0 +1,193 @@
...
+            Op1 => "Op1",
+            Op2_0f => "Op2",
+            Op3_0f_38 | Op3_0f_3a => "Op3",
+            Mp1_66 | Mp1_f3 | Mp1_f2 => "Mp1",
+            Mp2_66_0f | Mp2_f3_0f | Mp2_f2_0f => "Mp2",
+            Mp3_66_0f_38 | Mp3_f3_0f_38 | Mp3_f2_0f_38 => "Mp3",
+            Mp3_66_0f_3a | Mp3_f3_0f_3a | Mp3_f2_0f_3a => "Mp3",
+        }
+    }
+}
```
abrown: I think the approach of factoring out `EncodingBits` and `OpcodePrefix` is great for code clarity! I wish there were some way to avoid the multiple back-and-forth conversions for `OpcodePrefix` and the manual bit-fiddling in `EncodingBits`--I mean, it's better that it's all in one place but I can't be sure that I didn't miss a typo as I look through it. What do you think of https://crates.io/crates/bitflags or https://crates.io/crates/bitfields? Or could we write some tests to make it more clear that we have covered all the cases?

sstangl: Yeah, I looked at some external crates built for this purpose, but wasn't sure if we wanted to add an external dependency just for this, since that code is likely to be very static.

cranelift-codegen/shared/src/isa/x86/encoding_bits.rs
sstangl: Quite unhelpfully, `pp` and `mm` are two bits each of the `OpcodePrefix`. I have no idea what they stand for, so I couldn't document them myself. As far as I can tell, the `OpcodePrefix` bits are structured as they are to enable these `pp` and `mm` lookups to index into tables, and they have no other meaning.

Probably we could get rid of them entirely and just teach the `OpcodePrefix` itself how to index into these tables. Then at least that detail would be part of the encoding structure, and not known generally.

cranelift-codegen/shared/src/isa/x86/encoding_bits.rs
sstangl: Oops :) I think I was looking at the `& 3` code in the original implementation and passed out for a second there.

cranelift-codegen/shared/src/isa/x86/encoding_bits.rs
abrown: Maybe a couple of doc tests then?

cranelift-codegen/shared/src/isa/x86/encoding_bits.rs
abrown: Makes sense to me

cranelift-codegen/src/isa/x86/binemit.rs
sstangl: Unfortunately it's not just checking the OpcodePrefix (which is `0xf00`) -- it also checks the `rrr` bits, as in `0x8c00` for `Mp2`. It's hard to give a name to that. We'll be rewriting all these methods eventually as part of recipe refactoring, so it should be OK to put that off.

sstangl:
I uploaded a new version that uses the `packed_struct` crate. I believe the codegen will be worse, although I haven't checked, and it currently fails tests during binemit for reasons that I haven't been able to debug yet. However, it does avoid some redundancy, although u16s have to be manually packed/unpacked to/from u8 byte arrays.

Just submitting this patch to see whether this is preferred. It should at least be easier to review.

sstangl:
Fixed tests. Also added more tests to catch the error I made (flipped two bytes).

PR #1213, opened 2019-11-07
Try and assign directly to return registers; backtrack to use struct-return param

fitzgen:
> This has been discussed in issue #..., or if not, please tell us why here.

https://github.com/CraneStation/wasmtime/pull/467

> A short description of what this does, why it is needed; if the
> description becomes long, the matter should probably be discussed in an issue
> first.

Rather than trying to count number of return registers that would be used by a given set of return values, optimistically assign the return values to registers. If we later find that we can't fit them all in registers, then backtrack and introduce the use of a struct-return pointer parameter.

> This PR contains test cases, if meaningful.

Tests in wasmtime. If we compiled the wasm spec testsuite in cranelift, that would have caught the issue as well.

### DIFF
--- a/cranelift-codegen/src/isa/x86/abi.rs
+++ b/cranelift-codegen/src/isa/x86/abi.rs
@@ -224,7 +224,18 @@ pub fn legalize_signature(
         shared_flags,
         isa_flags,
     );
-    let mut rets2 = rets.clone();
+
+    let sig_is_multi_return = sig.is_multi_return();
+
+    // If this is a multi-value return and we don't have enough available return
+    // registers to fit all of the return values, we need to backtrack and start
+    // assigning locations all over again with a different strategy. In order to
+    // do that, we need a copy of the original assigner for the returns.
+    let backup_rets_for_struct_return = if sig_is_multi_return {
+        Some(rets.clone())
+    } else {
+        None
+    };
 
     if let Some(new_returns) = legalize_args(&sig.returns, &mut rets) {
         if sig.is_multi_return()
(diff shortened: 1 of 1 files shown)


cranelift-codegen/src/isa/x86/abi.rs
```diff
@@ -331,71 +224,72 @@ pub fn legalize_signature(
...
-            value_type: args.pointer_type,
-            purpose: ArgumentPurpose::StructReturn,
-            extension: ArgumentExtension::None,
-            location: ArgumentLoc::Unassigned,
-        };
-        match args.assign(&ret_ptr_param) {
-            ArgAction::Assign(ArgumentLoc::Reg(reg)) => {
-                ret_ptr_param.location = ArgumentLoc::Reg(reg);
-                sig.to_mut().params.push(ret_ptr_param);
+    if let Some(new_returns) = legalize_args(&sig.returns, &mut rets) {
```
bnjbvr: (light suggestion) instead of having `rets2` (which is not very descriptive a name), could we pass `&mut rets.clone()` here, and use `rets` where we used `rets2` below?

Otherwise, could we rename `rets2` to `rets_backup` and explain in a small comment why it's needed in the first place

cranelift-codegen/src/isa/x86/abi.rs
```diff
@@ -331,71 +224,72 @@ pub fn legalize_signature(
         shared_flags,
         isa_flags,
     );
+    let mut rets2 = rets.clone();
```
bnjbvr: It's unfortunate the clone happens even when one doesn't use multi-value returns. To make this slightly more zero-cost, could we clone only when `sig.is_multi_return` instead?

cranelift-codegen/src/isa/x86/abi.rs
bnjbvr: (or maybe `original_rets`, so it doesn't sound too weird with respect to the other comment.)

fitzgen:
Thanks for review! Will fix up with your suggestions and then merge once CI is green again :+1:

PR #1214, opened 2019-11-08
Fix build errors in x86 unwind info when building no_std.

### LINKED ISSUE #1203: Support no_std for Windows unwind information generation.
#### What is the feature or code improvement you would like to do in Cranelift?

With #1155 merged, the unwind information generation depends on `byteorder` and uses a trait that is explicitly gated by the `std` feature (`WriteBytesExt`).

The code in `isa/x86/unwind.rs` should be changed to not depend on this trait so that it the `byteorder` crate can be used with `no_std`.

#### What is the value of adding this in Cranelift?

Removes one more roadblock for `no_std` support.

#### Do you have an implementation plan, and/or ideas for data structures or algorithms to use?

Copy the tiny implementations of the convenience `write_xx` methods, as needed.

#### Have you considered alternative implementations? If so, how are they better or worse than your proposal

No alternatives considered.  To support `no_std`, we'll need to have the minimum implementation needed that doesn't depend on stdlib types.

peterhuene:
This PR fixes the build errors in the unwind info implementation for
the x86 ABI by changing `byteorder` to build `no_std`.

This copies two simple functions from the `WriteBytesExt` trait so that
we can easily write to a `Vec<u8>` with a particular endianness.

Fixes #1203.

cc @sunfishcode for review.

### DIFF
--- a/cranelift-codegen/Cargo.toml
+++ b/cranelift-codegen/Cargo.toml
@@ -22,7 +22,7 @@ log = { version = "0.4.6", default-features = false }
 serde = { version = "1.0.94", features = ["derive"], optional = true }
 smallvec = { version = "1.0.0" }
 thiserror = "1.0.4"
-byteorder = { version = "1.3.2" }
+byteorder = { version = "1.3.2", default-features = false }
 # It is a goal of the cranelift-codegen crate to have minimal external dependencies.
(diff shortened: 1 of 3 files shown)


peterhuene:
@sunfishcode no rush, but feel free to merge when you're able.

PR #1215, opened 2019-11-08
Hoist the stack alignment and Windows64 fastcall shadow stack space constants

bnjbvr:
Small refactoring.

I'll repeat a chat comment to @peterhuene about the possibility of not adding the shadow stack space on Win64 fastcall convention if the current function is a leaf function:

> When we're generating the function's prologue/epilogue, it's for the current function, not for a function we're calling as part of generated code. So I assume it would require synchronizing with all the callers of this function; we control direct callers (cranelift -> cranelift calls), but not external callers into cranelift code. For one thing, it would be an issue for Spidermonkey which always generates the shadow stack space when calling into a wasm callee from C++ / JIT code. In general for wasm, knowing the callee is a leaf function would imply having seen the body of the callee, thus disallowing streaming parallel compilation.

So I don't think we should do this in general, but maybe I'm overseeing something.

### DIFF
--- a/cranelift-codegen/src/isa/x86/abi.rs
+++ b/cranelift-codegen/src/isa/x86/abi.rs
@@ -34,6 +34,29 @@ static ARG_GPRS_WIN_FASTCALL_X64: [RU; 4] = [RU::rcx, RU::rdx, RU::r8, RU::r9];
 /// Return value registers for x86-64, when using windows fastcall
 static RET_GPRS_WIN_FASTCALL_X64: [RU; 1] = [RU::rax];
 
+/// The win64 fastcall ABI uses some shadow stack space, allocated by the caller, that can be used
+/// by the callee for temporary values.
+///
+/// [1] "Space is allocated on the call stack as a shadow store for callees to save" This shadow
+/// store contains the parameters which are passed through registers (ARG_GPRS) and is eventually
+/// used by the callee to save & restore the values of the arguments.
+///
+/// [2] https://blogs.msdn.microsoft.com/oldnewthing/20110302-00/?p=11333 "Although the x64 calling
+/// convention reserves spill space for parameters, you don’t have to use them as such"
+const WIN_SHADOW_STACK_SPACE: i32 = 32;
+
+/// Stack alignment requirement for functions.
+///
+/// 16 bytes is the perfect stack alignment, because:
+///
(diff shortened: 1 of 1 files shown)


peterhuene:
Also copying some of the Slack discussion here:

> In reality there should be two things considered for x64 calling convention when laying out the stack: every function can assume it has 32 bytes of shadow stack space (starting at rsp+8) allocated by the caller to use as it sees fit (generally for homing rcx, rdx, r8, and r9 in “debug” builds, but can be used even by functions that don’t use those registers for argument passing) and the frame being currently allocated must allocate 32 bytes for its callees.  the latter is only what I mean by what we can skip if we know the function doesn’t call

> For example, here's the disassembly for cranelift x64 code generation of an empty function.  These 32 bytes is what I’m talking about.

```asm
push        rbp  
mov         rbp,rsp  
sub         rsp,20h  
add         rsp,20h  
pop         rbp  
ret
```

> Granted, this is a “leaf” function in the Windows-sense, so it doesn’t need a prologue or epilogue at all, but even if it had an explicit stack slot, it wouldn’t need that additional 32 bytes because it has no callees to use it.

bnjbvr:
@peterhuene When compiling a single function, we have a signature for the current function and function being called. I'm now pretty sure the layout_stack function handles them the same way, causing the spurious shadow stack space allocation when laying out the stack for the case of the current function. There might be additional hidden bugs because of this, I'll investigate this.

bnjbvr:
Not merging yet: I'd like us to reset up the actions again so we have CI again, and this can be more thoroughly tested.

bnjbvr:
CI error seems to be when compiling a dependency, in `cargo fuzz` mode:

```
error[E0423]: expected function, tuple struct or tuple variant, found struct `ast::Name`
   --> /home/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/syntex_syntax-0.58.1/src/symbol.rs:146:27
    |
146 |                       name: ast::Name($index),
    |                             ^^^^^^^^^
...
165 | / declare_keywords! {
166 | |     // Invalid identifier
167 | |     (0,  Invalid,        "")
168 | |
...   |
231 | |     (56, CrateRoot, "{{root}}")
232 | | }
    | |_- in this macro invocation
```

Probably a dependency we'd need to upgrade, but very likely to be unrelated to this PR.

PR #1218, opened 2019-11-11
Add additional floating-point operations

abrown:
This is a follow-on PR to #1182 (wait for #1182 to merge) that adds SIMD implementations for `fabs` and `fneg`. Both of these are legalizations and to legalize `fabs` I needed the ability to do a vector shift with an immediate (i.e. `ushr_imm`) in order to create the proper mask for zeroing the FP sign bit. Because of this, I also included encodings for the x86 SIMD shift-with-immediate operations.

### DIFF
--- a/cranelift-codegen/meta/src/isa/x86/legalize.rs
+++ b/cranelift-codegen/meta/src/isa/x86/legalize.rs
@@ -36,6 +36,7 @@ pub(crate) fn define(shared: &mut SharedDefinitions, x86_instructions: &Instruct
     let fcvt_to_uint = insts.by_name("fcvt_to_uint");
     let fcvt_to_sint_sat = insts.by_name("fcvt_to_sint_sat");
     let fcvt_to_uint_sat = insts.by_name("fcvt_to_uint_sat");
+    let fabs = insts.by_name("fabs");
     let fmax = insts.by_name("fmax");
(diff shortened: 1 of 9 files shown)


cranelift-codegen/meta/src/isa/x86/legalize.rs
```diff
@@ -543,6 +548,31 @@ pub(crate) fn define(shared: &mut SharedDefinitions, x86_instructions: &Instruct
...
+    for ty in &[F32, F64] {
+        let fabs = fabs.bind(vector(*ty, sse_vector_size));
+        let lane_type_as_int = LaneType::int_from_bits(LaneType::from(*ty).lane_bits() as u16);
+        let vconst = vconst.bind(vector(lane_type_as_int, sse_vector_size));
+        let bitcast_to_float = raw_bitcast.bind(vector(*ty, sse_vector_size));
+        narrow.legalize(
+            def!(b = fabs(a)),
+            vec![
+                def!(c = vconst(ones)),
+                def!(d = ushr_imm(c, uimm8_one)), // Create a mask of all 1s except the MSB.
```
sstangl: In SpiderMonkey, for whatever reason, we load this from a constant pool. This is clearly the better option, particularly if we have a GVN pass.

PR #1222, opened 2019-11-13
Add legalization for popcnt.i128 via narrowing, fixes #1116

### LINKED ISSUE #1116: Legalize popcnt.i128

krk:
This PR lowers `popcnt.i128` by narrowing it to two `popcnt.i64` instructions and an addition.

### DIFF
--- a/cranelift-codegen/meta/src/shared/legalize.rs
+++ b/cranelift-codegen/meta/src/shared/legalize.rs
@@ -275,6 +275,17 @@ pub(crate) fn define(insts: &InstructionGroup, imm: &Immediates) -> TransformGro
         ],
     );
 
+    narrow.legalize(
+        def!(a = popcnt.I128(x)),
+        vec![
+            def!((xl, xh) = isplit(x)),
+            def!(e1 = popcnt(xl)),
+            def!(e2 = popcnt(xh)),
+            def!(e3 = iadd(e1, e2)),
+            def!(a = uextend(e3)),
+        ],
+    );
+
     // TODO(ryzokuken): benchmark this and decide if branching is a faster
     // approach than evaluating boolean expressions.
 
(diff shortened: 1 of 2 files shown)


krk:
ping @bnjbvr

cranelift-codegen/meta/src/shared/legalize.rs
```diff
@@ -275,6 +275,17 @@ pub(crate) fn define(insts: &InstructionGroup, imm: &Immediates) -> TransformGro
         ],
     );
 
+    narrow.legalize(
+        def!(a = popcnt.I128(x)),
```
bnjbvr: This is even valid for types other than `I128` when narrowing is needed, for instance for i64 on i686. Can you remove the `.I128` please? (I think it should just work because of our type inference; otherwise some in-flight work is required, and let's forget about this at the moment, see #1119 for context.)

If it does indeed work, can you also add a test for an `popcnt.i64` on i686, please?

krk: It does not compile when I remove the `.I128`:

```rust
error[E0308]: mismatched types
    --> /home/kerem/src/cranelift/target/debug/build/cranelift-codegen-f09ff84b74476b73/out/legalizer.rs:2138:60
     |
2138 |                     (Some(a), Some(b)) => a.wider_or_equal(b),
     |                                                            ^ expected struct `ir::types::Type`, found enum `core::option::Option`
     |
     = note: expected type `ir::types::Type`
                found type `core::option::Option<ir::types::Type>`
```

bnjbvr: Ah, then there are missing bits from this WIP PR mentioned in #1119, we'll worry about this later.

bnjbvr:
CI failure is the fuzzing failure we've seen for a couple days and that should disappear once landing on master. Landing. Thanks again for the patch.

PR #1223, opened 2019-11-13
Bump version to 0.50.0

abrown:
@sunfishcode, I think this commit didn't get a PR the last time `publish-all.sh` was run.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "cranelift-tools"
 authors = ["The Cranelift Project Developers"]
-version = "0.49.0"
+version = "0.50.0"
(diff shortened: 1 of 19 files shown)


abrown:
But, thinking about how this will affect Git history: crates.io has cranelift `0.50` crates already that are several commits behind `master` so if we merge this the Git history will not match the crates.io history. Another option is just to publish a `0.50.1` or something.

bnjbvr:
We'll probably want to publish a 0.51.0 with an accompanying pull request.

abrown:
Sounds good; I'm closing this.

sunfishcode:
I've now published a 0.51, with an accompanying pull request, so hopefully everything is sorted out here now.

Issue #1216, opened 2019-11-08
Check annotation failure in GitHub PRs


abrown:
- What are the steps to reproduce the issue? 

Take a look at the files for a PR, e.g., https://github.com/CraneStation/cranelift/pull/1211/files. I see something like the following at the bottom of the page:

![image](https://user-images.githubusercontent.com/541880/68495639-7a68c500-0205-11ea-95aa-6732112955e2.png)

- What do you expect to happen? What does actually happen? Does it panic, and
  if so, with which assertion?

This file was not changed in the PR so I would not expect this annotation failure. Also, it is not entirely clear what the failure is about. I suspect that with the switch to actions GitHub is now doing something extra for us but when I search around I don't see a clear path to fix this (@alexcrichton, any ideas?).

- Which Cranelift version / commit hash / branch are you using?

n/a 

- If relevant, can you include some extra information about your environment?
  (Rust version, operating system, architecture...)

n/a

bjorn3:
That annotation means that a command like `cargo test` exited with an error.

abrown:
I don't see that: when I look at https://github.com/CraneStation/cranelift/pull/1211/checks?check_run_id=293402517, e.g., there don't seem to be any error exits.

alexcrichton:
Github Actions I think seems to support adding "annotations" from CI to so CI can do something like issue a warning for a specific line in a specific file. I think the intention here is that it's immediately visible in the UI where a diagnostic from CI is pointed, rather than just parsing CI logs yourself.

It also seems to have a default annotation to say something failed, I'm not entirely sure what it's about. I'd turn it off if I knew how (since it's not too useful), but I don't know how :(

peterhuene:
I took a look at this and currently fuzzing is failing nightly CI (which runs on macOS) due to this Rust bug:

https://github.com/rust-lang/rust/issues/66140

Because the CI is marked as "continue on error", the "Run fuzz_module" step is marked as passing rather than failing.

I think there's a bug here in GitHub's annotations feature that is just showing the check failure at line 1 in the CI yaml file rather than the line for the "Run fuzz_module" command.

Our options to fix this:

* Do nothing, put up with the annotation, and let it correct itself when nightly is fixed.
* Switch to another platform for the nightly CI check.
* Disable the fuzz step for nightly until it is fixed.

alexcrichton:
[...]

PR #1224, opened 2019-11-13
Temporarily disable fuzzing until #1216 is resolved

abrown:
As discussed in #1216, fuzzing is temporarily broken. This change disables the GitHub workflow steps that run fuzzing until this changes.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -109,12 +109,13 @@ jobs:
         cargo fuzz run fuzz_translate_module fuzz/corpus/fuzz_translate_module/$fuzz_module
       env:
         RUST_BACKTRACE: 1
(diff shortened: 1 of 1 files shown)


peterhuene:
The annotation is still appearing for the nightly build.  Don't we need to disable its fuzzing step?

abrown:
Good catch; yes, now that should be disabled as well. I'm not sure that the bug you linked to in #1216 is causing the failures, though (https://github.com/rust-lang/rust/issues/66140). I see issues compiling `syntex_syntax`.

peterhuene:
Yeah, looks like something else is now breaking.  I don't know if it's hiding the original issue or not.

alexcrichton:
👍

PR #1229, opened 2019-11-14
Add legalization for bitrev.i128 via narrowing, fixes #1116

### LINKED ISSUE #1116: Legalize popcnt.i128

krk:
This PR lowers bitrev.i128 by narrowing it to two bitrev.i64 instructions and an addition.

ping @bnjbvr

### DIFF
--- a/cranelift-codegen/meta/src/shared/legalize.rs
+++ b/cranelift-codegen/meta/src/shared/legalize.rs
@@ -848,9 +848,9 @@ pub(crate) fn define(insts: &InstructionGroup, imm: &Immediates) -> TransformGro
         def!(a = bitrev.I128(x)),
         vec![
             def!((xl, xh) = isplit(x)),
-            def!(eh = bitrev(xl)),
-            def!(el = bitrev(xh)),
-            def!(a = iconcat(el, eh)),
+            def!(yh = bitrev(xl)),
+            def!(yl = bitrev(xh)),
+            def!(a = iconcat(yl, yh)),
         ],
     );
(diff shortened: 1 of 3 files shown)


cranelift-codegen/meta/src/shared/legalize.rs
```diff
@@ -844,6 +844,16 @@ pub(crate) fn define(insts: &InstructionGroup, imm: &Immediates) -> TransformGro
         ],
     );
 
+    narrow.legalize(
+        def!(a = bitrev.I128(x)),
+        vec![
+            def!((xl, xh) = isplit(x)),
+            def!(e1 = bitrev(xl)),
```
julian-seward1: Nano-nit: the naming is a bit inconsistent.  Maybe rename `e1` to `eh` and `e2` to `el`?  Then it would read a bit better.

julian-seward1:
Random drive-by comment that occurred to me while looking at this.  The legalisation for `bitrev.i64` is long and not very pretty.  On a target that has a byte-level swap instruction, wouldn't it be possible to use that instruction and hence omit the stages associated with the masks `0xff00_ff00_ff00_ff00`, `0xffff_0000_ffff_0000` and the 32 bit left/right shifts?  That would pretty much halve the length.  On x86_64, isn't there a `bswapq` instruction for that?

bjorn3:
>  On x86_64, isn't there a bswapq instruction for that?

https://github.com/bytecodealliance/cranelift/issues/794

PR #1231, opened 2019-11-14
Print return values of run tests if type is not boolean.

krk:
If return value of a run test is not boolean, an error is created with
the message "Function return type is not boolean, return value is {}".

- This was very briefly discussed on gitter with @sunfishcode.
- This PR provides a way to introspect into the run test variable values, helpful while writing tests and debugging.

### DIFF
--- a/cranelift-codegen/src/ir/constant.rs
+++ b/cranelift-codegen/src/ir/constant.rs
@@ -8,21 +8,16 @@
 //! - ensuring alignment of constants within the pool,
 //! - bucketing constants by size.
 
-#![allow(trivial_numeric_casts)]
-
 use crate::ir::immediates::{IntoBytes, V128Imm};
 use crate::ir::Constant;
 use crate::HashMap;
 use alloc::collections::BTreeMap;
-use alloc::string::String;
 use alloc::vec::Vec;
 use core::fmt;
 use core::iter::FromIterator;
-use core::mem;
 use core::slice::Iter;
 use core::str::{from_utf8, FromStr};
 use cranelift_entity::EntityRef;
-use std::convert::TryInto;
 
 /// This type describes the actual constant data. Note that the bytes stored in this structure are
 /// expected to be in little-endian order; this is due to ease-of-use when interacting with
@@ -54,72 +49,6 @@ impl From<&[u8]> for ConstantData {
 impl From<V128Imm> for ConstantData {
     fn from(v: V128Imm) -> Self {
         Self(v.to_vec())
-    }
-}
-
-macro_rules! try_into_int {
-    ($type: ty) => {
-        impl TryInto<$type> for ConstantData {
-            type Error = String;
-
-            fn try_into(self) -> Result<$type, Self::Error> {
-                if self.len() == (mem::size_of::<$type>()) {
-                    let v = self.into_vec().into_iter().rev();
-                    let mut r: $type = 0;
-                    for b in v {
-                        let (mut shifted, _) = r.overflowing_shl(8);
-                        shifted |= b as $type;
-                        r = shifted;
-                    }
-                    Ok(r)
-                } else {
-                    Err(format!(
-                        "Incorrect vector size: {}, expected {}",
-                        self.len(),
-                        mem::size_of::<$type>()
-                    ))
-                }
-            }
-        }
-    };
-}
-
-try_into_int!(u8);
-try_into_int!(u16);
-try_into_int!(u32);
-try_into_int!(u64);
-try_into_int!(u128);
-
-impl TryInto<f32> for ConstantData {
-    type Error = String;
-
-    fn try_into(self) -> Result<f32, Self::Error> {
-        let bits = self.try_into();
-        match bits {
-            Ok(v) => Ok(f32::from_bits(v)),
-            Err(e) => Err(e),
-        }
-    }
-}
-
-impl TryInto<f64> for ConstantData {
-    type Error = String;
-
-    fn try_into(self) -> Result<f64, Self::Error> {
-        let bits = self.try_into();
-        match bits {
-            Ok(v) => Ok(f64::from_bits(v)),
-            Err(e) => Err(e),
-        }
-    }
-}
-
-impl TryInto<bool> for ConstantData {
-    type Error = String;
-
-    fn try_into(self) -> Result<bool, Self::Error> {
-        // Only check the least significant byte regardless of size.
-        Ok(self.into_vec()[0] != 0)
     }
 }
 
--- a/cranelift-codegen/src/ir/constant.rs
+++ b/cranelift-codegen/src/ir/constant.rs
@@ -8,16 +8,21 @@
 //! - ensuring alignment of constants within the pool,
 //! - bucketing constants by size.
 
+#![allow(trivial_numeric_casts)]
+
 use crate::ir::immediates::{IntoBytes, V128Imm};
 use crate::ir::Constant;
 use crate::HashMap;
 use alloc::collections::BTreeMap;
+use alloc::string::String;
 use alloc::vec::Vec;
 use core::fmt;
 use core::iter::FromIterator;
+use core::mem;
 use core::slice::Iter;
 use core::str::{from_utf8, FromStr};
 use cranelift_entity::EntityRef;
+use std::convert::TryInto;
 
 /// This type describes the actual constant data. Note that the bytes stored in this structure are
 /// expected to be in little-endian order; this is due to ease-of-use when interacting with
@@ -49,6 +54,72 @@ impl From<&[u8]> for ConstantData {
 impl From<V128Imm> for ConstantData {
     fn from(v: V128Imm) -> Self {
         Self(v.to_vec())
+    }
+}
+
+macro_rules! try_into_int {
+    ($type: ty) => {
+        impl TryInto<$type> for ConstantData {
+            type Error = String;
+
+            fn try_into(self) -> Result<$type, Self::Error> {
+                if self.len() == (mem::size_of::<$type>()) {
+                    let v = self.into_vec().into_iter().rev();
+                    let mut r: $type = 0;
+                    for b in v {
+                        let (mut shifted, _) = r.overflowing_shl(8);
+                        shifted |= b as $type;
+                        r = shifted;
+                    }
+                    Ok(r)
+                } else {
+                    Err(format!(
+                        "Incorrect vector size: {}, expected {}",
+                        self.len(),
+                        mem::size_of::<$type>()
+                    ))
+                }
+            }
+        }
+    };
+}
+
+try_into_int!(u8);
+try_into_int!(u16);
+try_into_int!(u32);
+try_into_int!(u64);
+try_into_int!(u128);
+
+impl TryInto<f32> for ConstantData {
+    type Error = String;
+
+    fn try_into(self) -> Result<f32, Self::Error> {
+        let bits = self.try_into();
+        match bits {
+            Ok(v) => Ok(f32::from_bits(v)),
+            Err(e) => Err(e),
+        }
+    }
+}
+
+impl TryInto<f64> for ConstantData {
+    type Error = String;
+
+    fn try_into(self) -> Result<f64, Self::Error> {
+        let bits = self.try_into();
+        match bits {
+            Ok(v) => Ok(f64::from_bits(v)),
+            Err(e) => Err(e),
+        }
+    }
+}
+
+impl TryInto<bool> for ConstantData {
+    type Error = String;
+
+    fn try_into(self) -> Result<bool, Self::Error> {
+        // Only check the least significant byte regardless of size.
+        Ok(self.into_vec()[0] != 0)
     }
 }
 
--- a/cranelift-filetests/src/function_runner.rs
+++ b/cranelift-filetests/src/function_runner.rs
@@ -1,5 +1,6 @@
 use core::fmt::{self, Display, Formatter};
 use core::mem;
+use core::str::FromStr;
 use cranelift_codegen::binemit::{NullRelocSink, NullStackmapSink, NullTrapSink};
 use cranelift_codegen::ir::{types, AbiParam, ArgumentPurpose, ConstantData, Function};
 use cranelift_codegen::isa::TargetIsa;
@@ -46,6 +47,18 @@ impl Display for Operator {
             Operator::EQ => "==",
             Operator::NEQ => "!=",
         })
+    }
+}
+
+impl FromStr for Operator {
+    type Err = String;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        match s {
+            "==" => Ok(Operator::EQ),
+            "!=" => Ok(Operator::NEQ),
+            _ => Err(format!("Unsupported operator {}", s)),
+        }
     }
 }
 
@@ -317,7 +330,7 @@ impl FunctionRunnerBuilder {
         let expected = parse_constant_data(parts[parts.len() - 1], ty).unwrap();
         parts.drain(parts.len() - 1..);
 
-        let operator = Self::parse_operator(parts[parts.len() - 1]).unwrap();
+        let operator: Operator = parts[parts.len() - 1].parse().unwrap();
         parts.drain(parts.len() - 1..);
 
         if parts.len() != 1 {
@@ -341,14 +354,6 @@ impl FunctionRunnerBuilder {
             "Expected \"%fn()\" <operator> <value>\", found \"{}\"",
             parts.join(" ")
         );
-    }
-
-    fn parse_operator(s: &str) -> Result<Operator, String> {
-        match s {
-            "==" => Ok(Operator::EQ),
-            "!=" => Ok(Operator::NEQ),
-            _ => Err(format!("Unsupported operator {}", s)),
-        }
     }
 }
 
--- a/cranelift-filetests/src/function_runner.rs
+++ b/cranelift-filetests/src/function_runner.rs
@@ -110,12 +110,6 @@ impl Value {
             return Err(format!("Expected a non-vector type, not {}", ty));
         }
 
-        fn boolean_to_vec(value: bool) -> Vec<u8> {
-            let mut buffer = vec![0; 1];
-            buffer[0] = if value { 1 } else { 0 };
-            buffer
-        }
-
         let mut p = Parser::new(s);
 
         let constant_data = match ty {
@@ -152,12 +146,7 @@ impl Value {
                 let data = ConstantDataU64::try_into(data).unwrap();
                 Self::F64(f64::from_bits(data))
             }
-            b if b.is_bool() => {
-                let data = ConstantData::default()
-                    .append(boolean_to_vec(p.match_bool("Expected a boolean").unwrap()));
-                let data = ConstantDataU8::try_into(data).unwrap();
-                Self::Bool(data != 0)
-            }
+            b if b.is_bool() => Self::Bool(p.match_bool("Expected a boolean").unwrap()),
             _ => {
                 return Err(format!(
                     "Expected a type of: float, int or bool, found {}.",
--- a/cranelift-filetests/src/function_runner.rs
+++ b/cranelift-filetests/src/function_runner.rs
@@ -345,12 +345,7 @@ fn return_type(func: &Function) -> Result<types::Type, String> {
     }
     if normal_rets.len() == 2 {
         // Only two i64 return arguments supported and they are evaluated as a i128.
-        let both_i64 = normal_rets
-            .into_iter()
-            .map(|p| p.value_type)
-            .all(|ty| ty == types::I64);
-
-        if !both_i64 {
+        if !(normal_rets[0].value_type == types::I64 && normal_rets[1].value_type == types::I64) {
             return Err(String::from(
                 "Function return type with two parameters can only return i64, i64.",
             ));
--- a/cranelift-filetests/src/function_runner.rs
+++ b/cranelift-filetests/src/function_runner.rs
@@ -86,7 +86,6 @@ macro_rules! try_into_int {
     };
 }
 
-try_into_int!(u8, ConstantDataU8);
 try_into_int!(u16, ConstantDataU16);
 try_into_int!(u32, ConstantDataU32);
 try_into_int!(u64, ConstantDataU64);
@@ -114,20 +113,24 @@ impl Value {
 
         let constant_data = match ty {
             types::I8 => {
-                let data = p.match_constant_data(ty).unwrap();
-                Self::I8(ConstantDataU8::try_into(data).unwrap())
+                let data = p.match_uimm8("Expected a 8-bit unsigned integer").unwrap();
+                Self::I8(data.into())
             }
             types::I16 => {
                 let data = p.match_constant_data(ty).unwrap();
                 Self::I16(ConstantDataU16::try_into(data).unwrap())
             }
             types::I32 => {
-                let data = p.match_constant_data(ty).unwrap();
-                Self::I32(ConstantDataU32::try_into(data).unwrap())
+                let data = p
+                    .match_uimm32("Expected a 32-bit unsigned integer")
+                    .unwrap();
+                Self::I32(data.into())
             }
             types::I64 => {
-                let data = p.match_constant_data(ty).unwrap();
-                Self::I64(ConstantDataU64::try_into(data).unwrap())
+                let data = p
+                    .match_uimm64("Expected a 64-bit unsigned integer")
+                    .unwrap();
+                Self::I64(data.into())
             }
             types::I128 => {
                 let data = p.match_constant_data(ty).unwrap();
--- a/cranelift-filetests/src/function_runner.rs
+++ b/cranelift-filetests/src/function_runner.rs
@@ -341,7 +341,7 @@ fn return_type(func: &Function) -> Result<types::Type, String> {
         .filter(|r| r.purpose == ArgumentPurpose::Normal)
         .collect();
     if normal_rets.len() > 2 || normal_rets.len() == 0 {
-        return Err(String::from("Function return type is not supported"));
+        return Err(String::from("Number of return values not supported"));
     }
     if normal_rets.len() == 2 {
         // Only two i64 return arguments supported and they are evaluated as a i128.
--- a/cranelift-filetests/src/function_runner.rs
+++ b/cranelift-filetests/src/function_runner.rs
@@ -171,10 +171,7 @@ impl FunctionRunner {
         }
     }
 
-    fn invoke<T>(code_page: &Mmap) -> T
-    where
-        T: std::fmt::Display,
-    {
+    fn invoke<T>(code_page: &Mmap) -> T {
         let callable_fn: fn() -> T = unsafe { mem::transmute(code_page.as_ptr()) };
         callable_fn()
     }
--- a/cranelift-filetests/src/function_runner.rs
+++ b/cranelift-filetests/src/function_runner.rs
@@ -1,5 +1,3 @@
-#![allow(trivial_numeric_casts)]
-
 use core::fmt::{self, Display, Formatter};
 use core::mem;
 use core::str::FromStr;
--- a/cranelift-filetests/src/function_runner.rs
+++ b/cranelift-filetests/src/function_runner.rs
@@ -156,14 +156,16 @@ impl FunctionRunner {
         let func = self.function.clone();
 
         if !(func.signature.params.is_empty() && func.signature.returns.len() == 1) {
-            return Err(String::from(
-                "Functions must have a signature like: () -> ty",
+            return Err(format!(
+                "Functions must have a signature like: () -> ty, found {}",
+                func.name
             ));
         }
 
         if func.signature.call_conv != self.isa.default_call_conv() {
-            return Err(String::from(
-                "Functions only run on the host's default calling convention; remove the specified calling convention in the function signature to use the host's default.",
+            return Err(format!(
+                "Functions only run on the host's default calling convention; remove the specified calling convention in the function signature to use the host's default, in {}",
+                func.name
             ));
         }
 
@@ -199,13 +201,16 @@ impl FunctionRunner {
         match &self.action {
             FunctionRunnerAction::Test => {
                 if !ret_value_type.is_bool() {
-                    return Err(String::from("Function return type is not boolean"));
+                    return Err(format!(
+                        "Function {} return type is not boolean",
+                        context.func.name
+                    ));
                 }
                 let result = Self::invoke::<bool>(&code_page);
                 if result {
                     return Ok(());
                 }
-                return Err(format!("Failed: {}", context.func.name.to_string()));
+                return Err(format!("Failed: {}", context.func.name));
             }
             FunctionRunnerAction::Print => Self::check(
                 code_page,
--- a/cranelift-filetests/src/function_runner.rs
+++ b/cranelift-filetests/src/function_runner.rs
@@ -7,8 +7,7 @@ use cranelift_codegen::isa::TargetIsa;
 use cranelift_codegen::{settings, Context};
 use cranelift_native::builder as host_isa_builder;
 use cranelift_reader::Parser;
-use memmap::Mmap;
-use memmap::MmapMut;
+use memmap::{Mmap, MmapMut};
 
 /// Run a function on a host
 pub struct FunctionRunner {
--- a/cranelift-filetests/src/function_runner.rs
+++ b/cranelift-filetests/src/function_runner.rs
@@ -1,34 +1,139 @@
+use core::fmt::{self, Display, Formatter};
 use core::mem;
 use cranelift_codegen::binemit::{NullRelocSink, NullStackmapSink, NullTrapSink};
-use cranelift_codegen::ir::Function;
+use cranelift_codegen::ir::{types, AbiParam, ArgumentPurpose, ConstantData, Function};
 use cranelift_codegen::isa::TargetIsa;
 use cranelift_codegen::{settings, Context};
 use cranelift_native::builder as host_isa_builder;
+use cranelift_reader::parse_constant_data;
+use memmap::Mmap;
 use memmap::MmapMut;
+use std::convert::TryInto;
 
 /// Run a function on a host
 pub struct FunctionRunner {
     function: Function,
     isa: Box<dyn TargetIsa>,
+    action: FunctionRunnerAction,
+}
+
+/// Build a function runner
+pub struct FunctionRunnerBuilder {
+    function: Function,
+    isa: Option<Box<dyn TargetIsa>>,
+    action: FunctionRunnerAction,
+}
+
+#[derive(Clone, Debug)]
+pub enum FunctionRunnerAction {
+    Print,
+    Test,
+    TestWithResult {
+        operator: Operator,
+        expected: ConstantData,
+    },
+}
+
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+pub enum Operator {
+    EQ,
+    NEQ,
+}
+
+impl Display for Operator {
+    fn fmt(&self, f: &mut Formatter) -> fmt::Result {
+        f.write_str(match self {
+            Operator::EQ => "==",
+            Operator::NEQ => "!=",
+        })
+    }
 }
 
 impl FunctionRunner {
-    /// Build a function runner from a function and the ISA to run on (must be the host machine's ISA)
-    pub fn new(function: Function, isa: Box<dyn TargetIsa>) -> Self {
-        Self { function, isa }
+    /// Build a function runner from an action, a function and the ISA to run on (must be the host machine's ISA).
+    pub fn new(function: Function, isa: Box<dyn TargetIsa>, action: FunctionRunnerAction) -> Self {
+        Self {
+            function,
+            isa,
+            action,
+        }
     }
 
-    /// Build a function runner using the host machine's ISA and the passed flags
-    pub fn with_host_isa(function: Function, flags: settings::Flags) -> Self {
-        let builder = host_isa_builder().expect("Unable to build a TargetIsa for the current host");
-        let isa = builder.finish(flags);
-        Self::new(function, isa)
+    fn invoke<T>(code_page: &Mmap) -> T
+    where
+        T: std::fmt::Display,
+    {
+        let callable_fn: fn() -> T = unsafe { mem::transmute(code_page.as_ptr()) };
+        callable_fn()
     }
 
-    /// Build a function runner using the host machine's ISA and the default flags for this ISA
-    pub fn with_default_host_isa(function: Function) -> Self {
-        let flags = settings::Flags::new(settings::builder());
-        Self::with_host_isa(function, flags)
+    fn check_assertion<T>(
+        value: T,
+        operator: Operator,
+        expected: T,
+        print_only: bool,
+    ) -> Result<(), String>
+    where
+        T: std::fmt::Display,
+        T: std::cmp::PartialEq,
+        T: Copy,
+    {
+        if print_only {
+            return Err(format!("Return value is: {}", value));
+        }
+
+        let ok = match operator {
+            Operator::EQ => value == expected,
+            Operator::NEQ => value != expected,
+        };
+
+        if ok {
+            return Ok(());
+        }
+        return Err(format!(
+            "Check failed, expected {} {} {}",
+            expected, operator, value,
+        ));
+    }
+
+    fn check(
+        code_page: memmap::Mmap,
+        ret_value_type: types::Type,
+        operator: Operator,
+        expected: ConstantData,
+        print_only: bool,
+    ) -> Result<(), String> {
+        macro_rules! gen_match {
+            ($x:ident, $($y: path => $z: ty), +) => {
+                match $x {
+                    $($y => {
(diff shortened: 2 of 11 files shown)


abrown:
Good idea; I've needed something like this in the past. @nbp and I had talked about extending the `; run` syntax indicating that a function is supposed to be run with something like `run: %fn(3, 5) == 42`. With multiple lines we could then test different arguments to a function and verify their results, presumably then printing out errors like `expected 42, found 99`. Is there any way this could towards that direction?

abrown:
For context: https://github.com/bytecodealliance/cranelift/pull/872

krk:
I have initially suggested adding a new opcode for printing the argument similar to what `dbg!` does in rust, @sunfishcode suggested that it would be too much work.

If the intention is to print results of comparisons instead of a single value, maybe the function can return a multilane value and we could print the data as two values that are half width of the return type - which would be a workaround.

Extending the run comment `run: %fn(3, 5) == 42` to parse expressions and print both values brings the possibility of inventing a new language that could have more features like other operators, other types of functions etc. I think it would be overly complicated for me to implement at this stage.

On the other hand, one might say that a new `; print "value 1", v1` type of comment command would seem idiomatic w.r.t. the SSA expressions and variable names used in the code. If this is relatively easier to implement than a new `dbg!` opcode, I can try to implement that. Optionally, result of this `;print` command could be available to `check` commands, if that makes sense.

abrown:
> I think it would be overly complicated for me to implement at this stage

Makes sense, let's just go with the simple print idea then. I suspect that implementing `; print "value 1", v1` might be just as tricky as adding a new `print` instruction (we would have to mess around inside the function). To keep it simpler and still move towards the comparison idea, could we:
 - keep `; run` with its current logic of only running boolean-returning functions and testing their result (and eventually move this to `; run: %fn(2) > 42`)--optionally we could rename `; run` to `; run: test` for consistency
 - add `; run: print` which would use much of your current code to run a function and then print out the results if it can

What do you think of that?

krk:
Would it mean that `; run: print` would make the tests fail always, as it is the case currently?

krk:
Sure thing, it makes sense to modify the `run` comment now.

* I have renamed `;run` to `;run: test` and added print and assertion commands.
* Assertion only supports equality and inequality: `; run: %fn() == 42` and `; run: %fn() !=  0x2a`.
* Printing always fails the test and prints the return value: `; run: print`.

I do not like my `parse_options` function, is there an idiomatic way to refactor it into?

abrown:
Nice work! I'll take a closer look on Monday.

cranelift-filetests/src/function_runner.rs
```diff
@@ -75,14 +311,41 @@ impl FunctionRunner {
...
+                    ret_value_type,
+                    Operator::EQ,
+                    u128::from(true),
+                    self.options.print,
+                ),
+                _ => Err(format!(
+                    "Function arguments more than supported count: {}",
+                    self.options.args.len()
+                )),
+            },
```
abrown: This was a bit hard to read but I think I get it now; it might be more clear with an enum like I mentioned above:
```
match action {
  Print => println!(...),
  Test => check(...),
  TestWithResult => check(...),
}
```

abrown (commented):
Thanks for doing this! See my comments for some improvements but I think this approach makes sense. I will use these new features for sure 😄. If anything I said didn't make sense I'll be on the Mozilla IRC if you want to discuss.

bnjbvr (commented):
Just wanted to add: could we keep the previous behavior that `; run` is equivalent to `; run: test`, so we don't have to modify all the existing test cases, and remove `; run: test` too? (or find a better sub-directive name after the `:`, since "test" is quite overloaded in the context of testing)

cranelift-filetests/src/function_runner.rs
krk: This simplified things a lot!

cranelift-filetests/src/function_runner.rs
```diff
@@ -1,47 +1,283 @@
...
+        return Err(format!(
+            "Check failed: 0x{:X} ({0}, f32: {1}, f64: {2})\n{3:?}\n0x{4:X} ({4}, f32: {5}, f64: {6})",
+            v,
+            f32::from_bits(v as u32),
+            f64::from_bits(v as u64),
+            operator,
+            expected,
+            f32::from_bits(expected as u32),
+            f64::from_bits(expected as u64)
+        ));
```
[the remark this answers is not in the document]
krk: I used `ConstantData` https://github.com/bytecodealliance/cranelift/blob/5d98701e5cae03160ed1b5a36f5798b3544f43cd/cranelift-codegen/src/ir/constant.rs#L28 as the type for the `expected` value and reused parts of `parse_literals_to_constant_data` in `parser.rs`.

cranelift-filetests/src/function_runner.rs
```diff
@@ -1,47 +1,283 @@
...
+        }
+    }
+
+    fn parse_u128(s: &str) -> Result<u128, ParseIntError> {
+        if s.starts_with("0x") {
+            // Parse it as a u128 in hexadecimal form.
+            u128::from_str_radix(&s[2..], 16)
+        } else {
+            u128::from_str_radix(s, 10)
+        }
```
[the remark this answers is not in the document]
krk: Good catches, I have added tests for supported types in `filetests/function_runner` and added bool and float parsing to `parser::parse_constant_data`.

cranelift-filetests/src/function_runner.rs
```diff
@@ -1,47 +1,283 @@
...
 
 impl FunctionRunner {
     /// Build a function runner from a function and the ISA to run on (must be the host machine's ISA)
-    pub fn new(function: Function, isa: Box<dyn TargetIsa>) -> Self {
-        Self { function, isa }
+    pub fn new(function: Function, isa: Box<dyn TargetIsa>, comment: String) -> Self {
+        Self {
+            function,
+            isa,
+            options: Self::parse_options(comment),
```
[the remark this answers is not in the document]
krk: I have introduced a builder `FunctionRunnerBuilder` to unify it with other `with` functions.

cranelift-filetests/src/function_runner.rs
```diff
@@ -76,13 +179,175 @@ impl FunctionRunner {
...
+            "Expected \"%fn()\" <operator> <value>\", found \"{}\"",
+            parts.join(" ")
+        );
+    }
+
+    fn parse_operator(s: &str) -> Result<Operator, String> {
+        match s {
+            "==" => Ok(Operator::EQ),
+            "!=" => Ok(Operator::NEQ),
+            _ => Err(format!("Unsupported operator {}", s)),
```
abrown: A minor re-factoring suggestion: elsewhere in cranelift I've seen this type of thing as `impl FromStr for Operator { ... }` so that we can just call `.parse()` up on line 320. Not a big deal but it keeps the operator implementation together.

cranelift-filetests/src/function_runner.rs
```diff
@@ -76,13 +179,175 @@ impl FunctionRunner {
...
+    /// Use the default host ISA.
+    pub fn with_default_host_isa(self) -> Self {
+        let flags = settings::Flags::new(settings::builder());
+        self.with_host_isa(flags)
+    }
+
+    /// Build a function runner from an action, a function and the ISA to run on.
+    pub fn finish(self) -> FunctionRunner {
+        FunctionRunner::new(self.function, self.isa.unwrap(), self.action)
+    }
```
abrown: Nit: elsewhere in the project we are calling this `build`.

cranelift-filetests/src/function_runner.rs
```diff
@@ -76,13 +179,175 @@ impl FunctionRunner {
...
-            Ok(())
-        } else {
-            Err(format!("Failed: {}", context.func.name.to_string()))
+        // Merge two i64 to i128 if necessary.
+        let ret_value_type = return_type(&context.func)?;
+
+        match &self.action {
+            FunctionRunnerAction::Test => {
+                if !ret_value_type.is_bool() {
+                    return Err(String::from("Function return type is not boolean"));
```
abrown: Let's add the `context.func.name` here as well so that we can more easily find it.

cranelift-filetests/src/function_runner.rs
```diff
@@ -76,13 +179,175 @@ impl FunctionRunner {
...
+                }
+                let result = Self::invoke::<bool>(&code_page);
+                if result {
+                    return Ok(());
+                }
+                return Err(format!("Failed: {}", context.func.name.to_string()));
```
abrown: Re-factoring nit: I'm not really used to seeing `return` much in Rust 😄:
```
if Self::invoke::<bool>(&code_page) {
  Ok(())
} else {
  Err(format!("Failed: {}", context.func.name.to_string()))
}
```

cranelift-filetests/src/function_runner.rs
```diff
@@ -76,13 +179,175 @@ impl FunctionRunner {
...
+            }
+            FunctionRunnerAction::Print => Self::check(
+                code_page,
+                ret_value_type,
+                Operator::EQ,
+                ConstantData::default().append(1),
+                true,
+            ),
```
abrown: Perhaps we should not overload `check`; why not just `invoke` here and `println!` the returned value?

cranelift-codegen/src/ir/constant.rs
```diff
@@ -52,6 +57,72 @@ impl From<V128Imm> for ConstantData {
...
+
+impl TryInto<bool> for ConstantData {
+    type Error = String;
+
+    fn try_into(self) -> Result<bool, Self::Error> {
+        // Only check the least significant byte regardless of size.
+        Ok(self.into_vec()[0] != 0)
+    }
+}
+
```
abrown: I don't think we should modify `ConstantData` for what we are doing; it is designed for constant pool stuff and adding these conversions might make it harder to decouple later from test functionality. Why not build our own `Value` enum:
```
enum Value {
  Bool(bool),
  F32(f32),
  F64(f64),
  ... 
}
```
I wouldn't worry too much about 128-bit types just yet until we get the basics in place. Then I would add a way to parse into this (using Rust's `FromStr/parse`) from text and a type:
```
impl Value {
  fn parse_as(text: &str, type: Type) -> Result<Self, ...> {
    match type {
      B8 | B16 | ... => Ok(Value::Bool(text.parse()?)),
      F32 => Ok(Value::F32(text.parse()?)),
      ...
}
```
Then I would `#[derive(Eq, PartialEq)]` on the enum so we can compare them to each other and add some `From` implementations so that `invoke` can get the raw value (e.g. f32), wrap it in a `Value` with `Value::from` (or `into`), and then compare it against the parsed `Value`s that we get from the comments. (Hopefully that explains the other comments I've made; this isn't the only way to do things but it doesn't involve touching other components like `ConstantData` and it seems natural to Rust--from the little I know about Rust).

cranelift-filetests/src/function_runner.rs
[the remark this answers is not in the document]
abrown: See my comments about using something like `Value` instead.

cranelift-reader/src/parser.rs
```diff
@@ -864,6 +870,37 @@ impl<'a> Parser<'a> {
...
+            F32 => consume!(ty, self.match_ieee32("Expected a 32-bit float")?),
+            F64 => consume!(ty, self.match_ieee64("Expected a 64-bit float")?),
+            b if b.is_bool() => {
+                consume!(ty, boolean_to_vec(self.match_bool("Expected a boolean")?))
+            }
+            _ => return err!(self.loc, "Expected a type of: float, int, bool"),
+        };
+        Ok(constant_data)
+    }
+
```
abrown: See my parsing comments about `Value`; we may not need to change the parser.

docs/testing.rst
```diff
@@ -397,3 +397,31 @@ Example::
...
+    test run
+
+    function %trivial_test() -> i16 {
+    ebb0:
+        v0 = iconst.i16 42
+        return v0
+    }
+    ; run: print
+
+To assert on the return value of the function, add a ``; run: %fn() <op> <expected>`` directive
```
abrown: Is it `%fn` or `%trivial_test` here?

abrown: Never mind, looks like it is `%fn` 👍

filetests/function_runner/types.clif
```diff
@@ -0,0 +1,88 @@
...
+    return v0
+}
+; run: %fn() == 0x1.23
+
+function %test_f64() -> f64 {
+ebb0:
+    v0 = f64const 0x1000.0001
+    return v0
+}
+; run: %fn() == 0x1000.0001
```
abrown: Nice work on testing

cranelift-codegen/src/ir/constant.rs
abrown: Let me change my tune a bit: I now see down below that you implemented some parsing logic in `parser.rs`; if you want to use that for the `FromStr` on `Value` instead of the built-in Rust parsing that seems OK. We could use methods like `Parser::new(text).match_bool(...)` instead and `Value` may have to use different inner types (e.g. `F32(Ieee32)` instead of `F32(f32)`)--but that seems fine. The underlying thing that I'm getting at is that we likely don't want to touch `ConstantData` in this case.

cranelift-filetests/src/function_runner.rs
krk: `cranelift-codegen/src/isa/mod.rs` calls it `finish`.

https://github.com/bytecodealliance/cranelift/blob/da996fe5ee4ea32a57156ba93f6813b052964183/cranelift-codegen/src/isa/mod.rs#L151

abrown (approved):
Looks good to me; nice work. We could probably simplify parsing more but we can do that when we finish out the `%fn(42) == 2` syntax (do we have an issue for that, BTW?). @bnjbvr, I'm good to merge but giving you a chance to take a look. In the meantime, @krk, any idea why CI is failing?

abrown:
(I will be out next week on vacation and it looks like @bnjbvr may be as well so it might not get merged until after Thanskgiving week).

krk:
Only on windows there is a verifier error:

```
2019-11-21T22:50:54.2726499Z FAIL filetests\function_runner\types.clif: run(%test_i128):
2019-11-21T22:50:54.2727654Z Verifier errors
2019-11-21T22:50:54.2727967Z 263 tests
2019-11-21T22:50:54.2728213Z thread 'filetests' panicked at 'test harness: "1 failure"', src\libcore\result.rs:1165:5
2019-11-21T22:50:54.2728470Z stack backtrace:
2019-11-21T22:50:54.2728728Z    0: backtrace::backtrace::trace_unsynchronized
2019-11-21T22:50:54.2728940Z              at C:\Users\VssAdministrator\.cargo\registry\src\github.com-1ecc6299db9ec823\backtrace-0.3.37\src\backtrace\mod.rs:66
2019-11-21T22:50:54.2729193Z    1: std::sys_common::backtrace::_print_fmt
2019-11-21T22:50:54.2729455Z              at /rustc/4560ea788cb760f0a34127156c78e2552949f734\/src\libstd\sys_common\backtrace.rs:76
2019-11-21T22:50:54.2729709Z    2: std::sys_common::backtrace::_print::{{impl}}::fmt
2019-11-21T22:50:54.2729958Z              at /rustc/4560ea788cb760f0a34127156c78e2552949f734\/src\libstd\sys_common\backtrace.rs:60
2019-11-21T22:50:54.2730196Z    3: core::fmt::write
2019-11-21T22:50:54.2730388Z              at /rustc/4560ea788cb760f0a34127156c78e2552949f734\/src\libcore\fmt\mod.rs:1030
2019-11-21T22:50:54.2730634Z    4: std::io::Write::write_fmt<test::Sink>
2019-11-21T22:50:54.2731057Z              at /rustc/4560ea788cb760f0a34127156c78e2552949f734\src\libstd\io\mod.rs:1412
2019-11-21T22:50:54.2731568Z    5: std::io::impls::{{impl}}::write_fmt<Write>
2019-11-21T22:50:54.2731827Z              at /rustc/4560ea788cb760f0a34127156c78e2552949f734\/src\libstd\io\impls.rs:141
2019-11-21T22:50:54.2732043Z    6: std::sys_common::backtrace::_print
2019-11-21T22:50:54.2732298Z              at /rustc/4560ea788cb760f0a34127156c78e2552949f734\/src\libstd\sys_common\backtrace.rs:64
2019-11-21T22:50:54.2732735Z    7: std::sys_common::backtrace::print
2019-11-21T22:50:54.2732982Z              at /rustc/4560ea788cb760f0a34127156c78e2552949f734\/src\libstd\sys_common\backtrace.rs:49
2019-11-21T22:50:54.2733226Z    8: std::panicking::default_hook::{{closure}}
2019-11-21T22:50:54.2733467Z              at /rustc/4560ea788cb760f0a34127156c78e2552949f734\/src\libstd\panicking.rs:196
2019-11-21T22:50:54.2733672Z    9: std::panicking::default_hook
2019-11-21T22:50:54.2733917Z              at /rustc/4560ea788cb760f0a34127156c78e2552949f734\/src\libstd\panicking.rs:207
2019-11-21T22:50:54.2734390Z   10: std::panicking::rust_panic_with_hook
2019-11-21T22:50:54.2734598Z              at /rustc/4560ea788cb760f0a34127156c78e2552949f734\/src\libstd\panicking.rs:473
2019-11-21T22:50:54.2734849Z   11: std::panicking::continue_panic_fmt
2019-11-21T22:50:54.2735087Z              at /rustc/4560ea788cb760f0a34127156c78e2552949f734\/src\libstd\panicking.rs:380
2019-11-21T22:50:54.2735324Z   12: std::panicking::rust_begin_panic
2019-11-21T22:50:54.2735566Z              at /rustc/4560ea788cb760f0a34127156c78e2552949f734\/src\libstd\panicking.rs:307
2019-11-21T22:50:54.2735767Z   13: core::panicking::panic_fmt
2019-11-21T22:50:54.2736005Z              at /rustc/4560ea788cb760f0a34127156c78e2552949f734\/src\libcore\panicking.rs:85
2019-11-21T22:50:54.2736242Z   14: core::result::unwrap_failed
2019-11-21T22:50:54.2736470Z              at /rustc/4560ea788cb760f0a34127156c78e2552949f734\/src\libcore\result.rs:1165
2019-11-21T22:50:54.2736724Z   15: core::result::Result<core::time::Duration, alloc::string::String>::expect
2019-11-21T22:50:54.2736986Z              at /rustc/4560ea788cb760f0a34127156c78e2552949f734\src\libcore\result.rs:960
2019-11-21T22:50:54.2737227Z   16: filetests::filetests
2019-11-21T22:50:54.2737452Z              at .\tests\filetests.rs:4
2019-11-21T22:50:54.2737679Z   17: filetests::filetests::{{closure}}
2019-11-21T22:50:54.2737902Z              at .\tests\filetests.rs:2
2019-11-21T22:50:54.2738191Z   18: core::ops::function::FnOnce::call_once<closure-0,()>
2019-11-21T22:50:54.2738481Z              at /rustc/4560ea788cb760f0a34127156c78e2552949f734\src\libcore\ops\function.rs:227
2019-11-21T22:50:54.2738729Z   19: alloc::boxed::{{impl}}::call_once<(),FnOnce<()>>
2019-11-21T22:50:54.2738972Z              at /rustc/4560ea788cb760f0a34127156c78e2552949f734\src\liballoc\boxed.rs:922
2019-11-21T22:50:54.2739206Z   20: panic_unwind::__rust_maybe_catch_panic
2019-11-21T22:50:54.2739443Z              at /rustc/4560ea788cb760f0a34127156c78e2552949f734\/src\libpanic_unwind\lib.rs:80
2019-11-21T22:50:54.2739652Z   21: std::panicking::try
2019-11-21T22:50:54.2739887Z              at /rustc/4560ea788cb760f0a34127156c78e2552949f734\src\libstd\panicking.rs:271
2019-11-21T22:50:54.2740131Z   22: std::panic::catch_unwind
2019-11-21T22:50:54.2740359Z              at /rustc/4560ea788cb760f0a34127156c78e2552949f734\src\libstd\panic.rs:394
2019-11-21T22:50:54.2740600Z   23: test::run_test::run_test_inner::{{closure}}
2019-11-21T22:50:54.2741000Z              at /rustc/4560ea788cb760f0a34127156c78e2552949f734\/src\libtest\lib.rs:1413
2019-11-21T22:50:54.2741270Z note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```

bnjbvr:
I'm happy to give it a look once the Windows failure has been solved. Let us know if you need any help on this!

krk:
I think incorrect machine code is generated for Windows x86-64 ABI. 

Return value in the failing test is 128 bits wide and in System V ABI it is expected to be in `rax` and `rdx` whereas Microsoft x64 ABI expects it in `rax` and the stack.

@bnjbvr When `%test_i128` is removed, tests would pass; otherwise I need some pointers to continue :)

Disassembly for `%test_i128` generated on Windows, which is same as it is generated on 64-bit Ubuntu except the calling convention annotation:

```diff
-function %test_i128(i64 fp [%rbp]) -> i64 [%rax], i64 [%rdx], i64 fp [%rbp] fast {
+function %test_i128(i64 fp [%rbp]) -> i64 [%rax], i64 [%rdx], i64 fp [%rbp] system_v {
```

```asm
function %test_i128(i64 fp [%rbp]) -> i64 [%rax], i64 [%rdx], i64 fp [%rbp] fast {
    ss0 = incoming_arg 16, offset -16
 
                                ebb0(v3: i64 [%rbp]):
[RexOp1pushq#50]                    x86_push v3
[RexOp1copysp#8089]                 copy_special %rsp -> %rbp
[RexOp1pu_iq#80b8,%rax]             v0 = iconst.i64 0x4242_3535_1234_5678
[RexOp1pu_iq#80b8,%rcx]             v1 = iconst.i64 0xabcd_ef01_beef_b411
[-,-]                               v2 = iconcat v0, v1
>>>[RexOp1rmov#8089]                   regmove v1, %rcx -> %rdx
[RexOp1popq#58,%rbp]                v4 = x86_pop.i64
[Op1ret#c3]                         return v0, v1, v4
}

Disassembly of 31 bytes:
   0:   40 55                   push    rbp
   2:   48 89 e5                mov     rbp, rsp
   5:   48 b8 78 56 34 12 35 35 42 42
                                movabs  rax, 0x4242353512345678
   f:   48 b9 11 b4 ef be 01 ef cd ab
                                movabs  rcx, 0xabcdef01beefb411
>>>  19:   48 89 ca                mov     rdx, rcx
  1c:   40 5d                   pop     rbp
  1e:   c3                      ret
```

`regmove` is generated in the RA coloring phase:

```rust
DEBUG cranelift_codegen::timing::details      > timing: Starting RA coloring, (during Register allocation)
 DEBUG cranelift_codegen::regalloc::coloring   > Coloring for:
function %test_i128() -> i64 [%rax], i64 [%rdx] system_v {
                                ebb0:
[RexOp1pu_iq#80b8]                  v0 = iconst.i64 0x4242_3535_1234_5678
[RexOp1pu_iq#80b8]                  v1 = iconst.i64 0xabcd_ef01_beef_b411
[-]                                 v2 = iconcat v0, v1
[Op1ret#c3]                         return v0, v1
}

 DEBUG cranelift_codegen::regalloc::coloring   > Coloring ebb0:
 DEBUG cranelift_codegen::regalloc::coloring   > Start ebb0 with entry-diversion set to
      { }
 DEBUG cranelift_codegen::regalloc::coloring   > Coloring v0 = iconst.i64 0x4242_3535_1234_5678
    from [ GPR: acdb--sd89012345 FPR: 0123456789012345 FLAG: f ]
 DEBUG cranelift_codegen::regalloc::coloring   >     glob [ GPR: acdb--sd89012345 FPR: 0123456789012345 FLAG: f ]
 DEBUG cranelift_codegen::regalloc::coloring   >     color v0 -> %rax
 DEBUG cranelift_codegen::regalloc::coloring   > Coloring v1 = iconst.i64 0xabcd_ef01_beef_b411
    from [ GPR: -cdb--sd89012345 FPR: 0123456789012345 FLAG: f ]
 DEBUG cranelift_codegen::regalloc::coloring   >     glob [ GPR: acdb--sd89012345 FPR: 0123456789012345 FLAG: f ]
 DEBUG cranelift_codegen::regalloc::coloring   >     color v1 -> %rcx
 DEBUG cranelift_codegen::regalloc::coloring   > Coloring return v0, v1
    from [ GPR: --db--sd89012345 FPR: 0123456789012345 FLAG: f ]
 DEBUG cranelift_codegen::regalloc::solver     > reassign_in(v0:GPR, %rax -> %rax)
 DEBUG cranelift_codegen::regalloc::solver     > reassign_in(v1:GPR, %rcx -> %rdx)
 DEBUG cranelift_codegen::regalloc::coloring   >     kill v0 in %rax (local GPR)
 DEBUG cranelift_codegen::regalloc::coloring   >     kill v1 in %rcx (local GPR)
 DEBUG cranelift_codegen::regalloc::coloring   >     glob [ GPR: acdb--sd89012345 FPR: 0123456789012345 FLAG: f ]
 DEBUG cranelift_codegen::regalloc::solver     > collect_moves: [v1:GPR(%rcx -> %rdx)]
>>> DEBUG cranelift_codegen::regalloc::solver     > move #0: v1:GPR(%rcx -> %rdx)
 DEBUG cranelift_codegen::timing::details      > timing: Ending RA coloring
```

bnjbvr:
Uh, it's an unfortunate conflict with the multi-value returns that emulate what SystemV does, which is incompatible with Win64 here. We can comment out the test in the meanwhile and open an issue about this problem.

Otherwise, it might be needed to investigate deeper in `cranelift-codegen/src/isa/x86/abi.rs` how to pass multiple values with respect to the target ABI convention.

cranelift-filetests/src/function_runner.rs
```diff
@@ -1,53 +1,301 @@
...
+
+impl Value {
+    fn parse_as(s: &str, ty: types::Type) -> Result<Self, String> {
+        if ty.is_vector() {
+            return Err(format!("Expected a non-vector type, not {}", ty));
+        }
+
+        macro_rules! consume {
+            ( $ty:ident, $match_fn:expr ) => {{
+                ConstantData::default().append($match_fn)
```
bnjbvr: Note: $ty is unused. Could we remove this macro and instead inline it where it's used?

cranelift-filetests/src/function_runner.rs
```diff
@@ -1,53 +1,301 @@
...
+            }};
+        }
+
+        fn boolean_to_vec(value: bool) -> Vec<u8> {
```
bnjbvr: This function is used once, can we inline it at its single use, and then remove it instead?

cranelift-filetests/src/function_runner.rs
```diff
@@ -1,53 +1,301 @@
...
+            types::I16 => {
+                let data = p.match_constant_data(ty).unwrap();
+                Self::I16(ConstantDataU16::try_into(data).unwrap())
+            }
+            types::I32 => {
+                let data = p.match_constant_data(ty).unwrap();
+                Self::I32(ConstantDataU32::try_into(data).unwrap())
+            }
+            types::I64 => {
+                let data = p.match_constant_data(ty).unwrap();
```
bnjbvr: It seems we could avoid using ConstantData for types that aren't SIMD, and have something more straightforward avoiding al the ConstantDataU{...} variants.

cranelift-filetests/src/function_runner.rs
```diff
@@ -1,53 +1,301 @@
...
+                let data = ConstantDataU64::try_into(data).unwrap();
+                Self::F64(f64::from_bits(data))
+            }
+            b if b.is_bool() => {
+                let data = consume!(
+                    ty,
+                    boolean_to_vec(p.match_bool("Expected a boolean").unwrap())
+                );
+                let data = ConstantDataU8::try_into(data).unwrap();
+                Self::Bool(data != 0)
```
bnjbvr: Could this just consume the result of `match_bool.unwrap()` instead?

cranelift-filetests/src/function_runner.rs
```diff
@@ -1,53 +1,301 @@
...
     }
 
-    /// Build a function runner using the host machine's ISA and the passed flags
-    pub fn with_host_isa(function: Function, flags: settings::Flags) -> Self {
-        let builder = host_isa_builder().expect("Unable to build a TargetIsa for the current host");
-        let isa = builder.finish(flags);
-        Self::new(function, isa)
+    fn invoke<T>(code_page: &Mmap) -> T
+    where
+        T: std::fmt::Display,
```
bnjbvr: It this requirement actually needed here?

cranelift-filetests/src/function_runner.rs
```diff
@@ -1,53 +1,301 @@
...
-            && func.signature.returns.len() == 1
-            && func.signature.returns.first().unwrap().value_type.is_bool())
-        {
-            return Err(String::from(
-                "Functions must have a signature like: () -> boolean",
+
+        if !(func.signature.params.is_empty() && func.signature.returns.len() == 1) {
+            return Err(format!(
+                "Functions must have a signature like: () -> ty, found {}",
+                func.name
```
bnjbvr: The error message seems to be incorrect; maybe we can tell which function doesn't respect this, and what is its actual signature?

cranelift-filetests/src/function_runner.rs
```diff
@@ -76,14 +324,159 @@ impl FunctionRunner {
...
+
+fn return_type(func: &Function) -> Result<types::Type, String> {
+    let normal_rets: Vec<&AbiParam> = func
+        .signature
+        .returns
+        .iter()
+        .filter(|r| r.purpose == ArgumentPurpose::Normal)
+        .collect();
+    if normal_rets.len() > 2 || normal_rets.len() == 0 {
+        return Err(String::from("Function return type is not supported"));
```
bnjbvr: This should probably read "Number of return values not supported".

cranelift-filetests/src/function_runner.rs
```diff
@@ -76,14 +324,159 @@ impl FunctionRunner {
...
+    }
+    if normal_rets.len() == 2 {
+        // Only two i64 return arguments supported and they are evaluated as a i128.
+        let both_i64 = normal_rets
+            .into_iter()
+            .map(|p| p.value_type)
+            .all(|ty| ty == types::I64);
```
bnjbvr: It seems it would be straightforward to check the type of the two array values at once here.

cranelift-filetests/src/function_runner.rs
bnjbvr: Also, it would be worth merging this check in the `return_type` function, and call this function sooner, to avoid compiling if we can't check the result.

cranelift-filetests/src/function_runner.rs
bnjbvr: Yeah, in this particular case it's this file which is not consistent with the Rust style, so we should fix it (later).

bnjbvr (approved):
Looks good to me. There are a few simplifications that could be done, either in this PR or as a follow-up PR. Thanks!

cranelift-filetests/src/function_runner.rs
krk: No, remnant from debugging, will remove.

cranelift-filetests/src/function_runner.rs
krk: I have removed ConstantData use for I8, I32 and I64.

cranelift-filetests/src/function_runner.rs
```diff
@@ -1,53 +1,280 @@
+#![allow(trivial_numeric_casts)]
```
bjorn3: Could you please move this to the actual function doing this?

cranelift-filetests/src/function_runner.rs
```diff
@@ -1,53 +1,280 @@
...
+use core::str::FromStr;
 use cranelift_codegen::binemit::{NullRelocSink, NullStackmapSink, NullTrapSink};
-use cranelift_codegen::ir::Function;
+use cranelift_codegen::ir::{types, AbiParam, ArgumentPurpose, ConstantData, Function};
 use cranelift_codegen::isa::TargetIsa;
 use cranelift_codegen::{settings, Context};
 use cranelift_native::builder as host_isa_builder;
+use cranelift_reader::Parser;
+use memmap::Mmap;
 use memmap::MmapMut;
```
bjorn3: `use memmap::{Mmap, MmapMut};`

cranelift-filetests/src/function_runner.rs
```diff
@@ -1,53 +1,280 @@
...
+macro_rules! try_into_int {
+    ($type: ty, $name: ident) => {
+        struct $name {}
+        impl $name {
+            fn try_into(d: ConstantData) -> Result<$type, String> {
+                if d.len() == (mem::size_of::<$type>()) {
+                    let v = d.into_vec().into_iter().rev();
+                    let mut r: $type = 0;
+                    for b in v {
+                        let (mut shifted, _) = r.overflowing_shl(8);
```
bjorn3: This can't overflow, right?

cranelift-filetests/src/function_runner.rs
```diff
@@ -1,53 +1,280 @@
...
+
+        let mut p = Parser::new(s);
+
+        let constant_data = match ty {
+            types::I8 => {
+                let data = p.match_uimm8("Expected a 8-bit unsigned integer").unwrap();
+                Self::I8(data.into())
+            }
+            types::I16 => {
+                let data = p.match_constant_data(ty).unwrap();
```
bjorn3: match_uimm16

cranelift-filetests/src/function_runner.rs
```diff
@@ -1,53 +1,280 @@
...
+            types::I64 => {
+                let data = p
+                    .match_uimm64("Expected a 64-bit unsigned integer")
+                    .unwrap();
+                Self::I64(data.into())
+            }
+            types::I128 => {
+                let data = p.match_constant_data(ty).unwrap();
+                let data = ConstantDataU128::try_into(data).unwrap();
+                Self::I128(data.rotate_left(64))
```
bjorn3: ???

cranelift-filetests/src/function_runner.rs
```diff
@@ -1,53 +1,280 @@
...
+            action,
+        }
     }
 
-    /// Build a function runner using the host machine's ISA and the passed flags
-    pub fn with_host_isa(function: Function, flags: settings::Flags) -> Self {
-        let builder = host_isa_builder().expect("Unable to build a TargetIsa for the current host");
-        let isa = builder.finish(flags);
-        Self::new(function, isa)
+    fn invoke<T>(code_page: &Mmap) -> T {
```
bjorn3: unsafe

cranelift-filetests/src/function_runner.rs
```diff
@@ -1,53 +1,280 @@
...
+    {
+        let ok = match operator {
+            Operator::EQ => value == expected,
+            Operator::NEQ => value != expected,
+        };
+
+        if ok {
+            return Ok(());
+        }
+        return Err(format!(
```
bjorn3: Unnecessary `return`.

cranelift-filetests/src/function_runner.rs
```diff
@@ -1,53 +1,280 @@
...
+        expected: Value,
+    ) -> Result<(), String> {
+        macro_rules! gen_match {
+            ($x:ident, $($y: path => $z: ty, $t: path), +) => {
+                match $x {
+                    $($y => {
+                        if let $t(v) = expected {
+                            Self::check_assertion(Self::invoke::<$z>(&code_page), operator, v)
+                        } else {
+                            panic!()
```
bjorn3: Please use a descriptive message.

cranelift-filetests/src/function_runner.rs
```diff
@@ -76,15 +303,159 @@ impl FunctionRunner {
...
+                Self::check(code_page, ret_value_type, *operator, *expected)
+            }
+        }
+    }
+}
+
+fn return_type(func: &Function) -> Result<types::Type, String> {
+    if !(func.signature.params.is_empty() && func.signature.returns.len() == 1) {
+        return Err(format!(
+            "Functions must have a signature like: () -> ty, found {} for {}",
```
bjorn3: Maybe "function {} must have ..."?

bjorn3 (changes requested):

cranelift-filetests/src/function_runner.rs
krk: I have removed it instead.

cranelift-filetests/src/function_runner.rs
krk: I cannot find it, also there is no `Uimm16` type. Should I create them for this PR?

cranelift-filetests/src/function_runner.rs
krk: Swaps left and right qwords, if causes more problems than it solves, I can remove I128 support from the PR.

krk:
ping @abrown @bnjbvr @bjorn3

cranelift-filetests/src/function_runner.rs
bjorn3: Makes sense

cranelift-filetests/src/function_runner.rs
bjorn3: `ConstantData` is stored in little endian format. `ConstantDataU128` converts it to a native endian `u128`, so no swap should be necessary.

bnjbvr:
I'm back from vacation, I'll try to take a look at it this week.

cranelift-filetests/src/function_runner.rs
```diff
@@ -76,17 +298,161 @@ impl FunctionRunner {
...
+        if comment == "print" {
+            // Print test return value, if return type is not b1, always fail the test.
+            return FunctionRunnerAction::Print;
+        }
+
+        // ";   run:  %fn() > 42 " is now "%fn() > 42".
+        let mut parts: Vec<&str> = comment.split_ascii_whitespace().collect();
+
+        if parts.len() < 3 {
+            panic!("Invalid syntax for test assertion, expected \"%fn() <operator> <value>\", found \"{}\"", comment);
```
fitzgen: We shouldn't panic on parse errors, we should return a fallible `ParseResult` like other parsing functions do.

Additionally, it seems a little strange that we have this parsing code in the `cranelift-filetests` crate, when AFAICT all the other parsing code is in the `cranelift-reader` crate. Unless there is a good reason not to, we should match existing conventions and keep the parsing code inside `cranelift-reader`.

cranelift-reader/src/parser.rs
```diff
@@ -609,8 +610,8 @@ impl<'a> Parser<'a> {
         }
     }
 
-    // Match and consume either a hexadecimal Uimm128 immediate (e.g. 0x000102...) or its literal list form (e.g. [0 1 2...])
-    fn match_constant_data(&mut self, controlling_type: Type) -> ParseResult<ConstantData> {
+    /// Match and consume either a hexadecimal Uimm128 immediate (e.g. 0x000102...) or its literal list form (e.g. [0 1 2...])
+    pub fn match_constant_data(&mut self, controlling_type: Type) -> ParseResult<ConstantData> {
```
fitzgen: I believe that moving that parsing code into this `cranelift-reader` crate would let us avoid making all these methods public as well.

cranelift-filetests/src/function_runner.rs
fitzgen: Was this still going to happen?

cranelift-filetests/src/function_runner.rs
fitzgen: It isn't clear to me what the resolution is here, if any.

cranelift-filetests/src/function_runner.rs
```diff
@@ -1,53 +1,275 @@
...
+            _ => Err(format!("Unsupported operator {}", s)),
+        }
+    }
+}
+
+macro_rules! try_into_int {
+    ($type: ty, $name: ident) => {
+        struct $name {}
+        impl $name {
+            fn try_into(d: ConstantData) -> Result<$type, String> {
```
fitzgen: Instead of making an ad-hoc `try_into` method, we should follow convention and implement the `TryFrom` trait, which gives us a `try_into` via the blanket impl `TryInto<T> for U` where `U: TryFrom<T>`. For details, see: https://doc.rust-lang.org/stable/std/convert/trait.TryInto.html and https://doc.rust-lang.org/stable/std/convert/trait.TryFrom.html

If I'm reading this code right, it would be something like this:

```rust
struct $name(ConstantData);

impl TryFrom<$name> for $type {
    // ...
}
```

But this also makes me wonder why we don't implement `TryFrom<ConstantData> for $type` directly instead of creating all these temporary `ConstantData$Type` types. Is that something that you explored at all? I think it would be the best course to take, unless it doesn't work for some reason or I am overlooking something else.

fitzgen (commented):
Hi @krk -- thanks for this PR!

I'm not *super* familiar with this area of code, but since Benjamin tagged me for review (I assume because he's got a lot on his plate at the moment) I've taken a look and left a few comments.

Overall, this looks good, but I think there are a few things we could clean up a bit before landing, and it seems like there are still some unresolved discussions from earlier rounds of review that we should figure out before merging this as well.

If any of my comments are unclear, please let me know! I'm happy to clarify.

Cheers!

alexcrichton:

krk:
Sorry for abandoning this PR. 

It seems that functionality is being added: https://github.com/bytecodealliance/wasmtime/pull/1223

Issue #1192, opened 2019-11-01
Too many raw_bitcasts in SIMD code


abrown:
- What is the feature or code improvement you would like to do in Cranelift?

During translation from Wasm to CLIF, a combination of Wasm's `v128` type and Cranelift's current type system forces us to add many `raw_bitcast` instructions between operations. For example, this Wasm code:
```
  (func (export "add-sub") (param v128 v128 v128) (result v128)
    (i16x8.add (i16x8.sub (local.get 0) (local.get 1))(local.get 2)))
```
Translates to this CLIF code:
```
function u0:4(i64 vmctx [%rdi], i8x16 [%xmm0], i8x16 [%xmm1], i64 fp [%rbp]) -> i8x16 [%xmm0], i64 fp [%rbp] system_v {
    ss0 = incoming_arg 16, offset -16

                                ebb0(v0: i64 [%rdi], v1: i8x16 [%xmm0], v2: i8x16 [%xmm1], v12: i64 [%rbp]):
[RexOp1pushq#50]                    x86_push v12
[RexOp1copysp#8089]                 copy_special %rsp -> %rbp
@00a6 [null_fpr#00,%xmm0]           v4 = raw_bitcast.i16x8 v1
@00a6 [Mp2vconst_optimized#5ef,%xmm2] v11 = vconst.i16x8 0x00
@00a6 [Mp2fa#5f9,%xmm2]             v5 = isub v11, v4
@00a6 [null_fpr#00,%xmm2]           v6 = raw_bitcast.i8x16 v5
@00aa [null_fpr#00,%xmm2]           v7 = raw_bitcast.i16x8 v6
@00aa [null_fpr#00,%xmm1]           v8 = raw_bitcast.i16x8 v2
@00aa [Mp2fa#5fd,%xmm2]             v9 = iadd v7, v8
@00aa [null_fpr#00,%xmm2]           v10 = raw_bitcast.i8x16 v9
@00ac [-]                           fallthrough ebb1(v10)

                                ebb1(v3: i8x16 [%xmm2]):
@00ac [Op2frmov#428]                regmove v3, %xmm2 -> %xmm0
[RexOp1popq#58,%rbp]                v13 = x86_pop.i64 
@00ac [Op1ret#c3]                   return v3, v13
}
```

This issue is to discuss if and how to remove these extra bitcasts.

- What is the value of adding this in Cranelift?

[...]

PR #1236, opened 2019-11-16
Relax verification to allow I8X16 to act as a default vector type

abrown:
- [x] This has been discussed in issue #1192.

This change uses I8X16 as a default type to represent Wasm's V128 type. Values of Wasm's V128 type can be used in any instruction of similar width (e.g. `i32x4.add`). Cranelift must build signatures for Wasm functions using V128 and, lacking a special type to do so, Cranelift uses I8X16 for V128. To avoid excessive bitcasting, however, we relax the verifier to allow I8X16 to type-check as the other 128-bit types. This implies that:
 - SIMD instructions must have explicit types (e.g. iadd.i32x4) to avoid type inference deciding a wrong type and emitting the wrong instruction
 - The verifier will be unable to catch unintended errors in sequences like: `v1 = iadd.i8x16 v0, v0; v2 = isub.i32x4 v1, v0` (presumably we want `v0` and `v1` to be `I8X16` in fact and would want the verifier to tell us that we should not add them as `I32X4`s).

With this change, we still might choose to bitcast between SIMD instructions. Consider the Wasm snippet  `i64x2.add (f32x4.sub (local.get 0) (local.get 1))` for the following three options:
 - we retain all the current bitcasts between SIMD instructions when translating from Wasm; this means that the Wasm snippet is valid and compilable
 - we remove all of the current bitcasts between SIMD instructions when translating from Wasm; this makes the Wasm snippet invalid and compilation would fail with a verifier error
 - we retain the bitcasts but make this a configurable feature; in strict mode the snippet above would fail to compile but in relaxed mode it would succeed.

### DIFF
--- a/cranelift-codegen/src/verifier/mod.rs
+++ b/cranelift-codegen/src/verifier/mod.rs
@@ -64,6 +64,7 @@ use crate::flowgraph::{BasicBlock, ControlFlowGraph};
 use crate::ir;
 use crate::ir::entities::AnyEntity;
 use crate::ir::instructions::{BranchInfo, CallInfo, InstructionFormat, ResolvedConstraint};
+use crate::ir::types::I8X16;
 use crate::ir::{
     types, ArgumentLoc, Ebb, FuncRef, Function, GlobalValue, Inst, InstructionData, JumpTable,
     Opcode, SigRef, StackSlot, StackSlotKind, Type, Value, ValueDef, ValueList, ValueLoc,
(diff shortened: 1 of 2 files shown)


bjorn3:
Can we instead have a separate type for this? (`v128`?)

abrown:
> Can we instead have a separate type for this? (v128?)

I considered that but as I looked into it it seemed to me that the type would be a weird one: it would have to be able to be coerced into any 128-bit wide vector AND every 128-bit wide vector would have to be able to coerce itself into such a `v128`. And I'm not sure I understand all the changes necessary to `Type`, `TypeSet`, `ValueType`, `ValueTypeSet` (am I forgetting some?) to do this. If you do, and it is simple, I think that would be a less hidden approach than the verifier relaxation.

abrown:
@bnjbvr, I'm going to close this in favor of #1272 and (possibly, eventually) #1251.

PR #1249, opened 2019-11-20
Remove macros from verifier; fixes #1248

### LINKED ISSUE #1248: Remove error macros in the verifier
- What is the feature or code improvement you would like to do in Cranelift?

Remove `report!`, `fatal!`, and `nonfatal!` macros from the verifier (`cranelift-codegen/src/verifier/mod.rs`).

- What is the value of adding this in Cranelift?

Macros are hard to troubleshoot and sometimes, as in #1226, we need the flexibility to change what is reported on the error.

- Do you have an implementation plan, and/or ideas for data structures or
  algorithms to use?

I propose we remove the `report!`, `fatal!`, and `nonfatal!` macros and replace them with methods on `VerifierErrors`. The verifier code already threads a `VerifierErrors` object through all of the verifier methods so we could make better use of that object.

One good thing about the macros was that they allowed variadic arguments, e.g. `fatal!(errors, inst, "there was an error in {}", ebb)`. Moving to methods on `VerifierErrors` prevents this shorthand but I believe the alternative is not terrible: `errors.fatal(inst, format!("there was an error in {}", ebb))`. We can discuss this further in the comments but I would suggest that, though slightly more verbose, makes it easier to work with verifier errors.

- Have you considered alternative implementations? If so, how are they better
  or worse than your proposal?

We could leave the code as-is.

abrown:
This removes `report!`, `fatal!`, and `nonfatal!` from the verifier code and replaces them with methods on `VerifierErrors`. In order to maintain similar ease-of-use, `VerifierError` is expanded with several `From` implementations that convert a tuple to a verifier error.

- [x] This has been discussed in issue #1248.
- [x] A short description of what this does, why it is needed. _See above_.

### DIFF
--- a/cranelift-codegen/src/verifier/mod.rs
+++ b/cranelift-codegen/src/verifier/mod.rs
@@ -85,46 +85,6 @@ pub use self::cssa::verify_cssa;
 pub use self::liveness::verify_liveness;
 pub use self::locations::verify_locations;
 
-/// Report an error.
-///
-/// The first argument must be a `&mut VerifierErrors` reference, and the following
-/// argument defines the location of the error and must implement `Into<AnyEntity>`.
-/// Finally, subsequent arguments will be formatted using `format!()` and set
-/// as the error message.
-macro_rules! report {
-    ( $errors: expr, $loc: expr, $msg: tt ) => {
-        $errors.0.push(crate::verifier::VerifierError {
-            location: $loc.into(),
-            context: None,
-            message: String::from($msg),
-        })
-    };
-
-    ( $errors: expr, $loc: expr, $fmt: tt, $( $arg: expr ),+ ) => {
-        $errors.0.push(crate::verifier::VerifierError {
-            location: $loc.into(),
-            context: None,
-            message: format!( $fmt, $( $arg ),+ ),
-        })
-    };
-}
-
-/// Diagnose a fatal error, and return `Err`.
-macro_rules! fatal {
-    ( $( $arg: expr ),+ ) => ({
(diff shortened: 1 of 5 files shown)


cranelift-codegen/src/verifier/mod.rs
```diff
@@ -202,6 +206,23 @@ impl VerifierErrors {
...
+    pub fn fatal(&mut self, error: impl Into<VerifierError>) -> VerifierStepResult<()> {
+        self.report(error);
+        Err(())
+    }
+
+    /// Report a non-fatal error and return `Ok`.
+    pub fn nonfatal(&mut self, error: impl Into<VerifierError>) -> VerifierStepResult<()> {
+        self.report(error);
+        Ok(())
+    }
```
abrown: This is what is replacing the macros.

cranelift-codegen/src/verifier/mod.rs
```diff
@@ -151,6 +111,50 @@ fn format_context(context: &Option<String>) -> String {
...
+    fn from(items: (L, M)) -> Self {
+        let (location, message) = items;
+        Self {
+            location: location.into(),
+            context: None,
+            message: message.into(),
+        }
+    }
+}
+
```
abrown: These convenience implementations make it possible to be a bit less verbose... I was trying to make this as easy-to-use and concise as the macros and this was as close as I could get.

abrown:
@bnjbvr, I annotated the two spots with the fundamental changes of this PR; all of the rest should be grep substitutions.

PR #1251, opened 2019-11-21
Add X128 type to represent WebAssembly's V128 type

abrown:
WebAssembly has a generic 128-bit wide type that can be used in any SIMD operation (e.g., `i32x4.add`). Cranelift's type system is more strict (i.e., only an I32X4 value can be used for an `i32x4.add`). To avoid many bitcasts when translating WebAssembly's SIMD operations, we add the X128 type which acts as a super-type for all 128-bit wide types.

- [x] This has been discussed in a previous issue:

The original problem is described at #1192 and an alternate implementation at #1236 (it relaxes the verifier and uses `I8X16` as a default stand-in type for `V128`).

This adds a new type `X128` to act as a super type for 128-bit wide vector types.

### DIFF
--- a/cranelift-codegen/src/ir/types.rs
+++ b/cranelift-codegen/src/ir/types.rs
@@ -31,6 +31,22 @@ pub const INVALID: Type = Type(0);
 // 512-bit SIMD vectors.
 include!(concat!(env!("OUT_DIR"), "/types.rs"));
 
+/// A V128 type--a supertype of any vector of 128-bit length.
+///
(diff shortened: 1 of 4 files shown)


abrown:
cc @bjorn3; as a heads up, @bnjbvr, this does not include any of the configuration settings we discussed. I can add those later once we decide if this is the approach to take or if we prefer #1236.

abrown:
@bnjbvr, I can't reset this to a WIP PR so please ignore it for review; I'm going to be working on #1272 in the short term.

bnjbvr:
Sounds good! Unassigning me from review then.

PR #1253, opened 2019-11-23
More i8 legalizations

### LINKED ISSUE #433: Missing encodings: bint.i8, stack_store for i8
```
function u0:1(i64 [%rdi]) fast {
                                ebb0(v0: i64):
                                    v4 = iconst.i32 0
                                    v1 = ireduce.i8 v4
                                    v2 = icmp_imm.i8 eq v1, 0
[-]                                 v3 = bint.i8 v2
                                    ^~~~~~ verifier inst2: v3 is a real GPR value defined by a ghost instruction

                                    v5 = uextend.i32 v3
                                    istore8.i32 v5, v0
                                    return
}
```

Not important for me (just found it while reducing above case):

`stack_store` for i8

```
                                    v8 = ireduce.i8 v17
[-]                                 stack_store v8, ss3
                                    ^~~~~~ verifier inst8: Store must have an encoding
```

bjorn3:
  first.
  This adds several legalizations for small integers.

Fixes #433.
cc #466 and #1117

### DIFF
--- a/cranelift-codegen/meta/src/shared/legalize.rs
+++ b/cranelift-codegen/meta/src/shared/legalize.rs
@@ -139,6 +139,7 @@ pub(crate) fn define(insts: &InstructionGroup, imm: &Immediates) -> TransformGro
     expand.custom_legalize(trapnz, "expand_cond_trap");
     expand.custom_legalize(br_table, "expand_br_table");
     expand.custom_legalize(select, "expand_select");
+    widen.custom_legalize(select, "expand_select"); // small ints
 
     // Custom expansions for floating point constants.
     // These expansions require bit-casting or creating constant pool entries.
--- a/cranelift-codegen/meta/src/shared/legalize.rs
+++ b/cranelift-codegen/meta/src/shared/legalize.rs
@@ -149,6 +149,10 @@ pub(crate) fn define(insts: &InstructionGroup, imm: &Immediates) -> TransformGro
     expand.custom_legalize(insts.by_name("stack_load"), "expand_stack_load");
     expand.custom_legalize(insts.by_name("stack_store"), "expand_stack_store");
 
+    // Custom expansions for small stack memory acccess.
+    widen.custom_legalize(insts.by_name("stack_load"), "expand_stack_load");
+    widen.custom_legalize(insts.by_name("stack_store"), "expand_stack_store");
+
     // List of variables to reuse in patterns.
     let x = var("x");
     let y = var("y");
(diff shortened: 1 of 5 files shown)


filetests/isa/x86/stack-load-store8.clif
```diff
@@ -0,0 +1,19 @@
+test compile
```
abrown: Should this be `test legalizer`?

abrown:
I think this PR makes sense to me but I'm not confident about how the legalization groups work (e.g. `widen`): @bnjbvr, any issues with this or should I approve and merge?

abrown:
(note to self: if we do merge we should probably "squash and merge")

bnjbvr:
I think @julian-seward1 proposed to give a look at this PR; I'll let the two of you coordinate for reviews.

filetests/isa/x86/stack-load-store8.clif
bjorn3: I prefer `test compile` as it ensures that that all instructions after legalize do actually have encodings. For example https://github.com/bytecodealliance/cranelift/pull/1222 had a `test legalize` test which didn't actually compile.

abrown: Sounds good to me.

julian-seward1:
I have to say, I don't understand the legalisation machinery enough to comment meaningfully here.  I would however say that IMO we should try to avoid generating operations that read or write 16 bit registers on x86_{32,64} since that puts us at risk of partial register stalls.

bjorn3:
This makes some existing "macro" legalizations apply to small ints too. It also uses uextend to legalize brz.i8 to brnz.i32. There should be a faster way for this, but that can be implemented later.

cranelift-codegen/meta/src/shared/legalize.rs
```diff
@@ -612,6 +617,18 @@ pub(crate) fn define(insts: &InstructionGroup, imm: &Immediates) -> TransformGro
...
+        widen.legalize(
+            def!(brz.ty(x, ebb, vararg)),
+            vec![def!(a = uextend.I32(x)), def!(brz(a, ebb, vararg))],
+        );
+
+        widen.legalize(
+            def!(brnz.ty(x, ebb, vararg)),
+            vec![def!(a = uextend.I32(x)), def!(brnz(a, ebb, vararg))],
+        );
+    }
```
abrown: @julian-seward1, if this was what you were concerned about when you mentioned "partial register stalls," I think that is avoided here because of `uextend` which maps to a `MOV` with zero-extension. If you were concerned about using 16-bit registers on x86 in general, then it seems that is a bigger issue (should Cranelift then have an `I16` type?).

Issue #1176, opened 2019-10-28
Load/store R32/R64 types


yurydelendik:
Most likely global.set/.get or table.set/.get for reference types will be translated into regular memory load and store instructions. Currently, their implementations are missing at codegen side. The https://github.com/CraneStation/cranelift/pull/1073 performs the translation of the wasm, but CL validator fails when ref types and globals/tables operations are present (see e.g. [example](https://github.com/WebAssembly/wasm-c-api/blob/master/example/hostref.wat))

fitzgen:
The question is how to deal with barriers for e.g. generational gc or reference counting.

@sunfishcode and I had talked about how to handle this and one idea would be to have a trait method in `cranelift_wasm::FuncEnvironment` for translating table get/set to some snippet of IR that includes the barriers. This would mean that if one wasn't using `cranelift-wasm`, one would have to figure out their own method of introducing the barriers. This approach is flexible and maintains cranelift's current property of not calling back into the embedder when compiling IR to machine code.

Another variant of this is to introduce something akin to `FuncEnvironment` to `cranelift` proper, rather than only at the wasm translation layer. This would mean we are introducing embedder callbacks into compilation, and that seems like something that we would want to be careful about. But the conceptual advantage is that barriers are pretty low-level, and right now, after wasm translation the IR is very high level and then legalization passes bring it lower and lower level, so conceptually it would be congruent to legalize high-level reference type loads and stores into low-level loads and stores with barriers in the same phase.

The other idea we were oscillating back to was supporting just the three specific use cases we know we want to support: naive refcounting, spidermonkey gc, and a wasmtime gc. This avoids the full flexibility of the previous options and their bit of "architecture astronautics". I think this ran against the grain of some constraints that the .NET embedding has (+cc @peterhuene; or was it the Wasm C API?). I don't quite remember the details however. I vaguely remember coming away thinking that, yes, we would likely be better off with the fully general solution.

bnjbvr:
> cranelift_wasm::FuncEnvironment for translating table get/set to some snippet of IR that includes the barriers.

For what it's worth, that's the way I envisioned it too; it would probably be sufficient for Spidermonkey.

peterhuene:
[...]

PR #1258, opened 2019-11-26
Add hooks for implementing bulk-memory-operations

eqrion:
I have patches that add support for bulk-memory-operations in SpiderMonkey with Cranelift. These commits are the prerequisites for those patches.

The only major change is adding extensibility points to cranelift-wasm for translating bulk-memory-operations. These will be used in SM to translate the operations into function calls which will do the actual work.

### DIFF
--- a/cranelift-wasm/src/code_translator.rs
+++ b/cranelift-wasm/src/code_translator.rs
@@ -988,7 +988,9 @@ pub fn translate_operator<FE: FuncEnvironment + ?Sized>(
             return Err(wasm_unsupported!("proposed thread operator {:?}", op));
         }
         Operator::MemoryCopy => {
-            // The WebAssembly MVP only supports one linear memory and we assume it here.
+            // The WebAssembly MVP only supports one linear memory and
+            // wasmparser will ensure that the memory indices specified are
+            // zero.
             let heap_index = MemoryIndex::from_u32(0);
             let heap = state.get_heap(builder.func, 0, environ)?;
             let len = state.pop1();
@@ -997,7 +999,9 @@ pub fn translate_operator<FE: FuncEnvironment + ?Sized>(
             environ.translate_memory_copy(builder.cursor(), heap_index, heap, dest, src, len)?;
         }
         Operator::MemoryFill => {
-            // The WebAssembly MVP only supports one linear memory and we assume it here.
+            // The WebAssembly MVP only supports one linear memory and
+            // wasmparser will ensure that the memory index specified is
+            // zero.
             let heap_index = MemoryIndex::from_u32(0);
             let heap = state.get_heap(builder.func, 0, environ)?;
             let len = state.pop1();
@@ -1006,7 +1010,9 @@ pub fn translate_operator<FE: FuncEnvironment + ?Sized>(
             environ.translate_memory_fill(builder.cursor(), heap_index, heap, dest, val, len)?;
         }
         Operator::MemoryInit { segment } => {
-            // The WebAssembly MVP only supports one linear memory and we assume it here.
+            // The WebAssembly MVP only supports one linear memory and
+            // wasmparser will ensure that the memory index specified is
+            // zero.
             let heap_index = MemoryIndex::from_u32(0);
             let heap = state.get_heap(builder.func, 0, environ)?;
             let len = state.pop1();
@@ -1025,20 +1031,30 @@ pub fn translate_operator<FE: FuncEnvironment + ?Sized>(
         Operator::DataDrop { segment } => {
             environ.translate_data_drop(builder.cursor(), *segment)?;
         }
-        Operator::TableSize { table } => {
-            state.push1(environ.translate_table_size(builder.cursor(), *table)?);
+        Operator::TableSize { table: index } => {
+            let table = state.get_table(builder.func, *index, environ)?;
+            state.push1(environ.translate_table_size(
+                builder.cursor(),
+                TableIndex::from_u32(*index),
+                table,
+            )?);
         }
         Operator::TableCopy => {
-            // The WebAssembly MVP only supports one table and we assume it here.
-            let dest_index = 0;
-            let src_index = 0;
+            // The WebAssembly MVP only supports one table and wasmparser will
+            // ensure that the table index specified is zero.
+            let dst_table_index = 0;
+            let dst_table = state.get_table(builder.func, dst_table_index, environ)?;
+            let src_table_index = 0;
+            let src_table = state.get_table(builder.func, src_table_index, environ)?;
             let len = state.pop1();
             let src = state.pop1();
             let dest = state.pop1();
             environ.translate_table_copy(
                 builder.cursor(),
-                dest_index,
-                src_index,
+                TableIndex::from_u32(dst_table_index),
+                dst_table,
+                TableIndex::from_u32(src_table_index),
+                src_table,
                 dest,
                 src,
                 len,
@@ -1047,13 +1063,15 @@ pub fn translate_operator<FE: FuncEnvironment + ?Sized>(
         Operator::TableInit { segment } => {
             // The WebAssembly MVP only supports one table and we assume it here.
             let table_index = 0;
+            let table = state.get_table(builder.func, table_index, environ)?;
             let len = state.pop1();
             let src = state.pop1();
             let dest = state.pop1();
             environ.translate_table_init(
                 builder.cursor(),
                 *segment,
-                table_index,
+                TableIndex::from_u32(table_index),
+                table,
                 dest,
                 src,
                 len,
--- a/cranelift-wasm/src/code_translator.rs
+++ b/cranelift-wasm/src/code_translator.rs
@@ -987,18 +987,86 @@ pub fn translate_operator<FE: FuncEnvironment + ?Sized>(
         | Operator::Fence { .. } => {
             return Err(wasm_unsupported!("proposed thread operator {:?}", op));
         }
-        Operator::MemoryInit { .. }
-        | Operator::DataDrop { .. }
-        | Operator::MemoryCopy
-        | Operator::MemoryFill
-        | Operator::TableInit { .. }
-        | Operator::ElemDrop { .. }
-        | Operator::TableCopy
-        | Operator::TableGet { .. }
-        | Operator::TableSet { .. }
-        | Operator::TableGrow { .. }
-        | Operator::TableSize { .. } => {
-            return Err(wasm_unsupported!("proposed bulk memory operator {:?}", op));
+        Operator::MemoryCopy => {
+            // The WebAssembly MVP only supports one linear memory and we assume it here.
+            let heap_index = MemoryIndex::from_u32(0);
+            let heap = state.get_heap(builder.func, 0, environ)?;
+            let len = state.pop1();
+            let src = state.pop1();
+            let dest = state.pop1();
+            environ.translate_memory_copy(builder.cursor(), heap_index, heap, dest, src, len)?;
+        }
+        Operator::MemoryFill => {
+            // The WebAssembly MVP only supports one linear memory and we assume it here.
+            let heap_index = MemoryIndex::from_u32(0);
+            let heap = state.get_heap(builder.func, 0, environ)?;
+            let len = state.pop1();
+            let val = state.pop1();
+            let dest = state.pop1();
+            environ.translate_memory_fill(builder.cursor(), heap_index, heap, dest, val, len)?;
+        }
+        Operator::MemoryInit { segment } => {
+            // The WebAssembly MVP only supports one linear memory and we assume it here.
+            let heap_index = MemoryIndex::from_u32(0);
+            let heap = state.get_heap(builder.func, 0, environ)?;
+            let len = state.pop1();
+            let src = state.pop1();
+            let dest = state.pop1();
+            environ.translate_memory_init(
+                builder.cursor(),
+                heap_index,
+                heap,
+                *segment,
+                dest,
+                src,
+                len,
+            )?;
+        }
+        Operator::DataDrop { segment } => {
+            environ.translate_data_drop(builder.cursor(), *segment)?;
+        }
+        Operator::TableSize { table } => {
+            state.push1(environ.translate_table_size(builder.cursor(), *table)?);
+        }
+        Operator::TableCopy => {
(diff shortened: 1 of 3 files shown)


cranelift-codegen/src/ir/dfg.rs
```diff
@@ -660,6 +660,17 @@ impl DataFlowGraph {
         self.insts[inst].put_value_list(branch_values)
     }
 
+    /// Append new value arguments to an instruction.
+    ///
+    /// Panics if the instruction doesn't support arguments.
+    pub fn append_inst_args(&mut self, inst: Inst, new_args: &[Value]) {
```
bjorn3: You could use `func.dfg[inst]` instead of `func.dfg.insts[inst]`. That way you can avoid adding this method.

cranelift-wasm/src/code_translator.rs
```diff
@@ -987,18 +987,86 @@ pub fn translate_operator<FE: FuncEnvironment + ?Sized>(
...
+                dest,
+                src,
+                len,
+            )?;
+        }
+        Operator::DataDrop { segment } => {
+            environ.translate_data_drop(builder.cursor(), *segment)?;
+        }
+        Operator::TableSize { table } => {
+            state.push1(environ.translate_table_size(builder.cursor(), *table)?);
```
bjorn3: You could use the following instead:

```suggestion
            let table_data = &builder.cursor().func.tables[*table];
            let base = builder.cursor().ins().global_value(table_data.index_type, table_data.base_gv);
            let bound = builder.cursor().ins().global_value(table_data.index_type, table_data.bound_gv);
            state.push1(builder.cursor().ins().isub(bound, base));
```

Maybe even add an instruction for it, just like `table_addr`? (legalized at https://github.com/bytecodealliance/cranelift/blob/3317fc623c315073710fc1ed67c2d17d2466ab5d/cranelift-codegen/src/legalizer/table.rs#L14)

cranelift-wasm/src/code_translator.rs
```diff
@@ -987,18 +987,86 @@ pub fn translate_operator<FE: FuncEnvironment + ?Sized>(
...
+            environ.translate_memory_copy(builder.cursor(), heap_index, heap, dest, src, len)?;
+        }
+        Operator::MemoryFill => {
+            // The WebAssembly MVP only supports one linear memory and we assume it here.
+            let heap_index = MemoryIndex::from_u32(0);
+            let heap = state.get_heap(builder.func, 0, environ)?;
+            let len = state.pop1();
+            let val = state.pop1();
+            let dest = state.pop1();
+            environ.translate_memory_fill(builder.cursor(), heap_index, heap, dest, val, len)?;
```
bjorn3: This can use [`call_memset`](https://docs.rs/cranelift-frontend/0.51.0/cranelift_frontend/struct.FunctionBuilder.html#method.call_memset) / [`emit_small_memset`](https://docs.rs/cranelift-frontend/0.51.0/cranelift_frontend/struct.FunctionBuilder.html#method.emit_small_memset), right?

cranelift-wasm/src/code_translator.rs
```diff
@@ -987,18 +987,86 @@ pub fn translate_operator<FE: FuncEnvironment + ?Sized>(
...
+            let src = state.pop1();
+            let dest = state.pop1();
+            environ.translate_table_copy(
+                builder.cursor(),
+                dest_index,
+                src_index,
+                dest,
+                src,
+                len,
+            )?;
```
bjorn3: This can use `table_addr` combined with [`call_memset`](https://docs.rs/cranelift-frontend/0.51.0/cranelift_frontend/struct.FunctionBuilder.html#method.call_memset) / [`emit_small_memset`](https://docs.rs/cranelift-frontend/0.51.0/cranelift_frontend/struct.FunctionBuilder.html#method.emit_small_memset), right?

cranelift-wasm/src/code_translator.rs
```diff
@@ -987,18 +987,86 @@ pub fn translate_operator<FE: FuncEnvironment + ?Sized>(
...
+        }
+        Operator::ElemDrop { segment } => {
+            environ.translate_elem_drop(builder.cursor(), *segment)?;
+        }
+        Operator::TableGet { .. } | Operator::TableSet { .. } | Operator::TableGrow { .. } => {
+            return Err(wasm_unsupported!(
```
bjorn3: `table_addr` + `load.r32/64 for `table_get` and `table_addr` + `store.r32/64 for `table_set`?

cranelift-wasm/src/code_translator.rs
```diff
@@ -987,18 +987,86 @@ pub fn translate_operator<FE: FuncEnvironment + ?Sized>(
...
-        | Operator::TableSize { .. } => {
-            return Err(wasm_unsupported!("proposed bulk memory operator {:?}", op));
+        Operator::MemoryCopy => {
+            // The WebAssembly MVP only supports one linear memory and we assume it here.
+            let heap_index = MemoryIndex::from_u32(0);
+            let heap = state.get_heap(builder.func, 0, environ)?;
+            let len = state.pop1();
+            let src = state.pop1();
+            let dest = state.pop1();
+            environ.translate_memory_copy(builder.cursor(), heap_index, heap, dest, src, len)?;
```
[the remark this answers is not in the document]
eqrion: Ah, I wasn't aware of those. There are two complications here however.

The first is that the spec requires us to not write anything if the source or destination ranges are OOBs. We can't use the normal bounds checking code because it relies on signal handling to catch accesses that are above `wasm memory.size` but below `cl heap.bound_gv`. Our signal handling won't work inside of the libc::memmove because it ignores faults outside of wasm code. We also wouldn't have a guarantee that no writing would be performed before a fault happened.

If we could access the canonical `wasm memory.size` from the JIT than might be able to do a bounds check against that. The difficulty here is that while the [`size`](https://searchfox.org/mozilla-central/rev/cce8b90aece0f42e5025e45282de16066eeaa662/js/src/wasm/WasmInstance.cpp#368) is at a fixed offset before the heap base, it depends on whether the memory is shared or not. And when the memory is shared, we need to perform an atomic access which doesn't appear to be implemented yet. (although I'm new and may have missed it)

The second complication is that, IIUC, memmove on shared memory is UB in the C/C++ sense. The [implementation](https://searchfox.org/mozilla-central/rev/cce8b90aece0f42e5025e45282de16066eeaa662/js/src/wasm/WasmInstance.cpp#508) in SM uses a JIT'ed version to proactively avoid any issues. Although the original author that introduced this said he might have been too paranoid, so this might not be an issue.

If there is a way around these, that would be great as it would allow other users to share the impl. However this seemed to easiest way to get interoperability with SM.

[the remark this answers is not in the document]
sunfishcode: I agree, it makes sense to hand this off to the `environ` here. The `environ` can then decide whether it wants to do `call_memmove` or similar (after checking the bounds).

cranelift-codegen/src/ir/trapcode.rs
```diff
@@ -52,6 +52,9 @@ pub enum TrapCode {
     /// This trap is resumable.
     Interrupt,
 
+    /// Signal an error that was reported in VM code.
+    ThrowReported,
```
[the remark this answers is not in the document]
eqrion: SM handles errors in VM functions by [reporting an error and returning a sentinel result value](https://searchfox.org/mozilla-central/rev/cce8b90aece0f42e5025e45282de16066eeaa662/js/src/wasm/WasmInstance.cpp#544), [then trapping in wasm code](https://searchfox.org/mozilla-central/rev/cce8b90aece0f42e5025e45282de16066eeaa662/js/src/jit/MacroAssembler.cpp#3111) if the sentinel was received. This trap code is used here to indicate that the reported error should be used for the trap reason.

I wasn't sure whether it made sense to use `User` here or not, but I don't think other embedders will need this so I'll switch to using `User`.

cranelift-codegen/src/ir/dfg.rs
eqrion: Nice, I'll drop this commit then.

cranelift-wasm/src/code_translator.rs
eqrion: I'll take a look at doing this.

cranelift-wasm/src/code_translator.rs
eqrion: This has the same issue as [above](https://github.com/bytecodealliance/cranelift/pull/1258#discussion_r351027233).

cranelift-wasm/src/code_translator.rs
eqrion: I'm not an expert with the representation of tables, but the implementation of `table.copy` is [non-trivial](https://searchfox.org/mozilla-central/rev/cce8b90aece0f42e5025e45282de16066eeaa662/js/src/wasm/WasmTable.cpp#259). There's some GC barriers, type conversions, and rooting going on.

cranelift-wasm/src/code_translator.rs
eqrion: I was actually hoping to get to these when I started on ref-types support in SM/CL, as they're technically under that spec.

One relevant issue seems to be #1176. Additionally, like with `table.copy` we seem to have [different code](https://searchfox.org/mozilla-central/rev/cce8b90aece0f42e5025e45282de16066eeaa662/js/src/wasm/WasmInstance.cpp#857) for `funcref` vs. `anyref`. I do want to get these operators implemented soon though.

cranelift-wasm/src/code_translator.rs
eqrion: It seems like this would work, however on further thought, I think that we should leave this to the environ as well for consistency. If we could implement most of the bulk-memory operators in an embedding neutral way then I wouldn't mind. But if this is the only one, it seems like there could be some confusion.

eqrion:
I dropped the two commits that weren't needed. Also I forgot that I should like to the associated [SpiderMonkey patch](https://phabricator.services.mozilla.com/D54778).

cranelift-wasm/src/environ/spec.rs
```diff
@@ -266,6 +266,81 @@ pub trait FuncEnvironment {
...
+    /// The `index` provided identifies the linear memory to query, and `heap` is the heap reference
+    /// returned by `make_heap` for the same index.
+    fn translate_memory_copy(
+        &mut self,
+        pos: FuncCursor,
+        index: MemoryIndex,
+        heap: ir::Heap,
+        dst: ir::Value,
+        src: ir::Value,
+        len: ir::Value,
```
bnjbvr: I wonder if in the interest of having a nice API, we could introduce a structure to pass parameters to these functions, so there's no doubt about the ordering (since Rust doesn't have named parameters). It might not be a big deal, since IDE auto-completion should be able to show the signature at any call site, and it would add some small overhead too.

julian-seward1: I see the attraction, but I think there are too many variants to make this worthwhile.  At least for these cases I'd guess we're pretty safe, in that any confusion of the roles of (here) `heap`, `dst`, `src` and `len` is highly likely to cause the generated code to crash during testing.

cranelift-wasm/src/code_translator.rs
```diff
@@ -987,18 +987,86 @@ pub fn translate_operator<FE: FuncEnvironment + ?Sized>(
...
-        | Operator::TableInit { .. }
-        | Operator::ElemDrop { .. }
-        | Operator::TableCopy
-        | Operator::TableGet { .. }
-        | Operator::TableSet { .. }
-        | Operator::TableGrow { .. }
-        | Operator::TableSize { .. } => {
-            return Err(wasm_unsupported!("proposed bulk memory operator {:?}", op));
+        Operator::MemoryCopy => {
+            // The WebAssembly MVP only supports one linear memory and we assume it here.
```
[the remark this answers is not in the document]
julian-seward1: Yeah, +1 for Benjamin's comment.  Or at least, ensure that we don't silently produce incorrect code as the result of this assumption.  Adding an assertion would cause the system to fail during compilation, which is ugly but effective.  Is it easy to instead cause a validation error to be propagated?  If if's really easy to do that, I'd vote for it instead.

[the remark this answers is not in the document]
eqrion: All operators's that have a reserved index in the binary format such as memory instructions (before multi-memory) and table instructions (before reftypes) have an error result in [wasmparser](https://github.com/bytecodealliance/wasmparser/blob/b3f6ed69039372f14ea940216bef43ee8f154f55/src/binary_reader.rs#L1274) if the index is not 0. So it should be safe to assume the index is 0 here. I'll update the comments to indicate this.

cranelift-wasm/src/environ/spec.rs
eqrion: I'm a bit torn here as well, but I think I agree that adding an argument struct for each of these methods would be too much overhead and the likelyhood of an issue escaping tests is minimal.

cranelift-wasm/src/code_translator.rs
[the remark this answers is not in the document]
bnjbvr: Would that mean that if `wasmparser` changed behavior, non-zero indexes could silently make their way here? If so, wouldn't that be something we'd like to guard against?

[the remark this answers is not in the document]
eqrion: Yes, I just started looking over the patch again and there's some improvements we need to make.

`MemoryCopy`, `MemoryFill`, and `MemoryInit` don't encode a memory index and won't until multi-memory is supported. Once that happens we'll be forced by the compiler to update all of these sites, as wasmparser will start providing it as an immediate in the Operator enum. I don't think it's likely that wasmparser will start allowing non-zero memory indices but not start passing them along in the Operator enum.

However one change needed is that the normal load/store instructions do call `get_heap(MemoryIndex(0))?` which ensures that there is a memory in this module. We need to do that for these instructions as well.

`TableCopy` and `TableInit` are the same as above. Except that after [PR](https://github.com/bytecodealliance/wasmparser/pull/151) wasmparser will start giving us indices. Once that happens, we can update `make_table` to ensure that we only use 0 if reference types is not implemented.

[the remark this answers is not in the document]
eqrion: I mispoke, it looks like the `MemoryCopy...` instructions do call `get_heap`. Just the table instructions need to be updated.

Issue #1261, opened 2019-12-01
cratelift-codegen no longer builds with no_std


IcyDefiance:
I've tracked down a few things that are preventing this from happening.

1. `packed_struct` requires std by default
2. There are a few uses of std in `cranelift-codegen` and `cranelift-codegen-shared` that can be replaced with core or alloc.
3. `thiserror` requires std, because the Error trait doesn't exist in core or alloc.

The first two are trivial to fix, and I'll create a PR with those changes in a second.

However, there are 11 uses of the `Display` and `Into<result::CodegenError>` traits that prevent me from placing `thiserror` behind a feature gate, and I'm not sure what to do about them.

<details>
<summary>Here are the 11 things that rely on those traits, as reported by rustc (click to expand)</summary>

    $ cargo build --no-default-features --features core
    Compiling cranelift-codegen v0.51.0 (/home/icydefiance/Documents/code/cranelift/cranelift-codegen)
    error[E0599]: no method named `to_string` found for type `verifier::VerifierError` in the current scope
    --> cranelift-codegen/src/print_errors.rs:218:36
        |
    218 |     writeln!(w, "; error: {}", err.to_string())?;
        |                                    ^^^^^^^^^ method not found in `verifier::VerifierError`
        | 
    ::: cranelift-codegen/src/verifier/mod.rs:138:1
        |
    138 | pub struct VerifierError {
        | ------------------------ method `to_string` not found for this
        |
        = help: items from traits can only be used if the trait is implemented and in scope
        = note: the following trait defines an item `to_string`, perhaps you need to implement it:
                candidate #1: `alloc::string::ToString`

    error[E0599]: no method named `to_string` found for type `result::CodegenError` in the current scope
    --> cranelift-codegen/src/print_errors.rs:227:13
        |
    227 |         err.to_string()
        |             ^^^^^^^^^ method not found in `result::CodegenError`
        | 
    ::: cranelift-codegen/src/result.rs:12:1
        |
    12  | pub enum CodegenError {
        | --------------------- method `to_string` not found for this
        |
        = help: items from traits can only be used if the trait is implemented and in scope
        = note: the following trait defines an item `to_string`, perhaps you need to implement it:
                candidate #1: `alloc::string::ToString`

    error[E0277]: `verifier::VerifierError` doesn't implement `core::fmt::Display`
    --> cranelift-codegen/src/verifier/mod.rs:235:33
        |
    235 |             writeln!(f, "- {}", err)?;
[...]

PR #1262, opened 2019-12-01
Partially fix no_std build for cratelift-codegen

IcyDefiance:
* Add a feature gate for the std feature in `packed_struct`
* Replace a few std references with core and alloc

See #1261 for more info. I'm intentionally avoiding the use of an issue keyword here, because this PR only fixes points 1 and 2 in that issue.

@bnjbvr The placeholder text says to ping you if I don't know who can review this, and it doesn't look like I can assign a reviewer at all, so here you go.

### DIFF
--- a/cranelift-codegen/shared/src/lib.rs
+++ b/cranelift-codegen/shared/src/lib.rs
@@ -19,6 +19,11 @@
         clippy::use_self
     )
 )]
+#![no_std]
+
+#[allow(unused_imports)] // #[macro_use] is required for no_std
+#[macro_use]
+extern crate alloc;
 
 pub mod condcodes;
 pub mod constant_hash;
--- a/cranelift-codegen/shared/src/condcodes.rs
+++ b/cranelift-codegen/shared/src/condcodes.rs
@@ -313,7 +313,7 @@ impl FromStr for FloatCC {
 #[cfg(test)]
 mod tests {
     use super::*;
-    use std::string::ToString;
+    use alloc::string::ToString;
 
     static INT_ALL: [IntCC; 12] = [
         IntCC::Equal,
--- a/cranelift-codegen/shared/src/constant_hash.rs
+++ b/cranelift-codegen/shared/src/constant_hash.rs
@@ -10,7 +10,8 @@
 //! This module provides build meta support for lookups in these tables, as well as the shared hash
 //! function used for probing.
 
-use std::iter;
+use alloc::vec::Vec;
+use core::iter;
 
 /// A primitive hash function for matching opcodes.
 pub fn simple_hash(s: &str) -> usize {
@@ -57,6 +58,7 @@ pub fn generate_table<'cont, T, I: iter::Iterator<Item = &'cont T>, H: Fn(&T) ->
 #[cfg(test)]
 mod tests {
     use super::{generate_table, simple_hash};
+    use alloc::string::ToString;
 
     #[test]
     fn basic() {
--- a/cranelift-codegen/shared/src/isa/x86/encoding_bits.rs
+++ b/cranelift-codegen/shared/src/isa/x86/encoding_bits.rs
@@ -1,6 +1,6 @@
 //! Provides a named interface to the `u16` Encoding bits.
 
-use std::ops::RangeInclusive;
+use core::ops::RangeInclusive;
 
 /// Named interface to the `u16` Encoding bits, representing an opcode.
 ///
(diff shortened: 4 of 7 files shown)


bnjbvr:
Thanks. Out of curiosity, does this mean that the cargo build step is done with std even in non-std builds? If not, doesn't this cause compilation errors in the codegen-shared crate?

IcyDefiance:
That was a fun question to investigate. You *can* use std during the build step (e.g. inside codegen-meta), but it turns out there's a catch.

If you enable a feature on a build-dependency, it will be enabled everywhere. (rust-lang/cargo#5730) For that reason, I had to add another feature gate to codegen-meta just to disable codegen-shared/std.

---

While doing that, I realized that I also forgot to disable codegen-shared/std inside codegen. Fixing that exposed a few more errors, so I fixed those and then added `#![no_std]` to codegen-shared for good measure.

cranelift-codegen/shared/src/lib.rs
```diff
@@ -19,6 +19,15 @@
...
 )]
+#![no_std]
+
+#[allow(unused_imports)] // #[macro_use] is required for no_std
+#[macro_use]
+extern crate alloc;
+
+#[cfg(feature = "std")]
+#[macro_use]
+extern crate std;
```
bjorn3: It should be possible to never use std in cranelift_codegen_shared, right?

cranelift-codegen/shared/Cargo.toml
```diff
@@ -9,5 +9,10 @@ readme = "README.md"
...
 
 [dependencies]
-packed_struct = "0.3"
-packed_struct_codegen = "0.3"
+packed_struct = { version =  "0.3", default-features = false }
+packed_struct_codegen = { version =  "0.3", default-features = false }
+
+[features]
+default = ["std"]
+std = ["packed_struct/std", "packed_struct_codegen/std"]
```
bjorn3: Are those features actually used?

cranelift-codegen/shared/src/lib.rs
IcyDefiance: Right now, yes. There's one unit test that uses std, and it can be replaced with alloc. I can get rid of that if you want; you'll just have to add it back if you write another unit test or some feature-gated code that requires it.

cranelift-codegen/shared/Cargo.toml
IcyDefiance: Nope, I just thought it might be better to have the full feature set available in an std environment. I can get rid of it once I'm done with work today.

IcyDefiance:
^ Those checks failed because github made a breaking change to the master branch of their checkout action. #1268 will fix it.

bnjbvr:
Hey @IcyDefiance can you please rebase? (we removed the "packed_struct" dependency recently.) Then I'll make sure this is triaged for review. Thanks!

IcyDefiance:
Cool, that simplifies this PR a bit. I don't need to modify any Cargo.toml files anymore. One more reference to std was added to codegen-shared since the last review, but that was easy to replace with core.

abrown:
@IcyDefiance, I recently saw https://github.com/bytecodealliance/wasmtime/pull/554 and wonder if the same applies here?

IcyDefiance:
Well, I hope not. I would like to make something similar to [nebulet](https://github.com/nebulet/nebulet) (an OS that runs entirely in ring 0, relying on a jitter for sandboxing), combined with a couple other crazy ideas, so if the same change is applied to cranelift-wasm and its dependencies it would be kind of unfortunate.

That said, it's not up to me, and I don't expect some crazy experimental use case to guide that sort of decision. It'll probably be a while before I switch back to that project anyway. I'm just saying it would be pretty cool.

alexcrichton:
I would personally think that this shouldn't be merged to stay consistent with https://github.com/bytecodealliance/wasmtime/pull/554. As mentioned on that PR it doesn't mean we'll *never* add support for this, just that this probably isn't the way we'll implement it if we do implement it one day. If it's not needed at this time as well then we've got time to figure out how best to organize this.

bjorn3:
Nebulet already uses Cranelift, while there is no `no-std` user of Wasmtime.

ratmice:
Hmm, that features are enabled when building seems weird is there another catch for when `host != target`? I haven't yet wrapped my head around what this means for the proposal in the wasmtime PR mentioned above which recommended moving to the no_std_compat crate :confused:

PR #1266, opened 2019-12-03
Fix #1247: Tweak enable_verifier doc comment so it's true

### LINKED ISSUE #1247: enable_verifier default disagrees with docs
Currently we state that the verifier is disabled by default, but it is in fact enabled by default. See: https://github.com/bytecodealliance/cranelift/blob/afc1e0ebee55ec66b86c2267c086cf69eaa4d8a8/cranelift-codegen/meta/src/shared/settings.rs#L19-L28

This default was switched about two years ago, and has been carried forward: https://github.com/bytecodealliance/cranelift/commit/774e8fc637bc6461ef0465d282fba637fe5399d4

The verifier still catches bugs IME, and it seems like a fine default, but the docs for the flag are wrong!

bnjbvr:
Fix the doc comment observed in #1247. @iximeow PTAL, thanks!

### DIFF
--- a/cranelift-codegen/meta/src/shared/settings.rs
+++ b/cranelift-codegen/meta/src/shared/settings.rs
@@ -21,8 +21,8 @@ pub(crate) fn define() -> SettingGroup {
         r#"
         Run the Cranelift IR verifier at strategic times during compilation.
 
-        This makes compilation slower but catches many bugs. The verifier is
-        disabled by default, except when reading Cranelift IR from a text file.
+        This makes compilation slower but catches many bugs. The verifier is always enabled by
+        default, which is useful during development.
         "#,
         true,
     );


cranelift-codegen/meta/src/shared/settings.rs
```diff
@@ -21,8 +21,9 @@ pub(crate) fn define() -> SettingGroup {
         r#"
         Run the Cranelift IR verifier at strategic times during compilation.
 
-        This makes compilation slower but catches many bugs. The verifier is
-        disabled by default, except when reading Cranelift IR from a text file.
+        This makes compilation slower but catches many bugs. The verifier is always enabled by
+        default, which is useful during development. It is recommended to manually disable it in
```
iximeow: I think `It is recommended to manually disable it in production environments.` is even a bit strong. I'd be lightly surprised if folks disabled verification in AOT contexts, for example. In fact, I can't think of guidance to offer other than the general "turn off optional features when you're tuning Cranelift for codegen speed". What do you think about just removing the last sentence here?

bnjbvr: Good point!

bnjbvr:
Thanks for the quick review, @iximeow ! I assume you've aborted the Github actions since this is just a comment tweak, which makes sense. Merging.

bnjbvr:
Actually no, it's the failure fixed by #1268.

PR #1270, opened 2019-12-04
Ensure builds are deterministic in CI

bnjbvr:
Not discussed in an issue, but this has happened enough times that it would be nice to address once and for all.

This introduces a script that has a high probability of failing if the
Rust source code generated by the meta crate is not consistent accross
build script runs.

It also adds a new CI job to run it on each push, on a single platform.

### DIFF
--- a/ci/ensure_deterministic_build.sh
+++ b/ci/ensure_deterministic_build.sh
@@ -1,0 +1,41 @@
+#!/bin/bash
+
+# This script makes sure that the meta crate deterministically generate files
+# with a high probability.
+# The current directory must be set to the repository's root.
+
+set -e
+
+BUILD_SCRIPT=$(find -wholename "./target/debug/build/cranelift-codegen-*/build-script-build")
+
+# First, run the script to generate a reference comparison.
+rm -rf /tmp/reference
+mkdir /tmp/reference
+OUT_DIR=/tmp/reference TARGET=x86_64 $BUILD_SCRIPT
+
+# To make sure the build script doesn't depend on the current directory, we'll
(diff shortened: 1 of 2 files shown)


bnjbvr:
The new CI step failed (because #1269 hasn't been merged yet), so this PR works!

bjorn3:
Can you please move the source dir in between tests? That would detect if it depends on the source dir.

bnjbvr:
@bjorn3 I've implemented this, fun idea. To make it locally reproducible without moving the content of the local clone, I now copy the `target/` directory content on the first run, then this is moved around.

The move makes the whole script run slower, so I've reduced the number of iterations from 100 to 30. Local testing shows that different content was generated after at most 3 runs, so it should be sufficient.

ci/ensure_deterministic_build.sh
```diff
@@ -0,0 +1,41 @@
...
+rm -rf /tmp/reference
+mkdir /tmp/reference
+OUT_DIR=/tmp/reference TARGET=x86_64 $BUILD_SCRIPT
+
+# To make sure the build script doesn't depend on the current directory, we'll
+# change the current working directory on every iteration. Make this easy to
+# reproduce this locally by first copying the target/ directory into an initial
+# temporary directory (and not move and lose the local clone's content).
+rm -rf /tmp/src0
+mkdir /tmp/src0
```
bjorn3: Can you use `mktemp` instead?

bnjbvr: I find using `mkdir` is simple enough, works fine as is, and I wouldn't want to spend too much time on this test script :p.

PR #1272, opened 2019-12-04
Bitcasting at control flow exits

abrown:
- [x] This has been discussed in issue #1192.
- [x] A short description of what this does, why it is needed: I have other PRs (#1236 and #1251) to solve the underlying SIMD type issues in different ways; however, they still fall short because I have not yet found a good way to retain the necessary type information to inform cranelift that, e.g., `iadd` needs to operate on `i32x4` and not the default `i8x16`. This PR adds more `raw_bitcasts` so that function signatures returning vectors will have some bitcasts before `return` to convert the actual vector type (e.g. `i32x4`) into the expected signature type (e.g. the default `i8x16`). I propose we merge something like this first so I can get more SIMD spec tests running and circle back to a better solution to #1192 afterwards.

### DIFF
--- a/cranelift-wasm/src/code_translator.rs
+++ b/cranelift-wasm/src/code_translator.rs
@@ -1845,17 +1845,24 @@ fn pop2_with_bitcast(
     (bitcast_a, bitcast_b)
 }
 
-/// A helper for bitcasting a sequence of arguments; to avoid SIMD type issues, this will bitcast
-/// vectors to the type expected in their next use. This modifies the arguments in place.
-fn bitcast_arguments(args: &mut [Value], expected_types: &[Type], builder: &mut FunctionBuilder) {
-    assert_eq!(args.len(), expected_types.len());
+/// A helper for bitcasting a sequence of values (e.g. function arguments). If a value is a
+/// vector type that does not match its expected type, this will modify the value in place to point
+/// to the result of a `raw_bitcast`. This conversion is necessary to translate Wasm code that
+/// uses `V128` as function parameters (or implicitly in EBB parameters) and still use specific
+/// CLIF types (e.g. `I32X4`) in the function body.
+pub fn bitcast_arguments(
+    arguments: &mut [Value],
+    expected_types: &[Type],
+    builder: &mut FunctionBuilder,
+) {
+    assert_eq!(arguments.len(), expected_types.len());
     for (i, t) in expected_types.iter().enumerate() {
         if t.is_vector() {
             assert!(
-                builder.func.dfg.value_type(args[i]).is_vector(),
+                builder.func.dfg.value_type(arguments[i]).is_vector(),
                 "unexpected type mismatch"
             );
-            args[i] = optionally_bitcast_vector(args[i], *t, builder)
+            arguments[i] = optionally_bitcast_vector(arguments[i], *t, builder)
         }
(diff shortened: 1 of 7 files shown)


cranelift-wasm/src/code_translator.rs
```diff
@@ -270,9 +271,14 @@ pub fn translate_operator<FE: FuncEnvironment + ?Sized>(
...
             if !builder.is_unreachable() || !builder.is_pristine() {
                 let return_count = frame.num_return_values();
-                builder
-                    .ins()
-                    .jump(frame.following_code(), state.peekn(return_count));
+                let args = bitcast_arguments(
+                    state.peekn(return_count),
+                    &builder.func.signature.returns.clone(), // TODO this use of signature
+                    // returns may be incorrect; what if this End is in a different control flow
+                    // structure?
```
abrown: This PR is still a WIP because 1) I'm not yet confident that I've covered all of the possible cases where a vector should be bitcast to another type and 2) in situations like the above (`Operator::End`), I am pretending that I can bitcast to the function return types but in fact the `End` may be the end of some other block, not necessarily the end of a function (though adding this makes more SIMD spec tests pass).

sunfishcode: I think this might be on the right track, except that you should use the block's signature, rather than the function's signature.

cranelift-wasm/src/code_translator.rs
```diff
@@ -1741,3 +1753,29 @@ fn pop2_with_bitcast(
...
     (bitcast_a, bitcast_b)
 }
+
+/// A helper for bitcasting a sequence of arguments; to avoid SIMD type issues, this will bitcast
+/// vectors to the type expected in their next use.
+fn bitcast_arguments(
+    args: &[Value],
+    params: &[AbiParam],
+    builder: &mut FunctionBuilder,
+) -> Vec<Value> {
```
sunfishcode: Would it be possible to make this function take `args` by `&mut [Value]` and mutate it in place, instead of returning a new `Vec`? At the places where this is called, it's passing the result from `peekn`, so it'd require creating a `peekn_mut` or so. Assuming that works, that would avoid a dynamic allocation for each `End`.

abrown:
@sunfishcode, I pushed some commits that should resolve your comments above. I am still thinking through your comment about control flow merge points: shouldn't control flow divergences merge back into a common EBB? If that is the case, I think my 114ca62 commit might solve that problem--I look at the parameters of the EBB we jump to at an `Operator::End` and bitcast any vectors to the expected type in that next EBB.

But perhaps you meant something else: I believe we will have to bitcast in some way in all the cases listed as [control instructions in the spec]; e.g., before a `call` we will have to bitcast to `I8X16` which is what the called function will expect type-wise. Is that more what you meant?

[control instructions in the spec]: https://webassembly.github.io/spec/core/bikeshed/index.html#control-instructions%E2%91%A0

abrown:
@sunfishcode, I think this is ready for another look. I am open to adding any new test cases you can think of (perhaps `simd-call.wat` and `simd-icall.wat` should be merged into `call.wat` and `icall.wat`?).

bnjbvr:
(Unflagging me for review, since @sunfishcode is already flagged in and did review a first time.)

PR #1276, opened 2019-12-06
Truncate over-large parameters to splat

abrown:
- [x] This has not been discussed in an issue as the bug is described here.
- [x] A short description of what this does, why it is needed: while running the SIMD spec tests in wasmtime, I noticed incorrect results due to `i32`-sized integers passed to a splat to `i8`-sized lanes. The spec was undefined on what to do with the extra bits, so I opened [an issue](https://github.com/WebAssembly/simd/issues/149) and [associated PR](https://github.com/WebAssembly/simd/pull/151) on the spec repository. This change implements the general consensus to just truncate the extra bits.
- [x] This PR contains test cases, if meaningful: see `wasmtime` SIMD spec tests.

### DIFF
--- a/cranelift-wasm/src/code_translator.rs
+++ b/cranelift-wasm/src/code_translator.rs
@@ -1093,15 +1093,16 @@ pub fn translate_operator<FE: FuncEnvironment + ?Sized>(
             // the v128.const is typed in CLIF as a I8x16 but raw_bitcast to a different type before use
             state.push1(value)
         }
-        Operator::I8x16Splat
-        | Operator::I16x8Splat
-        | Operator::I32x4Splat
+        Operator::I8x16Splat | Operator::I16x8Splat => {
+            let reduced = builder.ins().ireduce(type_of(op).lane_type(), state.pop1());
+            let splatted = builder.ins().splat(type_of(op), reduced);
+            state.push1(splatted)
+        }
+        Operator::I32x4Splat
         | Operator::I64x2Splat
         | Operator::F32x4Splat
         | Operator::F64x2Splat => {
-            let value_to_splat = state.pop1();
-            let ty = type_of(op);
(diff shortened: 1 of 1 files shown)


abrown:
> The only thing I don't understand is why truncation isn't also applicable to a 64x2 -> 32x4 case. Is that implicit?

In the spec they do pass `i32` to the `i8x16.splat` and `i16x8.splat` but they do not do the same for `i32x4.splat`, e.g.--in that case they only ever pass an `i32` and no larger. See the list of splat instructions: https://github.com/WebAssembly/simd/blob/master/proposals/simd/SIMD.md#create-vector-with-identical-lanes.

sstangl:
> In the spec they do pass `i32` to the `i8x16.splat` and `i16x8.splat` but they do not do the same for `i32x4.splat`, e.g.--in that case they only ever pass an `i32` and no larger. See the list of splat instructions: https://github.com/WebAssembly/simd/blob/master/proposals/simd/SIMD.md#create-vector-with-identical-lanes.

Is it worth asserting that those combinations can't occur?

abrown:
@sstangl, I thought the asserts were a good idea and started to add them but then remembered that the Cranelift type system might already be doing this. In the [splat definition], the value to splat is constrained to be the same type as the lane type of the returned vector. So if the value to splat doesn't match the lane type, we see the type error I was originally observing. I think we are ok with the asserts; your thoughts?

[splat definition]: https://github.com/bytecodealliance/cranelift/blob/cd856883cfd21e8ff1f11f9feaa35633fd2efe0b/cranelift-codegen/meta/src/shared/instructions.rs#L1647-L1661

abrown:
@sstangl, I'm going to merge and if you feel like we need assertions above and beyond the Cranelift type-checking then I can add them in a subsequent PR.

PR #1277, opened 2019-12-06
fix some clippy warnings

llogiq:
This PR removes a few needless clones and adds `# Safety` sections to docs of unsafe functions.

### DIFF
--- a/cranelift-codegen/src/isa/mod.rs
+++ b/cranelift-codegen/src/isa/mod.rs
@@ -22,7 +22,6 @@
 //! ```
 //! # extern crate cranelift_codegen;
 //! # #[macro_use] extern crate target_lexicon;
-//! # fn main() {
 //! use cranelift_codegen::isa;
 //! use cranelift_codegen::settings::{self, Configurable};
 //! use std::str::FromStr;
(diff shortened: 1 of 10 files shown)


cranelift-codegen/src/isa/mod.rs
```diff
@@ -22,7 +22,6 @@
 //! ```
 //! # extern crate cranelift_codegen;
 //! # #[macro_use] extern crate target_lexicon;
```
abrown: Should we remove these two commented-out lines as well?

cranelift-entity/src/boxed_slice.rs
```diff
@@ -27,6 +27,10 @@ where
 {
     /// Create a new slice from a raw pointer. A safer way to create slices is
     /// to use `PrimaryMap::into_boxed_slice()`.
+    ///
+    /// # Safety
+    ///
+    /// relies on `raw` pointing to a valid slice of `V`s
```
abrown: Before @bnjbvr gets to this: he'll say this should be capitalized and punctuated.

cranelift-codegen/src/isa/mod.rs
llogiq: Those are not commented out, they are just marked as invisible on rustdoc render.

llogiq:
Added the punctuation and made it a complete sentence.

llogiq:
Added the punctuation and made it a complete sentence.

cranelift-codegen/src/isa/mod.rs
abrown: Cool, didn't know about that, 👍

PR #1278, opened 2019-12-06
Fix legalization of `icmp ugt`

abrown:
- [x] This has not been discussed in a separate issue.
- [x] A short description of what this does, why it is needed: when I implemented the legalization of `icmp ugt` for SIMD, I incorrectly used the same pattern that I did for `icmp uge` (`pmax + pcmpeq`). I realized that this logic was incorrect when running the SIMD spec tests in wasmtime.


abrown:
@sstangl, ping?

Issue #1271, opened 2019-12-04
Remove dependency on packed_struct?


bnjbvr:
`packed_struct` has a few issues:

- it requires `serde` as a dependency, and an old version, transitively depending on an old version of `syn`, causing it to be rebuilt an additional time when Cranelift is embedded in Firefox (apparently `syn` is built 4 or 5 times in Firefox these days).
- It uses Serde to allow serde serializing / deserializing of structures defined with it, which is not useful to us, and can't be disabled through a compile feature.
- it seems the project isn't very active, see https://github.com/hashmismatch/packed_struct.rs/issues/40, date of last commit, date of last open pull requests with no activity, etc.

So overall, it seems using this dependency doesn't give a lot of value, while depending on external code that isn't being maintained (no blame here). While we could either fork the project, and/or propose to take ownership and maintain the existing one, there is another path where we could just reimplement the interesting bits (pun unintended) of this crate and use these in our code.

@sstangl @abrown what do you think?

abrown:
Discussed on IRC and I agree that as a dependency `packed_struct` is causing a few issues. I will try an implementation based on what we discussed:
 - add `read(bit_range)` and `write(bit_range, value)` inline methods to `EncodingBits`
 - create inline accessor methods for reading and writing the bit ranges stored in `EncodingBits`

glandium:
FYI, another issue is that the chain of having to build syn, serde_derive and serde before packed_struct and thus cranelift-codegen has a noticeable effect on Firefox build times. See https://glandium.org/files/gkrust-cargo-timing.html

FWIW, "simply" reverting 047e1eed82080299ed160344d2ee11a4af2de8eb makes my 5 minutes build on a Threadripper 3970X into a 4 minutes and a half build. That's a 10% win on the entirety of Firefox!

PR #1282, opened 2019-12-10
Remove packed_struct dependency

abrown:
- [x] This has been discussed in issue #1271.
This re-factors `EncodingBits` to remove the dependency on `packed_struct`; the result is more lines of code, which is unfortunate.

### DIFF
--- a/cranelift-codegen/shared/src/isa/x86/encoding_bits.rs
+++ b/cranelift-codegen/shared/src/isa/x86/encoding_bits.rs
@@ -1,6 +1,6 @@
 //! Provides a named interface to the `u16` Encoding bits.
 
-use packed_struct::prelude::*;
+use std::ops::RangeInclusive;
 
 /// Named interface to the `u16` Encoding bits, representing an opcode.
 ///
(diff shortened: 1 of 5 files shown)


abrown:
@sstangl, @bnjbvr: I pushed this so you could observe [this failure]:
```
thread 'main' panicked at 'trying to insert different recipes with a same name (Op1ldWithIndex)', cranelift-codegen/meta/src/isa/x86/encodings.rs:46:13
```

[this failure]: https://github.com/bytecodealliance/cranelift/pull/1282/checks?check_run_id=342756653#step:6:94

I've looked over the code I changed and I don't exactly understand why the following assertion should fail due to my changes: https://github.com/bytecodealliance/cranelift/blob/6e2bffc4a1dea09832bb3f854ee70fde1855f6cb/cranelift-codegen/meta/src/isa/x86/encodings.rs#L46-L52

In fact, it looks as if there is a case where the already-added recipe and the to-be-added recipe are not equal in which case we skip adding the to-be-added recipe and return the already-added recipe--this seems very counter-intuitive and possibly wrong. For reference, those recipes compare to each other on all fields except the name:

https://github.com/bytecodealliance/cranelift/blob/afc1e0ebee55ec66b86c2267c086cf69eaa4d8a8/cranelift-codegen/meta/src/cdsl/recipes.rs#L144-L157

There might be something else wrong with the `EncodingBits` implementation but that `assert!` in `recipe_add` seems fishy.

abrown:
> I don't exactly understand why the following assertion

Never mind, I pushed a new version with a corrected constructor. Hopefully you all can see any other potential issues in the bit-fiddling logic (`write` and `read`).

cranelift-codegen/shared/src/isa/x86/encoding_bits.rs
```diff
@@ -90,55 +130,103 @@ impl EncodingBits {
...
+            0b0111 => Mp2_f2_0f,
+            0b1000 => Op3_0f_38,
+            0b1001 => Mp3_66_0f_38,
+            0b1010 => Mp3_f3_0f_38,
+            0b1011 => Mp3_f2_0f_38,
+            0b1100 => Op3_0f_3a,
+            0b1101 => Mp3_66_0f_3a,
+            0b1110 => Mp3_f3_0f_3a,
+            0b1111 => Mp3_f2_0f_3a,
+            _ => panic!("invalid OpcodePrefix"), // TODO TryInto?
```
sstangl: This doesn't need a TODO, I think.

cranelift-codegen/shared/src/isa/x86/encoding_bits.rs
```diff
@@ -27,61 +27,101 @@ use packed_struct::prelude::*;
...
     }
 
     /// Extracts the MM bits of the OpcodePrefix.
     #[inline]
     pub fn mm(self) -> u8 {
-        (self.prefix.to_primitive() >> 2) & 0x3
+        (self.prefix().to_primitive() >> 2) & 0x3
+    }
+
+    /// Bits for the ModR/M byte for certain opcodes. TODO rename to MOD_RM?
```
sstangl: Possibly, but it's called `rrr` absolutely everywhere in the Recipe code. Probably OK to keep for now, just for symmetry.

cranelift-codegen/shared/src/isa/x86/encoding_bits.rs
```diff
@@ -27,61 +27,101 @@ use packed_struct::prelude::*;
...
     /// Constructs a new EncodingBits from parts.
     pub fn new(op_bytes: &[u8], rrr: u16, rex_w: u16) -> Self {
-        EncodingBits {
-            opcode_byte: op_bytes[op_bytes.len() - 1],
-            prefix: OpcodePrefix::from_opcode(op_bytes),
-            rrr: (rrr as u8).into(),
-            rex_w: (rex_w as u8).into(),
-        }
+        assert!(
+            op_bytes.len() > 0,
```
sstangl: Usually this is `assert!(!op_bytes.is_empty())`. I vaguely recall that Clippy can whine about this.

cranelift-codegen/shared/src/isa/x86/encoding_bits.rs
```diff
@@ -27,61 +27,101 @@ use packed_struct::prelude::*;
...
+
+    /// Convenience method for writing bits to specific range.
+    #[inline]
+    fn write(&mut self, range: RangeInclusive<u16>, value: u16) {
+        let size = range.end() - range.start() + 1; // Calculate the number of bits in the range.
+        let mask = (1 << size) - 1; // Generate a bit mask.
+        debug_assert!(
+            value <= mask,
+            "The written value should have fewer than {} bits.",
+            size
```
sstangl: Oh, I didn't realize that assertions had an implicit format! Learned something new :)

cranelift-codegen/shared/src/isa/x86/encoding_bits.rs
```diff
@@ -27,61 +27,101 @@ use packed_struct::prelude::*;
...
+        );
+        let value = (value & mask) << *range.start(); // Place the value in the correct locations.
+        let complement = !value; // Create the bitwise complement for the clear mask.
+        self.0 &= complement; // Clear the bits in `range`.
+        self.0 |= value; // Modify the bits in `range`.
```
sstangl: This logic works if it is only used to construct an Encoding and none of the bits overlap. However, it is possible that someone comes by later on and decides to make a setter to, for example, flip the `REX.W` bit. If they used this `write()` method, they would be unable to unset the bit once set.

To avoid that surprising behavior, the `complement` must be the complement of `mask << *range.start()` instead of the complement of `(value & mask) << *range.start()`.

I do understand that the ranges are controlled here, but I'd still recommend checking the invariants as the docs suggest.

cranelift-codegen/shared/src/isa/x86/encoding_bits.rs
```diff
@@ -27,61 +27,101 @@ use packed_struct::prelude::*;
...
+    }
+
+    /// Convenience method for reading bits from a specific range.
+    #[inline]
+    fn read(self, range: RangeInclusive<u16>) -> u8 {
+        let size = range.end() - range.start() + 1; // Calculate the number of bits in the range.
```
sstangl: The `RangeInclusive<u16>` doesn't actually enforce that `start <= end`. It merely guarantees that if given a range with an empty inclusive range, as an iterator it will not yield any elements.

However, if given such a range, the documentation for `RangeInclusive::start()` and `RangeInclusive::end()` say that calling `start()` and `end()` yield *unspecified values*. The documentation says that before using those function, the existence of a valid range must be checked explicitly via `is_empty()`.

bnjbvr:
Removing myself from the reviewers, since Sean has taken care of it, thanks! Just a side comment: can you remove the dependencies section in the meta-shared crate Cargo.toml, and add a comment in there that we should really strive to not having any dependencies in this crate? (Since it blocks both the meta and non-meta codegen crates to compile, and these are slow to build)

cranelift-codegen/shared/src/isa/x86/encoding_bits.rs
abrown: Maybe it's a Rust version issue or something (I'm on 1.39.0) but when I add the assert I see the following error:

```
error[E0034]: multiple applicable items in scope
  --> cranelift-codegen/shared/src/isa/x86/encoding_bits.rs:70:24
   |
70 |         assert!(!range.is_empty());
   |                        ^^^^^^^^ multiple `is_empty` found
   |
   = note: candidate #1 is defined in an impl for the type `std::ops::RangeInclusive<_>`
   = note: candidate #2 is defined in an impl of the trait `std::iter::ExactSizeIterator` for the type `std::ops::RangeInclusive<u16>`
   = help: to disambiguate the method call, write `std::iter::ExactSizeIterator::is_empty(range)` instead
```

When I take the compiler's advice I still see the error and adding feature flags as in https://github.com/rust-lang/rust/issues/48111#issuecomment-445132965 does not seem to work either. I may have to create a new issue for this...

cranelift-codegen/shared/src/isa/x86/encoding_bits.rs
abrown: Good catch!

cranelift-codegen/shared/src/isa/x86/encoding_bits.rs
abrown: Opened https://github.com/bytecodealliance/cranelift/issues/1284.

philipc: Both `is_empty` methods are still unstable. Maybe use `range.len()` instead of calculating the size yourself?

abrown: Does it have `len`? I can't seem to find it...

philipc: I haven't tried, but it should have it via the `ExactSizeIterator` impl.

abrown: Thanks! I think that works.

bnjbvr:
Thanks for doing this, @abrown !

PR #1285, opened 2019-12-13
Fix SIMD float comparison encoding

abrown:
The Intel manual uses `CMPNLT` and `CMPNLE` to denote not-less-than and not-less-than-or-equals. These were translated previously to `FloatCC::GreaterThan` and `FloatCC::GreaterThanOrEqual` but should be correctly translated to `FloatCC::UnorderedOrGreaterThanOrEqual` and `FloatCC::UnorderedOrGreaterThan`. This change adds the necessary legalizations to make use of these new encodings.

### DIFF
--- a/filetests/isa/x86/simd-comparison-run.clif
+++ b/filetests/isa/x86/simd-comparison-run.clif
@@ -231,3 +231,17 @@ ebb0:
     return v8
 }
 ; run
+
+function %fcmp_gt_nans_f32x4() -> b1 {
+ebb0:
+    v0 = vconst.f32x4 [NaN 0x42.0 -NaN NaN]
+    v1 = vconst.f32x4 [NaN NaN 0x42.0 Inf]
+    v2 = fcmp gt v0, v1
+    ; now check that the result v2 is all zeroes
+    v3 = vconst.i32x4 0x00
+    v4 = raw_bitcast.i32x4 v2
+    v5 = icmp eq v3, v4
+    v8 = vall_true v5
+    return v8
+}
+; run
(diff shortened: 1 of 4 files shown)


filetests/isa/x86/simd-comparison-run.clif
```diff
@@ -217,3 +217,17 @@ ebb0:
     return v8
 }
 ; run
+
+function %fcmp_lt_nans_f32x4() -> b1 {
+ebb0:
+    v0 = vconst.f32x4 [NaN NaN NaN NaN]
+    v1 = vconst.f32x4 [NaN NaN NaN NaN]
```
julian-seward1: I'm a bit concerned that this test isn't broad enough.  And that having the same operand-pair in all 4 lanes might be a missed opportunity.  Would it be possible to use it to compare not only a NaN-pair, but also a NaN vs a non-NaN, and the other way round?  Get more coverage from just this one test, I mean.  I looked through the rest of the file but wasn't sure that I sure NaN vs non-NaN tests for fcmp_lt anywhere else.  (I may have missed them).

abrown: That's a good point; let me expand that test.

filetests/isa/x86/simd-comparison-run.clif
```diff
@@ -231,3 +231,17 @@ ebb0:
     return v8
 }
 ; run
+
+function %fcmp_gt_nans_f32x4() -> b1 {
+ebb0:
+    v0 = vconst.f32x4 [NaN 0x42.0 -NaN NaN]
+    v1 = vconst.f32x4 [NaN NaN 0x42.0 Inf]
```
abrown: @julian-seward1, it's not exhaustive but this is more what you're looking for?

abrown:
@julian-seward1, can you take another look?

filetests/isa/x86/simd-comparison-run.clif
julian-seward1: Yes, that's better.  Thanks.

PR #1286, opened 2019-12-13
Fix legalization of SIMD `fneg`

abrown:
Previously `fsub` was used but this fails when negating -0.0 and +0.0 in the SIMD spec tests; using more instructions, this change uses shifts to create a constant for flipping the most significant bit of each lane with `bxor`.

### DIFF
--- a/cranelift-codegen/meta/src/isa/x86/legalize.rs
+++ b/cranelift-codegen/meta/src/isa/x86/legalize.rs
@@ -40,7 +40,6 @@ pub(crate) fn define(shared: &mut SharedDefinitions, x86_instructions: &Instruct
     let fmax = insts.by_name("fmax");
     let fmin = insts.by_name("fmin");
     let fneg = insts.by_name("fneg");
-    let fsub = insts.by_name("fsub");
     let iadd = insts.by_name("iadd");
     let icmp = insts.by_name("icmp");
     let iconst = insts.by_name("iconst");
@@ -48,6 +47,7 @@ pub(crate) fn define(shared: &mut SharedDefinitions, x86_instructions: &Instruct
     let ineg = insts.by_name("ineg");
     let insertlane = insts.by_name("insertlane");
     let ishl = insts.by_name("ishl");
+    let ishl_imm = insts.by_name("ishl_imm");
     let isub = insts.by_name("isub");
     let popcnt = insts.by_name("popcnt");
     let raw_bitcast = insts.by_name("raw_bitcast");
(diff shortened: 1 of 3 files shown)


filetests/isa/x86/simd-arithmetic-legalize.clif
```diff
@@ -39,13 +39,17 @@ function %fneg_legalized() {
...
     v2 = vconst.f64x2 [0x1.0 0x2.0]
     v3 = fneg v2
-    ; check: v5 = vconst.f64x2 0x00
-    ; nextln: v3 = fsub v5, v2
+    ; check: v7 = vconst.i64x2 0xffffffffffffffffffffffffffffffff
+    ; nextln: v8 = ishl_imm v7, 63
+    ; nextln: v9 = raw_bitcast.f64x2 v8
+    ; nextln: v3 = bxor v2, v9
 
     return
```
bnjbvr: Can you add tests for -0, 0, maybe -Infinity, +Infinity, and one NaN value, please?

cranelift-codegen/meta/src/isa/x86/legalize.rs
```diff
@@ -550,9 +550,18 @@ pub(crate) fn define(shared: &mut SharedDefinitions, x86_instructions: &Instruct
...
+        let lane_type_as_int = LaneType::int_from_bits(LaneType::from(*ty).lane_bits() as u16);
+        let uimm8_shift = Literal::constant(&imm.uimm8, lane_type_as_int.lane_bits() as i64 - 1);
+        let vconst = vconst.bind(vector(lane_type_as_int, sse_vector_size));
+        let bitcast_to_float = raw_bitcast.bind(vector(*ty, sse_vector_size));
         narrow.legalize(
             def!(b = fneg(a)),
-            vec![def!(c = vconst(u128_zeroes)), def!(b = fsub(c, a))],
+            vec![
+                def!(c = vconst(ones)),
+                def!(d = ishl_imm(c, uimm8_shift)), // Create a mask of all 1s except the MSB.
```
bnjbvr: Isn't it the opposite? (all 0s but the MSB?)

cranelift-codegen/meta/src/isa/x86/legalize.rs
```diff
@@ -550,9 +550,18 @@ pub(crate) fn define(shared: &mut SharedDefinitions, x86_instructions: &Instruct
 
     for ty in &[F32, F64] {
         let fneg = fneg.bind(vector(*ty, sse_vector_size));
+        let lane_type_as_int = LaneType::int_from_bits(LaneType::from(*ty).lane_bits() as u16);
+        let uimm8_shift = Literal::constant(&imm.uimm8, lane_type_as_int.lane_bits() as i64 - 1);
+        let vconst = vconst.bind(vector(lane_type_as_int, sse_vector_size));
+        let bitcast_to_float = raw_bitcast.bind(vector(*ty, sse_vector_size));
```
bnjbvr: I'm pretty sure the type inference could deal with an unbound raw_bitcast here, since its type would be conditioned by the input type of `fneg` itself. Explicit is nice though, so as you prefer.

filetests/isa/x86/simd-arithmetic-legalize.clif
abrown: Sure, but the most complete testing (which I believe covers those cases) is already present in the spec tests.

abrown: I added a `run` test for `[0x0.0 -0x0.0 -Inf Inf]` but I can't use `NaN` in there yet because the comparison requires `ueq` (unordered or equal) which is not yet implemented.

PR #1287, opened 2019-12-13
Fix `scalar_to_vector`: move not wide enough for 64-bit values

abrown:
Previously, the use of `enc_x86_64` emitted two encodings for `scalar_to_vector.i64`: a 32-bit move to an XMM register and a 64-bit move to an XMM register. Due to the ordering in `enc_x86_64`, the 32-bit move was used when legalizing `splat`, resulting in missing 4 bytes when splatting 64-bit values. This change makes 64-bit values always use the REX.W encoding, leaving lower-width values with three encodings: 32-bit, 64-bit no REX, and 64-bit with REX.W.

### DIFF
--- a/cranelift-codegen/meta/src/isa/x86/encodings.rs
+++ b/cranelift-codegen/meta/src/isa/x86/encodings.rs
@@ -1746,10 +1746,13 @@ pub(crate) fn define(
         } else {
             let template = rec_frurm.opcodes(&MOVD_LOAD_XMM);
             if ty.lane_bits() < 64 {
-                // no 32-bit encodings for 64-bit widths
                 e.enc32(instruction.clone(), template.clone());
+                e.enc_x86_64(instruction, template);
+            } else {
+                // No 32-bit encodings for 64-bit widths.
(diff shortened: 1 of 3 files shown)


cranelift-codegen/meta/src/isa/x86/encodings.rs
```diff
@@ -1746,10 +1746,12 @@ pub(crate) fn define(
         } else {
             let template = rec_frurm.opcodes(&MOVD_LOAD_XMM);
             if ty.lane_bits() < 64 {
-                // no 32-bit encodings for 64-bit widths
                 e.enc32(instruction.clone(), template.clone());
+                e.enc_x86_64(instruction, template);
+            } else {
+                // no 32-bit encodings for 64-bit widths
```
bnjbvr: nit: please capitalize the first letter and end the sentence with a dot.

Can this also assert that the size if 64 in this case?

abrown:
@bnjbvr, good catch on the commit message. I updated the commit message to what I think is the correct description of this change.

PR #1289, opened 2019-12-14
Fix report when output register anontation is missing

### LINKED ISSUE #814: [test_binemit] Signal when the test lacks register constraints
When writing a binemit test, if we test the generation of an assembly instruction which outputs registers (GPR or FPR), and we forget about setting the output register constraint, then test_binemit doesn't find any encodings, because constraints can never be satisfied on unknown value locations. Instead, we should report that an output register annotation is missing.

Y-Nak:
Hi, This PR resolves #814
r? @bnjbvr

- [x] This has been discussed in issue #814
- [x] Fix issue #814 and a little bit of refactoring.

### DIFF
--- a/cranelift-filetests/src/test_binemit.rs
+++ b/cranelift-filetests/src/test_binemit.rs
@@ -231,18 +231,8 @@ impl SubTest for TestBinEmit {
                 // Send legal encodings into the emitter.
                 if enc.is_legal() {
                     // Generate a better error message if output locations are not specified.
-                    if let Some(&v) = func
-                        .dfg
-                        .inst_results(inst)
-                        .iter()
-                        .find(|&&v| !func.locations[v].is_assigned())
-                    {
-                        return Err(format!(
-                            "Missing register/stack slot for {} in {}",
-                            v,
-                            func.dfg.display_inst(inst, isa)
-                        ));
-                    }
+                    validate_annotation(&func, inst, isa, false)?;
+
                     let before = sink.offset;
                     isa.emit_inst(&func, inst, &mut divert, &mut sink);
                     let emitted = sink.offset - before;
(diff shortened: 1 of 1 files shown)


cranelift-filetests/src/test_binemit.rs
```diff
@@ -337,3 +316,41 @@ impl SubTest for TestBinEmit {
         Ok(())
     }
 }
+
+// validate registers/stack slots are correctly annotated
```
bjorn3: Please use a tripple slash here.

Y-Nak:
OK.
And really sorry I forgot branching from master.
Can I close this PR and recreate PR?

bjorn3:
There are no conflicts between master and this PR, so it isn't a problem that this isn't based on the latest commit on master.

Y-Nak:
OK, thanks.
I fixed.

cranelift-filetests/src/test_binemit.rs
```diff
@@ -337,3 +316,41 @@ impl SubTest for TestBinEmit {
         Ok(())
     }
 }
+
+/// validate registers/stack slots are correctly annotated
```
bnjbvr: nit: Please capitalize the first letter, and end sentence with a dot.

cranelift-filetests/src/test_binemit.rs
```diff
@@ -337,3 +316,41 @@ impl SubTest for TestBinEmit {
...
+    {
+        Err(format!(
+            "Need register/stack slot annotation for {} in {}",
+            v,
+            func.dfg.display_inst(inst, isa)
+        ))
+    } else if validate_args {
+        if let Some(&v) = func
+            .dfg
+            .inst_args(inst)
```
bnjbvr: Alternatively, to reduce code duplication, could the function either look at args *or* results, but not both at the same time? In the second call site, it would be called twice instead of once, which is fine.

Then you can do something like `let values = if validate_args { func.dfg.inst_args(inst) } else { func.dfg.inst_results(inst) };` and check the values' locations.

cranelift-filetests/src/test_binemit.rs
```diff
@@ -337,3 +316,41 @@ impl SubTest for TestBinEmit {
         Ok(())
     }
 }
+
+/// validate registers/stack slots are correctly annotated
+fn validate_annotation(
```
bnjbvr: Can this be named `validate_location_annotations` to be more complete, please?

Issue #1288, opened 2019-12-14
cranelift-module should document that `finalize_definitions` needs to be called before `finish`


jyn514:
- What are the steps to reproduce the issue? Can you include a CLIF test case,
  ideally reduced with the `bugpoint` clif-util command?

1. `git clone https://github.com/jyn514/rcc && cd rcc`
2.  `git checkout cranelift-object`
3. `cargo run <<< 'int i = 1; int main(void) { return i; }'; ./a.out; echo $?`

- What do you expect to happen? What does actually happen? Does it panic, and
  if so, with which assertion?

I expect this to return 1. Instead it returns (on my machine) 64.
I feel reasonably confident this is not a bug in my code because the exact same code works fine using cranelift-faerie (you can reproduce by checking out the `master` branch).
CLIF (can reproduced with `cargo run -- --debug-asm`):

```
function u0:0() -> i32 system_v {
    gv0 = symbol colocated u1:0

ebb0:
    v0 = global_value.i64 gv0
    v1 = load.i32 v0
    return v1
}
```

Generated assembly. Note how the `lea` points to the next instruction which is not correct.

```
00000000000005fa <main>:
 5fa:	40 55                	rex push %rbp
 5fc:	48 89 e5             	mov    %rsp,%rbp
 5ff:	48 8d 05 00 00 00 00 	lea    0x0(%rip),%rax        # 606 <main+0xc>
 606:	40 8b 00             	rex mov (%rax),%eax
 609:	40 5d                	rex pop %rbp
 60b:	c3                   	retq   
 60c:	0f 1f 40 00          	nopl   0x0(%rax)
```
- Which Cranelift version / commit hash / branch are you using?

`cranelift-object 0.51.0 (git+https://github.com/bytecodealliance/cranelift#ec787eb281bb2e18e191508c17abe694e91f0677)`

- If relevant, can you include some extra information about your environment?
  (Rust version, operating system, architecture...)

Rust version: `rustc 1.39.0 (4560ea788 2019-11-04)`
System info: `Linux 5.0.0-37-generic` on Ubuntu 18.04 x86_64 with glibc.

jyn514:
I know the CLIF is not super helpful but I'm not sure how to show the data/global objects I've sent to cranelift. If anyone knows, I would love to add that!

jyn514:
Note that I've reproduced this bug on Windows as well so it's not something specific to the object file format.

bjorn3:
[...]

PR #1290, opened 2019-12-14
[module] Finalize definitions for the end-user

jyn514:
- [x] This has been discussed in issue #1288 
  first.

This removes the need for end-users of `cranelift-module` to call `module.finalize_definitions` before `module.finish` is called. It is fine for `finalize_definitions` to remain `pub` because the implementation calls `.drain()` on all relevant data, so it will never be finalized twice.

N/A, no new features.

I am not sure who should review this. r? @bnjbvr

### DIFF
--- a/cranelift-module/src/module.rs
+++ b/cranelift-module/src/module.rs
@@ -714,7 +714,8 @@ where
     /// Consume the module and return the resulting `Product`. Some `Backend`
     /// implementations may provide additional functionality available after
     /// a `Module` is complete.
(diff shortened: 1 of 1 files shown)


bnjbvr:
Thanks! Merging then.

iximeow:
just an appreciation comment: I tripped across `finalize_definitions` not being called when first working out #902, and found that my use of #902 had become incorrect because cranelift is nicer by default :) thanks!

Issue #1273, opened 2019-12-05
How do I use get_libcall_funcref?


jyn514:
I'm looking at the documentation for https://docs.rs/cranelift-codegen/0.51.0/cranelift_codegen/ir/fn.get_libcall_funcref.html and I'm not sure how to get an `Inst` to pass to it. It looks like `Inst` is returned by `InstBuilder.call`, but that requires me to have a `FuncRef` with the same signature already. Am I missing something obvious? It would seem to make more sense to have it take a `Signature` ...

Context: I'm trying to call `memcpy` from libc to fix https://github.com/jyn514/rcc/issues/71.

bjorn3:
You could use [`FunctionBuilder::emit_small_memcpy`](https://docs.rs/cranelift-frontend/0.51.0/cranelift_frontend/struct.FunctionBuilder.html#method.emit_small_memcpy) instead.

`get_libcall_funcref` is only used to legalize some float insts to libcalls:

https://github.com/bytecodealliance/cranelift/blob/6f4ff3153449600a5ad531b0622917c900f13c0d/cranelift-codegen/src/legalizer/libcall.rs#L10-L40

https://github.com/bytecodealliance/cranelift/blob/cabbacbe63477fcc7d1f1e36717d84f08f749a98/cranelift-codegen/src/ir/libcall.rs#L84-L102

jyn514:
emit_small_memcpy is exactly what I was looking for, thanks :)

Would it be possible to make `get_libcall_funcref` a private method so other developers aren't confused in the future? (or `pub(crate)`

PR #1291, opened 2019-12-14
Make `get_libcall_funcref` `pub(crate)`

jyn514:
- [x] This has been discussed in issue #1273
  first.

Since `get_libcall_funcref` is only used internally by the verifier,
it doesn't make sense to have it be public. This will encourage users to
look elsewhere for `memcpy` (they should be looking at
https://docs.rs/cranelift-frontend/0.51.0/cranelift_frontend/struct.FunctionBuilder.html#method.emit_small_memcpy)

N/A

I am not sure who should review this PR. r? @bnjbvr

### DIFF
--- a/cranelift-codegen/src/ir/mod.rs
+++ b/cranelift-codegen/src/ir/mod.rs
@@ -14,7 +14,7 @@ pub mod immediates;
 pub mod instructions;
 pub mod jumptable;
 pub mod layout;
-mod libcall;
+pub(crate) mod libcall;
 mod memflags;
 mod progpoint;
 mod sourceloc;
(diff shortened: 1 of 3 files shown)


cranelift-codegen/src/ir/mod.rs
```diff
@@ -49,7 +49,8 @@ pub use crate::ir::instructions::{
 };
 pub use crate::ir::jumptable::JumpTableData;
 pub use crate::ir::layout::Layout;
-pub use crate::ir::libcall::{get_libcall_funcref, get_probestack_funcref, LibCall};
+pub(crate) use crate::ir::libcall::get_libcall_funcref;
```
bnjbvr: Makes sense to me. Neither of Spidermonkey or Wasmtime seems to be using it, so it seems fine to make pub(crate) only. In this case, can you even remove it from this list, and have the verifier directly import it please?

jyn514: Sure, I made the change. That requires making libcall `pub(crate)`, not sure if that's intended.

PR #1294, opened 2019-12-17
Fix build failure in cranelift-codegen

philipc:
#1278 renamed `ones` to `u128_ones`, but #1286 added a new usage of `ones`.
```
error[E0425]: cannot find value `ones` in this scope
   --> cranelift-codegen/meta/src/isa/x86/legalize.rs:564:33
    |
564 |                 def!(c = vconst(ones)),
    |                                 ^^^^ not found in this scope
```
r? @abrown


abrown:
Thanks, @philipc!

PR #1295, opened 2019-12-17
Update wasmparser to 0.45.0

abrown:
Adds many new operators and a few API changes.

- [x] This has not been discussed in any issue.
- [x] A short description of what this does, why it is needed: see title.
- [x] This PR contains test cases, if meaningful: uses existing tests.

### DIFF
--- a/cranelift-wasm/Cargo.toml
+++ b/cranelift-wasm/Cargo.toml
@@ -11,7 +11,7 @@ keywords = ["webassembly", "wasm"]
 edition = "2018"
 
 [dependencies]
-wasmparser = { version = "0.39.2", default-features = false }
+wasmparser = { version = "0.45.0", default-features = false }
 cranelift-codegen = { path = "../cranelift-codegen", version = "0.51.0", default-features = false }
 cranelift-entity = { path = "../cranelift-entity", version = "0.51.0" }
 cranelift-frontend = { path = "../cranelift-frontend", version = "0.51.0", default-features = false }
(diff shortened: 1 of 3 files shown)


cranelift-wasm/Cargo.toml
```diff
@@ -26,8 +26,8 @@ target-lexicon = "0.9"
 
 [features]
 default = ["std", "basic-blocks"]
-std = ["cranelift-codegen/std", "cranelift-frontend/std", "wasmparser/std"]
-core = ["hashbrown", "cranelift-codegen/core", "cranelift-frontend/core", "wasmparser/core"]
+std = ["cranelift-codegen/std", "cranelift-frontend/std"]
+core = ["hashbrown", "cranelift-codegen/core", "cranelift-frontend/core"]
```
abrown: I wasn't sure about removing these but it seemed that wasmparser no longer had these features?

cranelift-wasm/src/sections_translator.rs
```diff
@@ -318,10 +319,7 @@ pub fn parse_element_section<'data>(
                 elems.into_boxed_slice(),
             )?
         } else {
-            return Err(wasm_unsupported!(
-                "unsupported passive elements section: {:?}",
-                kind
-            ));
+            return Err(wasm_unsupported!("unsupported passive elements section",));
```
abrown: `kind` is no longer `Debug`--should it be?

abrown:
cc @yurydelendik

cranelift-wasm/Cargo.toml
yurydelendik: lgtm

cranelift-wasm/src/sections_translator.rs
yurydelendik: maybe, but it is not useful here

PR #1298, opened 2019-12-18
Add a DynRex recipe type for x86, decreasing the number of recipes

sstangl:
- [x] This has been discussed in https://github.com/bytecodealliance/cranelift/pull/1173. The original approach was to avoid using Templates, but it was extremely complicated. This approach is much better to reduce the number of recipes.
- [x] This patch adds a third mode for templates: REX inference is requestable at template instantiation time. This reduces the number of recipes by removing rex()/nonrex() redundancy for many instructions. In the common case of `i32_i64` and `b32_b64` encodings, this inferrable-REX mode is used.
- [x] This PR contains test cases, if meaningful. (It's covered by the filetests.)

I understand that this is work will likely be overridden by what Chris is working on. However, I believe this is still a good idea to land, because:

1. It is a strict decrease in the number of recipes.
2. It makes more of the REX-related code explicit, and therefore visible. Removing implicit logic makes the code easier to reason about, and therefore easier to refactor later.
3. Because the recipe count decreases, there should be a small performance improvement due to faster encoding -- there are fewer options from which to select.
4. The transformation is fairly mechanical and I believe not difficult to review.

This invalidates the work in https://github.com/bytecodealliance/cranelift/pull/1173. That approach did not work because it tried to avoid using Templates. Removing templates at the same time as making this change resulted in too many moving parts -- it was overly complicated and error-prone.

### DIFF
--- a/cranelift-codegen/meta/src/isa/x86/recipes.rs
+++ b/cranelift-codegen/meta/src/isa/x86/recipes.rs
@@ -61,7 +61,7 @@ impl<'builder> RecipeGroup<'builder> {
         self.templates
             .iter()
             .find(|recipe| recipe.name() == name)
-            .unwrap_or_else(|| panic!("unknown tail recipe name: {}. Try recipe?", name))
+            .unwrap_or_else(|| panic!("unknown template name: {}. Try recipe?", name))
     }
 }
 
@@ -248,7 +248,7 @@ impl<'builder> Template<'builder> {
     pub fn nonrex(&self) -> Self {
         assert!(
             self.rex_kind != RexRecipeKind::AlwaysEmitRex,
-            "Tail recipe requires REX prefix."
+            "Template requires REX prefix."
         );
         let mut copy = self.clone();
         copy.rex_kind = RexRecipeKind::NeverEmitRex;
@@ -257,7 +257,7 @@ impl<'builder> Template<'builder> {
     pub fn rex(&self) -> Self {
         assert!(
             self.rex_kind != RexRecipeKind::NeverEmitRex,
-            "Tail recipe requires no REX prefix."
+            "Template requires no REX prefix."
         );
         if let Some(prefixed) = &self.when_prefixed {
             let mut ret = prefixed.rex();
@@ -272,10 +272,9 @@ impl<'builder> Template<'builder> {
         copy
     }
     pub fn infer_rex(&self) -> Self {
-        // TODO: when_prefixed is a hack that hurts clarity; get rid of it.
         assert!(
             self.rex_kind != RexRecipeKind::NeverEmitRex,
-            "Tail recipe requires no REX prefix."
+            "Template requires no REX prefix."
         );
         assert!(
             self.when_prefixed.is_none(),
@@ -519,7 +518,7 @@ pub(crate) fn define<'shared>(
                 ),
             regs,
         )
-        .inferred_rex_compute_size("size_with_inferred_rex_for_two_in_regs"),
+        .inferred_rex_compute_size("size_with_inferred_rex_for_inreg0_inreg1"),
     );
 
     // XX /r with operands swapped. (RM form).
@@ -536,7 +535,7 @@ pub(crate) fn define<'shared>(
                 ),
             regs,
         )
-        .inferred_rex_compute_size("size_with_inferred_rex_for_two_in_regs"),
+        .inferred_rex_compute_size("size_with_inferred_rex_for_inreg0_inreg1"),
     );
 
     // XX /r with FPR ins and outs. A form.
@@ -602,7 +601,7 @@ pub(crate) fn define<'shared>(
                 ),
             regs,
         )
-        .inferred_rex_compute_size("size_with_inferred_rex_for_one_in_reg"),
+        .inferred_rex_compute_size("size_with_inferred_rex_for_inreg0"),
     );
 
     // XX /r, but for a unary operator with separate input/output register, like
@@ -621,7 +620,7 @@ pub(crate) fn define<'shared>(
                 ),
             regs,
         )
-        .inferred_rex_compute_size("size_with_inferred_rex_for_1in_1out"),
+        .inferred_rex_compute_size("size_with_inferred_rex_for_inreg0_outreg0"),
     );
 
     // Same as umr, but with FPR -> GPR registers.
@@ -741,7 +740,7 @@ pub(crate) fn define<'shared>(
                 ),
             regs,
         )
-        .inferred_rex_compute_size("size_with_inferred_rex_for_1in_1out"),
+        .inferred_rex_compute_size("size_with_inferred_rex_for_inreg0_outreg0"),
     );
 
     // XX /r, RM form, FPR -> GPR.
@@ -889,7 +888,7 @@ pub(crate) fn define<'shared>(
                     ),
                 regs,
             )
-            .inferred_rex_compute_size("size_with_inferred_rex_for_one_in_reg"),
+            .inferred_rex_compute_size("size_with_inferred_rex_for_inreg0"),
         );
 
         recipes.add_template_recipe(
@@ -934,7 +933,7 @@ pub(crate) fn define<'shared>(
                     ),
                 regs,
             )
-            .inferred_rex_compute_size("size_with_inferred_rex_for_one_in_reg"),
+            .inferred_rex_compute_size("size_with_inferred_rex_for_inreg0"),
         );
     }
 
@@ -1430,7 +1429,7 @@ pub(crate) fn define<'shared>(
                 .operands_in(vec![gpr, gpr])
                 .inst_predicate(has_no_offset.clone())
                 .clobbers_flags(false)
-                .compute_size("size_plus_maybe_sib_or_offset_for_in_reg_1")
+                .compute_size("size_plus_maybe_sib_or_offset_for_inreg_1")
                 .emit(
                     r#"
                         if !flags.notrap() {
@@ -1458,7 +1457,7 @@ pub(crate) fn define<'shared>(
                     .operands_in(vec![abcd, gpr])
                     .inst_predicate(has_no_offset.clone())
                     .clobbers_flags(false)
-                    .compute_size("size_plus_maybe_sib_or_offset_for_in_reg_1")
+                    .compute_size("size_plus_maybe_sib_or_offset_for_inreg_1")
                     .emit(
                         r#"
                         if !flags.notrap() {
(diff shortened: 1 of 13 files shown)


abrown:
Do you see a future where this can be extended for [VEX/EVEX prefixes](https://github.com/bytecodealliance/cranelift/issues/1244)?

sstangl:
Yes, in theory. The only reason I would be reticent to do so is that Chris Fallin is working on a prototype for deleting the entire Recipes system, and so I don't know what the future will look like. However if it's not a ton of work and makes the emission-conditions more explicit, that's likely a good idea.

cranelift-codegen/src/isa/x86/enc_tables.rs
```diff
@@ -99,6 +125,136 @@ fn size_plus_maybe_sib_or_offset_for_in_reg_1(
...
+    let reg0 = divert.reg(func.dfg.inst_args(inst)[0], &func.locations);
+    if needs_rex_rb(reg0) {
+        return sizing.base_size + 1;
+    }
+
+    // Emit a REX prefix if REX.R is needed to encode the second register.
+    sizing.base_size + additional_size_if(1, inst, divert, func, needs_rex_rb)
+}
+
+/// Infers whether a dynamic REX prefix will be emitted, based on the inreg1.
```
iximeow: Can this comment be adjusted to indicate that this refers specifically to the second operand, `inreg1`, for `mulx`? Reading this independently had me thinking the operand index was off by one

cranelift-codegen/src/isa/x86/enc_tables.rs
```diff
@@ -99,6 +125,136 @@ fn size_plus_maybe_sib_or_offset_for_in_reg_1(
...
+) -> u8 {
+    // Emit a REX prefix if REX.W is requested by the encoding.
+    let bits = EncodingBits::from(enc.bits());
+    if u8::from(bits.rex_w()) > 0 {
+        return sizing.base_size + 1;
+    }
+    sizing.base_size + additional_size_if(1, inst, divert, func, needs_rex_rb)
+}
+
+/// Infers whether a dynamic REX prefix will be emitted, based on the inreg2.
```
iximeow: Same here, but for `div`'s GPR argument. Also thought that the index was off by one, but `2` made me go check the use first :)

cranelift-codegen/src/isa/x86/enc_tables.rs
```diff
@@ -16,9 +16,20 @@ use crate::isa::{self, TargetIsa};
...
+use cranelift_codegen_shared::isa::x86::EncodingBits;
+
 include!(concat!(env!("OUT_DIR"), "/encoding-x86.rs"));
 include!(concat!(env!("OUT_DIR"), "/legalize-x86.rs"));
 
+/// Whether the REX prefix is needed for encoding extended registers (via REX.R/B).
+///
+/// Normal x86 instructions have only 3 bits for encoding a register.
+/// The REX prefix adds REX.R and REX.B bits, interpreted as fourth bits.
+pub fn needs_rex_rb(reg: RegUnit) -> bool {
```
iximeow: small quibble: this PR doesn't touch REX prefixes for instructions with memory operands, but this is also a correct check for if the index register in a memory operand needs `X`. How do you feel about naming this simply `is_extended_reg`?

cranelift-codegen/src/isa/x86/enc_tables.rs
```diff
@@ -99,6 +125,136 @@ fn size_plus_maybe_sib_or_offset_for_in_reg_1(
...
+    // Emit a REX prefix if REX.W is requested by the encoding.
+    let bits = EncodingBits::from(enc.bits());
+    if u8::from(bits.rex_w()) > 0 {
+        return sizing.base_size + 1;
+    }
+
+    // Emit a REX prefix if REX.B is needed to encode the first register.
+    let reg0 = divert.reg(func.dfg.inst_args(inst)[0], &func.locations);
+    if needs_rex_rb(reg0) {
+        return sizing.base_size + 1;
```
iximeow: This applies to all the `size_with_inferred_rex*` here - how do you feel about structuring bodies like:
```
let needs_rex = (bits.rex_w() != 0) || // it'd be handy if `rex_w` could be obtained as a bool, alas
  test_input(0, inst, divert, func, needs_rex_rb) ||
  test_input(1, inst, divert, func, needs_rex_rb);

sizing.base_size + if needs_rex { 1 } else { 0 }
```
? Where `input_test` and `output_test` are `additional_size_if` but returning only the predicate? Admittedly I'm not a huge fan of `additional_size_if` because it's a bit eager in turning "we need a prefix" into "it will be so-and-so many bytes" (which would be complicating with @abrown's mention of VEX/EVEX!), so having a generic "is this true for this operand" which we can use to later decide "we do want a REX prefix, and that's one byte", seems nice. I also personally like avoiding the early returns with hardcoded sizes, YMMV!

sstangl: I like it. Will make that change.

sstangl:
Much cleaner now :slightly_smiling_face: 

> In light of plans to just delete all of Recipes (yay), do you plan to extend this for inferring REX on memory-operand encodings, or leave it at this? I have a few comments but overall I'm a big fan of getting encodings more precise with less effort.

I don't have any plans currently. I'll see what Benjamin says.

cranelift-codegen/src/isa/x86/enc_tables.rs
```diff
@@ -96,7 +119,100 @@ fn size_plus_maybe_sib_or_offset_for_in_reg_1(
     divert: &RegDiversions,
     func: &Function,
 ) -> u8 {
-    sizing.base_size + additional_size_if(1, inst, divert, func, needs_sib_byte_or_offset)
+    let needs_sib_or_offset = test_input(1, inst, divert, func, needs_sib_byte_or_offset);
+    sizing.base_size + if needs_sib_or_offset { 1 } else { 0 }
+}
+
+/// Infers whether a dynamic REX prefix will be emitted, for use with one in-reg.
```
bnjbvr: nit: here and below, can you expand "in-regs" to "input regs" in comments, please? (and if it's pre-existing, replace it in other places in this file too)

cranelift-codegen/src/isa/x86/enc_tables.rs
```diff
@@ -96,7 +119,100 @@ fn size_plus_maybe_sib_or_offset_for_in_reg_1(
...
+///
+/// A REX prefix is known to be emitted if either:
+///  1. The EncodingBits specify that REX.W is to be set.
+///  2. Registers are used that require REX.R or REX.B bits for encoding.
+fn size_with_inferred_rex_for_one_in_reg(
```
bnjbvr: nit: for naming consistency with the other functions, how about `size_with_inferred_rex_for_inreg0`?

cranelift-codegen/src/isa/x86/enc_tables.rs
```diff
@@ -96,7 +119,100 @@ fn size_plus_maybe_sib_or_offset_for_in_reg_1(
...
+        || test_input(0, inst, divert, func, is_extended_reg);
+    sizing.base_size + if needs_rex { 1 } else { 0 }
+}
+
+/// Infers whether a dynamic REX prefix will be emitted, for use with two in-regs.
+///
+/// A REX prefix is known to be emitted if either:
+///  1. The EncodingBits specify that REX.W is to be set.
+///  2. Registers are used that require REX.R or REX.B bits for encoding.
+fn size_with_inferred_rex_for_two_in_regs(
```
bnjbvr: ditto "two_in_regs" => "inreg_0_1" or "2inregs"? (or conversely, use the longer form for all if you think it's more readable)

cranelift-codegen/src/isa/x86/enc_tables.rs
```diff
@@ -96,7 +119,100 @@ fn size_plus_maybe_sib_or_offset_for_in_reg_1(
...
+    divert: &RegDiversions,
+    func: &Function,
+) -> u8 {
+    let needs_rex = (EncodingBits::from(enc.bits()).rex_w() != 0)
+        || test_input(2, inst, divert, func, is_extended_reg);
+    sizing.base_size + if needs_rex { 1 } else { 0 }
+}
+
+/// Infers whether a dynamic REX prefix will be emitted, based on a single
+/// in-reg and a single out-reg.
```
bnjbvr: nit: please expand "out-reg" in comments

cranelift-codegen/meta/src/isa/x86/recipes.rs
```diff
@@ -227,36 +268,63 @@ impl<'builder> Template<'builder> {
             return ret;
         }
         let mut copy = self.clone();
-        copy.rex = true;
+        copy.rex_kind = RexRecipeKind::AlwaysEmitRex;
+        copy
+    }
+    pub fn infer_rex(&self) -> Self {
+        // TODO: when_prefixed is a hack that hurts clarity; get rid of it.
```
bnjbvr: :+1: Can you open an issue for this and write the issue number here please?

cranelift-codegen/meta/src/isa/x86/recipes.rs
```diff
@@ -212,12 +246,19 @@ impl<'builder> Template<'builder> {
         copy
     }
     pub fn nonrex(&self) -> Self {
-        assert!(!self.requires_prefix, "Tail recipe requires REX prefix.");
+        assert!(
+            self.rex_kind != RexRecipeKind::AlwaysEmitRex,
+            "Tail recipe requires REX prefix."
```
bnjbvr: oops, preexisting: these are called Templates since the migration to Rust, not TailRecipes anymore. (a few times below too)

cranelift-codegen/src/isa/x86/enc_tables.rs
```diff
@@ -16,9 +16,20 @@ use crate::isa::{self, TargetIsa};
...
 
+use cranelift_codegen_shared::isa::x86::EncodingBits;
+
 include!(concat!(env!("OUT_DIR"), "/encoding-x86.rs"));
 include!(concat!(env!("OUT_DIR"), "/legalize-x86.rs"));
 
+/// Whether the REX prefix is needed for encoding extended registers (via REX.RXB).
+///
+/// Normal x86 instructions have only 3 bits for encoding a register.
+/// The REX prefix adds REX.R, REX,X, and REX.B bits, interpreted as fourth bits.
```
bjorn3: *REZ.X

cranelift-codegen/src/isa/x86/binemit.rs
```diff
@@ -101,6 +116,20 @@ fn put_rexop2<CS: CodeSink + ?Sized>(bits: u16, rex: u8, sink: &mut CS) {
     sink.put1(bits as u8);
 }
 
+/// Emit two-byte opcode: 0F XX with inferred REX prefix.
+fn put_dynrexop2<CS: CodeSink + ?Sized>(bits: u16, rex: u8, sink: &mut CS) {
+    debug_assert_eq!(
+        bits & 0x0f00,
+        0x0400,
```
bjorn3: Doc says 0F (as emitted), yet compared with 04.

sstangl: That assertion is copied from 'put_rexopt2()`. These assertions should definitely not open-code the bit-patterns, but should be changed to use the new `EncodingBits` struct.

PR #1301, opened 2019-12-19
Merge emit_small_memcpy and emit_small_memmove

bjorn3:
  here. This is mostly a cleanup.
  first. They share a lot of code. Also by performing all loads before all stores, the performance of the simple-raytracer cg_clif benchmark increases by 1%.
- [x] This PR contains test cases, if meaningful. N/A
  `bnjbvr`. <strike>The list of suggested reviewers on the right can help you.</strike> (no list for non members)

### DIFF
--- a/cranelift-frontend/src/frontend.rs
+++ b/cranelift-frontend/src/frontend.rs
@@ -633,7 +633,7 @@ impl<'a> FunctionBuilder<'a> {
     ///
     /// The following properties must hold to prevent UB:
     ///
-    /// * `src_align` and `dest_align` are an upper-bound on the alignment of `src` cq `dest`.
+    /// * `src_align` and `dest_align` are an upper-bound on the alignment of `src` respectively `dest`.
     /// * If `non_overlapping` is true, then this must be correct.
     pub fn emit_small_memory_copy(
         &mut self,
--- a/cranelift-frontend/src/frontend.rs
+++ b/cranelift-frontend/src/frontend.rs
@@ -1,7 +1,6 @@
 //! A frontend for building Cranelift IR from other languages.
 use crate::ssa::{Block, SSABuilder, SideEffects};
 use crate::variable::Variable;
-use alloc::vec::Vec;
 use cranelift_codegen::cursor::{Cursor, FuncCursor};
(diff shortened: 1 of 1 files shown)


bjorn3:
I found a bug with cg_clif. Will investigate if it is caused by this or a change to cg_clif itself.

cranelift-frontend/src/frontend.rs
```diff
@@ -665,16 +672,27 @@ impl<'a> FunctionBuilder<'a> {
...
+            if non_overlapping {
+                self.call_memcpy(config, dest, src, size_value);
+            } else {
+                self.call_memmove(config, dest, src, size_value);
+            }
             return;
         }
 
         let mut flags = MemFlags::new();
         flags.set_aligned();
```
bjorn3: Pre-existing, but this needs to compare `{src,dest}_align` with `access_size`.

bjorn3:
> I found a bug with cg_clif. Will investigate if it is caused by this or a change to cg_clif itself.

This was a bug in cg_clif itself.

cranelift-frontend/src/frontend.rs
```diff
@@ -628,15 +627,23 @@ impl<'a> FunctionBuilder<'a> {
...
 
-    /// Optimised memcpy for small copies.
-    pub fn emit_small_memcpy(
+    /// Optimised memcpy or memmove for small copies.
+    ///
+    /// # Codegen safety
+    ///
+    /// The following properties must hold to prevent UB:
+    ///
+    /// * `src_align` and `dest_align` are an upper-bound on the alignment of `src` cq `dest`.
```
eqrion: `src cq dest`

Is `cq` a typo or am I missing an abbreviation?

bjorn3: It seems that `cq` is a Dutch abbreviation: https://nl.m.wikipedia.org/wiki/Casu_quo. I also used it wrong. :man_facepalming: I meant: `src_align` is the upper-bound on the alignment of `src` and `dest_align` is the upper-bound on the alignment of `dest`. I replaced it with `respectively`.

PR #1304, opened 2019-12-20
Encode stack load store

bjorn3:
- [x] This has been discussed in issue #597, or if not, please tell us why
  here.
  first.
  This adds a `stack_load` and `stack_store` encoding for 32bit and 64bit integers. This saves one `lea` instruction for each `stack_load` and `stack_store`.

<strike>
This currently doesn't encode `stack_load` and `stack_store` correctly.

```
[RexOp1spst_id#8089]                stack_store v0, ss1
[...]
@0000 [RexOp1spaddr_ld_id#808b,%rax] v20 = stack_load.i64 ss1
```

results in

```
  4a3df4:       48 89 bc 24 40 00 00    mov    %rdi,0x40(%rsp)
  4a3dfb:       00
[...]
  4a3e04:       48 8b 84 24 40 00 00    mov    0x40(%rsp),%rax
  4a3e0b:       00
```
</strike>

### DIFF
--- a/cranelift-codegen/meta/src/isa/x86/encodings.rs
+++ b/cranelift-codegen/meta/src/isa/x86/encodings.rs
@@ -2191,7 +2191,10 @@ fn define_entity_ref(
 
     // Stack accesses.
     e.enc32(stack_addr.bind(I32), rec_spaddr_ld_id.opcodes(&LEA));
-    e.enc64(stack_addr.bind(I64), rec_spaddr_ld_id.opcodes(&LEA).rex().w());
+    e.enc64(
+        stack_addr.bind(I64),
+        rec_spaddr_ld_id.opcodes(&LEA).rex().w(),
+    );
     e.enc_i32_i64_explicit_rex(stack_load, rec_spaddr_ld_id.opcodes(&MOV_LOAD));
     e.enc_i32_i64_explicit_rex(stack_store, rec_spst_id.opcodes(&MOV_STORE));
 }
--- a/cranelift-codegen/meta/src/isa/x86/encodings.rs
+++ b/cranelift-codegen/meta/src/isa/x86/encodings.rs
@@ -2191,7 +2191,7 @@ fn define_entity_ref(
 
     // Stack accesses.
     e.enc32(stack_addr.bind(I32), rec_spaddr_ld_id.opcodes(&LEA));
-    e.enc64(stack_addr.bind(I64), rec_spaddr_ld_id.opcodes(&LEA));
+    e.enc64(stack_addr.bind(I64), rec_spaddr_ld_id.opcodes(&LEA).rex().w());
     e.enc_i32_i64_explicit_rex(stack_load, rec_spaddr_ld_id.opcodes(&MOV_LOAD));
     e.enc_i32_i64_explicit_rex(stack_store, rec_spst_id.opcodes(&MOV_STORE));
 }
--- a/cranelift-codegen/meta/src/isa/x86/encodings.rs
+++ b/cranelift-codegen/meta/src/isa/x86/encodings.rs
@@ -2089,6 +2089,8 @@ fn define_entity_ref(
     // Shorthands for instructions.
     let func_addr = shared.by_name("func_addr");
     let stack_addr = shared.by_name("stack_addr");
+    let stack_load = shared.by_name("stack_load");
+    let stack_store = shared.by_name("stack_store");
     let symbol_value = shared.by_name("symbol_value");
 
     // Shorthands for recipes.
@@ -2102,8 +2104,8 @@ fn define_entity_ref(
     let rec_gvaddr8 = r.template("gvaddr8");
     let rec_pcrel_fnaddr8 = r.template("pcrel_fnaddr8");
     let rec_pcrel_gvaddr8 = r.template("pcrel_gvaddr8");
-    let rec_spaddr4_id = r.template("spaddr4_id");
-    let rec_spaddr8_id = r.template("spaddr8_id");
+    let rec_spaddr_ld_id = r.template("spaddr_ld_id");
+    let rec_spst_id = r.template("spst_id");
 
     // Predicates shorthands.
     let all_ones_funcaddrs_and_not_is_pic =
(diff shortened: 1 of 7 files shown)


bjorn3:
Could somebody help me with the encoding problem?

bjorn3:
I fixed the encoding problem. (I forgot a `.rex().w()`)

bjorn3:
```
Benchmark #1: ./raytracer_cg_clif
  Time (mean ± σ):      9.283 s ±  0.134 s    [User: 9.224 s, System: 0.025 s]
  Range (min … max):    9.124 s …  9.591 s    20 runs
 
Benchmark #2: ./raytracer_cg_clif_opt
  Time (mean ± σ):      7.071 s ±  0.098 s    [User: 7.023 s, System: 0.021 s]
  Range (min … max):    6.997 s …  7.381 s    20 runs
 
Benchmark #3: ./raytracer_cg_clif_opt_stack_enc
  Time (mean ± σ):      6.912 s ±  0.063 s    [User: 6.872 s, System: 0.018 s]
  Range (min … max):    6.861 s …  7.113 s    20 runs
 
Benchmark #4: ./raytracer_cg_clif_stack_enc
  Time (mean ± σ):      8.966 s ±  0.109 s    [User: 8.915 s, System: 0.023 s]
  Range (min … max):    8.861 s …  9.343 s    20 runs
 
Benchmark #5: ./raytracer_cg_llvm
  Time (mean ± σ):      7.252 s ±  0.201 s    [User: 7.183 s, System: 0.029 s]
  Range (min … max):    7.086 s …  7.663 s    20 runs
 
  Warning: Statistical outliers were detected. Consider re-running this benchmark on a quiet PC without any interferences from other programs. It might help to use the '--warmup' or '--prepare' options.
 
Summary
  './raytracer_cg_clif_opt_stack_enc' ran
    1.02 ± 0.02 times faster than './raytracer_cg_clif_opt'
    1.05 ± 0.03 times faster than './raytracer_cg_llvm'
    1.30 ± 0.02 times faster than './raytracer_cg_clif_stack_enc'
    1.34 ± 0.02 times faster than './raytracer_cg_clif'
```

This is a performance win of a few percent.

cranelift-codegen/meta/src/isa/x86/encodings.rs
```diff
@@ -2174,12 +2176,14 @@ fn define_entity_ref(
...
     );
 
-    // Stack addresses.
-    //
-    // TODO: Add encoding rules for stack_load and stack_store, so that they
-    // don't get legalized to stack_addr + load/store.
-    e.enc32(stack_addr.bind(I32), rec_spaddr4_id.opcodes(&LEA));
-    e.enc64(stack_addr.bind(I64), rec_spaddr8_id.opcodes(&LEA).rex().w());
+    // Stack accesses.
+    e.enc32(stack_addr.bind(I32), rec_spaddr_ld_id.opcodes(&LEA));
```
bnjbvr: Is it possible to use the 8-bytes displacement on x86 32 bits, though? It seems like a possible regression, and one that's hard to test, since we only test that the generated machine code matches our encodings.

(One possible way to resolve this would be to just remove this line, in light of the plan to have a new isel/codegen mechanism, since x86 32 bits is badly supported anyways.)

cranelift-codegen/meta/src/isa/x86/recipes.rs
```diff
@@ -1378,41 +1386,44 @@ pub(crate) fn define<'shared>(
...
                 r#"
-                    let sp = StackRef::sp(stack_slot, &func.stack_slots);
-                    let base = stk_base(sp.base);
-                    {{PUT_OP}}(bits, rex2(base, out_reg0), sink);
-                    modrm_sib_disp32(out_reg0, sink);
-                    sib_noindex(base, sink);
-                    let imm : i32 = offset.into();
-                    sink.put4(sp.offset.checked_add(imm).unwrap() as u32);
-                "#,
+                        sink.trap(TrapCode::StackOverflow, func.srclocs[inst]);
```
bnjbvr: Why is this needed here and not in the load case?

bnjbvr (commented):
Looks good, a couple questions below. There's also a rebase conflict you'll need to manually solve. Thanks!

bnjbvr:
(I tried it on our Spidermonkey wasm benchmark suite, and it's between a slowdown down to 1.8% and a speedup up to 2%, depending on the benchmark.)

cranelift-codegen/meta/src/isa/x86/recipes.rs
bjorn3: I followed the example of the spill and fill encodings. Stack loads should always have a stack store before them to prevent reading undefined bytes, so any stack overflow would have already been catched by the stack store. Unlike the fill instruction, a stack load of undefined bytes is possible, so maybe I should add a StackOverflow trap marker to load anyway.

cranelift-codegen/meta/src/isa/x86/encodings.rs
bjorn3: I can add the 8bit displacement back.

bjorn3:
I am curious about what would cause the slowdown. What kind of operations does that 1.8% slowdown benchmark perform. Does it use the same stack slot a lot? Or is is the generated code a bit bigger after this PR?

cranelift-codegen/meta/src/isa/x86/encodings.rs
iximeow: I'm not sure if @bnjbvr meant 8-byte or 8-bit displacement, but to be clear, x86 does not support 8-byte displacements.

I'm suspect of the `spaddr4` encoding for x86 32-bit that was present since while it used `modrm_sib_disp8(out_reg0, sink);` to prepare for an 8-bit displacement, it proceeded to write four bytes of displacement out with  `sink.put4(sp.offset.checked_add(imm).unwrap() as u32);`

I think adding keeping an 8-bit displacement form would be good though, as it would be useful for functions with smaller stack regions.

bjorn3: Yeah, that put4 was the reason I removed it.

alexcrichton:

PR #1317, opened 2020-01-07
Improve support for reference-types

eqrion:
This is the Cranelift part of the work to add reference-types to Spidermonkey+Cranelift (see [bug 1574865](https://bugzilla.mozilla.org/show_bug.cgi?id=1574865)). I will post the Spidermonkey patches there soon.

I also have two additional commits that are blocked on bytecodealliance/wasmparser#168, but this is the majority of the work.

### DIFF
--- a/cranelift-wasm/src/code_translator.rs
+++ b/cranelift-wasm/src/code_translator.rs
@@ -965,8 +965,8 @@ pub fn translate_operator<FE: FuncEnvironment + ?Sized>(
             let val = builder.ins().is_null(arg);
             state.push1(val);
         }
-        Operator::RefFunc { .. } => {
-            return Err(wasm_unsupported!("proposed ref operator {:?}", op))
+        Operator::RefFunc { function_index } => {
+            state.push1(environ.translate_ref_func(builder.cursor(), *function_index)?);
         }
         Operator::AtomicNotify { .. }
         | Operator::I32AtomicWait { .. }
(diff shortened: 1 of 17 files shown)


eqrion:
I've found an issue through more extensive testing, this isn't ready for review.

eqrion:
> I've found an issue through more extensive testing, this isn't ready for review.

That was a false alarm, I was getting a timeout in a non-standard configuration. Upon further investigation, it looks like it's just caused by a very slow implementation of barriers. So this should be ready for review.

cranelift-wasm/src/code_translator.rs
```diff
@@ -1077,56 +1077,73 @@ pub fn translate_operator<FE: FuncEnvironment + ?Sized>(
...
+            environ.translate_table_fill(builder.cursor(), *table, dest, val, len)?;
+        }
+        Operator::TableInit {
+            segment,
+            table: table_index,
+        } => {
             // The WebAssembly MVP only supports one table and we assume it here.
-            let table_index = 0;
-            let table = state.get_table(builder.func, table_index, environ)?;
+            let table = state.get_table(builder.func, *table_index, environ)?;
```
fitzgen: The comment above this line seems like it is now out of date.

eqrion:
> Code looks good to me, but I think it should be possible to have some tests, right? I don't think we actually need to manufacture any reference types to have tests (which would require askign how we're going to manage their lifetimes) and can instead have test functions that take `r32` and `r64` parameters and put them in tables and all that. I think we should be able to do that even in the dummy environment.

That seems plausible to me. I'm not very familiar with the testing architecture, so I'll need to do some investigation. I'll try to do that today or tomorrow.

eqrion:
I've now added a basic test for r32 and r64, removed the out-of-date comments, and updated the documentation on `StackMap` to more thoroughly explain the representation.

PR #1320, opened 2020-01-07
Refactor unwind; add FDE support.

yurydelendik:
Currently, codegen produces unwind information only for Windows platform. We need to produce this type of information for all platforms to properly produce backtrace. See https://github.com/bytecodealliance/wasmtime/pull/759

TODO (and requires feedback):
- [x] use feature to cfg existing unwind logic
- [x] agree on location for FDE/UNWIND_INFO logic
- [x] borrow reg mapper from #902

cc @peterhuene @iximeow

### DIFF
--- a/cranelift-codegen/Cargo.toml
+++ b/cranelift-codegen/Cargo.toml
@@ -20,6 +20,7 @@ hashbrown = { version = "0.6", optional = true }
 target-lexicon = "0.10"
 log = { version = "0.4.6", default-features = false }
 serde = { version = "1.0.94", features = ["derive"], optional = true }
+gimli = { version = "0.19.0", optional = true }
 smallvec = { version = "1.0.0" }
 thiserror = "1.0.4"
 byteorder = { version = "1.3.2", default-features = false }
@@ -32,7 +33,7 @@ byteorder = { version = "1.3.2", default-features = false }
 cranelift-codegen-meta = { path = "meta", version = "0.54.0" }
 
 [features]
-default = ["std", "basic-blocks"]
+default = ["std", "basic-blocks", "unwind"]
 
 # The "std" feature enables use of libstd. The "core" feature enables use
 # of some minimal std-like replacement libraries. At least one of these two
@@ -46,6 +47,9 @@ core = ["hashbrown"]
 # This enables some additional functions useful for writing tests, but which
 # can significantly increase the size of the library.
 testing_hooks = []
+
+# This enables unwind info generation functionality.
+unwind = ["gimli"]
 
 # ISA targets for which we should build.
 # If no ISA targets are explicitly enabled, the ISA target for the host machine is enabled.
--- a/cranelift-codegen/Cargo.toml
+++ b/cranelift-codegen/Cargo.toml
@@ -20,7 +20,7 @@ hashbrown = { version = "0.6", optional = true }
 target-lexicon = "0.10"
 log = { version = "0.4.6", default-features = false }
 serde = { version = "1.0.94", features = ["derive"], optional = true }
-gimli = { version = "0.19.0", optional = true }
+gimli = { version = "0.19.0", default-features = false, features = ["write"], optional = true }
 smallvec = { version = "1.0.0" }
 thiserror = "1.0.4"
 byteorder = { version = "1.3.2", default-features = false }
--- a/cranelift-codegen/src/binemit/mod.rs
+++ b/cranelift-codegen/src/binemit/mod.rs
@@ -156,6 +156,7 @@ pub trait CodeSink {
 }
 
 /// Type of the frame unwind information.
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
 pub enum FrameUnwindKind {
     /// Windows fastcall unwinding (as in .pdata).
     Fastcall,
--- a/cranelift-codegen/src/binemit/mod.rs
+++ b/cranelift-codegen/src/binemit/mod.rs
@@ -163,19 +163,25 @@ pub enum FrameUnwindKind {
     Libunwind,
 }
 
+/// Offset in frame unwind information buffer.
+pub type FrameUnwindOffset = usize;
+
 /// Sink for frame unwind information.
 pub trait FrameUnwindSink {
     /// Get the current position.
-    fn offset(&self) -> CodeOffset;
+    fn len(&self) -> FrameUnwindOffset;
 
     /// Add bytes to the code section.
     fn bytes(&mut self, _: &[u8]);
 
+    /// Reserves bytes in the buffer.
+    fn reserve(&mut self, _len: usize) {}
+
     /// Add a relocation entry.
-    fn reloc(&mut self, _: Reloc, _: CodeOffset);
+    fn reloc(&mut self, _: Reloc, _: FrameUnwindOffset);
 
     /// Specified offset to main structure.
-    fn set_entry_offset(&mut self, _: CodeOffset);
+    fn set_entry_offset(&mut self, _: FrameUnwindOffset);
 }
 
 /// Report a bad encoding error.
--- a/cranelift-codegen/src/binemit/mod.rs
+++ b/cranelift-codegen/src/binemit/mod.rs
@@ -155,6 +155,29 @@ pub trait CodeSink {
     fn add_stackmap(&mut self, _: &[Value], _: &Function, _: &dyn TargetIsa);
 }
 
+/// Type of the frame unwind information.
+pub enum FrameUnwindKind {
+    /// Windows fastcall unwinding (as in .pdata).
+    Fastcall,
+    /// FDE entry for libunwind (similar to .eh_frame format).
+    Libunwind,
+}
+
+/// Sink for frame unwind information.
+pub trait FrameUnwindSink {
+    /// Get the current position.
+    fn offset(&self) -> CodeOffset;
+
+    /// Add bytes to the code section.
+    fn bytes(&mut self, _: &[u8]);
+
+    /// Add a relocation entry.
+    fn reloc(&mut self, _: Reloc, _: CodeOffset);
+
+    /// Specified offset to main structure.
+    fn set_entry_offset(&mut self, _: CodeOffset);
+}
+
 /// Report a bad encoding error.
 #[cold]
 pub fn bad_encoding(func: &Function, inst: Inst) -> ! {
--- a/cranelift-codegen/src/context.rs
+++ b/cranelift-codegen/src/context.rs
@@ -10,8 +10,8 @@
 //! single ISA instance.
 
 use crate::binemit::{
-    relax_branches, shrink_instructions, CodeInfo, MemoryCodeSink, RelocSink, StackmapSink,
-    TrapSink,
+    relax_branches, shrink_instructions, CodeInfo, FrameUnwindKind, FrameUnwindSink,
+    MemoryCodeSink, RelocSink, StackmapSink, TrapSink,
 };
 use crate::dce::do_dce;
 use crate::dominator_tree::DominatorTree;
@@ -201,8 +201,13 @@ impl Context {
     ///
     /// Only some calling conventions (e.g. Windows fastcall) will have unwind information.
     /// This is a no-op if the function has no unwind information.
-    pub fn emit_unwind_info(&self, isa: &dyn TargetIsa, mem: &mut Vec<u8>) {
-        isa.emit_unwind_info(&self.func, mem);
+    pub fn emit_unwind_info(
+        &self,
+        isa: &dyn TargetIsa,
+        kind: FrameUnwindKind,
+        sink: &mut dyn FrameUnwindSink,
+    ) {
+        isa.emit_unwind_info(&self.func, kind, sink);
     }
 
     /// Run the verifier on the function.
--- a/cranelift-codegen/src/isa/x86/abi.rs
+++ b/cranelift-codegen/src/isa/x86/abi.rs
@@ -1,11 +1,14 @@
 //! x86 ABI implementation.
 
 use super::super::settings as shared_settings;
+#[cfg(feature = "unwind")]
 use super::fde::emit_fde;
 use super::registers::{FPR, GPR, RU};
 use super::settings as isa_settings;
+#[cfg(feature = "unwind")]
 use super::unwind::UnwindInfo;
 use crate::abi::{legalize_args, ArgAction, ArgAssigner, ValueConversion};
+#[cfg(feature = "unwind")]
 use crate::binemit::{FrameUnwindKind, FrameUnwindSink};
 use crate::cursor::{Cursor, CursorPosition, EncCursor};
 use crate::ir;
@@ -948,6 +951,7 @@ fn insert_common_epilogue(
     }
 }
 
+#[cfg(feature = "unwind")]
 pub fn emit_unwind_info(
     func: &ir::Function,
     isa: &dyn TargetIsa,
--- a/cranelift-codegen/src/isa/x86/abi.rs
+++ b/cranelift-codegen/src/isa/x86/abi.rs
@@ -1,10 +1,12 @@
 //! x86 ABI implementation.
 
 use super::super::settings as shared_settings;
+use super::fde::emit_fde;
 use super::registers::{FPR, GPR, RU};
 use super::settings as isa_settings;
 use super::unwind::UnwindInfo;
 use crate::abi::{legalize_args, ArgAction, ArgAssigner, ValueConversion};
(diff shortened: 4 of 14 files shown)


cranelift-codegen/src/binemit/mod.rs
```diff
@@ -155,6 +155,29 @@ pub trait CodeSink {
     fn add_stackmap(&mut self, _: &[Value], _: &Function, _: &dyn TargetIsa);
 }
 
+/// Type of the frame unwind information.
+pub enum FrameUnwindKind {
+    /// Windows fastcall unwinding (as in .pdata).
+    Fastcall,
+    /// FDE entry (as in .eh_frame or .debug_frame).
+    EhFrame,
```
bjorn3: Maybe Seh and Dwarf?

cranelift-codegen/src/isa/x86/fde.rs
```diff
@@ -0,0 +1,237 @@
...
+    fn write_at(&mut self, offset: usize, bytes: &[u8]) -> Result<()> {
+        self.vec.write_at(offset, bytes)
+    }
+    fn write_address(&mut self, address: Address, size: u8) -> Result<()> {
+        match address {
+            Address::Constant(_) => self.vec.write_address(address, size),
+            Address::Symbol { symbol, addend } => {
+                assert_eq!(symbol, 0);
+                self.relocs
+                    //.push(FDERelocEntry(addend, self.vec.len(), size));
```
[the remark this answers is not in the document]
yurydelendik: Yes. Currently I'm just looking feedback on API design -- I'll tie ends with FDE support in a day or two.

cranelift-codegen/src/isa/x86/abi.rs
```diff
@@ -947,10 +949,31 @@ fn insert_common_epilogue(
...
+            }
+            sink.bytes(&mem);
+        }
+        FrameUnwindKind::EhFrame => {
+            if func.frame_layout.is_some() {
+                let (data, entry, relocs) = emit_fde(func, isa);
+                sink.bytes(&data);
+                sink.set_entry_offset(entry as u32);
+                for (off, r) in relocs {
+                    sink.reloc(r, off);
```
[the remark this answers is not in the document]
yurydelendik: Not sure if it is applied here: the FDE entry is for `__register_frame`

cranelift-codegen/src/binemit/mod.rs
yurydelendik: Changed to `Libunwind`

cranelift-codegen/src/isa/x86/abi.rs
```diff
@@ -947,10 +949,31 @@ fn insert_common_epilogue(
...
+    kind: FrameUnwindKind,
+    sink: &mut dyn FrameUnwindSink,
+) {
+    match kind {
+        FrameUnwindKind::Fastcall => {
+            let mut mem = Vec::new();
+            // Assumption: RBP is being used as the frame pointer
+            // In the future, Windows fastcall codegen should usually omit the frame pointer
+            if let Some(info) = UnwindInfo::try_from_func(func, isa, Some(RU::rbp.into())) {
+                info.emit(&mut mem);
```
peterhuene: I think we'll want to pass the sink on down into `UnwindInfo::emit` so we're not double buffering these bytes (once here and then presumably in the caller's sink impl).

peterhuene: Also, I'm totally open to renaming `UnwindInfo` so it's more specific to Windows so this gets less confusing.

cranelift-codegen/src/context.rs
```diff
@@ -201,8 +201,13 @@ impl Context {
     ///
     /// Only some calling conventions (e.g. Windows fastcall) will have unwind information.
     /// This is a no-op if the function has no unwind information.
-    pub fn emit_unwind_info(&self, isa: &dyn TargetIsa, mem: &mut Vec<u8>) {
-        isa.emit_unwind_info(&self.func, mem);
+    pub fn emit_unwind_info(
+        &self,
+        isa: &dyn TargetIsa,
+        kind: FrameUnwindKind,
```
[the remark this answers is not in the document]
yurydelendik: Right, but now I need to specify in which format the data was produced. The caller now have to react on type of data it receives.

cranelift-codegen/src/isa/x86/abi.rs
yurydelendik: Not an issue atm.

cranelift-codegen/src/context.rs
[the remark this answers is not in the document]
peterhuene: I see.  Are you planning on having a sink implementation that differs based on the unwind kind then?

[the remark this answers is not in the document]
yurydelendik: That's depend on the consumer, but in case of wasmtime, I planned to have one.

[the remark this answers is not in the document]
peterhuene: 👍

cranelift-codegen/src/isa/x86/fde.rs
```diff
@@ -0,0 +1,255 @@
...
+                self.relocs.push((self.vec.len().try_into().unwrap(), rt));
+                self.vec.write_udata(0, size)
+            }
+        }
+    }
+}
+
+const RETURN_ADDRESS_REGISTER: Register = X86_64::RA;
+
+fn map_reg(isa: &dyn TargetIsa, reg: RegUnit) -> Register {
```
bjorn3: Please check that the isa is x86_64 for both `map_reg` and `RETURN_ADDRESS_REGISTER`.

yurydelendik: okay, I probably just need to implement that for X86 too.

cranelift-codegen/Cargo.toml
```diff
@@ -20,6 +20,7 @@ hashbrown = { version = "0.6", optional = true }
 target-lexicon = "0.9"
 log = { version = "0.4.6", default-features = false }
 serde = { version = "1.0.94", features = ["derive"], optional = true }
+gimli = { version = "0.19.0", optional = true }
```
philipc: You should change this to only enable the "write" feature (also see https://github.com/gimli-rs/gimli/pull/466 which reduces the number of dependencies for this).
```suggestion
gimli = { version = "0.19.0", default-features = false, features = ["write"], optional = true }
```

yurydelendik:
> I also think a follow-up PR to make test_unwind.rs more generic so that it supports FDE generation in addition to Windows unwind information would be ideal.

Yeah, currently I added "test_fde.rs" -- it will match how wasmtime will use the functionality, by using FrameUnwindKind parameter.

peterhuene:
👍 

I think in a future PR we might want to unify the `unwind` and `fde` subtests so that it'll print out the right thing depending on the given function's ABI and calling convention.

I feel that `unwind` is too generic a term to mean  just Windows unwind information, so it might lead to confusing in the long run.  Perhaps something to revisit when we further abstract things.

iximeow:
sorry for the late comments - I was out last week and as a result wasn't checking `cranelift` too closely. I do have a few comments, but I expect them to be subsequent PRs, and would like to discuss a bit first.

cranelift-codegen/src/isa/x86/fde.rs
```diff
@@ -0,0 +1,354 @@
...
+        self.vec.write_at(offset, bytes)
+    }
+    fn write_address(&mut self, address: Address, size: u8) -> Result<()> {
+        match address {
+            Address::Constant(_) => self.vec.write_address(address, size),
+            Address::Symbol { .. } => {
+                assert_eq!(address, FUNCTION_ENTRY_ADDRESS);
+                let rt = match size {
+                    4 => Reloc::Abs4,
+                    8 => Reloc::Abs8,
```
iximeow: I don't think 8-byte absolute relocations are necessarily available for all targets - specifically, as far as I know, MachO objects only permit up to dword-size relocations. This is one of the reasons I used the address encoding `SDATA4` in #902. Are qword relocations available in MachO targets? If so, I also prefer absolute relocations, but I'd like to not exclude macos here.

cranelift-codegen/src/isa/x86/fde.rs
```diff
@@ -0,0 +1,354 @@
...
+    }
+}
+
+fn return_address_reg(isa: &dyn TargetIsa) -> Register {
+    assert!(isa.name() == "x86" && isa.pointer_bits() == 64);
+    X86_64::RA
+}
+
+fn map_reg(isa: &dyn TargetIsa, reg: RegUnit) -> Register {
+    assert!(isa.name() == "x86" && isa.pointer_bits() == 64);
```
iximeow: checking that `pointer_bits()` is 64 is a good call, I'd neglected to think about 32-bit targets. That said, I'll put up a PR to extend these to support i686 register numbers (`gimli` already has constants for them!).

With 32-bit register numbers, the assert would just be `assert!(isa.name() == "x86")`, which seems redundant given that this is in isa/**x86**/fde.rs.

cranelift-codegen/src/isa/x86/fde.rs
```diff
@@ -0,0 +1,354 @@
+//! Support for FDE data generation.
```
iximeow: other than the register mapping, and possibly code/data alignment, `FDE` generation is architecture-agnostic. I assume you rather this code in `cranelift-codegen` for use in wasmtime, but I'd like to have a very clear path to supporting other architectures. Is there another reason this would be in `isa/x86/`? If not I'd like to move it to somewhere like `cranelift-codegen/src/binemit/unwind.rs` that is more clearly independent of a specific architecture.

cranelift-codegen/src/isa/x86/fde.rs
```diff
@@ -0,0 +1,354 @@
...
+                    }
+                };
+                self.relocs.push((self.vec.len().try_into().unwrap(), rt));
+                self.vec.write_udata(0, size)
+            }
+        }
+    }
+}
+
+fn return_address_reg(isa: &dyn TargetIsa) -> Register {
```
iximeow: Also on the point of being architecture agnostic, I'd specifically wanted `DwarfRegMapper` to clearly indicate how to add support for non-x86_64 architectures. You mentioned wanting `cranelift-module` changed for other reasons, is there something unsuitable about `struct DwarfRegMapper` in #902? If it was sufficient, and we're just deferring non-x86 thought for the moment, I'd like to revive `DwarfRegMapper` in a subsequent PR.

cranelift-codegen/src/isa/x86/fde.rs
```diff
@@ -0,0 +1,354 @@
...
+        _ => {
+            panic!("unsupported register bank: {}", bank.name);
+        }
+    }
+}
+
+fn to_cfi(
+    isa: &dyn TargetIsa,
+    change: &FrameLayoutChange,
+    cfa_def_reg: &mut Register,
```
iximeow: I strongly prefer `isa::RegUnit` here for debugging reasons - going backwards from `Register` to `isa::RegUnit`. Does something necessitate this being `gimli::Register`?

cranelift-codegen/src/isa/x86/fde.rs
```diff
@@ -0,0 +1,354 @@
...
+        FrameLayoutChange::RegAt { reg, cfa_offset } => {
+            assert!(cfa_offset % -8 == 0);
+            let cfa_offset = *cfa_offset as i32;
+            let mapped = map_reg(isa, *reg);
+            CallFrameInstruction::Offset(mapped, cfa_offset)
+        }
+        FrameLayoutChange::ReturnAddressAt { cfa_offset } => {
+            assert!(cfa_offset % -8 == 0);
+            let cfa_offset = *cfa_offset as i32;
+            CallFrameInstruction::Offset(X86_64::RA, cfa_offset)
```
iximeow: quibble: if `X86_64::RA` were replaced with `return_address_reg(isa)`, `to_cfi` would not have any x86-specific details anymore

cranelift-codegen/src/isa/x86/fde.rs
```diff
@@ -0,0 +1,354 @@
...
+        func.signature.call_conv == CallConv::Fast
+            || func.signature.call_conv == CallConv::Cold
+            || func.signature.call_conv == CallConv::SystemV
+    );
+
+    assert!(func.frame_layout.is_some(), "expected func.frame_layout");
+    let frame_layout = func.frame_layout.as_ref().unwrap();
+
+    let mut ebbs = func.layout.ebbs().collect::<Vec<_>>();
+    ebbs.sort_by_key(|ebb| func.offsets[*ebb]); // Ensure inst offsets always increase
```
iximeow: :+1:! I'd not thought about basic blocks possibly being out of order in the layout.

filetests/isa/x86/windows_systemv_x64_fde.clif
```diff
@@ -0,0 +1,54 @@
+test fde
```
iximeow: :+1: :+1:! I was extremely unsure how to test that #902 generated expected unwind information, other than checking that Lucet could unwind correctly with those patches. This is great, thank you.

cranelift-codegen/src/isa/x86/fde.rs
```diff
@@ -0,0 +1,354 @@
...
+
+    let (bytes, relocs) = eh_frame.clone().into_vec_and_relocs();
+
+    let unwind_start = sink.len();
+    sink.bytes(&bytes);
+
+    for (off, r) in relocs {
+        sink.reloc(r, off + unwind_start);
+    }
+    let fde_offset = unsafe { ptr::read::<u32>(bytes.as_ptr() as *const u32) } as usize + 4;
```
iximeow: I'm not sure if this just snuck by in review or not, but I'd at the very least want an explanation next to this pointer read with why it will be safe. I'd like to replace this with `u32::from_le_bytes(bytes.as_slice()[..8].try_into())` - since the checks here aren't in a particularly hot path it seems like that shouldn't be an issue?

cranelift-codegen/src/isa/x86/fde.rs
yurydelendik: This serialization targets specific use case of libunwind (in context of JIT), and it requires FDE in specific format. This PR will not address other use cases.

cranelift-codegen/src/isa/x86/fde.rs
yurydelendik: > the assert would just be assert!(isa.name() == "x86"), which seems redundant given that this is in isa/x86/fde.rs.

yes, I just wanted to address the reviewer's comment to complete satisfaction and note that function is indeed x86 specific.

cranelift-codegen/src/isa/x86/fde.rs
yurydelendik: yep, that's a bug

yurydelendik:
> I do have a few comments, but I expect them to be subsequent PRs

@iximeow thank you for review, I'll try to get back to them shortly after bytecodealliance/wasmtime#759 . As @peterhuene noted, there will be more refactoring connected with generalizing unwind information.

cranelift-codegen/src/isa/x86/fde.rs
iximeow: good to know, thank you - I'll be revising #902 to support writing .eh_frame to object files, and I'll make sure to preserve this relocation style.

cranelift-codegen/src/isa/x86/fde.rs
```diff
@@ -0,0 +1,354 @@
...
+            fde.add_instruction((*addr) as u32, cfi);
+        }
+    }
+
+    frames.add_fde(cie_id, fde);
+
+    let mut eh_frame = EhFrame::from(FDEWriter::new());
+    frames.write_eh_frame(&mut eh_frame).unwrap();
+
+    let (bytes, relocs) = eh_frame.clone().into_vec_and_relocs();
```
philipc: Use `eh_frame.0` instead of `eh_frame.clone()`.

cranelift-codegen/src/isa/x86/fde.rs
philipc: This whole thing is fighting against the gimli API a bit. You're using `FrameTable` to write a single CIE and FDE, and then trying to figure out the FDE offset.

How about if gimli made the CIE/FDE write methods public, and then you can call those methods instead of using `FrameTable`, and get the FDE offset simply as `bytes.len()` prior to calling the FDE write?

bnjbvr:
Random drive-by comment for a future PR: can you explicit what the acronym FDE means, at the top of the fde.rs file, for instance?

PR #1324, opened 2020-01-08
Fix clippy warnings in EncodingBits

abrown:
Prior to this I would see the following on `cargo build`:

```
$ cargo build
warning: identical conversion
  --> cranelift-codegen/shared/src/isa/x86/encoding_bits.rs:63:26
   |
63 |         debug_assert_eq!(u8::from(self.rrr()), 0);
   |                          ^^^^^^^^^^^^^^^^^^^^ help: consider removing `u8::from()`: `self.rrr()`
   |
   = note: `#[warn(clippy::identity_conversion)]` on by default
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#identity_conversion

warning: using `clone` on a `Copy` type
  --> cranelift-codegen/shared/src/isa/x86/encoding_bits.rs:64:23
   |
64 |         let mut enc = self.clone();
   |                       ^^^^^^^^^^^^ help: try removing the `clone` call: `self`
   |
   = note: `#[warn(clippy::clone_on_copy)]` on by default
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#clone_on_copy

warning: using `clone` on a `Copy` type
  --> cranelift-codegen/shared/src/isa/x86/encoding_bits.rs:73:23
   |
73 |         let mut enc = self.clone();
   |                       ^^^^^^^^^^^^ help: try removing the `clone` call: `self`
   |
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#clone_on_copy

    Finished dev [unoptimized + debuginfo] target(s) in 0.04s

```

### DIFF
--- a/cranelift-codegen/shared/src/isa/x86/encoding_bits.rs
+++ b/cranelift-codegen/shared/src/isa/x86/encoding_bits.rs
@@ -59,20 +59,18 @@ impl EncodingBits {
 
     /// Returns a copy of the EncodingBits with the RRR bits set.
     #[inline]
-    pub fn with_rrr(self, rrr: u8) -> Self {
-        debug_assert_eq!(u8::from(self.rrr()), 0);
-        let mut enc = self.clone();
-        enc.write(RRR, rrr.into());
-        enc
+    pub fn with_rrr(mut self, rrr: u8) -> Self {
+        debug_assert_eq!(self.rrr(), 0);
+        self.write(RRR, rrr.into());
+        self
     }
 
     /// Returns a copy of the EncodingBits with the REX.W bit set.
     #[inline]
-    pub fn with_rex_w(self) -> Self {
+    pub fn with_rex_w(mut self) -> Self {
         debug_assert_eq!(self.rex_w(), 0);
-        let mut enc = self.clone();
-        enc.write(REX_W, 1);
-        enc
+        self.write(REX_W, 1);
+        self
     }
 
     /// Returns the raw bits.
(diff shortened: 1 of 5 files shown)


cranelift-codegen/shared/src/isa/x86/encoding_bits.rs
```diff
@@ -60,8 +60,8 @@ impl EncodingBits {
     /// Returns a copy of the EncodingBits with the RRR bits set.
     #[inline]
     pub fn with_rrr(self, rrr: u8) -> Self {
```
bjorn3: You can use `mut self` here.

abrown: Good call 👍

cranelift-codegen/src/simple_gvn.rs
```diff
@@ -124,7 +124,8 @@ pub fn do_simple_gvn(func: &mut Function, domtree: &mut DominatorTree) {
             use crate::scoped_hash_map::Entry::*;
             match visible_values.entry(key) {
                 Occupied(entry) => {
-                    debug_assert!(domtree.dominates(*entry.get(), inst, &func.layout));
+                    let dominates = domtree.dominates(*entry.get(), inst, &func.layout);
+                    debug_assert!(dominates);
```
bnjbvr: Out of curiosity, why was it split into two lines? It seems `dominates` would be dead in a non-debug build.

abrown: clippy was complaining about using something mutable inside the `debug_assert!` (IIRC it was `func.layout`); moving the call out of the macro fixed the warning but honestly I didn't give it too much thought. Is there a better way?

philipc: Seems like a false positive to me, see [this reduced case](https://github.com/rust-lang/rust-clippy/issues/4737#issuecomment-573296739), and it's probably better to ignore clippy here if it's wrong.

PR #1325, opened 2020-01-08
When splitting a const, insert prior to the terminal branch group.

### LINKED ISSUE #1159: isplit legalization breaks with basic blocks
```
test compile
target x86_64

function u0:0(i128, i128, i64) -> i128 system_v {
                                ebb0(v0: i128, v1: i128, v2: i64):
                                    trap user0

                                ebb1:
@0005                               v17 = iconst.i128 0
@0006                               v20 = iconst.i128 0
@000a                               trap user0

                                ebb79:
@00dc                               v425 = iconst.i64 0
@00dc                               v426 = icmp_imm eq v425, 1
@00dc                               brnz v426, ebb80
@00dc                               jump ebb85(v20, v17)

                                ebb80:
@00e0                               trap user0

                                ebb85(v462: i128, v874: i128):
@00ed                               trap user0
}

### PR DESCRIPTION
**Pull request by sstangl**
Closes #1159 

Given code like the following, on x86_64, which does not have i128 registers:

    ebb0(v0: i64):
        v1 = iconst.i128 0
        v2 = icmp_imm eq v0, 1
        brnz v2, ebb1
        jump ebb2(v1)

It would be split to:

    ebb0(v0: i64):
        v1 = iconst.i128 0
        v2 = icmp_imm eq v0, 1
        brnz v2, ebb1
        v3, v4 = isplit.i128 v1
        jump ebb2(v3, v4)

But that fails basic-block invariants. This patch changes that to:

    ebb0(v0: i64):
        v1 = iconst.i128 0
        v2 = icmp_imm eq v0, 1
        v3, v4 = isplit.i128 v1
        brnz v2, ebb1
        jump ebb2(v3, v4)

- [x] This has been discussed in issue #1159 .

I decided to move the `isplit` before the terminal branch group to keep the invariants simple: in that issue it was suggested that we solve this by allowing `isplit` instructions within the branch group, but in prior patches it was common need to have to inspect multiple branches in the same terminal group. If we allowed `isplit` there, many places throughout the code would have to learn how to ignore those instructions. Moving the `isplit` prior to all branches is a lower cognitive burden.

- [x] A short description of what this does, why it is needed.

### DIFF
--- a/cranelift-codegen/src/legalizer/split.rs
+++ b/cranelift-codegen/src/legalizer/split.rs
@@ -189,6 +189,17 @@ fn perform_repairs(pos: &mut FuncCursor, cfg: &ControlFlowGraph, mut repairs: Ve
 
             // Split the old argument, possibly causing more repairs to be scheduled.
             pos.goto_inst(inst);
+            #[cfg(feature = "basic-blocks")]
+            {
+                let inst_ebb = pos.func.layout.inst_ebb(inst).expect("inst in ebb");
+
+                // Insert split values prior to the terminal branch group.
+                let dfg = &pos.func.dfg;
+                let canonical = pos.func.layout.canonical_branch_inst(dfg, inst_ebb);
+                if let Some(first_branch) = canonical {
+                    pos.goto_inst(first_branch);
+                }
+            }
             let (lo, hi) = split_value(pos, old_arg, repair.concat, &mut repairs);
 
             // The `lo` part replaces the original argument.
@@ -248,8 +259,8 @@ fn split_value(
             }
         }
         ValueDef::Param(ebb, num) => {
-            // This is an EBB parameter. We can split the parameter value unless this is the entry
-            // block.
+            // This is an EBB parameter.
+            // We can split the parameter value unless this is the entry block.
             if pos.func.layout.entry_block() != Some(ebb) {
                 reuse = Some(split_ebb_param(pos, ebb, num, value, concat, repairs));
             }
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by bjorn3** [comment]
You could use `iconst.i64` +`iconcat` instead of `iconst.i128` for the test.

File: cranelift-codegen/src/legalizer/split.rs [hunk 1]
```diff
@@ -189,6 +189,17 @@ fn perform_repairs(pos: &mut FuncCursor, cfg: &ControlFlowGraph, mut repairs: Ve
 
             // Split the old argument, possibly causing more repairs to be scheduled.
             pos.goto_inst(inst);
+            #[cfg(feature = "basic-blocks")]
+            {
+                let inst_ebb = pos.func.layout.inst_ebb(inst).expect("inst in ebb");
+
+                // Insert split values prior to the terminal branch group.
+                let dfg = &pos.func.dfg;
```
bnjbvr: Is this split into its own variable to avoid borrowing pos.func twice? If not, since it's used only once can we inline it at its use?

filetests/verifier/isplit-bb.clif
```diff
@@ -0,0 +1,28 @@
+test compile
```
bnjbvr: Can you add test annotations to check that the compiled form is what we expect, here, please?

bnjbvr:
Thank you! Merging right now, I'll address the remaining nits as soon as possible.

filetests/verifier/isplit-bb.clif
bnjbvr: Actually we can't check for the split instruction being present in the output, because we can reuse a pair created by concat, it seems.

PR #1332, opened 2020-01-10
Try to reduce CI times with a Rust `*.wat` parser

alexcrichton:
This commit moves the cranelift tests and tools from the `wabt` crate on
crates.io (which compiles the wabt C++ codebase) to the `wat` crate on
crates.io which is a Rust parser for the `*.wat` format. This was
motivated by me noticing that release builds on Windows are ~5 minutes
longer than Linux builds, and local timing graphs showed that `wabt-sys`
was by far the longest build step in the build process.

This commit changes the `clif-util` binary where the `--enable-simd`
flag is no longer respected with the text format as input, since the
`wat` crate has no feature gating. This was already sort of not
respected, though, since `--enable-simd` wasn't consulted for binary
inputs which `clif-util` supports as well. If this isn't ok though then
it should be ok to close this PR!

### DIFF
--- a/src/clif-util.rs
+++ b/src/clif-util.rs
@@ -107,24 +107,6 @@ fn add_debug_flag<'a>() -> clap::Arg<'a, 'a> {
         .help("Enable debug output on stderr/stdout")
 }
 
-fn add_enable_simd_flag<'a>() -> clap::Arg<'a, 'a> {
-    Arg::with_name("enable-simd")
-        .long("enable-simd")
-        .help("Enable WASM's SIMD operations")
-}
-
-fn add_enable_multi_value<'a>() -> clap::Arg<'a, 'a> {
-    Arg::with_name("enable-multi-value")
-        .long("enable-multi-value")
-        .help("Enable WASM's multi-value support")
-}
-
-fn add_enable_reference_types_flag<'a>() -> clap::Arg<'a, 'a> {
-    Arg::with_name("enable-reference-types")
-        .long("enable-reference-types")
-        .help("Enable WASM's reference types operations")
-}
-
 fn add_just_decode_flag<'a>() -> clap::Arg<'a, 'a> {
     Arg::with_name("just-decode")
         .short("t")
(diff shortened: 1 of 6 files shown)


src/wasm.rs
```diff
@@ -103,25 +94,17 @@ fn handle_module(
...
+    let module_binary = if path.to_str() == Some("-") {
+        let stdin = std::io::stdin();
+        let mut buf = Vec::new();
+        stdin
+            .lock()
+            .read_to_end(&mut buf)
+            .map_err(|e| e.to_string())?;
+        wat::parse_bytes(&buf).map_err(|err| format!("{:?}", err))?.into()
+    } else {
+        wat::parse_file(path).map_err(|err| format!("{:?}", err))?
```
alexcrichton: It's worth noting here that `wat` itself has the fallbacks already where if the input file is a wasm binary it won't attempt to parse it as text, it'll just return the raw binary that was read. When I removed the call to `read_to_end` here it made that a dead function, so I inlined the logic for handling `-`

src/clif-util.rs
```diff
@@ -321,9 +321,6 @@ fn main() {
                     rest_cmd.is_present("print-size"),
                     rest_cmd.is_present("time-passes"),
                     rest_cmd.is_present("value-ranges"),
-                    rest_cmd.is_present("enable-simd"),
-                    rest_cmd.is_present("enable-multi-value"),
-                    rest_cmd.is_present("enable-reference-types"),
```
abrown: I think then that we should remove the `add_enable_simd_flag`, `add_enable_multi_value`, `add_enable_reference_types_flag` functions.

alexcrichton: Done!

abrown:
What's the speedup using `wat`?

alexcrichton:
Oh wow that had a much bigger effect than I would have thought!

Comparing a [recent PR build run](https://github.com/bytecodealliance/cranelift/runs/383950038) with [this PR's build run](https://github.com/bytecodealliance/cranelift/runs/384183813) the time decreased from 12m51s  to 9m10s, a 30% time reduction!

PR #1335, opened 2020-01-10
Allow 'clif-util run' to read stdin as intended

### LINKED ISSUE #1004: clif-util no longer reads from stdin by default
```sh
$ cargo run -- run
    Finished dev [unoptimized + debuginfo] target(s) in 0.11s
     Running `target/debug/clif-util run`
Unable to read file: IO error for operation on -: No such file or directory (os error 2)
```

It looks like it's trying to read from the literal file '-' instead of stdin.

zummenix:
This enables `clif-util run -` to work with stdin (seems it was intended initially).

Now errors in `iterate_files` are propagated further.

Closes #1004

### DIFF
--- a/src/run.rs
+++ b/src/run.rs
@@ -10,9 +10,19 @@ use target_lexicon::Triple;
 use walkdir::WalkDir;
 
 pub fn run(files: Vec<String>, flag_print: bool) -> Result<(), String> {
+    let stdin_exist = files.iter().find(|file| *file == "-").is_some();
+    let filtered_files = files
+        .iter()
+        .filter(|file| *file != "-")
+        .map(|file| file.to_string())
+        .collect::<Vec<String>>();
     let mut total = 0;
     let mut errors = 0;
-    for file in iterate_files(files) {
+    let mut special_files: Vec<PathBuf> = vec![];
+    if stdin_exist {
+        special_files.push("-".into());
+    }
+    for file in iterate_files(filtered_files).chain(special_files) {
         total += 1;
         match run_single_file(&file) {
             Ok(_) => {


zummenix:
@bnjbvr Please assign a reviewer.

abrown:
I just merged a PR not 30 seconds ago that may have already done this. Is your change different than https://github.com/bytecodealliance/cranelift/pull/1332/files#r365446123?

zummenix:
@abrown 
AFAIK, yes, they are different.

src/run.rs
```diff
@@ -56,14 +68,14 @@ fn iterate_files(files: Vec<String>) -> impl Iterator<Item = PathBuf> {
...
         })
-        .map(|f| {
-            f.expect("This should not happen: we have already filtered out the errors")
-                .into_path()
+        .map(|r| match r {
+            Ok(d) => Ok(d.into_path()),
+            Err(e) => match e.path() {
+                Some(p) => Ok(p.into()),
+                None => Err(e),
+            },
```
abrown: I was a bit confused reading this but tell me if I understand this correctly: if we pass `"-"` inside `files` then `WalkDir` will give us an error; however, we _can_ turn `"-"` into a valid `PathBuf` by doing `e.path().into()`?

zummenix: Yes, that's correct. We propagate all paths that `WalkDir` wasn't able to deal with to `read_single_file` function. I agree this is a bit clever and confusing but I wasn't able to find a better way.

abrown: What about the following approach: why don't we separate out any strings that we know to have special semantics and chain them on the end of the iterated files? E.g.:
```
let (normal_files, special_files) = drain_specials(files);
for file in iterate_files(normal_files).chain(special_files)
```
Where `drain_specials` could be something like (there's surely a better way):
```
/// This could use `drain_filter` instead but it is a nightly-only experimental API.
fn drain_specials(mut files: Vec<String>) -> (Vec<String>, Vec<String>) {
    let mut normals = vec![];
    let mut specials = vec![];
    for s in files.drain(..) {
        match s.as_str() {
            "-" => specials.push(s),
            _ => normals.push(s),
        }
    }
    (normals, specials)
}
```
The reason I propose something like this is to avoid the catch-and-then-use of `e.path().into()` which feels hard to debug (and to read too, as you mention).

zummenix: Does it make sense to have multiple stdin inputs? For example with `clif-util run - -` user fills in first input and then second. If not, then we can check if stdin arg exist, filter out all the `-` from `files` and `chain` the `-` at the end as you have proposed.

abrown: Yeah, to me deduplicating special things like `-` makes sense.

PR #1336, opened 2020-01-10
Add CI to auto-publish to crates.io on tags

alexcrichton:
This commit updates the publication process for the cranelift suite of
crates. the `publish-all.sh` script was removed in favor of a new Rust
script located at `ci/publish.rs` This script is based off the one in
the [wasm-bindgen repository][wbg]. Additionally publication
instructions have been added to `CONTRIBUTING.md`.

The new process of publication looks like:

* Run `rustc ci/publish.rs && ./publish bump`
* Commit the change and get it merged (either push it or go through a PR)
* Push a tag with the versions bumped

The CI for the last step, the tag push, will automatically publish the
crates once CI has finished.

[wbg]: https://github.com/rustwasm/wasm-bindgen/blob/master/publish.rs


alexcrichton:
I've tested this out locally with [my own suite of cranelift crates](https://crates.io/crates/alexcrichton-cranelift). The publication step [looks like this](https://github.com/alexcrichton/cranelift/runs/384203195) and the PR step [looks like this](https://github.com/alexcrichton/cranelift/runs/384097249).

Note that to merge this we'll need to add a crates.io token to the secret environment variables of this repository. I think the best thing to do is to set up a new user which only controls these crates. That way the token only has access to these crates, no other crates (such as another users's crates if we used one of our own personal tokens). For this purpose I've set up a user https://github.com/cranelift-auto-publish which now has a crates.io account. If everyone's ok with this setup I'll configure the registry tokens and such before this merges.

abrown:
I feel like the two-step process is what is tripping us up right now; is there any way to do the version bump and tagging all in the same PR?

alexcrichton:
Heh well technically CI allows us to do anything so everything is in the realm of possibility! It's largely just a question of what we want this to look like.

We could possibly forgo the tags entirely, and have a scheme where every merge to master compares versions against crates.io and if they changed the crate is published. I personally like the system of tags myself because it's pretty lightweight but still has a deliberate "this is the thing I'm releasing step", but I'm down for implementing whatever!

abrown:
I'll defer to @sunfishcode and @bnjbvr since they will likely be the ones doing this type of stuff 😄.

ci/publish.rs
```diff
@@ -0,0 +1,242 @@
...
+    format!("{}.{}.0", major, minor + 1)
+}
+
+fn publish(krate: &Crate, dry_run: bool, any_failed: &mut bool) {
+    if !CRATES_TO_PUBLISH.iter().any(|s| *s == krate.name) {
+        return;
+    }
+
+    // Skip the simplejit crate since it tries to generate a lock file and may
+    // point to unpublished dependencies. Only skip it in dry run mode though.
```
sunfishcode: To my knowledge, cranelift-simplejit doesn't do anything special wrt cargo, and doesn't have any unusual dependencies. Can you point to what it is about this crate that causes problems?

ci/publish.rs
```diff
@@ -0,0 +1,242 @@
...
+            line.contains("dependencies")
+        } else {
+            is_deps
+        };
+
+        for other in crates {
+            if !is_deps || !line.starts_with(&format!("{} ", other.name)) {
+                continue;
+            }
+            if !line.contains(&other.version) {
```
sunfishcode: This is unlikely, but could you append and prepend `"` to the version string, so that `1.1.1` doesn't match `1.1.11`?

.github/workflows/main.yml
```diff
@@ -147,3 +148,26 @@ jobs:
...
+      run: rustup update stable && rustup default stable
+
+    - name: Configure env vars
+      run: |
+        echo ::set-env name=NO_DRY_RUN::1
+        echo ::set-env name=CARGO_REGISTRY_TOKEN::${{ secrets.CARGO_REGISTRY_TOKEN }}
+      if: github.event_name == 'push' && startsWith(github.event.ref, 'refs/tags')
+
+    - name: Publish
+      run: rustc ci/publish.rs && ./publish publish
```
sunfishcode: Do we need to do anything special to tell it to only publish if this is the master branch, and not some random feature branch?

CONTRIBUTING.md
```diff
@@ -151,3 +151,19 @@ review, although it is always preferred to have one.
...
+Publication of Cranelift crates require half manual intervention and half
+automation. The process to release a new version of Cranelift looks like:
+
+1. `cd` into the `cranelift` checkout root
+2. Execute `rustc ci/publish.rs`
+3. Execute `./publish bump`
+4. Review the changes (`git diff`), commit the changes, and send a PR.
+5. Upon merge, someone with push access needs to execute the following commands:
+  * `git tag vA.B.C abcdefghijk` - where `abcdefghijk` is the sha of the merge
+  * `git push origin-rw --tags`
```
sunfishcode: Would it be difficult to add a `publish` subcommand to do this part too? Assuming the merge doesn't get a new sha, it seems like we should be able to pick up the sha from the top of the local repo.

sunfishcode (commented):
Do we need to special-case the `witx` crate, which we pull in as a submodule and don't publish though this repo?

ci/publish.rs
alexcrichton: It's sort of a weird part of Cargo, but if Cargo for some crates decides to put a `Cargo.lock` in the package, meaning it needs to resolve the crate graph. If a crate depends on something not published on crates.io, however, as would be the case in the "dry run" mode which happens on PRs, then this step will fail because `Cargo.lock` cannot be generated. I hit this locally when I prefixed all the crates with `alexcrichton-*` and nothing existed on crates.io.

Cargo decides to put `Cargo.lock` in crates that contains examples or binaries, and simplejit happened to be the only cranelift crate that met that criteria (had an example)

.github/workflows/main.yml
alexcrichton: This'll only actually do things for tags (previous step enables), and only for tags that start with a `v` (I'm not great on regexes). Do you think it should be limited further though?

CONTRIBUTING.md
alexcrichton: Sure yeah, I'll edit this in

alexcrichton:
I'm gonna close this and we can recreate it later if still necessary on the wasmtime repo.

PR #1337, opened 2020-01-11
Correctly count the number of wasm parameters.

sunfishcode:
Following up on #1329, this further replaces `num_normal_params` with a function
which calls `is_wasm_parameter` to correctly count the number of wasm
parameters a function has.

- [X] This has been discussed in issue #1329.
- [X] See the commit message above.
- [X] No user-visible behavior, and we'll mainly test this in wasmtime.

### DIFF
--- a/cranelift-wasm/src/environ/dummy.rs
+++ b/cranelift-wasm/src/environ/dummy.rs
@@ -201,8 +201,8 @@ impl<'dummy_environment> TargetEnvironment for DummyFuncEnvironment<'dummy_envir
 }
 
 impl<'dummy_environment> FuncEnvironment for DummyFuncEnvironment<'dummy_environment> {
-    fn is_wasm_parameter(&self, func: &ir::Function, index: usize) -> bool {
-        func.signature.params[index].purpose == ir::ArgumentPurpose::Normal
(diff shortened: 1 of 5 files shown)


cranelift-wasm/src/environ/dummy.rs
```diff
@@ -201,8 +201,8 @@ impl<'dummy_environment> TargetEnvironment for DummyFuncEnvironment<'dummy_envir
 }
 
 impl<'dummy_environment> FuncEnvironment for DummyFuncEnvironment<'dummy_environment> {
-    fn is_wasm_parameter(&self, func: &ir::Function, index: usize) -> bool {
-        func.signature.params[index].purpose == ir::ArgumentPurpose::Normal
+    fn is_wasm_parameter(&self, signature: &ir::Signature, index: usize) -> bool {
+        signature.params[index].purpose == ir::ArgumentPurpose::Normal
```
bnjbvr: I also realized it would be nicer for embedders that this implementation be the default impl in FuncEnvironment (eg I doubt spidermonkey will require special behavior here). Can you move the code there, please?

sunfishcode: Done!

Issue #1170, opened 2019-10-25
Integrate clippy checking in CI build


abrown:
Please try to describe precisely what you would like to do in Cranelift and/or
expect from it. You can answer the questions below if they're relevant and
delete this text before submitting. Thanks for opening an issue!

- What is the feature or code improvement you would like to do in Cranelift?
After merging #1169 (and seeing #1168), it occurred to me that we should output clippy warnings as a part of `test-all.sh`. I'm not sure we should fail the build (though this could be an option) but this would make non-clippy-approved additions more visible.

- What is the value of adding this in Cranelift?
It should result in cleaner, more standard code in the long run.

- Do you have an implementation plan, and/or ideas for data structures or
  algorithms to use?
Add `cargo clippy` to `test-all.sh`.

- Have you considered alternative implementations? If so, how are they better
  or worse than your proposal?
No.

abrown:
@bnjbvr, this seems like a relatively no-brainer to me but I guess the question is: do we fail the build on clippy warnings or not?

fitzgen:
In general, I feel like it is good to fix clippy warnings (or ignore them in specific cases) but rarely useful to hard gate merging PRs on being clippy clean.

peterhuene:
I agree that it shouldn't be required, but perhaps an optional CI leg that may fail and bring some visibility to the accumulation of the warnings?  It looks like it's been some time since they were last addressed and I think it's important for upstream crates to be free from linter warnings so that downstream crates don't get inundated with unfixable warnings.

bnjbvr:
Agreed it's a nice idea in general and that it shouldn't block PR from being merged, though. We could add an optional job in the CI, that would have a `continueOnError` directive, as we've recently done for the nightly fuzzing job.

That being said, I question just a bit the utility of such CI jobs: there's no real incentive to fix them, apart from wanting to maintain great quality. It's a nice source of easily-fixable bugs, though.

Another approach would be to remove the `clippy.toml` file that's at the top-level: most IDEs see it and enable clippy warnings by default in the IDE. Removing the file would hide IDE warnings, and thus remove the cognitive load associated to them.

My personal preference goes to a `continueOnError` CI job right now.

abrown:
[...]

PR #1340, opened 2020-01-13
Fix remaining clippy warnings

abrown:
- [x] This has been discussed in issue #1170
- [x] A short description of what this does, why it is needed: this fixes the remaining clippy warnings that I could see from running `cargo clean && cargo clippy` (is there a better way to do this?)
- [x] This PR contains no test cases; no functionality should change
- [x] A reviewer from the core maintainer team has been assigned for this PR

### DIFF
--- a/cranelift-codegen/src/isa/x86/fde.rs
+++ b/cranelift-codegen/src/isa/x86/fde.rs
@@ -10,7 +10,6 @@ use gimli::write::{
     FrameDescriptionEntry, FrameTable, Result, Writer,
 };
 use gimli::{Encoding, Format, LittleEndian, Register, X86_64};
-use std::ptr;
 
 pub type FDERelocEntry = (FrameUnwindOffset, Reloc);
 
@@ -196,7 +195,7 @@ pub fn emit_fde(func: &Function, isa: &dyn TargetIsa, sink: &mut dyn FrameUnwind
             assert!(last_offset <= address_offset);
             if let Some(cmds) = frame_layout.instructions.get(&inst) {
                 for cmd in cmds.iter() {
-                    changes.push((address_offset, cmd.clone()));
+                    changes.push((address_offset, *cmd));
                 }
             }
             last_offset = address_offset;
@@ -252,7 +251,9 @@ pub fn emit_fde(func: &Function, isa: &dyn TargetIsa, sink: &mut dyn FrameUnwind
     for (off, r) in relocs {
         sink.reloc(r, off + unwind_start);
     }
-    let fde_offset = unsafe { ptr::read::<u32>(bytes.as_ptr() as *const u32) } as usize + 4;
+
+    let cie_len = u32::from_le_bytes(bytes.as_slice()[..4].try_into().unwrap());
+    let fde_offset = cie_len as usize + 4;
     sink.set_entry_offset(unwind_start + fde_offset);
 
     // Need 0 marker for GCC unwind to end FDE "list".
--- a/cranelift-codegen/meta/src/cdsl/instructions.rs
+++ b/cranelift-codegen/meta/src/cdsl/instructions.rs
@@ -1346,7 +1346,7 @@ mod test {
         let type1 = TypeSetBuilder::new().ints(8..64).build();
         let in1 = OperandKindFields::TypeVar(TypeVar::new("a", "...", type1));
         let inst = build_fake_instruction(vec![in1], vec![]);
-        inst.bind(LaneType::IntType(I32));
+        inst.bind(LaneType::Int(I32));
     }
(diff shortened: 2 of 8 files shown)


abrown:
@philipc, based on your comment https://github.com/bytecodealliance/cranelift/pull/1324#discussion_r365509394 I added 1195612 as well.

abrown:
@yurydelendik, the clippy error comes at line 255:
```
error: casting from `*const u8` to a more-strictly-aligned pointer (`*const u32`) (1 < 4 bytes)
   --> cranelift-codegen/src/isa/x86/fde.rs:255:48
    |
255 |     let fde_offset = unsafe { ptr::read::<u32>(bytes.as_ptr() as *const u32) } as usize + 4;
    |                                                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: `#[deny(clippy::cast_ptr_alignment)]` on by default
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#cast_ptr_alignment
```
How do you think this should be fixed? Do we know that `bytes` is aligned to a 4-byte boundary or can we get there?

abrown:
Or perhaps https://doc.rust-lang.org/std/vec/struct.Vec.html#method.align_to and assert that the prefix has zero length?

yurydelendik:
> error: casting from `*const u8` to a more-strictly-aligned pointer (`*const u32`) (1 < 4 bytes)
>    --> cranelift-codegen/src/isa/x86/fde.rs:255:48
> How do you think this should be fixed? Do we know that `bytes` is aligned to a 4-byte boundary or can we get there?

There was a thread probably related to this: https://github.com/bytecodealliance/cranelift/pull/1320#discussion_r366016546 -- `u32::from_le_bytes(bytes.as_slice()[..8].try_into()` may not need alignment?

abrown:
@yurydelendik, @iximeow: is 0252c77 what you meant? The tests will likely fail (they fail on my system) so I don't think I understood exactly what was intended here.

iximeow:
@abrown my point was to remove the `unsafe ptr::read` entirely, and replace it with the `from_le_bytes` expression. mentally conflating the pointer size and `u32` size is why i wrote the wrong slice length `[..8]` instead of `[..4]` :)

```
let fde_offset = u32::from_le_bytes(bytes.as_slice()[..4]
    .try_into()
    .expect("4 bytes at the beginning of the allocated FDE") as usize + 4;
```
is what i was thinking should be a good substitute. i'm pretty sure we can actually be confident there will be a `u32` of bytes there, so the `expect` ought to never fail.

cranelift-codegen/src/isa/x86/fde.rs
```diff
@@ -252,7 +251,13 @@ pub fn emit_fde(func: &Function, isa: &dyn TargetIsa, sink: &mut dyn FrameUnwind
...
         sink.reloc(r, off + unwind_start);
     }
-    let fde_offset = unsafe { ptr::read::<u32>(bytes.as_ptr() as *const u32) } as usize + 4;
+
+    let fde_first_u32 = u32::from_le_bytes(
+        bytes.as_slice()[..4]
+            .try_into()
+            .expect("4 bytes at the beginning of the allocated FDE"),
+    );
+    let fde_offset = fde_first_u32 as usize + 4;
```
abrown: @iximeow, @yurydelendik: here it is! (With an extra variable assignment)

iximeow: @yurydelendik Do you recall what this u32 specifically is? It's a count of FDEs or something, right? It'd be nice if we could use that name instead of `fde_first_u32`.

yurydelendik: As discussed [in previous PR](https://github.com/bytecodealliance/cranelift/pull/1320#discussion_r366607644), gimli does not provide right API to get all data needed. It is a field (u32) that show length of the CIE, so the len and 4 u32 bytes is offset of the FDE structure that follows CIE.

abrown: Should I rename `fde_first_u32` as `cie_len`?

philipc: Nit: the expect message is wrong. That `try_into` isn't checking there were 4 bytes at the beginning of `bytes`. It's checking that the slice passed to `try_into` is 4 bytes, which will always be true because we indexed with a range of `[..4]`. If there were less than 4 bytes, then there would be a panic on the slice indexing, not on the `try_into().expect()`. IMO using `unwrap` is fine in this case.

yurydelendik:
Thank you for the patch

PR #1343, opened 2020-01-14
Fix narrow_icmp_imm

jmkrauz:
  here.
This is a really small change.
- [ ] A short description of what this does, why it is needed; if the
  first.
This fixes small bug caused by inaccurate use of parentheses in `narrow_icmp_imm`.
No tests.
@bnjbvr


jmkrauz:
I have added a legalizer test. At first I wanted to create a run test for `icmp_imm.i128`, but I have noticed that `imm` in `ir::InstructionData` has type `ir::immediates::Imm64`, so it won't work for I128.

PR #1346, opened 2020-01-14
Add 8x16 shift legalizations for x86 SIMD

abrown:
- [x] A short description of what this does, why it is needed: I did not implement the `i8x16` shifts previously because there was discussion about whether they should be included in the spec (https://github.com/WebAssembly/simd/issues/117). I still feel they should not be included due to the large overhead they impose on x86 but I am (rather unwillingly) implementing them here so we can get additional performance measurements.

### DIFF
--- a/cranelift-codegen/src/isa/x86/enc_tables.rs
+++ b/cranelift-codegen/src/isa/x86/enc_tables.rs
@@ -6,7 +6,7 @@ use crate::cursor::{Cursor, FuncCursor};
 use crate::flowgraph::ControlFlowGraph;
 use crate::ir::condcodes::{FloatCC, IntCC};
 use crate::ir::types::*;
-use crate::ir::{self, Function, Inst, InstBuilder};
+use crate::ir::{self, Function, Inst, InstBuilder, MemFlags};
 use crate::isa::constraints::*;
 use crate::isa::enc_tables::*;
 use crate::isa::encoding::base_size;
@@ -16,6 +16,7 @@ use crate::isa::{self, TargetIsa};
 use crate::predicates;
 use crate::regalloc::RegDiversions;
 
+use core::iter;
 use cranelift_codegen_shared::isa::x86::EncodingBits;
 
 include!(concat!(env!("OUT_DIR"), "/encoding-x86.rs"));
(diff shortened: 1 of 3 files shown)


cranelift-codegen/src/isa/x86/enc_tables.rs
```diff
@@ -1236,3 +1237,64 @@ fn convert_ineg(
...
+                // Since x86 does not have a lowering for 8x16 shifts and some of the SIMD spec
+                // committee wants to include this instruction, this non-optimal lowering is
+                // provided (see discussion at https://github.com/WebAssembly/simd/issues/117).
+                // It will use a 16x8 shift and mask off the necessary bits from a table of
+                // constants. This may cause latency due to memory access if the constant table is
+                // not in cache.
+                let shift_index = pos.ins().bitcast(I64X2, arg1);
+                let bitcast_vector = pos.ins().raw_bitcast(I16X8, arg0);
+                let shifted_vector = pos.ins().x86_psll(bitcast_vector, shift_index);
+                let mask_base = pos.ins().iconst(I64, 42); // FIXME this needs to be the base pointer of the constant table
```
abrown: @sunfishcode, @bnjbvr: I don't know how to get a pointer to the table of constants but I need something like that in order to load the mask. Is there a different way to do this? I looked at `binemit.rs::const_disp4` and it is not clear to me if/how at this point (when I'm doing legalizations) I can get an offset to the constant table.

abrown: @sunfishcode, here's my current thinking on this:
 - I considered whether having some sort of symbol (unrelated to globals) in the CL IR like `$func.rodata` would make this work; if that symbol could have a known address at compile time then I could use that and constant offsets to get at runtime-determined constants.
 - But can the address of that symbol be known at compile time? I think it can if it is function-related (not global) and relative to the PC: 1) once we know how long the function will be, [calculate all of the offsets of jump tables, constants, and (now) symbols](https://github.com/bytecodealliance/cranelift/blob/cabbacbe63477fcc7d1f1e36717d84f08f749a98/cranelift-codegen/src/binemit/relaxation.rs#L130-L143), and 2) use functions like `jt_disp4`, `const_disp4`, and (now) `symbol_disp4` in `binemit.rs` to pull out those calculated offsets when we emit the code.
 - Looking at this from distance, we would be doing the same type of thing three times and the code is fragile due to the distance between when the offsets are assigned (`relax_branches`), referenced in code (`binemit.rs`), and actually populated with data (`emit_function`). Perhaps we add an abstraction to `Function` like `RelativeOffsets` that would map any of these things (jump table, constant, symbol) to their offset; it would only be usable after some new counting pass (removing that logic from `relax_branches`), would be queryable when encoding instructions and would know how to lay out the actual data in `emit_function`

alexcrichton:

PR #1349, opened 2020-01-14
Improve issue template: make questions headings, add space for writing, make questions single-line

abrown:


philipc:
I didn't know the questions themselves were meant to be left in the issue text; I thought it was saying to delete them. But if they are, how about using short headings and leaving the verbose questions out of the resulting issue? I find that they make the issues too cluttered. So something like:
```
#### Feature

<!-- What is the feature or code improvement you would like to do in Cranelift? -->

#### Benefit

<!-- What is the value of adding this in Cranelift? -->

#### Implementation

<!-- Do you have an implementation plan, and/or ideas for data structures or algorithms to use? -->

#### Alternatives

<!-- Have you considered alternative implementations? If so, how are they better or worse than your proposal? -->
```

bnjbvr:
Thanks! It's a good idea to revisit it. I like @philipc's template, in that it's a bit shorter and still includes detailed questions that would get removed once the issue is submitted. What do you think, @abrown ?

abrown:
@philipc, @bnjbvr sounds good to me!

PR #1351, opened 2020-01-15
Add initial Cranelift interpreter functionality

abrown:
- [x] This has been discussed in meetings and chat messages.
- [x] A short description of what this does, why it is needed: this implements a rudimentary interpreter for Cranelift IR; only a subset of instructions are currently implemented and implementation gaps are marked with `TODO` or `FIXME`. This PR is a RFC on the interpreter design; any feedback is welcome.

### DIFF
--- a/cranelift-interpreter/src/frame.rs
+++ b/cranelift-interpreter/src/frame.rs
@@ -1,0 +1,123 @@
+//! Implements a call frame (activation record) for the Cranelift interpreter.
+
+use crate::value::Value;
+use cranelift_codegen::ir::{Function, Value as ValueRef};
+use std::collections::HashMap;
+
+/// Holds the mutable elements of an interpretation. At some point I thought about using
+/// Cell/RefCell to do field-level mutability, thinking that otherwise I would have to
+/// pass around a mutable object (for inst and registers) and an immutable one (for function,
+/// could be self)--in the end I decided to do exactly that but perhaps one day that will become
+/// untenable.
+#[derive(Debug)]
+pub struct Frame<'a> {
+    /// The currently executing function.
+    pub function: &'a Function,
+    /// The current mapping of SSA value-references to their actual values.
+    registers: HashMap<ValueRef, Value>,
+}
+
+impl<'a> Frame<'a> {
+    /// Construct a new frame for a function. This allocates a slot in the hash map for each SSA
+    /// `Value` (renamed as `ValueRef` here) which should mean that no additional allocations are
+    /// needed while interpreting the frame.
+    pub fn new(function: &'a Function) -> Self {
+        Self {
+            function,
+            registers: HashMap::with_capacity(function.dfg.num_values()),
+        }
+    }
+
+    pub fn with_parameters(mut self, parameters: &[ValueRef], values: &[Value]) -> Self {
+        for (n, v) in parameters.iter().zip(values) {
+            self.registers.insert(*n, v.clone());
+        }
+        self
+    }
+
+    /// Retrieve the actual value associated with an SSA reference.
+    #[inline]
+    pub fn get(&self, name: &ValueRef) -> &Value {
+        self.registers
+            .get(name)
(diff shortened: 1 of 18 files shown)


cranelift-interpreter/src/interpreter.rs
```diff
@@ -0,0 +1,296 @@
...
+
+    fn iconst(&self, frame: &mut Frame, imm: Imm64, r: ValueRef) {
+        frame.set(r, Value::Int(imm.into()));
+    }
+
+    fn bconst(&self, frame: &mut Frame, imm: bool, r: ValueRef) {
+        frame.set(r, Value::Bool(imm));
+    }
+
+    //noinspection RsUnresolvedReference
```
bjorn3: What is the purpose of this?

cranelift-interpreter/src/lib.rs
```diff
@@ -0,0 +1,10 @@
+//! Cranelift IR interpreter.
+//!
+//! This module is a WIP project for interpreting Cranelift IR.
```
bjorn3: I am afraid this won't be removed when it isn't WIP anymore.

cranelift-interpreter/src/value.rs
```diff
@@ -0,0 +1,45 @@
...
+//!
+//! Each variant of `Value` may implement traits of `std::ops` for use in `interpreter`; e.g.,
+//! `Value::Int` has an implementation (a match arm) of `Add`.
+
+use std::ops::{Add, Sub};
+
+#[derive(Clone, Debug, PartialEq)]
+pub enum Value {
+    Bool(bool),
+    Int(i64),
```
bjorn3: I think this should be `u64` instead. In any case it is important to not forget correct zero/sign extension.

bjorn3 (changes requested):

cranelift-interpreter/src/interpreter.rs
```diff
@@ -0,0 +1,296 @@
...
+
+    fn block(&self, frame: &mut Frame, ebb: Ebb) -> Result<ControlFlow, Trap> {
+        debug!("Block: {}", ebb);
+        for inst in frame.function.layout.ebb_insts(ebb) {
+            match self.inst(frame, inst)? {
+                ControlFlow::Continue => continue,
+                ControlFlow::ContinueAt(ebb, old_names) => {
+                    let new_names = frame.function.dfg.ebb_params(ebb);
+                    frame.rename(&old_names, new_names);
+                    return self.block(frame, ebb); // TODO check that TCO happens
```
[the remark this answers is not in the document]
abrown: tail-call optimization

cranelift-interpreter/src/interpreter.rs
abrown: IDE hint; will remove

cranelift-interpreter/src/value.rs
abrown: You're probably right; let me think some more about this.

cranelift-interpreter/src/frame.rs
```diff
@@ -0,0 +1,123 @@
...
+
+/// Holds the mutable elements of an interpretation. At some point I thought about using
+/// Cell/RefCell to do field-level mutability, thinking that otherwise I would have to
+/// pass around a mutable object (for inst and registers) and an immutable one (for function,
+/// could be self)--in the end I decided to do exactly that but perhaps one day that will become
+/// untenable.
+#[derive(Debug)]
+pub struct Frame<'a> {
+    /// The currently executing function.
+    pub function: &'a Function,
```
abrown: Is this better as a `FuncRef`? I used `&'a Function` here because it was convenient but it may be more consistent and (possibly?) efficient to use a 4-byte `FuncRef` here instead. (I probably need to look again at how efficient it is to use `FuncRef` to index into an `Environment`).

bjorn3:
Close this?

PR #1352, opened 2020-01-16
Use `is_wasm_parameter` in translating wasm calls

alexcrichton:
Added in #1329 it's now possible for multiple parameters to be non-wasm
parameters, so the previous `param_types` method is no longer suitable
for acquiring all wasm-related parameters, rather then `FuncEnvironment`
must be consulted. This removes usage of `param_types()` as a method
from the wasm translation and instead adds a custom method inline for
filtering the parameters based on `is_wasm_parameter`.

### DIFF
--- a/cranelift-wasm/src/code_translator.rs
+++ b/cranelift-wasm/src/code_translator.rs
@@ -1945,10 +1945,7 @@ pub fn bitcast_arguments(
 /// A helper to extract all the `Type` listings of each variable in `params`
 /// for only parameters the return true for `is_wasm`, typically paired with
 /// `is_wasm_return` or `is_wasm_parameter`.
-pub fn wasm_param_types(
-    params: &[ir::AbiParam],
-    is_wasm: impl Fn(usize) -> bool,
-) -> Vec<Type> {
+pub fn wasm_param_types(params: &[ir::AbiParam], is_wasm: impl Fn(usize) -> bool) -> Vec<Type> {
     let mut ret = Vec::with_capacity(params.len());
     for (i, param) in params.iter().enumerate() {
         if is_wasm(i) {
--- a/cranelift-wasm/src/code_translator.rs
+++ b/cranelift-wasm/src/code_translator.rs
@@ -1947,7 +1947,7 @@ pub fn bitcast_arguments(
 /// `is_wasm_return` or `is_wasm_parameter`.
 pub fn wasm_param_types(
     params: &[ir::AbiParam],
-    mut is_wasm: impl FnMut(usize) -> bool,
+    is_wasm: impl Fn(usize) -> bool,
 ) -> Vec<Type> {
     let mut ret = Vec::with_capacity(params.len());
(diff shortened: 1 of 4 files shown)


abrown:
I don't understand a couple of things: 
 - from #1329, why isn't `is_wasm_parameter` called `is_normal_parameter`?
 - why add a new method instead of just changing `param_types` and `return_types` to use this `is_wasm_parameter/is_normal_parameter`?

sunfishcode:
It's `is_wasm_parameter` because it isn't counting the number of `ArgumentPurpose::Normal` parameters. Wasmtime is now passing extra arguments which are `Normal` but are not wasm parameters, so we added `is_wasm_parameter` to the `FuncEnvironment` trait to let it tell cranelift-wasm which parameters need to be translated as wasm locals.

I think we can just remove the `param_types` function altogether now. We could remove `return_types` too, if we added a similar `is_wasm_return` to `FuncEnvironment`, which we don't need right now, but would keep things consistent.

alexcrichton:
Sure thing, updated!

cranelift-wasm/src/code_translator.rs
```diff
@@ -1932,3 +1941,19 @@ pub fn bitcast_arguments(
         }
     }
 }
+
+/// A helper to extract all the `Type` listings of each variable in `params`
+/// for only parameters the return true for `is_wasm`, typically paired with
+/// `is_wasm_return` or `is_wasm_parameter`.
+pub fn wasm_param_types(
+    params: &[ir::AbiParam],
+    mut is_wasm: impl FnMut(usize) -> bool,
```
sunfishcode: Can you comment on why this is `mut` and `Mut`? The ultimate function should be a non-mutating predicate.

alexcrichton: Oh this is mostly just a Rust idioms thing where you idiomatically take `FnMut` since it's the most flexible. For brevity and simplicity though I've switched this to `Fn`, since `FnMut` wasn't actually needed.

cranelift-wasm/src/environ/spec.rs
```diff
@@ -144,6 +144,12 @@ pub trait FuncEnvironment: TargetEnvironment {
         signature.params[index].purpose == ir::ArgumentPurpose::Normal
     }
 
+    /// Is the given return of the given function a wasm-level parameter, as
+    /// opposed to a hidden parameter added for use by the implementation?
+    fn is_wasm_return(&self, signature: &ir::Signature, index: usize) -> bool {
+        signature.returns[index].purpose == ir::ArgumentPurpose::Normal
+    }
```
abrown: > It's is_wasm_parameter because it isn't counting the number of ArgumentPurpose::Normal parameters. 

@sunfishcode, so the above is a default implementation that embedders are supposed to override?

alexcrichton:
Would it be ok to get a release to land https://github.com/bytecodealliance/wasmtime/pull/789?

Issue #1357, opened 2020-01-19
Show existing entries in Switch


jyn514:
- What is the feature or code improvement you would like to do in Cranelift?

I would like [`Switch`](https://docs.rs/cranelift-frontend/0.56.0/cranelift_frontend/struct.Switch.html) to have a new method `entries()` added which returns an immutable list of the existing entries.

- What is the value of adding this in Cranelift?

This would allow me to check in the codegen module whether a `case` expression has already been added, instead of having to frontload it in the parser (which for various reasons would be possible but annoying).

- Do you have an implementation plan, and/or ideas for data structures or
  algorithms to use?

```rust
fn entries(&self) -> &HashMap<EntryIndex, Ebb> {
    &self.cases
}
```

- Have you considered alternative implementations? If so, how are they better
  or worse than your proposal?

Alternatively, Cranelift could `impl Deref<Target = HashMap<...>>` for Switch. This would also provide more methods at the same time.

Cranelift could also choose to do nothing at all.

jyn514:
Added in https://github.com/bytecodealliance/cranelift/pull/1358

PR #1358, opened 2020-01-19
add `entries()` API for Switch

jyn514:
- [x] This has been discussed in issue #1357
- [x] A short description of what this does, why it is needed;

This allows library users to see what entries already exist in a switch.

N/A

I am not sure who should review this PR. r? @bnjbvr

### DIFF
--- a/cranelift-frontend/src/switch.rs
+++ b/cranelift-frontend/src/switch.rs
@@ -258,6 +258,14 @@ impl Switch {
     }
 }
 
+impl core::ops::Deref for Switch {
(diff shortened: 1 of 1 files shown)


bjorn3:
I would prefer an `entries` method like you suggested in the issue instead. Otherwise it is easy to accidentially use a method from the internal `HashMap`. This will also polute autocompletion suggestions and the docs with every single `HashMap` method.

jyn514:
I changed it to entries, that makes sense. Users can still have all the `len()`, `is_empty()`, etc. functions by calling `entries()` first.

PR #1359, opened 2020-01-21
Crate directory reorganization + update fuzzing-related deps

fitzgen:
Crate directory reorg fixes #956

Second and third commits update to new fuzzer crates and also stop using a git dependency for `binaryen` since it's no longer necessary.


bnjbvr:
Thanks! I'll make sure to do a final call for comments on the Zulip and our Monday meeting, but I'm all in for doing this! In the meanwhile, it'll need to be rebased. Cheers!

bnjbvr (approved):
LGTM!

PR #1361, opened 2020-01-22
Update gimli to 0.20

bjorn3:
  here. Small change
  first. This updates the gimli dependency to the latest version. This removes a duplicate dependency from cg_clif.
- [x] This PR contains test cases, if meaningful. N/A


abrown:
I think there's another dependency on `gimli` in `cranelift-filetests` for some reason? Should we update that one as well?

bjorn3:
Updated in cranelift-filetests too.

PR #1362, opened 2020-01-23
Wasm: Typed select and nullref translation

eqrion:
The reftypes proposal has introduced a typed select instruction to ensure efficient validation in the future with complex subtyping relationships. In addition, a proper 'nullref' type has been added. Both of these changes are supported in `wasmparser` but there are some additional updates needed to `cranelift-wasm`.

### DIFF
--- a/cranelift-wasm/src/code_translator.rs
+++ b/cranelift-wasm/src/code_translator.rs
@@ -132,6 +132,13 @@ pub fn translate_operator<FE: FuncEnvironment + ?Sized>(
             let (arg1, arg2, cond) = state.pop3();
             state.push1(builder.ins().select(cond, arg1, arg2));
         }
+        Operator::TypedSelect { ty: _ } => {
+            // We ignore the explicit type parameter as it is only needed for
+            // validation, which we require to have been performed before
+            // translation.
+            let (arg1, arg2, cond) = state.pop3();
+            state.push1(builder.ins().select(cond, arg1, arg2));
+        }
         Operator::Nop => {
             // We do nothing
         }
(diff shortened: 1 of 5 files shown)


cranelift-wasm/src/code_translator.rs
```diff
@@ -132,6 +132,10 @@ pub fn translate_operator<FE: FuncEnvironment + ?Sized>(
             let (arg1, arg2, cond) = state.pop3();
             state.push1(builder.ins().select(cond, arg1, arg2));
         }
+        Operator::TypedSelect { ty: _ } => {
+            let (arg1, arg2, cond) = state.pop3();
+            state.push1(builder.ins().select(cond, arg1, arg2));
```
fitzgen: Just to make sure I understand correctly: the reason we don't care about `ty` is because we already validated the wasm (either in wasmtime or spidermonkey) and know that it is a supertype of both `arg1` and `arg2`, and the `builder.ins().select(..)` just treats them all reference types as `r32` or `r64`, right?

wasmtests/select.wat
```diff
@@ -0,0 +1,15 @@
...
+  	i32.const 24
+  	i32.const 1
+  	select
+  )
+
+  (func $typed-select (result anyref)
+  	ref.null
+  	ref.null
+  	i32.const 1
+  	select (result anyref)
```
fitzgen: We should also add a test like
```wat
(func (param anyref) (result anyref)
    ref.null
    local.get 0
    i32.const 1
    select (result anyref))
```

to exercise when the values are distinct subtypes of the typed select's result type. Would be even better if we had enough support for `ref.func` that we could have turn either a `nullref` or a `funcref` into an `anyref` but I don't think we are there yet, are we?

cranelift-wasm/src/code_translator.rs
eqrion: Yes, `wasmparser` itself contains validation code which will check that `arg1` and `arg2` are subtypes of the specified type. Spidermonkey uses its own equivalent code. And after validation, there's not a need for a stricter type in Cranelift IR, so it all becomes R32/R64.

I'll add a comment to this effect in the code here.

wasmtests/select.wat
eqrion: Good idea, I'll add that test.

I'm not sure if we can test `funcref` well. Right now the DummyEnvironment used in wasmtests will convert a wasm `ref.func` into a clif `r32/64.null`. So while it would test wasmparser's validation (which has it's own tests), the CL IR would be the same as the other test. (that's if I understand the purpose of wasmtests correctly)

Issue #1303, opened 2019-12-20
Complete the transition to Basic Blocks


sstangl:
Basic blocks have been default-enabled for a while. Speaking with @bnjbvr, the project has decided to keep moving forward on replacing Ebbs with Blocks.

The remaining points of work to finish that transition are:

- [x] Fixing https://github.com/bytecodealliance/cranelift/issues/1159 (thanks @bjorn3)
- [x] Removing the "basic-blocks" feature-gate on the relevant code.
- [ ] Mass-renaming "ebb" to "block", etc., throughout the code and the documentation.

However, I think we should keep something that is effectively Ebbs as an internal detail for the wasm IR to CLIF IR translation. They create a de-facto list of blocks that are jump targets, which is a useful optimization for the efficient algorithm we use to convert to SSA.

bjorn3:
https://github.com/bytecodealliance/cranelift/issues/1159 is not yet fixed.

sstangl:
Thanks, I'll fix that first.

eqrion:
I'm going write a commit to do the mass rename now. I believe the plan is to rename to `Block`.

So for example:
```
Ebb -> Block
EbbNode -> BlockNode
EbbData -> BlockData
ebb -> block
```

One potential issue is that `ebb` is used extensively throughout the codebase as a short variable name, and renaming to `block` will increase the identifier length by two. In my opinion this could increase readability, but might be too verbose for some peoples taste.

Let me know if anyone has objections or a different proposal for naming.

bjorn3:
Maybe use `bb` and `Bb`? Just dropping the `e` for extended.

bnjbvr:
`Bb` looks particularly weird, hence my suggestion of "block" (we have "value", after all). Somebody on IRC mentioned "blk"; I read it as "bulk", but it's nicely shorter than block, so...

eqrion:
I agree, that I prefer `Block` instead of `Bb` for the entity name. Rust's capitalization rules penalize two letter acronyms for types, and it's rare to type out the type name with inference.

For variable names though, I'm open to `bb`. I suppose it's kind of weird that technically expands to `basic_block` when it's referring to just a `Block`.

I think `blk` is a little ambiguous and I'm a bit averse to shortening names by dropping letters in the middle of an identifier, unless it's an acronym, as I think it hurts readability.

sstangl:
[...]

PR #1365, opened 2020-01-24
Mass rename Ebb and relatives to Block

eqrion:
Another component of #1303. From experience with mass-renames across Gecko, I've split the patches to be fine grained and marked whether the patch is automatic (i.e. find-replace) or manual. This helps in rebasing and reviewing, although the changes are quite large so those will still be challenging.

I'm opening this as a draft pull-request so that interested parties have a chance to look at this. It should be relatively good to go once consensus on naming is reached.


abrown:
@eqrion, this makes sense to me. Why don't we mark this ready for review?

eqrion:
> @eqrion, this makes sense to me. Why don't we mark this ready for review?

I just realized that I had forgotten to clean up the ordering of the commits. I'll do that first and then mark this ready for review.

abrown:
I think we will want to "squash and merge" this PR to make it atomic?

eqrion:
Yeah, this is probably ready for review. I see there are now conflicts. I think the best way forward is to review this as-is, and then when we're ready to merge I can rebase by re-creating the automatic rename commits.

Issue #1244, opened 2019-11-19
Support VEX and EVEX encodings


abrown:
- What is the feature or code improvement you would like to do in Cranelift?

Add VEX and EVEX encoding mechanisms for x86 code generation.

- What is the value of adding this in Cranelift?

Certain Wasm SIMD operations (e.g. [convert, trunc](https://github.com/WebAssembly/simd/blob/master/proposals/simd/SIMD.md#conversions)) only can be lowered with VEX and EVEX encodings. Many VEX and EVEX encodings are only available in later SIMD feature sets (e.g. AVX-512); adding the ability to use VEX and EVEX encodings does not preclude legalizations for x86 CPUs without later SIMD feature sets, but that is a separate issue.

- Do you have an implementation plan, and/or ideas for data structures or algorithms to use?

I think the relevant thing here is that it may impact how we go about resolving #1156.

- Have you considered alternative implementations? If so, how are they better or worse than your proposal?

Not having the ability to use VEX and EVEX encodings would force cranelift to use multi-instruction legalizations instead of optimal instructions during instruction selection.

bnjbvr:
We'll definitely support VEX and EVEX at some point.

For what it's worth, we added support for those in Spidermonkey a while ago but this resulted in hard to track down slowdowns. I seem to recall it was related to the fact that the VEX chip was slow to start up, such that the cost of starting it up would show up in profiles and sometimes regress performance, if the VEX instructions weren't used a lot. @sunfishcode might have more insights.

In the meanwhile, it seems fine to have multi instruction legalizations for the wasm opcodes without VEX, since we'd ideally would like to support SSE2 eventually...

sunfishcode:
The Bugzilla bug is https://bugzilla.mozilla.org/show_bug.cgi?id=1124751. Some OS's cause code using VEX prefixes to run significantly slower, so we'll need to determine which OS's and versions are ok, and test for them at runtime.

bjorn3:
Using this will be behind a `has_avx` flag, right? If so disabling it in Spidermonkey should prevent the hard to track down slowdowns.

abrown:
@sunfishcode, we talked about this in IRC a while back and I haven't tried to replicate the issue yet. In fact, I think it would be good to implement the VEX/EVEX encodings so that we _can_ try to find the OS's that cause slowdowns. The way we would use this initially (in truncate/convert) would be along the lines with what you observed on Linux so I would hope we would not see any regressions:

[...]

PR #1370, opened 2020-01-29
Add initial support for EVEX encodings

abrown:
- [x] This has been discussed in issue #1244.
- [x] A short description of what this does, why it is needed: since the Wasm SIMD spec has re-added instructions (e.g. `i64x2.mul`) that are optimally encoded in VEX/EVEX, it seems like we should support these encoding formats. For this PR, only EVEX was added (and not every addressing form) in order to implement i64x2 multiplication. I haven't yet test-run this on an AVX512DQ-compatible CPU but will soon; while implementing this, I discovered that there is no good way to indicate an encoding that is predicated on either of two settings flags (see TODO in `encodings.rs`).
- [x] A reviewer from the core maintainer team has been assigned for this PR (I'm tagging people that might be interested in this).

### DIFF
--- a/cranelift-codegen/meta/src/isa/x86/registers.rs
+++ b/cranelift-codegen/meta/src/isa/x86/registers.rs
@@ -3,17 +3,17 @@ use crate::cdsl::regs::{IsaRegs, IsaRegsBuilder, RegBankBuilder, RegClassBuilder
 pub(crate) fn define() -> IsaRegs {
     let mut regs = IsaRegsBuilder::new();
 
+    let builder = RegBankBuilder::new("FloatRegs", "xmm")
+        .units(32)
+        .track_pressure(true);
+    let float_regs = regs.add_bank(builder);
+
     let builder = RegBankBuilder::new("IntRegs", "r")
         .units(16)
         .names(vec!["rax", "rcx", "rdx", "rbx", "rsp", "rbp", "rsi", "rdi"])
         .track_pressure(true)
         .pinned_reg(15);
     let int_regs = regs.add_bank(builder);
-
-    let builder = RegBankBuilder::new("FloatRegs", "xmm")
-        .units(16)
-        .track_pressure(true);
-    let float_regs = regs.add_bank(builder);
 
     let builder = RegBankBuilder::new("FlagRegs", "")
         .units(1)
@@ -24,7 +24,10 @@ pub(crate) fn define() -> IsaRegs {
     let builder = RegClassBuilder::new_toplevel("GPR", int_regs);
     let gpr = regs.add_class(builder);
 
-    let builder = RegClassBuilder::new_toplevel("FPR", float_regs);
+    let builder = RegClassBuilder::new_toplevel("FPR32", float_regs);
+    let fpr32 = regs.add_class(builder);
+
+    let builder = RegClassBuilder::subclass_of("FPR", fpr32, 0, 16);
     let fpr = regs.add_class(builder);
 
     let builder = RegClassBuilder::new_toplevel("FLAG", flag_reg);
--- a/cranelift-codegen/src/isa/x86/binemit.rs
+++ b/cranelift-codegen/src/isa/x86/binemit.rs
@@ -61,6 +61,17 @@ fn rex3(rm: RegUnit, reg: RegUnit, index: RegUnit) -> u8 {
     BASE_REX | b | (x << 1) | (r << 2)
 }
 
+/// Encode the RXBR' bits of the EVEX P0 byte. For an explanation of these bits, see section 2.6.1
+/// in the Intel Software Development Manual, volume 2A. These bits can be used by different
+/// addressing modes (see section 2.6.2), requiring different `vex*` functions than this one.
+fn evex2(rm: RegUnit, reg: RegUnit) -> u8 {
+    let b = (!(rm >> 3) & 1) as u8;
+    let x = (!(rm >> 4) & 1) as u8;
+    let r = (!(reg >> 3) & 1) as u8;
+    let r_ = (!(reg >> 4) & 1) as u8;
+    0x00 | r_ | (b << 1) | (x << 2) | (r << 3)
+}
+
 /// Determines whether a REX prefix should be emitted.
 #[inline]
 fn needs_rex(bits: u16, rex: u8) -> bool {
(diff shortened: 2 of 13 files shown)


filetests/isa/x86/simd-arithmetic-binemit.clif
```diff
@@ -120,6 +120,21 @@ ebb0:
...
 
+function %imul_i64x2() {
+ebb0:
+    [-, %xmm1]    v0 = vconst.i64x2 [1 2]
+    [-, %xmm2]    v1 = vconst.i64x2 [2 2]
+    [-, %xmm1]    v2 = imul v0, v1 ; bin: 62 52 8d 08 40 ca
+    ; 62, mandatory EVEX prefix
+    ; 52 = 0101 0010, X and R' set (incorrectly I think) and mm set to OF38 (correctly)
+    ; 8d = 1000 1101, W set (correctly, vvvv set to 1 (correctly), bit 2 always set, pp set to 01 (correctly)
+    ; 08 = 0000 1000, V' set to 1 (incorrectly, I believe, due to packing of bits in RegUnit)
```
abrown: I don't think I understand how `RegUnit` is being packed, I'm afraid. EVEX (and VEX) support larger numbers of registers so I attempted to extract the next bit to use in the bits that EVEX uses to extend the set of possible registers (e.g. `XRBR'V'`)--this seems to give the wrong answer. How would I tell Cranelift that EVEX has more registers available?

cranelift-codegen/src/isa/x86/binemit.rs
```diff
@@ -206,6 +217,150 @@ fn put_rexmp3<CS: CodeSink + ?Sized>(bits: u16, rex: u8, sink: &mut CS) {
...
+    let mut p1 = enc.pp() | 0b100; // bit 2 is always set
+    p1 |= (vvvvv as u8 & 0b1111) << 3;
+    p1 |= (enc.rex_w() & 0b1) << 7;
+    sink.put1(p1);
+
+    let mut p2 = masking.aaa_bits();
+    p2 |= ((vvvvv as u8 >> 4) & 0b1) << 3;
+    p2 |= context.bits() << 4;
+    p2 |= masking.z_bit() << 7;
+    sink.put1(p2);
```
abrown: The [wikipedia article] is also helpful if you don't want to crack open the Intel manual; maybe I will link to it in the documentation.

[wikipedia article]: https://en.wikipedia.org/wiki/EVEX_prefix

cranelift-codegen/meta/src/isa/x86/registers.rs
```diff
@@ -24,7 +24,10 @@ pub(crate) fn define() -> IsaRegs {
     let builder = RegClassBuilder::new_toplevel("GPR", int_regs);
     let gpr = regs.add_class(builder);
 
-    let builder = RegClassBuilder::new_toplevel("FPR", float_regs);
+    let builder = RegClassBuilder::new_toplevel("FPR32", float_regs);
+    let fpr32 = regs.add_class(builder);
+
+    let builder = RegClassBuilder::subclass_of("FPR", fpr32, 0, 16);
```
abrown: @bnjbvr, this (and the `.units(32)` above) is causing test failures that I don't understand:
 - [`isa::x86::registers::tests::unit_encodings`](https://github.com/bytecodealliance/cranelift/blob/21516e5e988e70d1ff453de7c0b22ad58fe464a0/cranelift-codegen/src/isa/x86/registers.rs#L27) is failing because now the GPRs miraculously have more units so `%xmm0` starts at `Some(32)`
 - similarly the [`unit_names`](https://github.com/bytecodealliance/cranelift/blob/21516e5e988e70d1ff453de7c0b22ad58fe464a0/cranelift-codegen/src/isa/x86/registers.rs#L47) test thinks the reg index `16` is invalid

What is going on?

iximeow: I think you're running afoul of https://github.com/bytecodealliance/cranelift/blob/5602d1564c4121e4703843f23cc083dfb29c5592/cranelift-codegen/meta/src/cdsl/regs.rs#L240-L247 which bumps the start of `FPR` to be aligned with its size. In theory you could add the expanded FPR first, then GPR, and everything should be okay?

abrown: Aha!

abrown: Hm, too quick to celebrate: I tried moving the `RegBankBuilder::new("FloatRegs"...` with the newly-expanded 32 units up above the `RegBankBuilder::new("IntRegs"...` but now the meta build crashes [due to r15](https://github.com/bytecodealliance/cranelift/blob/5602d1564c4121e4703843f23cc083dfb29c5592/cranelift-codegen/meta/src/isa/x86/recipes.rs#L430) with a backtrace of:
```
thread 'main' panicked at 'attempt to subtract with overflow', cranelift-codegen/meta/src/cdsl/regs.rs:60:25
stack backtrace:
...
  15: cranelift_codegen_meta::cdsl::regs::RegBank::unit_by_name
             at cranelift-codegen/meta/src/cdsl/regs.rs:60
  16: cranelift_codegen_meta::cdsl::regs::IsaRegs::regunit_by_name
             at cranelift-codegen/meta/src/cdsl/regs.rs:410
  17: cranelift_codegen_meta::isa::x86::recipes::define
             at cranelift-codegen/meta/src/isa/x86/recipes.rs:430
  18: cranelift_codegen_meta::isa::x86::define
             at cranelift-codegen/meta/src/isa/x86/mod.rs:61
  19: cranelift_codegen_meta::isa::define::{{closure}}
             at cranelift-codegen/meta/src/isa/mod.rs:62
...
```

iximeow: I think the failing assert is in error, f.ex `let reg_xmm1 = Register::new(fpr, regs.regunit_by_name(fpr, "xmm1"));` currently asserts for the same reason on master. It looks like `unit_by_name` shouldn't even trying to subtract from the register number (`15` in `r15`), and happens to work out so long as the specified reg class starts at 0 - subtract by 0 won't underflow.

I think instead `unit_by_name` ought to be checking just that `as_num < self.units`, and return `self.first_unit + as_num` like it would below?

abrown: You're right. I pushed b0ee482 to fix that and b52fbdd to show what happens when I move the XMM registers first:
 - failures in `unit_names` and `unit_encodings` because, e.g., rax is no longer the first register; these can be fixed and are not a real problem
 - failures with unwinding (`test_*_alloc` and `windows_fastcall_x64_unwind.clif`): @peterhuene, it looks like the unwind info uses the register index directly; if I change a bunch of register indexes, does this completely break unwinding?
 - a failure in `simd-arithmetic-binemit.clif` is expected; that's what I'm currently working on
 - a failure in `pinned-reg.clif` like `panicked in worker #1: Couldn't clear fixed output interference for v1`; @bnjbvr, do you have any comment on this? Will changing the register indexes break regalloc?

iximeow: I was afraid there'd be some breakages like that :(

I noticed Windows unwinding using the regunit as an integer directly when fixing XMM preservation. Changing indices around will break it, but straightforward to fix: something like [this](https://github.com/bytecodealliance/cranelift/pull/1378/files#diff-c720e722fabb16edd5e9ce3554ee7f5dR85) but using `GPR.first`?

To the regalloc point, `pinned_reg` also accidentally depends on GPRs being first: [`is_pinned_reg`](https://github.com/bytecodealliance/cranelift/blob/master/cranelift-codegen/src/isa/registers.rs#L204-L209) tests against `pinned_reg` directly, where the ISA builder stores the provided `pinned_reg` [without translation](https://github.com/bytecodealliance/cranelift/blob/master/cranelift-codegen/meta/src/cdsl/regs.rs#L216-L220) and when building the x86 ISA description it's passed [the literal 15](https://github.com/bytecodealliance/cranelift/blob/master/cranelift-codegen/meta/src/isa/x86/registers.rs#L10). `pinned_reg` in the generated ISA should probably be offset from the start of its RegClass, so [`reg_bank.pinned_reg`](https://github.com/bytecodealliance/cranelift/blob/master/cranelift-codegen/meta/src/gen_registers.rs#L60) ought to be `reg_bank.pinned_reg + reg_bank.first_unit + reg_class.start`.

I think fixing dependence on `GPR.first == 0` would be a good PR independent of EVEX encoding, though I dunno how much of this will be replaced with New Backend Plans. (this is where stop talking and wait for Benjamin :D)

peterhuene: @iximeow I think your proposed fix for the GPRs in the Windows unwind information makes sense.

filetests/isa/x86/simd-arithmetic-binemit.clif
abrown: I've figured this out... finally 😄.

abrown:
@iximeow, thanks for all of the help! (And @peterhuene too!) I believe this is ready for an actual review--@iximeow, can you take a look? I rebased the commits into some semblance of order so they should be reviewable from oldest to newest (fun fact of the day: I had to redate them for GitHub to show them in order). Since we likely won't want to squash these things, I made sure the commits are bisectable by running `git rebase master --exec 'cargo test --all'` with no issues. 

I would like to be able to actually run this new instruction on AVX512 hardware with a `test run` filetest but I will try that in a future PR; currently I don't think we have the ability to skip `test run` filetests when the target doesn't match the host's CPU features (or do we?).

bnjbvr:
Closing in favor of https://github.com/bytecodealliance/wasmtime/pull/1192.

PR #1371, opened 2020-01-31
Add rounding average instructions for Wasm SIMD

abrown:
- [x] A short description of what this does, why it is needed: implements Wasm SIMD's lane-wise integer rounding average (see https://github.com/WebAssembly/simd/blob/master/proposals/simd/SIMD.md#lane-wise-integer-rounding-average).

### DIFF
--- a/cranelift-codegen/meta/src/isa/x86/opcodes.rs
+++ b/cranelift-codegen/meta/src/isa/x86/opcodes.rs
@@ -317,6 +317,12 @@ pub static PAND: [u8; 3] = [0x66, 0x0f, 0xdb];
 /// Bitwise AND NOT of xmm2/m128 and xmm1 (SSE2).
 pub static PANDN: [u8; 3] = [0x66, 0x0f, 0xdf];
 
+/// Average packed unsigned byte integers from xmm2/m128 and xmm1 with rounding (SSE2).
+pub static PAVGB: [u8; 3] = [0x66, 0x0f, 0xE0];
+
+/// Average packed unsigned word integers from xmm2/m128 and xmm1 with rounding (SSE2).
+pub static PAVGW: [u8; 3] = [0x66, 0x0f, 0xE3];
+
 /// Compare packed data for equal (SSE2).
 pub static PCMPEQB: [u8; 3] = [0x66, 0x0f, 0x74];
 
--- a/cranelift-codegen/meta/src/shared/instructions.rs
+++ b/cranelift-codegen/meta/src/shared/instructions.rs
@@ -548,6 +548,32 @@ fn define_simd_arithmetic(
         .operands_in(vec![x, y])
         .operands_out(vec![a]),
     );
+
+    let IxN = &TypeVar::new(
+        "IxN",
+        "A SIMD vector type containing integers",
+        TypeSetBuilder::new()
+            .ints(Interval::All)
+            .simd_lanes(Interval::All)
+            .includes_scalars(false)
+            .build(),
+    );
+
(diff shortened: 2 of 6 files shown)


cranelift-codegen/meta/src/shared/instructions.rs
```diff
@@ -2071,6 +2081,22 @@ pub(crate) fn define(
         .can_trap(true),
     );
 
+    let a = &Operand::new("a", IxN);
+    let x = &Operand::new("x", IxN);
+    let y = &Operand::new("y", IxN);
+
+    ig.push(
+        Inst::new(
+            "avgr",
```
bnjbvr: Should we use the more current abbreviation "avg", or just be explicit and call it "average" (it's not very long)? This is the first time I see this midway abbreviation for average, but I'm not the English native here.

filetests/isa/x86/simd-arithmetic-run.clif
```diff
@@ -265,3 +265,17 @@ ebb0:
     return v4
 }
 ; run
+
+function %average_rounding_i16x8() -> b1 {
+ebb0:
+    v0 = vconst.i16x8 [0 0 0 1 42 19 -1 -1]
+    v1 = vconst.i16x8 [0 1 2 4 42 18 -1 0]
+    v2 = vconst.i16x8 [0 1 1 3 42 19 -1 -32768] ; -1 (0xffff) + 0 + 1 == -32768 (0x8000)
```
bnjbvr: Can the parser understand the hexadecimal syntax? If so, we should use it here, and then we wouldn't need the comment anymore.

cranelift-codegen/meta/src/isa/x86/opcodes.rs
```diff
@@ -317,6 +317,12 @@ pub static PAND: [u8; 3] = [0x66, 0x0f, 0xdb];
 /// Bitwise AND NOT of xmm2/m128 and xmm1 (SSE2).
 pub static PANDN: [u8; 3] = [0x66, 0x0f, 0xdf];
 
+/// Average packed unsigned byte integers from xmm2/m128 and xmm1 with rounding (SSE2).
+pub static PAVGB: [u8; 3] = [0x66, 0x0f, 0xE0];
+
+/// Average packed unsigned byte integers from xmm2/m128 and xmm1 with rounding (SSE2).
```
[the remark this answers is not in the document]
bjorn3: I believe a word is 16 bits, not 8.

cranelift-codegen/meta/src/shared/instructions.rs
abrown: The complication here is that the semantics are actually "rounding average" instead of just "average"; I went along with the term ["avgr" used by the spec](fhttps://github.com/WebAssembly/simd/blob/master/proposals/simd/SIMD.md#lane-wise-integer-minimum). Do you want to expand it to "rounding_average" or "average_round"?

filetests/isa/x86/simd-arithmetic-run.clif
abrown: No, not yet for `i16`, `i32`, and `i64`; here's an issue to fix that: https://github.com/bytecodealliance/cranelift/issues/1401.

cranelift-codegen/meta/src/shared/instructions.rs
bnjbvr: Good point! Maybe then `avg_round` is a nice tradeoff? "avg" is standard for average, and we make the rounding part very explicit.

PR #1374, opened 2020-02-02
Add inline hints to some small functions

bjorn3:
  here.
- [ ] A short description of what this does, why it is needed; if the
  first.
  Before this PR, several short functions were not getting inlined into another crate, as they are neither generic, nor marked with an inline hint.
- [ ] This PR contains test cases, if meaningful. N/A

Keeping this as draft until I have identified more functions that could benefit from this.


bjorn3:
I think this will not have much of a difference on second thoughts

PR #1376, opened 2020-02-04
Add integer min/max instructions for Wasm SIMD

abrown:
The Wasm SIMD spec added [integer min/max instructions](https://github.com/WebAssembly/simd/blob/master/proposals/simd/SIMD.md#lane-wise-integer-minimum). For x86 codegen, Cranelift already had encodings for the instructions to emit (e.g. `x86_pmins`) but did not have shared instructions with which to translate the Wasm integer min/max. This change adds `[i|u]min` and `[i|u]max` to do just that and legalizes these new Cranelift instructions to the pre-existing x86 instructions. At some point we could remove the x86 instructions entirely and attach their encodings directly on `[i|u]min` and `[i|u]max`.

### DIFF
--- a/cranelift-codegen/meta/src/shared/instructions.rs
+++ b/cranelift-codegen/meta/src/shared/instructions.rs
@@ -481,6 +481,75 @@ fn define_control_flow(
     );
 }
 
+#[inline(never)]
+fn define_simd_arithmetic(
+    ig: &mut InstructionGroupBuilder,
+    formats: &Formats,
+    _: &Immediates,
+    _: &EntityRefs,
+) {
+    let Int = &TypeVar::new(
+        "Int",
+        "A scalar or vector integer type",
+        TypeSetBuilder::new()
+            .ints(Interval::All)
+            .simd_lanes(Interval::All)
+            .build(),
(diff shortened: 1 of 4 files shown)


cranelift-codegen/meta/src/shared/instructions.rs
```diff
@@ -2071,6 +2071,54 @@ pub(crate) fn define(
         .can_trap(true),
     );
 
+    ig.push(
```
bnjbvr: Forgot to tell in the previous review for average (and if you could apply it there too that'd be great), but as a way to make sure we stop regressing build-step compile time, can you put these in a different function marked as `#[inline(never)]` please? (No need to move all the existing instructions.) This for instance could go into a `simd_comparisons` function, or something appropriately named like this. Thanks!

abrown: I went with `define_simd_arithmetic` since it seems to be in that section in the [spec](https://github.com/WebAssembly/simd/blob/master/proposals/simd/SIMD.md#lane-wise-integer-minimum).

PR #1377, opened 2020-02-04
add documentation mentions of cranelift-object where appropriate

froydnj:
This change makes it slightly more obvious that `cranelift-object` can
be used in lieu of `cranelift-faerie`.

- [X] This has not been discussed in an issue.
- [X] A short description of what this does, why it is needed:

There are no mentions of `cranelift-object` in this documentation, and I was confused about what the difference was between `cranelift-object` and `cranelift-faerie`.  Ideally this PR makes things slightly less confusing; I would like to eliminate `cranelift-faerie` in a future PR.
- [X] This PR does not contain test cases, as it only modifies documentation.

### DIFF
--- a/cranelift-module/README.md
+++ b/cranelift-module/README.md
@@ -13,8 +13,11 @@ following `Backend` implementations:
 
  - `SimpleJITBackend`, provided by [cranelift-simplejit], which JITs
    code to memory for direct execution.
+ - `ObjectBackend`, provided by [cranelift-object], which emits native
+   object files.
  - `FaerieBackend`, provided by [cranelift-faerie], which emits native
    object files.
 
 [cranelift-simplejit]: https://crates.io/crates/cranelift-simplejit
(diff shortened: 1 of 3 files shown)


eqrion:
Thanks!

philipc:
> I would like to eliminate cranelift-faerie in a future PR.

There are users of cranelift-faerie (notably lucet, but possibly others too).

froydnj:
> There are users of cranelift-faerie (notably lucet, but possibly others too).

crates.io says `lucetc` and `graviton_backend` are the only consumers, but there may be others.

I am trying to determine how hard it is to add support for Windows to `lucetc`, and it seems easier to move to `cranelift-object` rather than try to add COFF support to `faerie`.  Switching `graviton_backend` over does not appear particularly hard, if for some reason that was the blocking factor to deleting `cranelift-faerie`.

Issue #1366, opened 2020-01-27
Windows (Fastcall) calling convention: Callee-saved XMM (FP) registers are not actually saved


hrydgard:
We're having issues, only on Windows, with floating point variables becoming corrupt across calls into Cranelift from Wasmer.

The following function `callee_saved_gprs` in the codegen is supposed to return a list of the callee-saved registers:

https://github.com/bytecodealliance/cranelift/blob/23e9bdb2d99fb8f554793cb87f1578cc543c355f/cranelift-codegen/src/isa/x86/abi.rs#L370

However, it, nor its callers, seem to concern themselves with saving the floating point callee-saved registers. Quoth https://docs.microsoft.com/en-us/cpp/build/x64-calling-convention?view=vs-2019#callercallee-saved-registers:

> The registers RBX, RBP, RDI, RSI, RSP, R12, R13, R14, R15, and XMM6-15 are considered nonvolatile and must be saved and restored by a function that uses them.

`callee_saved_gprs` only lists RBX, RBP, RDI, RSI, RSP, R12, R13, R14, R15, ignoring XMM6-15.

I believe this may be the cause of our troubles, it definitely looks wrong. What do you think?

sstangl:
Windows does indeed require restoring those registers on exit if changed.

IonMonkey implementation for reference: https://searchfox.org/mozilla-central/rev/cbd110d718bc89a499d3f996af24532abbf6f8ea/js/src/jit/x64/Trampoline-x64.cpp#72

hrydgard:
We've done more investigation, and manually saving and restoring the registers around calls into Cranelift using some inline assembly fixes our variable corruption issues. 

I don't know if this is actually a real security issue, since values are leaking from VM to host, and the host is in control of the VM's memory anyway. I guess it's possible that values could leak from host to VM if there are bugs in the register allocator, since nothing seems to wipe the XMM registers when entering the VM (though I haven't looked closely for that).

Either way, this clearly needs to be addressed!

sstangl:
> We've done more investigation, and manually saving and restoring the XMM registers around calls into Cranelift using some inline assembly fixes our variable corruption issues.

Well yes, that's logically equivalent to fixing Cranelift's Microsoft x64 calling convention implementation. Except in the case of Cranelift ever calling out to native code, in which case that boundary would also need to know to save/restore registers.

[...]

PR #1378, opened 2020-02-06
Preserve SIMD registers in Windows fastcall functions

iximeow:
- [x] This has been discussed in issue #1366 
- [x] A short description: Windows appears to have extended ABI constraints to require callees to save XMM6-XMM15 (see #1366 for links to reference material). This PR adds preserve/restore behavior for those registers.
- [x] This PR contains test cases. Ish. There is an additional test that would test the right behavior if this was totally ready to go.

This is a draft PR as there are two points where I think I should just ask for help rather than spin my wheels figuring out the Right Thing, I'll add comments below.

@sstangl if you get a chance, I think you might know the right things for the questions to follow!

### DIFF
--- a/cranelift-codegen/src/isa/x86/abi.rs
+++ b/cranelift-codegen/src/isa/x86/abi.rs
@@ -561,16 +561,20 @@ fn fastcall_prologue_epilogue(func: &mut ir::Function, isa: &dyn TargetIsa) -> C
     // The reserved stack area is composed of:
     //   return address + frame pointer + all callee-saved registers + shadow space
     //
-    // XMM registers are twice the size of general purpose registers, so count them
-    // twice for stack space accounting.
-    //
     // Pushing the return address is an implicit function of the `call`
     // instruction. Each of the others we will then push explicitly. Then we
     // will adjust the stack pointer to make room for the rest of the required
     // space for this frame.
     let word_size = isa.pointer_bytes() as usize;
-    let csr_stack_size = ((csrs.iter(GPR).len() + 2) * word_size
-        + csrs.iter(FPR).len() * types::F64X2.bytes() as usize) as i32;
+    let num_fprs = csrs.iter(FPR).len();
+    let mut csr_stack_size = ((csrs.iter(GPR).len() + 2) * word_size) as i32;
+
+    if num_fprs != 0 {
+        csr_stack_size += (num_fprs * types::F64X2.bytes() as usize) as i32;
+
+        // Ensure that csr stack space is 16-byte aligned for ideal SIMD value placement.
+        csr_stack_size = csr_stack_size & !0b1111 + 16;
+    }
 
     // TODO: eventually use the 32 bytes (shadow store) as spill slot. This currently doesn't work
     //       since cranelift does not support spill slots before incoming args
@@ -715,9 +719,17 @@ fn insert_common_prologue(
             // pushed CSRs, frame pointer.
             // Also, the size of a return address, implicitly pushed by a x86 `call` instruction,
             // also should be accounted for.
+            // If any FPR are present, count them as well as necessary alignment space.
             // TODO: Check if the function body actually contains a `call` instruction.
-            let total_stack_size = (csrs.iter(GPR).len() + 1 + 1) as i64 * word_size as i64
-                + csrs.iter(FPR).len() as i64 * types::F64X2.bytes() as i64;
+            let mut total_stack_size = (csrs.iter(GPR).len() + 1 + 1) as i64 * word_size as i64;
+
+            let num_fpr = csrs.iter(FPR).len();
+            if num_fpr > 0 {
+                total_stack_size += csrs.iter(FPR).len() as i64 * types::F64X2.bytes() as i64;
+
+                // Ensure that csr stack space is 16-byte aligned for ideal SIMD value placement.
+                total_stack_size = total_stack_size & !0b1111 + 16;
+            }
 
             insert_stack_check(pos, total_stack_size, stack_limit_arg);
         }
--- a/cranelift-codegen/src/isa/x86/abi.rs
+++ b/cranelift-codegen/src/isa/x86/abi.rs
@@ -613,7 +613,7 @@ fn fastcall_prologue_epilogue(func: &mut ir::Function, isa: &dyn TargetIsa) -> C
         // the low 128 bits here, but we may find that in practice we should preserve all of
         // YMM6-15 (or even ZMM6-15?)
         let csr_arg =
-            ir::AbiParam::special_reg(types::F64X2, ir::ArgumentPurpose::CalleeSaved, fp_csr);
+            ir::AbiParam::special_reg(types::F64, ir::ArgumentPurpose::CalleeSaved, fp_csr);
         func.signature.params.push(csr_arg);
         func.signature.returns.push(csr_arg);
     }
@@ -852,7 +852,7 @@ fn insert_common_prologue(
 
         for reg in csrs.iter(FPR) {
             // Append param to entry Block
-            let csr_arg = pos.func.dfg.append_block_param(block, types::F64X2);
+            let csr_arg = pos.func.dfg.append_block_param(block, types::F64);
 
             // Assign it a location
             pos.func.locations[csr_arg] = ir::ValueLoc::Reg(reg);
@@ -1048,7 +1048,7 @@ fn insert_common_epilogue(
 
         for reg in csrs.iter(FPR) {
             let value = pos.ins().load(
-                types::F64X2,
+                types::F64,
                 ir::MemFlags::trusted(),
                 stack_addr,
                 fpr_offset,
--- a/cranelift-codegen/src/isa/x86/abi.rs
+++ b/cranelift-codegen/src/isa/x86/abi.rs
@@ -857,7 +857,9 @@ fn insert_common_prologue(
             // Assign it a location
             pos.func.locations[csr_arg] = ir::ValueLoc::Reg(reg);
 
-            let reg_store_inst = pos.ins().store(ir::MemFlags::trusted(), csr_arg, stack_addr, fpr_offset);
+            let reg_store_inst =
+                pos.ins()
+                    .store(ir::MemFlags::trusted(), csr_arg, stack_addr, fpr_offset);
             fpr_offset += types::F64X2.bytes() as i32;
 
             if let Some(ref mut frame_layout) = pos.func.frame_layout {
@@ -1045,7 +1047,12 @@ fn insert_common_epilogue(
         pos.func.locations[stack_addr] = ir::ValueLoc::Reg(RU::rcx as u16);
 
         for reg in csrs.iter(FPR) {
-            let value = pos.ins().load(types::F64X2, ir::MemFlags::trusted(), stack_addr, fpr_offset);
+            let value = pos.ins().load(
+                types::F64X2,
+                ir::MemFlags::trusted(),
+                stack_addr,
+                fpr_offset,
+            );
             fpr_offset += types::F64X2.bytes() as i32;
 
             if let Some(ref mut cfa_state) = cfa_state.as_mut() {
--- a/cranelift-codegen/src/isa/x86/abi.rs
+++ b/cranelift-codegen/src/isa/x86/abi.rs
@@ -12,7 +12,8 @@ use crate::abi::{legalize_args, ArgAction, ArgAssigner, ValueConversion};
 use crate::binemit::{FrameUnwindKind, FrameUnwindSink};
 use crate::cursor::{Cursor, CursorPosition, EncCursor};
 use crate::ir;
-use crate::ir::immediates::{Imm64, Offset32};
+use crate::ir::entities::StackSlot;
+use crate::ir::immediates::Imm64;
 use crate::ir::stackslot::{StackOffset, StackSize};
 use crate::ir::types;
 use crate::ir::{
@@ -367,7 +368,7 @@ pub fn allocatable_registers(triple: &Triple, flags: &shared_settings::Flags) ->
     regs
 }
 
-/// Get the set of callee-saved registers general-purpose registers.
+/// Get the set of callee-saved general-purpose registers.
 fn callee_saved_gprs(isa: &dyn TargetIsa, call_conv: CallConv) -> &'static [RU] {
     match isa.triple().pointer_width().unwrap() {
         PointerWidth::U16 => panic!(),
@@ -579,7 +580,7 @@ fn fastcall_prologue_epilogue(func: &mut ir::Function, isa: &dyn TargetIsa) -> C
     // TODO: eventually use the 32 bytes (shadow store) as spill slot. This currently doesn't work
     //       since cranelift does not support spill slots before incoming args
 
-    func.create_stack_slot(ir::StackSlotData {
+    let csr_slot = func.create_stack_slot(ir::StackSlotData {
         kind: ir::StackSlotKind::IncomingArg,
         size: csr_stack_size as u32,
         offset: Some(-(WIN_SHADOW_STACK_SPACE + csr_stack_size)),
@@ -621,7 +622,7 @@ fn fastcall_prologue_epilogue(func: &mut ir::Function, isa: &dyn TargetIsa) -> C
     let entry_block = func.layout.entry_block().expect("missing entry block");
     let mut pos = EncCursor::new(func, isa).at_first_insertion_point(entry_block);
     let prologue_cfa_state =
-        insert_common_prologue(&mut pos, local_stack_size, reg_type, &csrs, isa);
+        insert_common_prologue(&mut pos, local_stack_size, reg_type, &csrs, &csr_slot, isa);
 
     // Reset the cursor and insert the epilogue
     let mut pos = pos.at_position(CursorPosition::Nowhere);
@@ -630,6 +631,7 @@ fn fastcall_prologue_epilogue(func: &mut ir::Function, isa: &dyn TargetIsa) -> C
         local_stack_size,
         reg_type,
         &csrs,
+        &csr_slot,
         isa,
         prologue_cfa_state,
     );
@@ -656,7 +658,7 @@ fn system_v_prologue_epilogue(func: &mut ir::Function, isa: &dyn TargetIsa) -> C
     // will adjust the stack pointer to make room for the rest of the required
     // space for this frame.
     let csr_stack_size = ((csrs.iter(GPR).len() + 2) * word_size) as i32;
-    func.create_stack_slot(ir::StackSlotData {
+    let csr_slot = func.create_stack_slot(ir::StackSlotData {
         kind: ir::StackSlotKind::IncomingArg,
         size: csr_stack_size as u32,
         offset: Some(-csr_stack_size),
@@ -686,7 +688,7 @@ fn system_v_prologue_epilogue(func: &mut ir::Function, isa: &dyn TargetIsa) -> C
     let entry_block = func.layout.entry_block().expect("missing entry block");
     let mut pos = EncCursor::new(func, isa).at_first_insertion_point(entry_block);
     let prologue_cfa_state =
-        insert_common_prologue(&mut pos, local_stack_size, reg_type, &csrs, isa);
+        insert_common_prologue(&mut pos, local_stack_size, reg_type, &csrs, &csr_slot, isa);
 
     // Reset the cursor and insert the epilogue
     let mut pos = pos.at_position(CursorPosition::Nowhere);
@@ -695,6 +697,7 @@ fn system_v_prologue_epilogue(func: &mut ir::Function, isa: &dyn TargetIsa) -> C
         local_stack_size,
         reg_type,
         &csrs,
+        &csr_slot,
         isa,
         prologue_cfa_state,
     );
@@ -709,6 +712,7 @@ fn insert_common_prologue(
     stack_size: i64,
     reg_type: ir::types::Type,
     csrs: &RegisterSet,
+    csr_slot: &StackSlot,
     isa: &dyn TargetIsa,
 ) -> Option<CFAState> {
     let word_size = isa.pointer_bytes() as isize;
@@ -832,33 +836,44 @@ fn insert_common_prologue(
         }
     }
 
-    for reg in csrs.iter(FPR) {
-        // Append param to entry EBB
-        let csr_arg = pos.func.dfg.append_ebb_param(ebb, types::F64X2);
+    if csrs.iter(FPR).len() != 0 {
+        let mut fpr_offset = 0;
 
-        // Assign it a location
-        pos.func.locations[csr_arg] = ir::ValueLoc::Reg(reg);
+        // `stack_store` is not directly encodable in x86_64 at the moment, so we'll need a base
+        // address. We are well after postopt could run, so load the CSR region base once here,
+        // instead of hoping that the addr/store will be combined later.
+        let stack_addr = pos.ins().stack_addr(types::I64, *csr_slot, 0);
 
-        let reg_store_inst = pos.ins().store(
-            ir::MemFlags::trusted(),
-            csr_arg,
-            panic!("get RSP value"),
-            Offset32::new(panic!("value-appropriate offset")),
-        );
+        // At this point we won't be using volatile registers for anything except for return
+        // At this point we have not saved any CSRs yet, and arguments are all in registers. So
+        // pick a location for `stack_addr` that is neither non-volatile nor an argument: RAX will
+        // do.
+        pos.func.locations[stack_addr] = ir::ValueLoc::Reg(RU::rax as u16);
 
-        if let Some(ref mut frame_layout) = pos.func.frame_layout {
-            let mut cfa_state = cfa_state
-                .as_mut()
-                .expect("cfa state exists when recording frame layout");
-            cfa_state.current_depth -= types::F64X2.bytes() as isize;
-            frame_layout.instructions.insert(
-                reg_store_inst,
-                vec![FrameLayoutChange::RegAt {
-                    reg,
-                    cfa_offset: cfa_state.current_depth,
-                }]
-                .into_boxed_slice(),
-            );
+        for reg in csrs.iter(FPR) {
+            // Append param to entry EBB
+            let csr_arg = pos.func.dfg.append_ebb_param(ebb, types::F64X2);
+
+            // Assign it a location
+            pos.func.locations[csr_arg] = ir::ValueLoc::Reg(reg);
+
+            let reg_store_inst = pos.ins().store(ir::MemFlags::trusted(), csr_arg, stack_addr, fpr_offset);
+            fpr_offset += types::F64X2.bytes() as i32;
+
+            if let Some(ref mut frame_layout) = pos.func.frame_layout {
+                let mut cfa_state = cfa_state
+                    .as_mut()
+                    .expect("cfa state exists when recording frame layout");
+                cfa_state.current_depth -= types::F64X2.bytes() as isize;
+                frame_layout.instructions.insert(
+                    reg_store_inst,
+                    vec![FrameLayoutChange::RegAt {
+                        reg,
+                        cfa_offset: cfa_state.current_depth,
+                    }]
+                    .into_boxed_slice(),
+                );
+            }
         }
     }
 
@@ -938,6 +953,7 @@ fn insert_common_epilogues(
     stack_size: i64,
     reg_type: ir::types::Type,
     csrs: &RegisterSet,
+    csr_slot: &StackSlot,
     isa: &dyn TargetIsa,
     cfa_state: Option<CFAState>,
 ) {
@@ -952,6 +968,7 @@ fn insert_common_epilogues(
                     pos,
                     reg_type,
                     csrs,
+                    csr_slot,
                     isa,
                     is_last,
                     cfa_state.clone(),
@@ -969,6 +986,7 @@ fn insert_common_epilogue(
     pos: &mut EncCursor,
     reg_type: ir::types::Type,
     csrs: &RegisterSet,
+    csr_slot: &StackSlot,
     isa: &dyn TargetIsa,
     is_last: bool,
     mut cfa_state: Option<CFAState>,
@@ -1013,25 +1031,35 @@ fn insert_common_epilogue(
         pos.func.dfg.append_inst_arg(inst, csr_ret);
     }
 
-    for reg in csrs.iter(FPR) {
-        let value = pos.ins().load(
-            types::F64X2,
-            ir::MemFlags::trusted(),
-            panic!("get RSP value"),
-            Offset32::new(panic!("value-appropriate offset")),
-        );
+    if csrs.iter(FPR).len() != 0 {
+        let mut fpr_offset = 0;
 
-        if let Some(ref mut cfa_state) = cfa_state.as_mut() {
-            // Note: don't bother recording a frame layout change because the popped value is
-            // still correct in memory, and won't be overwritten until we've returned where the
-            // current frame's layout would no longer matter. Only adjust `current_depth` for a
-            // consistency check later.
-            cfa_state.current_depth += types::F64X2.bytes() as isize;
+        // `stack_store` is not directly encodable in x86_64 at the moment, so we'll need a base
+        // address. We are well after postopt could run, so load the CSR region base once here,
+        // instead of hoping that the addr/store will be combined later.
+        let stack_addr = pos.ins().stack_addr(types::I64, *csr_slot, 0);
+
+        // At this point we won't be using volatile registers for anything except for return
+        // registers. Arbitrarily pick RCX as a volatile register that is not used for return
+        // values.
+        pos.func.locations[stack_addr] = ir::ValueLoc::Reg(RU::rcx as u16);
+
+        for reg in csrs.iter(FPR) {
+            let value = pos.ins().load(types::F64X2, ir::MemFlags::trusted(), stack_addr, fpr_offset);
+            fpr_offset += types::F64X2.bytes() as i32;
+
+            if let Some(ref mut cfa_state) = cfa_state.as_mut() {
+                // Note: don't bother recording a frame layout change because the popped value is
+                // still correct in memory, and won't be overwritten until we've returned where the
+                // current frame's layout would no longer matter. Only adjust `current_depth` for a
+                // consistency check later.
+                cfa_state.current_depth += types::F64X2.bytes() as isize;
+            }
+            pos.prev_inst();
+
+            pos.func.locations[value] = ir::ValueLoc::Reg(reg);
+            pos.func.dfg.append_inst_arg(inst, value);
         }
-        pos.prev_inst();
-
-        pos.func.locations[value] = ir::ValueLoc::Reg(reg);
-        pos.func.dfg.append_inst_arg(inst, value);
     }
 
     if let Some(ref mut frame_layout) = pos.func.frame_layout {
--- a/cranelift-codegen/src/isa/x86/abi.rs
+++ b/cranelift-codegen/src/isa/x86/abi.rs
@@ -626,8 +626,14 @@ fn fastcall_prologue_epilogue(func: &mut ir::Function, isa: &dyn TargetIsa) -> C
     // Set up the cursor and insert the prologue
     let entry_block = func.layout.entry_block().expect("missing entry block");
     let mut pos = EncCursor::new(func, isa).at_first_insertion_point(entry_block);
-    let prologue_cfa_state =
-        insert_common_prologue(&mut pos, local_stack_size, reg_type, &csrs, fpr_slot.as_ref(), isa);
+    let prologue_cfa_state = insert_common_prologue(
+        &mut pos,
+        local_stack_size,
+        reg_type,
+        &csrs,
+        fpr_slot.as_ref(),
+        isa,
+    );
 
     // Reset the cursor and insert the epilogue
     let mut pos = pos.at_position(CursorPosition::Nowhere);
@@ -886,7 +892,11 @@ fn insert_common_prologue(
         // `stack_store` is not directly encodable in x86_64 at the moment, so we'll need a base
         // address. We are well after postopt could run, so load the CSR region base once here,
         // instead of hoping that the addr/store will be combined later.
-        let stack_addr = pos.ins().stack_addr(types::I64, *fpr_slot.expect("if FPRs are preserved, a stack slot is allocated for them"), 0);
+        let stack_addr = pos.ins().stack_addr(
+            types::I64,
+            *fpr_slot.expect("if FPRs are preserved, a stack slot is allocated for them"),
+            0,
+        );
 
         // At this point we won't be using volatile registers for anything except for return
         // At this point we have not saved any CSRs yet, and arguments are all in registers. So
@@ -1010,7 +1020,11 @@ fn insert_common_epilogue(
         // `stack_store` is not directly encodable in x86_64 at the moment, so we'll need a base
         // address. We are well after postopt could run, so load the CSR region base once here,
         // instead of hoping that the addr/store will be combined later.
-        let stack_addr = pos.ins().stack_addr(types::I64, *fpr_slot.expect("if FPRs are preserved, a stack slot is allocated for them"), 0);
+        let stack_addr = pos.ins().stack_addr(
+            types::I64,
+            *fpr_slot.expect("if FPRs are preserved, a stack slot is allocated for them"),
+            0,
+        );
 
         // At this point we won't be using volatile registers for anything except for return
         // registers. Arbitrarily pick RBP as it won't hold an interesting value, and we're about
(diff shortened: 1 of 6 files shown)


cranelift-codegen/src/isa/x86/abi.rs
```diff
@@ -366,17 +367,18 @@ pub fn allocatable_registers(triple: &Triple, flags: &shared_settings::Flags) ->
...
         PointerWidth::U64 => {
             if call_conv.extends_windows_fastcall() {
-                // "registers RBX, RBP, RDI, RSI, RSP, R12, R13, R14, R15 are considered nonvolatile
-                //  and must be saved and restored by a function that uses them."
+                // "registers RBX, RBP, RDI, RSI, RSP, R12, R13, R14, R15, and XMM6-15 are
+                // considered nonvolatile and must be saved and restored by a function that uses
+                //  them."
                 // as per https://docs.microsoft.com/en-us/cpp/build/x64-calling-convention
-                // RSP & RSB are not listed below, since they are restored automatically during
+                // RSP & RBP are not listed below, since they are restored automatically during
```
iximeow: assuming this was a typo, `RSB` doesn't make sense to me as an x86-ism and the other automatically-restored register would be `RBP`.

cranelift-codegen/src/isa/x86/abi.rs
```diff
@@ -394,21 +396,60 @@ fn callee_saved_gprs(isa: &dyn TargetIsa, call_conv: CallConv) -> &'static [RU]
...
 
     let mut used = RegisterSet::empty();
     for value_loc in func.locations.values() {
         // Note that `value_loc` here contains only a single unit of a potentially multi-unit
         // register. We don't use registers that overlap each other in the x86 ISA, but in others
         // we do. So this should not be blindly reused.
         if let ValueLoc::Reg(ru) = *value_loc {
-            if !used.is_avail(GPR, ru) {
-                used.free(GPR, ru);
+            if GPR.contains(ru) {
```
iximeow: not checking that `GPR` (and `FPR`) contain the regunit in question could introduce errors if these RegClasses had differing widths. So far as I can see, x86 `RegClass` all have the same width of 1, so this happens to work out okay, but I think is strictly speaking a bug independent of SIMD register preservation.

cranelift-codegen/src/isa/x86/abi.rs
```diff
@@ -649,7 +716,8 @@ fn insert_common_prologue(
             // Also, the size of a return address, implicitly pushed by a x86 `call` instruction,
             // also should be accounted for.
             // TODO: Check if the function body actually contains a `call` instruction.
-            let total_stack_size = (csrs.iter(GPR).len() + 1 + 1) as i64 * word_size as i64;
+            let total_stack_size = (csrs.iter(GPR).len() + 1 + 1) as i64 * word_size as i64
+                + csrs.iter(FPR).len() as i64 * types::F64X2.bytes() as i64;
```
iximeow: `csrs.iter(FPR).len()` will be zero via the above assert, but I want the stack size to be the same expression in the three places we compute it here in `abi.rs`.

cranelift-codegen/src/isa/x86/abi.rs
```diff
@@ -752,6 +820,36 @@ fn insert_common_prologue(
...
+        // Append param to entry EBB
+        let csr_arg = pos.func.dfg.append_ebb_param(ebb, types::F64X2);
+
+        // Assign it a location
+        pos.func.locations[csr_arg] = ir::ValueLoc::Reg(reg);
+
+        let reg_store_inst = pos.ins().store(
+            ir::MemFlags::trusted(),
+            csr_arg,
+            panic!("get RSP value"),
```
iximeow: this is where I'd like input from someone who knows codegen APIs better than me (@sstangl?): SIMD registers can't be `x86_push`'d nor `x86_pop`'d (there is no push/pop encoding in the ISA for them, even), so my plan is to know there will be sufficient space reserved for them and simply `load`/`store` them. To do that, I would need to use `rsp` or `rbp` as a base, with an appropriate offset. Is there a nice way to get `function call frame base` as a value?

I don't want to load `rbp` here (even if it were possible) because future frame pointer elision would break this, so either `rsp` or `function call frame base` seem like better choices. Computing the offset from here ought to be straightforward.

filetests/isa/x86/windows_fastcall_x64.clif
```diff
@@ -32,6 +32,16 @@ ebb0(v0: i64, v1: i64, v2: i64, v3: i64, v4: i64):
 }
 ; check: function %five_args(i64 [%rcx], i64 [%rdx], i64 [%r8], i64 [%r9], i64 [32], i64 fp [%rbp]) -> i64 fp [%rbp] windows_fastcall {
 
+; check that we preserve xmm6 and above if we're using them locally
+function %float_callee_saves(f64, f64, f64, f64) windows_fastcall {
+ebb0(v0: f64, v1: f64, v2: f64, v3: f64):
+[-, %xmm6]  v4 = fadd v0, v1      ; bin: 66 0f db f3
```
iximeow: The `bin` expression here is wrong, but I've not gotten sufficiently far in this test to need to fix that yet. An earlier (more cumbersome) approach to preserving SIMD registers confirmed that this does get far enough to try generating stores to preserve values though, so that's a start.

cranelift-codegen/src/isa/x86/abi.rs
sstangl: I think [the IonMonkey-style approach](https://searchfox.org/mozilla-central/rev/cbd110d718bc89a499d3f996af24532abbf6f8ea/js/src/jit/x64/Trampoline-x64.cpp#72) works well: figure out how many registers to push, allocate stack by subtracting from `%rsp`, and `vmovdqa` into place in an arbitrary order of your choosing.

cranelift-codegen/src/isa/x86/abi.rs
sstangl: I think unfortunately that this isn't quite right -- in order to store the xmm registers, the destination must be 16-byte aligned. Since the number of GPRs pushed is non-constant, you might need padding bytes.

cranelift-codegen/src/isa/x86/abi.rs
```diff
@@ -509,17 +556,21 @@ fn fastcall_prologue_epilogue(func: &mut ir::Function, isa: &dyn TargetIsa) -> C
...
     }
 
-    let csrs = callee_saved_gprs_used(isa, func);
+    let csrs = callee_saved_regs_used(isa, func);
 
     // The reserved stack area is composed of:
     //   return address + frame pointer + all callee-saved registers + shadow space
     //
+    // XMM registers are twice the size of general purpose registers, so count them
+    // twice for stack space accounting.
```
sstangl: Is this comment out of date? It's using `types::F64X2.bytes() as usize` now which doesn't look like double-counting!

sstangl:
Also, just aside from reviews, thank you so much for writing this patch!

cranelift-codegen/src/isa/x86/abi.rs
iximeow: Yes, I think that would work well here too. What I don't know is how to get the value of `rsp` as a value I can provide here! Is there a way to write "give me an SSA value that is whatever is in `rsp` at this point"? The other option I can imagine is another pair of `x86_*` special instructions that directly encode `rsp`, and take the offset as an argument.

Sidenote: I see IonMonkey uses `vmovdqa` in favor of `movdqa` - is that actually what gets emitted? Is there a reason IonMonkey prefers the AVX form?

bjorn3: Would it be possible to use `stack_store`?

cranelift-codegen/src/isa/x86/abi.rs
iximeow: I'd hoped to store XMM registers before worrying about GPRs, so they would be a constant offset rather than conditionally requiring padding. I don't think that actually pans out though, and it's still variable on architecture width, so I'll adjust this to include some padding space if `csrs.iter(FPR).len() != 0`.

cranelift-codegen/src/isa/x86/abi.rs
iximeow: you are correct! I realized shortly after writing the comment that double-counting is wrong for 32-bit targets, but forgot to fix the comment too.

cranelift-codegen/src/isa/x86/abi.rs
iximeow: ah! yes, I think so. If not, `stack_addr` is the piece I was missing before, and should work too.

cranelift-codegen/src/isa/x86/abi.rs
```diff
@@ -648,8 +719,17 @@ fn insert_common_prologue(
...
+
+            let num_fpr = csrs.iter(FPR).len();
+            if num_fpr > 0 {
+                total_stack_size += csrs.iter(FPR).len() as i64 * types::F64X2.bytes() as i64;
+
+                // Ensure that csr stack space is 16-byte aligned for ideal SIMD value placement.
+                total_stack_size = total_stack_size & !0b1111 + 16;
+            }
 
             insert_stack_check(pos, total_stack_size, stack_limit_arg);
```
iximeow: Unrelated, but because I noticed: do we know of anyone using `StackLimit`? `total_stack_size` here undercounts the real amount of space used in this function, as it doesn't include locals.

bjorn3: All usages of `ArgumentPurpose::StackLimit` on github are either from Cranelift itself, or a gecko vendoring of Cranelift.

cranelift-codegen/src/isa/x86/abi.rs
```diff
@@ -366,17 +367,18 @@ pub fn allocatable_registers(triple: &Triple, flags: &shared_settings::Flags) ->
     regs
 }
 
-/// Get the set of callee-saved registers.
+/// Get the set of callee-saved registers general-purpose registers.
```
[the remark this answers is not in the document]
fitzgen: Apologies for the drive by nitpick :)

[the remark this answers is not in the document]
iximeow: good eye, this function went through some revisions too :)

cranelift-codegen/src/isa/x86/abi.rs
iximeow: it turns out `stack_load` and `stack_store` won't do - there's no direct encoding for them in x86 (yet, I hope) and `prologue_epilogue` is well past `legalize`.

`stack_addr` was surprising too: we're also past `regalloc` so I have to assign a location myself. The error not doing this (that I violated constraints) was quite confusing, because .. what constraints would copying `rsp` have? :D 

A third consideration I didn't realize: we're also well past `postopt`, so opportunity to merge a `stack_addr/{load,store}` has passed. To this end, I added only one `stack_addr` in the prologue, and one in the epilogue, and only if we actually preserve any FPRs in the first place. With `stack_load`/`stack_store` encodings this can be cleaned up some more, but this already took longer than I'd anticipated.

Anyway, it works now!

filetests/isa/x86/windows_fastcall_x64.clif
iximeow: It appears that `filetests` doesn't actually test `bin` expressions (anymore?). This continued expressing the wrong bytes even while the test passed.

I thought I saw an incorrect `bin` in a different filetest, but wrote it off at the time. I'll file a followup issue for this.

filetests/isa/x86/windows_fastcall_x64.clif
```diff
@@ -48,8 +48,8 @@ ebb0(v0: f64, v1: f64, v2: f64, v3: f64):
 ; nextln:                         store notrap aligned v8, v7
 ; nextln:                         store notrap aligned v9, v7+16
 ; check:                          v11 = stack_addr.i64 ss0
-; nextln: [Op2fldDisp8#410,%xmm7] v13 = load.f64x2 notrap aligned v11+16
```
iximeow: As I mention in the commit message, I do not know having this part of the check causes an error. I tried copying the style in f.ex `probestack-disabled.clif`, and I swear this passed once, but now yields an error like
```
> [RexOp1spaddr8_id#808d,%rcx]        v11 = stack_addr.i64 ss0               
                                      ^~~~~~~~~~~~~~~~~~~~~~~~            
Matched #7: \bv11 = stack_addr\.i64 ss0\b                                
> [Op2fldDisp8#410,%xmm7]             v13 = load.f64x2 notrap aligned v11+16
Missed #8: \[Op2fldDisp8\#410,%xmm7\] v13 = load\.f64x2 notrap aligned v11\+16\b
```
note that
```
got:      [Op2fldDisp8#410,%xmm7]             v13 = load.f64x2 notrap aligned v11+16
expected: [Op2fldDisp8#410,%xmm7] v13 = load.f64x2 notrap aligned v11+16
```
these are the same aside from whitespace. I tested having the same amount of whitespace just in case, and that didn't influence passing or not. Am I misunderstanding the `nextln` directive?

cranelift-codegen/src/isa/x86/abi.rs
```diff
@@ -903,6 +1033,42 @@ fn insert_common_epilogue(
...
+
+        // `stack_store` is not directly encodable in x86_64 at the moment, so we'll need a base
+        // address. We are well after postopt could run, so load the CSR region base once here,
+        // instead of hoping that the addr/store will be combined later.
+        let stack_addr = pos.ins().stack_addr(types::I64, *csr_slot, 0);
+
+        // At this point we won't be using volatile registers for anything except for return
+        // registers. Arbitrarily pick RBP as it won't hold an interesting value, and we're about
+        // to restore it at the end of the function anyway.
+        pos.func.locations[stack_addr] = ir::ValueLoc::Reg(RU::rbp as u16);
```
iximeow: This is brittle in the face of frame pointer elision. A more applicable register would be `r11`, but there are no rex encodings of `fstDisp8` or `fldDisp8`, so using `r11` as a base violates constraints that the registers must have index <=7.

iximeow:
Okay! This is properly ready for review now, after running into a few quirks.

cranelift-codegen/src/isa/x86/abi.rs
```diff
@@ -519,12 +567,20 @@ fn fastcall_prologue_epilogue(func: &mut ir::Function, isa: &dyn TargetIsa) -> C
...
     let word_size = isa.pointer_bytes() as usize;
-    let csr_stack_size = ((csrs.iter(GPR).len() + 2) * word_size) as i32;
+    let num_fprs = csrs.iter(FPR).len();
+    let mut csr_stack_size = ((csrs.iter(GPR).len() + 2) * word_size) as i32;
+
+    if num_fprs != 0 {
+        csr_stack_size += (num_fprs * types::F64X2.bytes() as usize) as i32;
+
+        // Ensure that csr stack space is 16-byte aligned for ideal SIMD value placement.
+        csr_stack_size = csr_stack_size & !0b1111 + 16;
```
hrydgard: `csr_stack_size = (csr_stack_size + 15) & !0b1111;`

would save extending when already aligned to 16 bytes (unless I'm misunderstanding something and that's desirable)

iximeow: that'll teach me for trying to do math. Good catch!

cranelift-codegen/src/isa/x86/abi.rs
```diff
@@ -544,8 +600,20 @@ fn fastcall_prologue_epilogue(func: &mut ir::Function, isa: &dyn TargetIsa) -> C
...
+        let csr_arg = ir::AbiParam::special_reg(reg_type, ir::ArgumentPurpose::CalleeSaved, gp_csr);
+        func.signature.params.push(csr_arg);
+        func.signature.returns.push(csr_arg);
+    }
+
+    for fp_csr in csrs.iter(FPR) {
+        // The calling convention described in
+        // https://docs.microsoft.com/en-us/cpp/build/x64-calling-convention only specifically
+        // discusses XMM6-XMM15, which are 128-bit registers. It may be sufficient to preserve only
+        // the low 128 bits here, but we may find that in practice we should preserve all of
```
jaykrell: You only need to preserve the lower 128 bits.
The upper 128 (or 384, whatever) are volatile.
The rationale is:
 - You cannot make registers nonvolatile after the fact.
 - The ABI cannot be revised in a compatible way, adding nonvolatile registers.
 - There are no unwind codes to describe how to restore anything not in the v1 ABI.

You can save more if you want, but it buys little/nothing.

Except, you know, you can have private calling conventions among known callers/callees.

Callers can assume more volatiles (ymm, zmm) are preserved, and can allow for nonvolatiles to be trashed (that the caller preserves).

Calling conventions are largely?entirely about separate compilation.

peterhuene: Speaking of unwind, can we update `unwind.rs` to emit the unwind codes for these CSRs as part of this PR?  The original implementation made some assumptions as to what instructions were present in the prologue.

iximeow: I've update Windows fastcall unwind generation to [emit XmmSave128](https://github.com/bytecodealliance/cranelift/pull/1378/files#diff-9fd98f9cfa3e5b06cfaaf0209f9546afR185-R208) and XmmSave128Far, though I'm suspect of the offset.

From reading https://docs.microsoft.com/en-us/cpp/build/exception-handling-x64?view=vs-2019#unwind-operation-code it looks like the offset for SaveXmm counts a number of 16-byte stack slots from the base of the call frame, so storing to `[rbp - 0x10]` would have an offset of 1, `[rbp - 0x20]` would be 2. In contrast, the callee-save code added here should be `[rsp - csr_stack_slot_offset + FPR_offset]` where `FPR_offset` is 0x10, 0x20, ... as appropriate. The offsets recorded are `FPR_offset`, and my suspicion is that they would be interpreted as offsets from the other side of callee-save space by Windows.

This is well beyond my experience with Windows to actually verify. If this looks right to you, I'd be happy to hear it - threading the size of `csr_stack_slot` through to `UnwindInfo` seems kind of annoying. I think it would best be a field on `Function`?

iximeow: I also updated the comment to be more strongly worded on preserving the low 128 bits, thanks for the feedback, @jaykrell.

jaykrell: To better understand, just look at C++ examples.
Here is a start. Change it up with alloca to get a frame pointer.
But ouch, keep in mind, compiler does some odd things that seem to violate the spec.
Like getting the rsp/rbp before adjustment, and recording offsets from after.
That does occur below unfortunately -- the use of rax.
That is an optimization and you don't have to do it. Please don't, really.
It is just trying to keep the offsets in the instructions to fit in 8 bits instead of 32 I think, to shrink instruction size, and looks unnecessary in this case, and wastes an instruction.

```
C:\s>type 1.c

__declspec(dllexport)
float (*f1)(float, float);

__declspec(dllexport)
float f2(float a, float b)
{
 float c = a + b;
 float d = a - b;
 float e = a * b;
 float f = a / b;

 return f1(a, b) + f1(c, d) + f1(c, d) - f1(e, f) + f1(e + f, e - f);
}

C:\s>cl /O2 /LD /Zi 1.c /link /noentry /incremental:no
Microsoft (R) C/C++ Optimizing Compiler Version 19.21.27702.2 for x64
Copyright (C) Microsoft Corporation.  All rights reserved.

1.c
Microsoft (R) Incremental Linker Version 14.21.27702.2
Copyright (C) Microsoft Corporation.  All rights reserved.

/out:1.dll
/dll
/implib:1.lib
/debug
/noentry
/incremental:no
1.obj
   Creating library 1.lib and object 1.exp

C:\s>link /dump /unwindinfo /disasm 1.dll
Microsoft (R) COFF/PE Dumper Version 14.21.27702.2
Copyright (C) Microsoft Corporation.  All rights reserved.


Dump of file 1.dll

File Type: DLL

f2:
; This instruction is ok but imho an unnecessary and possibly failed optimization.

  0000000180001000: 48 8B C4           mov         rax,rsp

  0000000180001003: 48 81 EC 98 00 00  sub         rsp,98h
                    00
; for example, this could/should be encoded as mov [rsp-80]

  000000018000100A: 0F 29 70 E8        movaps      xmmword ptr [rax-18h],xmm6

; for example, this could/should be encoded as mov [rsp-70]

  000000018000100E: 0F 29 78 D8        movaps      xmmword ptr [rax-28h],xmm7
  0000000180001012: 0F 28 F9           movaps      xmm7,xmm1
  0000000180001015: 44 0F 29 40 C8     movaps      xmmword ptr [rax-38h],xmm8
  000000018000101A: 44 0F 28 C0        movaps      xmm8,xmm0
  000000018000101E: 44 0F 29 48 B8     movaps      xmmword ptr [rax-48h],xmm9
  0000000180001023: 44 0F 28 C8        movaps      xmm9,xmm0
  0000000180001027: 44 0F 29 50 A8     movaps      xmmword ptr [rax-58h],xmm10
  000000018000102C: F3 44 0F 5C CF     subss       xmm9,xmm7
  0000000180001031: 44 0F 29 58 98     movaps      xmmword ptr [rax-68h],xmm11
  0000000180001036: 44 0F 28 D0        movaps      xmm10,xmm0
  000000018000103A: 44 0F 29 60 88     movaps      xmmword ptr [rax-78h],xmm12
  000000018000103F: F3 44 0F 58 D7     addss       xmm10,xmm7
  0000000180001044: 44 0F 28 E0        movaps      xmm12,xmm0
  0000000180001048: 44 0F 28 D8        movaps      xmm11,xmm0
  000000018000104C: F3 44 0F 59 E7     mulss       xmm12,xmm7
  0000000180001051: 41 0F 28 C9        movaps      xmm1,xmm9
  0000000180001055: 41 0F 28 C2        movaps      xmm0,xmm10
  0000000180001059: F3 44 0F 5E DF     divss       xmm11,xmm7
  000000018000105E: FF 15 A4 1F 00 00  call        qword ptr [f1]
  0000000180001064: 0F 28 F0           movaps      xmm6,xmm0
  0000000180001067: 0F 28 CF           movaps      xmm1,xmm7
  000000018000106A: 41 0F 28 C0        movaps      xmm0,xmm8
  000000018000106E: FF 15 94 1F 00 00  call        qword ptr [f1]
  0000000180001074: F3 0F 58 F0        addss       xmm6,xmm0
  0000000180001078: 41 0F 28 C9        movaps      xmm1,xmm9
  000000018000107C: 41 0F 28 C2        movaps      xmm0,xmm10
  0000000180001080: FF 15 82 1F 00 00  call        qword ptr [f1]
  0000000180001086: 0F 28 F8           movaps      xmm7,xmm0
  0000000180001089: 41 0F 28 CB        movaps      xmm1,xmm11
  000000018000108D: 41 0F 28 C4        movaps      xmm0,xmm12
  0000000180001091: F3 0F 58 FE        addss       xmm7,xmm6
  0000000180001095: FF 15 6D 1F 00 00  call        qword ptr [f1]
  000000018000109B: 41 0F 28 CC        movaps      xmm1,xmm12
  000000018000109F: F3 0F 5C F8        subss       xmm7,xmm0
  00000001800010A3: F3 41 0F 5C CB     subss       xmm1,xmm11
  00000001800010A8: F3 45 0F 58 DC     addss       xmm11,xmm12
  00000001800010AD: 41 0F 28 C3        movaps      xmm0,xmm11
  00000001800010B1: FF 15 51 1F 00 00  call        qword ptr [f1]

 ; blech, see below 
  00000001800010B7: 4C 8D 9C 24 98 00  lea         r11,[rsp+98h]
                    00 00
  00000001800010BF: 41 0F 28 73 E8     movaps      xmm6,xmmword ptr [r11-18h]
  00000001800010C4: F3 0F 58 C7        addss       xmm0,xmm7
  00000001800010C8: 0F 28 7C 24 70     movaps      xmm7,xmmword ptr [rsp+70h]
  00000001800010CD: 45 0F 28 43 C8     movaps      xmm8,xmmword ptr [r11-38h]
  00000001800010D2: 45 0F 28 4B B8     movaps      xmm9,xmmword ptr [r11-48h]
  00000001800010D7: 45 0F 28 53 A8     movaps      xmm10,xmmword ptr [r11-58h]
  00000001800010DC: 45 0F 28 5B 98     movaps      xmm11,xmmword ptr [r11-68h]
  00000001800010E1: 45 0F 28 63 88     movaps      xmm12,xmmword ptr [r11-78h]

; This violates the letter of the spec, alas.
; It is supposed to be add rsp,n, or lea rsp, n[frame]
  00000001800010E6: 49 8B E3           mov         rsp,r11
  00000001800010E9: C3                 ret

Function Table (1)

           Begin    End      Info      Function Name

  00000000 00001000 000010EA 0000211C  f2
    Unwind version: 1
    Unwind flags: None
    Size of prologue: 0x3F
    Count of codes: 16
    Unwind codes:
      3F: SAVE_XMM128, register=xmm12 offset=0x20
      36: SAVE_XMM128, register=xmm11 offset=0x30
      2C: SAVE_XMM128, register=xmm10 offset=0x40
      23: SAVE_XMM128, register=xmm9 offset=0x50
      1A: SAVE_XMM128, register=xmm8 offset=0x60
      12: SAVE_XMM128, register=xmm7 offset=0x70
      0E: SAVE_XMM128, register=xmm6 offset=0x80
      0A: ALLOC_LARGE, size=0x98

```

So, overall, mixed. The C++ examples help to understand, but they also confuse, because the C++ compiler violates the spec in ways that end up meaning the same thing at runtime, if you understand closely how unwinding works.
I would encourage you to follow the spec instead, though I suppose there might be measurable perf benefits otherwise.

You can do the link /dump /unwindinfo on any binary -- no need for symbols.

jaykrell: The NT unwinder is also open sourced, here:

https://github.com/dotnet/runtime/blob/master/src/coreclr/src/unwinder/amd64/unwinder_amd64.cpp#L895

at least an old maybe subsetted/altered form, but probably useful.

peterhuene: The offsets in the generated unwind information in the test do seem odd, especially compared to the above output form VC++, but I think that's a result of adjusting the stack pointer for the frame *after* we're writing in the SIMD CSRs.  In the test output, an offset of 0, 0x10, and 0x20 make sense given that the SP will be adjusted up during unwinding restores the XMM registers.

Perhaps we should adjust the SP first?

iximeow: If the generated unwind info looks okay, are you good with this change and revisiting in case it turns out not to be? I don't have anything resembling a useful Windows development environment, so I'm not even entirely sure what to look for being wrong in case the unwind information is wrong - I imagine "only" an incorrect value for a callee-save register in some parent frame? Given that overall this PR fixes an ABI error that results in dangerously wrong code, I _would_ like to get this figured out :)

peterhuene: I'm good with the change and open to revisiting it in the future.  I think the order of operations in our prologues is just a little "strange" compared to VC++'s prologues, but I don't necessarily see that as a problem with respect to unwind.

peterhuene (approved):
Reviewing the Windows unwind changes: 👍.  The rest makes sense to me, but we should probably get sign-off from another maintainer regarding those changes.

iximeow:
Thanks for reviewing! I'm always happy to get more eyes on ABI details - @fitzgen you took a look earlier, is this something you're comfortable reviewing? And if not, anyone want to volunteer?

filetests/isa/x86/windows_fastcall_x64_unwind.clif
```diff
@@ -147,23 +156,56 @@ block0(v0: i64, v1: i64):
...
+; nextln:     unwind_code_count_raw: 16,
 ; nextln:     frame_register: 5,
 ; nextln:     frame_register_offset: 0,
 ; nextln:     unwind_codes: [
 ; nextln:         UnwindCode {
-; nextln:             offset: 19,
+; nextln:             offset: 42,
 ; nextln:             op: SmallStackAlloc,
-; nextln:             info: 3,
+; nextln:             info: 5,
```
peterhuene: Sorry, I just caught this.  I would expect a value of **4** originally, indicating a 40 byte SP adjustment: 8 bytes to realign the stack from the even number of GPR pushes and 32 bytes for the callee spill space we never elide (even for leaf functions).  So there's a bug currently in how we're aligning the stack even when spilling just the GPRs.

With your changes, I would expect a value of 10, indicating 88 bytes: 8 bytes to realign the stack from the even number of GPR pushes, 48 bytes for the saved XMM registers, and 32 bytes for the callee spill space.

peterhuene (changes requested):
Retracting my previous approval until the SP adjustment comment above is resolved.

filetests/isa/x86/windows_fastcall_x64_unwind.clif
peterhuene: For comparison, the VC++ example above allocates 152 bytes for its frame: 8 bytes for stack alignment (upon entry stack is always off by 8 bytes due to call instruction pushing return address), 112 bytes for saved XMM registers, and 32 bytes for callee spill space (the function does call).

iximeow: Yep, there's definitely an error here. I hadn't checked the generated assembly on this case but
```
Disassembly of 356 bytes:                                                                                                                                                               [69/1840]
   0:   55                      push    rbp                        
   1:   48 89 e5                mov     rbp, rsp                   
   4:   53                      push    rbx                         
   5:   56                      push    rsi                         
   6:   57                      push    rdi                         
   7:   41 54                   push    r12                         
   9:   41 55                   push    r13                         
   b:   41 56                   push    r14                         
   d:   41 57                   push    r15                         
   f:   48 8d 84 24 10 00 00 00 lea     rax, [rsp + 0x10]           
  17:   f2 0f 11 30             movsd   qword ptr [rax], xmm6       
  1b:   f2 0f 11 78 10          movsd   qword ptr [rax + 0x10], xmm7
  20:   f2 44 0f 11 40 20       movsd   qword ptr [rax + 0x20], xmm8
```
is definitely not the prologue I'd want to see. It looks like the callee-save for fprs begins at where the stack was when created, rather than being a non-overlapping region laid out somewhere later on. Which makes sense, since this is so late in compilation. urgh.

cranelift-codegen/src/isa/x86/abi.rs
```diff
@@ -752,6 +838,49 @@ fn insert_common_prologue(
...
+            // Append param to entry Block
+            let csr_arg = pos.func.dfg.append_block_param(block, types::F64);
+
+            // Assign it a location
+            pos.func.locations[csr_arg] = ir::ValueLoc::Reg(reg);
+
+            let reg_store_inst =
+                pos.ins()
+                    .store(ir::MemFlags::trusted(), csr_arg, stack_addr, fpr_offset);
+            fpr_offset += types::F64X2.bytes() as i32;
```
fitzgen: Wait -- shouldn't `fpr_offset` be initialized to the offset chosen for the stack slot earlier? Right now it is initialized to 0. (Or alternatively but equivalently use the stack slot offset as an immediate for the `stack_addr` instruction?) Otherwise we are always reading xmm6-15 from stack frame offsets `0..`. Is this the source of the overlapping stack slot bug?

cranelift-codegen/src/isa/x86/abi.rs
```diff
@@ -903,6 +1035,39 @@ fn insert_common_epilogue(
         pos.func.dfg.append_inst_arg(inst, csr_ret);
     }
 
+    if csrs.iter(FPR).len() != 0 {
+        let mut fpr_offset = 0;
```
fitzgen: ditto

filetests/isa/x86/windows_fastcall_x64_unwind.clif
peterhuene: Was the unwind information "correct" in that RSP is adjusted by only 0x28 following saving the XMM registers?

iximeow: correct in that it was adjusted by 0x30 (5 * 8 == 40 == 0x28, but there's the implied extra 8 bytes bringing it to 0x30)

fitzgen (commented):
The overall patch looks good to me, but obviously the overlapping stack slot bug needs to be fixed before we land this.

I didn't look too much at the unwind code stuff, since Peter already did and is much more familiar with it than I am.

Once the bug is fixed, I'll give it a final once over, and then it should be good to land!

cranelift-codegen/src/isa/x86/abi.rs
```diff
@@ -585,7 +660,7 @@ fn system_v_prologue_epilogue(func: &mut ir::Function, isa: &dyn TargetIsa) -> C
     // will adjust the stack pointer to make room for the rest of the required
     // space for this frame.
     let csr_stack_size = ((csrs.iter(GPR).len() + 2) * word_size) as i32;
-    func.create_stack_slot(ir::StackSlotData {
+    let csr_slot = func.create_stack_slot(ir::StackSlotData {
         kind: ir::StackSlotKind::IncomingArg,
         size: csr_stack_size as u32,
         offset: Some(-csr_stack_size),
```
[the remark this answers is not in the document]
iximeow: I was just looking at this for the same reason, I think you're on the right track. These offsets are up from the bottom of the stack slot so zero is on the opposite side from where `push` starts placing registers. My expectation was the callee-save region would look like
```
size      | rbp
size - 8  | rbx
size - 16 | rsi
... other callee-save gprs
0x20      | xmm8
0x10      | xmm7
0x00      | xmm6
```
which is close to what the actual layout looks like, but the base isn't what I'd expected.

cranelift-codegen/src/isa/x86/abi.rs
iximeow: same as above, either this _is_ or is closely related to the overlapping stack slot issue. Either this will change or I'll update (probably with a comment describing CSR region layout) when it's fixed.

iximeow: (a nice benefit of picking the end of the CSR region is that it is defined to be aligned by ABI requirements, so we don't need to figure out if the last GPR save left the stack aligned or not)

filetests/isa/x86/windows_fastcall_x64_unwind.clif
peterhuene: Oh sorry, I meant 0x30 (the 5 in the unwind info indicating (8*5)+8; I seem to always forget the +8).  Thanks for confirming.

peterhuene: Also, judging from the above disassembly, The offsets from RSP for the XMM registers are off by 1 in the resulting unwind information it seems.  We're recording the offsets from RAX in the unwind info (scaled-by-16 values of 0, 1, and 2), but RAX itself is offset 0x10 from RSP.

peterhuene: I've deleted my last comment since it's probably just related to the overlapping stack slot issues.  Once we have the proper prologue generated, I'll verify that the unwind information properly describes it.

iximeow: Some notes as it took me a surprising amount of time to figure out what _exactly_ was going on. I didn't really understand `StackSlot` too well.

The gist of it is that `lea rax, [rsp + 0x10]` may look like an address of nothing in particular, but that's because it seems `StackSlot` used alongside adjusting the stack is a recipe for bad times. It's the instruction that would be correct to get the stack slot's start at the entry of the function without intervening changes. That is, the return address and `rbp` are preserved on the stack by Cranelift, and 0x10 down the stack from the base address of the call frame would be `rbx` and the start of the rest of the csr region. `rsp` isn't really "ready for use" until after the prologue though, so `stack_addr` doesn't seem appropriate to use as it currently exists for this purpose.

Subsequently, the xmm stores _would_ overwrite the return address, `rbp`, `rbx`, ..., were the stack slot address right - I misunderstood which direction offsets count from too.

I think what I can do is make a second stack slot that we can directly address for FPR save/lrestore, and requiring 16-byte alignment on that would keep offsets simple as well. I'll give it a try and we'll see how it goes :crossed_fingers:

filetests/isa/x86/windows_fastcall_x64.clif
```diff
@@ -44,14 +44,16 @@ block0(v0: f64, v1: f64, v2: f64, v3: f64):
 ; check: block0(v0: f64 [%xmm0], v1: f64 [%xmm1], v2: f64 [%xmm2], v3: f64 [%xmm3], v6: i64 [%rbp], v8: f64 [%xmm6], v9: f64 [%xmm7]):
 ; nextln:                         x86_push v6
 ; nextln:                         copy_special %rsp -> %rbp
+; nextln:                         adjust_sp_down_imm 64
```
iximeow: the missing stack adjustment in this test maybe should have been a giveaway that something was awry too. I'd read it as not necessary, assuming `ss0` would be some valid stack slot, but in reality this was testing that a store to `v7` would clobber the start of the CSR region :(

`ss0` is still the correct stack slot with this change because I add the FPR stack slot before the GPR stack slot. I've since updated this test to verify that.

filetests/isa/x86/windows_fastcall_x64_unwind.clif
iximeow: Surprise of surprises, the number is now not `3`, `4`, `5`, or `10`, but `12`! I can argue the correctness of 12, though. For `stack_slot` reasons I don't understand, the FPR preservation region is aligned more strictly than necessary. This grows what would be 48 bytes into 64. Plus eight bytes to realign the stack, and 32 bytes of shadow stack space, that makes for a 104-byte allocation. That would be 13 8-byte slots, minus one for the implied slot makes 12!

A similar wider-than-necessary FPR region is also present in the [`float_callee_save`](https://github.com/bytecodealliance/cranelift/blob/399fb63390f66592730f9922f89b980b16a1ad22/filetests/isa/x86/windows_fastcall_x64.clif#L49) test I added. I've tried to follow through what causes this in `layout_stack` and my best theory is that treating a stack slot as a collection of smaller units is not quite how it's intended to be used, and instead it tries to align the provided stack slots using their sizes as an alignment requirement. It looks like treating a stack slot as an aggregate of more permissible units might be a bit of a change, so I'm open to recommendations if taking ~16-48 extra bytes per call frame is a nonstarter.

cranelift-codegen/src/isa/x86/abi.rs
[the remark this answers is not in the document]
iximeow: update: it's not too late, but `layout_stack` requires slots of type `IncomingArg` to have a defined offset. That offset is used for the base of the stack layout so a `None` ends up being a panic.

I ended up making FPR save/restore go through a second stack slot of kind `SpillSlot` - it seems like either this or `ExplicitSlot` might make sense on account of stack store/load is a lot like a spill (if they weren't used, they wouldn't have to be "spilled" to the stack). I'm not entirely sure what `ExplicitSlot` ought to be used for, but it also looks applicable?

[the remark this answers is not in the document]
fitzgen: I'm not totally sure, but I *think* that `SpillSlot`s are intended to be reused/de-duped across spilled values with distinct live ranges. This should be possible to do in theory for calling convention related stack slots (same as with struct return) but I'm not sure that it works as of today nor that it wouldn't be buggy.

I think (again not totally sure, as they're mostly used in tests and pretty much all slot kinds other than incoming args are treated the same way) that explicit slots are intended as catch all slots used for random purposes.

It seems that other csr slots are all `IncomingArg` tho, so maybe its best to keep that slot kind and figure out what the explicit offset needs to be, rather than rely on `layout_stack`.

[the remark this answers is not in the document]
iximeow: I think we'd need to make some changes to `stack_layout` and `StackSlot` for "correct offset in `StackSlotKind::IncomingArg`" to be a question we can really answer. `stack_addr` and friends assume the prologue has already done its stack setup, and, seemingly, that `rsp` won't move through the body of the function. So for the address of an `IncomingArg` slot to be used _in_ the prologue, we'd need to defer actual resolution of stack offsets even after the instructions are added.

The other option is tracking the offset ourselves in the prologue, which I expect would look like tracking the amount we've moved SP with pushes/explicit adjustment, then acknowledging that the address from `stack_addr(IncomingArgSlot)` will be wrong (it'd be off by however much we've pushed) and correcting for that after the fact. Because floating point save/restore is past any GPR pushes, I think the offset will always be valid, just smaller than it "should" be.

I'd be more suspicious that I'm misunderstanding something, but I don't think we have other instances of wanting to do stack loads/stores in prologues/epilogues, so maybe this is actually a rough spot? :(

bjorn3:
Close this?

iximeow:
Yup, I was going through open conversations to figure out which unresolved conversations needed to move.

cranelift-codegen/src/isa/x86/abi.rs
jaykrell: The VIsual C++ code above is weird.
I do not recommend emulating it.
It is surely correct, and it might be optimized, but do not emulate it.
Visual C++ definitely does some strange things as optimizations.
In particular, it stores to the stack before adjusting rsp, and then has to misstate frame pointer offsets. "Normal" code only pushes prior to storing to the stack, I believe.
Maybe try looking at unoptimized output instead.
Or link /dump /disasm /unwindinfo ntoskrnl.exe to find simple stuff written in assembly.
Or read and follow the spec -- Visual C++ kinda skirts the spec.

PR #1381, opened 2020-02-07
Catch a few typos in comments

ggreif:
These were sitting in my tree for some time.


fitzgen:
Woops, looks like this has a conflict and needs a rebase.

ggreif:
> Woops, looks like this has a conflict and needs a rebase.

Resolved!

PR #1385, opened 2020-02-11
Add ineg legalization for scalar integer types

### LINKED ISSUE #1383: ineg instruction on ssa values: register affinity never assigned?
Thanks for opening a bug report on Cranelift! Please answer the questions below
if they're relevant and delete this text before submitting.

- My cranelift frontend is generating the following IL, and `module.define_function(...)` is failing with the "ghost value used by real instruction", which seems like maybe register allocation skipped over the `ineg` result??
```
function u0:0(i32, i64) -> i32 system_v {
    gv0 = symbol colocated u1:0
    sig0 = (i64, i64, i64, i64) -> i32 system_v
    fn0 = u0:1 sig0

block0(v0: i32, v1: i64):
    v2 = iconst.i32 1
    v3 = ineg v2
    return v3
}

error defining function "main": Compilation(
    Verifier(
        VerifierErrors(
            [
                VerifierError {
                    location: inst2,
                    context: None,
                    message: "v3 is a ghost value used by a real [Op1ret#c3] instruction",
                },
            ],
        ),
    ),
)
```
- I guess I thought ineg could be used on scalar registers, but I only see simd filetests using it?
- `cranelift = "0.58.0"`
- This is on Mac OS, using rust 1.41.0. I'm generating object files using the `cranelift-faerie` integration

peterdelevoryas:
- Closes #1383 
- Cranelift already has custom legalization for ineg on vector integer types, might as well add it for scalar integer types!
- This PR includes one new test case, it's relatively meager, maybe I could add cases for all the integer type variants? (i8, i16, i32, i64). Not sure if that's too much or not!
- I don't know who could review this, maybe `bnjbvr` or `bjorn3`?

I added some comments in the diff that refer to concerns I had

### DIFF
--- a/cranelift-codegen/meta/src/isa/x86/legalize.rs
+++ b/cranelift-codegen/meta/src/isa/x86/legalize.rs
@@ -319,6 +319,8 @@ pub(crate) fn define(shared: &mut SharedDefinitions, x86_instructions: &Instruct
         ],
     );
 
+    group.custom_legalize(ineg, "convert_ineg");
+
     group.build_and_add_to(&mut shared.transform_groups);
 
     let mut narrow = TransformGroupBuilder::new(
@@ -612,4 +614,17 @@ pub(crate) fn define(shared: &mut SharedDefinitions, x86_instructions: &Instruct
     narrow.custom_legalize(ineg, "convert_ineg");
 
     narrow.build_and_add_to(&mut shared.transform_groups);
+
+    let mut widen = TransformGroupBuilder::new(
+        "x86_widen",
+        r#"
+    Legalize instructions by widening.
+
+    Use x86-specific instructions if needed."#,
+    )
+    .isa("x86")
+    .chain_with(shared.transform_groups.by_name("widen").id);
+
+    widen.custom_legalize(ineg, "convert_ineg");
+    widen.build_and_add_to(&mut shared.transform_groups);
 }
--- a/cranelift-codegen/meta/src/isa/x86/mod.rs
+++ b/cranelift-codegen/meta/src/isa/x86/mod.rs
@@ -32,16 +32,15 @@ pub(crate) fn define(shared_defs: &mut SharedDefinitions) -> TargetIsa {
     let mut x86_32 = CpuMode::new("I32");
 
     let expand_flags = shared_defs.transform_groups.by_name("expand_flags");
-    let narrow_flags = shared_defs.transform_groups.by_name("narrow_flags");
-    let widen = shared_defs.transform_groups.by_name("widen");
+    let x86_widen = shared_defs.transform_groups.by_name("x86_widen");
     let x86_narrow = shared_defs.transform_groups.by_name("x86_narrow");
     let x86_expand = shared_defs.transform_groups.by_name("x86_expand");
 
     x86_32.legalize_monomorphic(expand_flags);
-    x86_32.legalize_default(narrow_flags);
+    x86_32.legalize_default(x86_narrow);
     x86_32.legalize_type(B1, expand_flags);
-    x86_32.legalize_type(I8, widen);
-    x86_32.legalize_type(I16, widen);
+    x86_32.legalize_type(I8, x86_widen);
+    x86_32.legalize_type(I16, x86_widen);
     x86_32.legalize_type(I32, x86_expand);
     x86_32.legalize_value_type(ReferenceType(R32), x86_expand);
     x86_32.legalize_type(F32, x86_expand);
@@ -50,8 +49,8 @@ pub(crate) fn define(shared_defs: &mut SharedDefinitions) -> TargetIsa {
     x86_64.legalize_monomorphic(expand_flags);
     x86_64.legalize_default(x86_narrow);
     x86_64.legalize_type(B1, expand_flags);
-    x86_64.legalize_type(I8, widen);
-    x86_64.legalize_type(I16, widen);
+    x86_64.legalize_type(I8, x86_widen);
+    x86_64.legalize_type(I16, x86_widen);
     x86_64.legalize_type(I32, x86_expand);
     x86_64.legalize_type(I64, x86_expand);
     x86_64.legalize_value_type(ReferenceType(R64), x86_expand);
--- a/cranelift-codegen/src/isa/x86/enc_tables.rs
+++ b/cranelift-codegen/src/isa/x86/enc_tables.rs
@@ -1213,7 +1213,7 @@ fn convert_insertlane(
     }
 }
 
-/// For SIMD negation, convert an `ineg` to a `vconst + isub`.
+/// For SIMD or scalar integer negation, convert `ineg` to `vconst + isub` or `iconst + isub`.
 fn convert_ineg(
     inst: ir::Inst,
     func: &mut ir::Function,
@@ -1229,10 +1229,16 @@ fn convert_ineg(
     } = pos.func.dfg[inst]
     {
         let value_type = pos.func.dfg.value_type(arg);
-        if value_type.is_vector() && value_type.lane_type().is_int() {
+        let zero_value = if value_type.is_vector() && value_type.lane_type().is_int() {
             let zero_immediate = pos.func.dfg.constants.insert(vec![0; 16].into());
-            let zero_value = pos.ins().vconst(value_type, zero_immediate); // this should be legalized to a PXOR
-            pos.func.dfg.replace(inst).isub(zero_value, arg);
-        }
+            pos.ins().vconst(value_type, zero_immediate) // this should be legalized to a PXOR
+        } else if value_type.is_int() {
+            pos.ins().iconst(value_type, 0)
+        } else {
+            panic!("Can't convert ineg of type {}", value_type)
+        };
+        pos.func.dfg.replace(inst).isub(zero_value, arg);
+    } else {
+        unreachable!()
     }
 }
--- a/filetests/isa/x86/legalize-custom.clif
+++ b/filetests/isa/x86/legalize-custom.clif
@@ -90,3 +90,44 @@ block0(v0: f32, v1: f32):
     ; check: $done(v2: f32):
     ; nextln: return v2
 }
+
+function %ineg_legalized_i8() {
+block0:
+    v0 = iconst.i8 1
+    v1 = ineg v0
+    ; check: v2 = iconst.i32 1
+    ; nextln: v0 = ireduce.i8 v2
+    ; nextln: v3 = iconst.i8 0
+    ; nextln: v4 = uextend.i32 v3
+    ; nextln: v5 = uextend.i32 v0
+    ; nextln: v6 = isub v4, v5
+    ; nextln: v1 = ireduce.i8 v6
+
+    return
+}
+
+function %ineg_legalized_i16() {
+block0:
+    v0 = iconst.i16 1
+    v1 = ineg v0
+    ; check: v2 = iconst.i32 1
+    ; nextln: v0 = ireduce.i16 v2
+    ; nextln: v3 = iconst.i16 0
+    ; nextln: v4 = uextend.i32 v3
+    ; nextln: v5 = uextend.i32 v0
+    ; nextln: v6 = isub v4, v5
+    ; nextln: v1 = ireduce.i16 v6
+
+    return
+}
+
+function %ineg_legalized_i32() {
+block0:
+    v0 = iconst.i32 1
+    v1 = ineg v0
+    ; check: v0 = iconst.i32 1
+    ; nextln: v2 = iconst.i32 0
+    ; nextln: v1 = isub v2, v0
+
+    return
+}
--- a/filetests/isa/x86/legalize-i64.clif
+++ b/filetests/isa/x86/legalize-i64.clif
@@ -338,3 +338,20 @@ block0(v1: i64):
     ; nextln: v10 = bor $b1, $c2
     return v10
 }
+
+function %ineg_legalized_i64() {
+block0:
+    v0 = iconst.i64 1
+    v1 = ineg v0
+    ; check: v2 = iconst.i32 1
+    ; nextln: v3 = iconst.i32 0
+    ; nextln: v0 = iconcat v2, v3
+    ; nextln: v5 = iconst.i32 0
+    ; nextln: v6 = iconst.i32 0
+    ; nextln: v4 = iconcat v5, v6
+    ; nextln: v7, v8 = isub_ifbout v5, v2
+    ; nextln: v9 = isub_ifbin v6, v3, v8
+    ; nextln: v1 = iconcat v7, v9
+
+    return
+}
--- a/filetests/isa/x86/legalize-ineg-x86_64.clif
+++ b/filetests/isa/x86/legalize-ineg-x86_64.clif
@@ -1,0 +1,13 @@
+; Test the custom legalization of ineg.i64 on x86_64.
+test legalizer
+target x86_64
+
+function %ineg_legalized_i64() {
+block0:
+    v0 = iconst.i64 1
+    v1 = ineg v0
+    ; check: v0 = iconst.i64 1
+    ; nextln: v2 = iconst.i64 0
+    ; nextln: v1 = isub v2, v0
+    return
+}


bjorn3:
Please also add one test for `i8`. Integer types <32bit often need extra legalizations.

peterdelevoryas:
> Please also add one test for `i8`. Integer types <32bit often need extra legalizations.

Ya ok I have to admit defeat: I added an i8 test, and I tried debugging it, but it makes no sense: the ineg legalization of the i8 returns "done", not "legalized", presumably because it thinks that the ineg is already legal?? https://pastebin.com/0E1Bs06A I added print statements that trace the result of `legalize_inst`, and for the i32 case it returns "legalized" and then it legalizes an iconst and an isub, presumably the two instructions that get inserted by `convert_ineg`, but for the ineg of the i8 value it doesn't occur. This must have something to do with the ireduce, I just can't figure out what yet.

cranelift-codegen/meta/src/isa/x86/legalize.rs
```diff
@@ -319,6 +319,8 @@ pub(crate) fn define(shared: &mut SharedDefinitions, x86_instructions: &Instruct
         ],
     );
 
+    group.custom_legalize(ineg, "convert_ineg");
```
bjorn3: This will need to be added to widen too to legalize i8 and i16.

cranelift-codegen/src/isa/x86/enc_tables.rs
```diff
@@ -1223,16 +1223,25 @@ fn convert_ineg(
     let mut pos = FuncCursor::new(func).at_inst(inst);
     pos.use_srcloc(inst);
 
+    // FIXME Is it necessary to do this? I thought the generated code only calls this function if
+    // the opcode is ineg, maybe this could just be an assert? I guess this is a very common
+    // pattern in this file, so maybe there's a good reason?
```
bjorn3: This is necessary to get the arg passes to ineg. Adding an `unreachable!()` as else would be nice though.

cranelift-codegen/meta/src/isa/x86/legalize.rs
peterdelevoryas: Ohhhh ok, I was trying to find widen and couldn't find it, let me check again

peterdelevoryas: ohhh I get it now, ok I'll make a widen group and add the custom legalization there just for x86, now in the shared-common one, since all this ineg stuff is in x86 right now.

cranelift-codegen/src/isa/x86/enc_tables.rs
```diff
@@ -1223,16 +1223,25 @@ fn convert_ineg(
...
-            let zero_value = pos.ins().vconst(value_type, zero_immediate); // this should be legalized to a PXOR
-            pos.func.dfg.replace(inst).isub(zero_value, arg);
-        }
+            pos.ins().vconst(value_type, zero_immediate) // this should be legalized to a PXOR
+        } else if value_type.is_int() {
+            pos.ins().iconst(value_type, 0)
+        } else {
+            // FIXME Should I panic here? Or maybe add something to the verifier? The ghost error
+            // in the verifier is kind of unintuitive.
+            panic!()
```
abrown: I think so; I see that type of thing up above. Something like the following would be good to get both the type and the legalization that failed: `panic!("Can't convert ineg of {}", valuetype)`.

cranelift-codegen/meta/src/isa/x86/legalize.rs
peterdelevoryas: Hey, I tried adding this to the widen TransformGroup, but that didn't seem to work. I saw a comment in another file about ordering being significant: maybe I need to add the widen group legalizations before or after the narrow/expand additions?

bjorn3: I think using x86_widen instead of widen at https://github.com/bytecodealliance/cranelift/blob/23e9bdb2d99fb8f554793cb87f1578cc543c355f/cranelift-codegen/meta/src/isa/x86/mod.rs#L36 should fix it.

cranelift-codegen/meta/src/isa/x86/legalize.rs
```diff
@@ -8,6 +8,20 @@ use crate::shared::Definitions as SharedDefinitions;
...
+        r#"
+    Legalize instructions by narrowing.
+
+    Use x86-specific instructions if needed."#,
+    )
+    .isa("x86")
+    .chain_with(shared.transform_groups.by_name("widen").id);
+
+    widen.custom_legalize(ineg, "convert_ineg");
+    widen.build_and_add_to(&mut shared.transform_groups);
```
peterdelevoryas: I thought this would be sufficient to get the transformation to apply to i8/i16, but it doesn't seem to be working? I'm getting stuck debugging this because I can't seem to find the place where we apply the transformations. Maybe this is in the generated Rust file?

peterdelevoryas: nvm, I think I see the spot where x86_narrow/etc gets tied to the x86 legalization, maybe I can fix this

cranelift-codegen/meta/src/isa/x86/legalize.rs
peterdelevoryas: That did fix it! And I added a test for i16 and that worked too. So, i8, i16, and i32 work. But, i64 doesn't: I think these changes made sure `convert_ineg` is added to widen, narrow, and expand, is there something missing still?

```
---- filetests stdout ----
FAIL filetests/isa/x86/legalize-custom.clif: legalizer(%ineg_legalized_i):
filecheck failed:
#0 regex: V=v\d+
#1 regex: BB=block\d+
#2 check: v2 = iconst.i32 1
#3 check: v3 = iconst.i32 0
#4 check: v0 = iconcat v2, v3
#5 nextln: v1 = isub v2, v0
> function %ineg_legalized_i() fast {
>                                 block0:
> [Op1pu_id#b8]                       v2 = iconst.i32 1
                                      ^~~~~~~~~~~~~~~~~
Matched #2: \bv2 = iconst\.i32 1\b
> [Op1pu_id#b8]                       v3 = iconst.i32 0
                                      ^~~~~~~~~~~~~~~~~
Matched #3: \bv3 = iconst\.i32 0\b
> [-]                                 v0 = iconcat v2, v3
                                      ^~~~~~~~~~~~~~~~~~~
Matched #4: \bv0 = iconcat v2, v3\b
> [-]                                 v1 = ineg v0
Missed #5: \bv1 = isub v2, v0\b
> [Op1ret#c3]                         return
> }
```

peterdelevoryas:
To summarize: everything is working except the i64 test I added, and I still have to remove the comments I added and add a better error message to the panic

cranelift-codegen/meta/src/isa/x86/legalize.rs
```diff
@@ -612,4 +614,17 @@ pub(crate) fn define(shared: &mut SharedDefinitions, x86_instructions: &Instruct
     narrow.custom_legalize(ineg, "convert_ineg");
 
     narrow.build_and_add_to(&mut shared.transform_groups);
+
+    let mut widen = TransformGroupBuilder::new(
+        "x86_widen",
+        r#"
+    Legalize instructions by narrowing.
```
abrown: "Legalize instructions by widening"?

filetests/isa/x86/legalize-custom.clif
```diff
@@ -90,3 +90,56 @@ block0(v0: f32, v1: f32):
...
+
+function %ineg_legalized_i8() {
+block0:
+    v0 = iconst.i8 1
+    v1 = ineg v0
+    ; check: v2 = iconst.i32 1
+    ; nextln: v0 = ireduce.i8 v2
+    ; nextln: v3 = iconst.i8 0
+    ; nextln: v4 = uextend.i32 v3
+    ; nextln: v5 = uextend.i32 v0
```
abrown: Not sure I get why v2 and v3 have different types initially: shouldn't they both start out as `i8`s and be extended? And we have to extend and then reduce because we don't have an encoding for `isub.i8`? (@peterdelevoryas, no need to answer; this is more for @bjorn3 as he is more familiar with this than I am)

filetests/isa/x86/legalize-custom.clif
```diff
@@ -90,3 +90,56 @@ block0(v0: f32, v1: f32):
...
+    return
+}
+
+function %ineg_legalized_i64() {
+block0:
+    v0 = iconst.i64 1
+    v1 = ineg v0
+    ; check: v2 = iconst.i32 1
+    ; check: v3 = iconst.i32 0
+    ; check: v0 = iconcat v2, v3
```
abrown: @bjorn3, I would not expect this `iconcat`; can't we directly encode an `iconst.i64`?

cranelift-codegen/meta/src/isa/x86/legalize.rs
peterdelevoryas: Ah woops thanks for catching that!

filetests/isa/x86/legalize-custom.clif
bjorn3: This file has `target i686` before `target x86_64`. How is that handled? Does it test both or only the first?

filetests/isa/x86/legalize-custom.clif
bjorn3: iconst.i8 is legalized as iconst.i32 + ireduce.i8. isub.i8 is legalized as uextend.i32 + isub.i32 + ireduce.i8.

filetests/isa/x86/legalize-custom.clif
abrown: Learn something new every day: apparently parsing tests [accumulates the targets](https://github.com/bytecodealliance/cranelift/blob/master/cranelift-reader/src/parser.rs#L992-L1025) and `test_tuples` will expand the different targets (in this case since the legalizer `needs_isa` we should have a test tuple for each `target` from [here](https://github.com/bytecodealliance/cranelift/blob/master/cranelift-filetests/src/runone.rs#L121-L125)). 

But then this doesn't make sense: how could these checks pass when `target x86_64` is set? Is that the build failure? No, it looks like the current build failure happens because `ineg` isn't getting legalized at all for `i64`?

peterdelevoryas: Yeah the build failure is happening because `ineg` still isn't getting legalized for `i64`, I still haven't figured out why that is. This must be because I need to extend another transform group right? That was the reason for `i8` and `i16` not getting legalized initially (I think)

bjorn3: There is also the `narrow` group for ints > pointer size.

peterdelevoryas:
Ok, so, I think the changes are finally done (`ineg` legalization tested on i686/x86_64 with all scalar integer types), one of my changes results in an unused import though

```
warning: unused import: `crate::ir::InstBuilder`
    --> /Users/peterd/src/peterd-cranelift/target/debug/build/cranelift-codegen-b0deee4e3070f0b5/out/legalize-x86.rs:1619:9
     |
1619 |     use crate::ir::InstBuilder;
     |         ^^^^^^^^^^^^^^^^^^^^^^
     |
     = note: `#[warn(unused_imports)]` on by default
```
This is the generated file:
```
/// Legalize instructions by narrowing.
///
/// Use x86-specific instructions if needed.
#[allow(unused_variables,unused_assignments,non_snake_case)]
pub fn x86_widen(
    inst: crate::ir::Inst,
    func: &mut crate::ir::Function,
    cfg: &mut crate::flowgraph::ControlFlowGraph,
    isa: &dyn crate::isa::TargetIsa,
) -> bool {
    use crate::ir::InstBuilder;
    use crate::cursor::{Cursor, FuncCursor};
    let mut pos = FuncCursor::new(func).at_inst(inst);
    pos.use_srcloc(inst);
    {
        match pos.func.dfg[inst].opcode() {
            ir::Opcode::Ineg => {
                convert_ineg(inst, func, cfg, isa);
                return true;
            }

            _ => {},
        }
    }
    crate::legalizer::widen(inst, func, cfg, isa)
}
```
I'm not sure exactly how to fix this, is there a way I could just add the custom legalization to the existing `widen` group without making a new `x86_widen` group? Maybe that would prevent the unused import?

cranelift-codegen/meta/src/isa/x86/legalize.rs
peterdelevoryas: Ok, this should be fixed now

cranelift-codegen/src/isa/x86/enc_tables.rs
peterdelevoryas: I added this panic message

cranelift-codegen/src/isa/x86/enc_tables.rs
peterdelevoryas: Added an `else { unreachable!() }` clause

cranelift-codegen/meta/src/isa/x86/mod.rs
```diff
@@ -32,13 +32,12 @@ pub(crate) fn define(shared_defs: &mut SharedDefinitions) -> TargetIsa {
...
     let expand_flags = shared_defs.transform_groups.by_name("expand_flags");
-    let narrow_flags = shared_defs.transform_groups.by_name("narrow_flags");
-    let widen = shared_defs.transform_groups.by_name("widen");
+    let widen = shared_defs.transform_groups.by_name("x86_widen");
     let x86_narrow = shared_defs.transform_groups.by_name("x86_narrow");
     let x86_expand = shared_defs.transform_groups.by_name("x86_expand");
 
     x86_32.legalize_monomorphic(expand_flags);
-    x86_32.legalize_default(narrow_flags);
+    x86_32.legalize_default(x86_narrow);
```
abrown: I want to make sure I understand this: x86_narrow is chained with narrow_flags so this just makes `x86_32` consistent with `x86_64`--i.e. was this necessary to get this patch to work?

cranelift-codegen/meta/src/isa/x86/mod.rs
```diff
@@ -32,13 +32,12 @@ pub(crate) fn define(shared_defs: &mut SharedDefinitions) -> TargetIsa {
     let mut x86_32 = CpuMode::new("I32");
 
     let expand_flags = shared_defs.transform_groups.by_name("expand_flags");
-    let narrow_flags = shared_defs.transform_groups.by_name("narrow_flags");
-    let widen = shared_defs.transform_groups.by_name("widen");
+    let widen = shared_defs.transform_groups.by_name("x86_widen");
```
abrown: Perhaps we should rename this to `x86_widen` (here and where used) to be very clear what is going on.

cranelift-codegen/meta/src/isa/x86/mod.rs
peterdelevoryas: That's correct, I think before this, the legalization of `ineg.i64` was not being applied on i686. I'll double check really quickly though and show the error here

cranelift-codegen/meta/src/isa/x86/mod.rs
peterdelevoryas: Oh yeah I can definitely do that!

abrown: Sounds good, I'll merge this after your push.

cranelift-codegen/meta/src/isa/x86/mod.rs
peterdelevoryas:
```
running 1 test
test filetests ... FAILED

failures:

---- filetests stdout ----
FAIL filetests/isa/x86/legalize-i64.clif: legalizer(%ineg_legalized_i):
filecheck failed:
#0 regex: V=v\d+
#1 check: v2 = iconst.i32 1
#2 nextln: v3 = iconst.i32 0
#3 nextln: v0 = iconcat v2, v3
#4 nextln: v5 = iconst.i32 0
#5 nextln: v6 = iconst.i32 0
#6 nextln: v4 = iconcat v5, v6
#7 nextln: v7, v8 = isub_ifbout v5, v2
#8 nextln: v9 = isub_ifbin v6, v3, v8
#9 nextln: v1 = iconcat v7, v9
> function %ineg_legalized_i() fast {
>                                 block0:
> [Op1pu_id#b8]                       v2 = iconst.i32 1
                                      ^~~~~~~~~~~~~~~~~
Matched #1: \bv2 = iconst\.i32 1\b
> [Op1pu_id#b8]                       v3 = iconst.i32 0
                                      ^~~~~~~~~~~~~~~~~
Matched #2: \bv3 = iconst\.i32 0\b
> [-]                                 v0 = iconcat v2, v3
                                      ^~~~~~~~~~~~~~~~~~~
Matched #3: \bv0 = iconcat v2, v3\b
> [-]                                 v1 = ineg v0
Missed #4: \bv5 = iconst\.i32 0\b
> [Op1ret#c3]                         return
> }
```

(with this diff from the current stuff)
```
diff --git a/cranelift-codegen/meta/src/isa/x86/mod.rs b/cranelift-codegen/meta/src/isa/x86/mod.rs
index 52d03f54..3fbd78b7 100644
--- a/cranelift-codegen/meta/src/isa/x86/mod.rs
+++ b/cranelift-codegen/meta/src/isa/x86/mod.rs
@@ -32,12 +32,13 @@ pub(crate) fn define(shared_defs: &mut SharedDefinitions) -> TargetIsa {
     let mut x86_32 = CpuMode::new("I32");

     let expand_flags = shared_defs.transform_groups.by_name("expand_flags");
+    let narrow_flags = shared_defs.transform_groups.by_name("narrow_flags");
     let widen = shared_defs.transform_groups.by_name("x86_widen");
     let x86_narrow = shared_defs.transform_groups.by_name("x86_narrow");
     let x86_expand = shared_defs.transform_groups.by_name("x86_expand");

     x86_32.legalize_monomorphic(expand_flags);
-    x86_32.legalize_default(x86_narrow);
+    x86_32.legalize_default(narrow_flags);
     x86_32.legalize_type(B1, expand_flags);
     x86_32.legalize_type(I8, widen);
     x86_32.legalize_type(I16, widen);
```

abrown:
@peterdelevoryas, thanks!

peterdelevoryas:
Hey no thank you @abrown, and everyone else, for helping with this! I really got stuck on the whole transform group thing, but I think I get the idea now. I've really been enjoying using Cranelift, and it was fun making these changes!

PR #1388, opened 2020-02-12
Infer REX prefix for SIMD `store` and `vconst` instructions

abrown:
- [x] This has been discussed in issues #1090 and #1379.
- [x] A short description of what this does, why it is needed: this demonstrates inferring REX prefixes for SIMD `store` and `vconst` instructions; if we think this is the right direction, we have some additional to do to infer REX prefixes for the following recipes:

__enc_32_64__: fstDisp8, fstDisp32, fld, fldDisp8, fldDisp32, fspillSib32, fregspill32, ffillSib32, fregfill32, frmov, furm, fa, fax, f_ib
__enc_32_64_maybe_isap__: r_ib_unsigned_r, fa_ib, fa, r_ib_unsigned_gpr, fcmp, icscc_fpr
__enc_both__: frurm, pfcmp, fa, furm

### DIFF
--- a/cranelift-codegen/src/isa/x86/enc_tables.rs
+++ b/cranelift-codegen/src/isa/x86/enc_tables.rs
@@ -123,6 +123,22 @@ fn size_plus_maybe_sib_or_offset_for_inreg_1(
     sizing.base_size + if needs_sib_or_offset { 1 } else { 0 }
 }
 
+/// Calculates the size while inferring if the first and second input registers (inreg0, inreg1)
+/// require a dynamic REX prefix and if the second input register (inreg1) requires a SIB or offset.
+fn size_plus_maybe_sib_or_offset_inreg1_plus_rex_prefix_for_inreg0_inreg1(
+    sizing: &RecipeSizing,
+    enc: Encoding,
+    inst: Inst,
+    divert: &RegDiversions,
+    func: &Function,
+) -> u8 {
+    let needs_rex = (EncodingBits::from(enc.bits()).rex_w() != 0)
+        || test_input(0, inst, divert, func, is_extended_reg)
+        || test_input(1, inst, divert, func, is_extended_reg);
+    size_plus_maybe_sib_or_offset_for_inreg_1(sizing, enc, inst, divert, func)
+        + if needs_rex { 1 } else { 0 }
+}
+
 /// Infers whether a dynamic REX prefix will be emitted, for use with one input reg.
 ///
 /// A REX prefix is known to be emitted if either:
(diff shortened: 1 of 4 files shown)


filetests/isa/x86/binary64.clif
```diff
@@ -1679,3 +1679,11 @@ block0:
     [-, %r10]   v0 = bconst.b64 true    ; bin: 41 ba 00000001
     return
 }
+
+function %V128() {
+block0:
+    [-,%r10]            v3 = iconst.i64 0x2102_0304_f1f2_f3f4 ; bin: 49 ba 21020304f1f2f3f4
+    [-, %xmm9]          v4 = vconst.i32x4 [0 1 2 3]  ; bin: 44 0f 10 0d 00000005 PCRelRodata4(23)
```
iximeow: :tada:

cranelift-codegen/src/isa/x86/enc_tables.rs
```diff
@@ -123,6 +123,21 @@ fn size_plus_maybe_sib_or_offset_for_inreg_1(
...
+/// prefix and if the second input register (inreg1) requires a SIB or offset.
+fn size_plus_maybe_sib_or_offset_inreg1_plus_rex_prefix_for_inreg0(
+    sizing: &RecipeSizing,
+    enc: Encoding,
+    inst: Inst,
+    divert: &RegDiversions,
+    func: &Function,
+) -> u8 {
+    let needs_rex = (EncodingBits::from(enc.bits()).rex_w() != 0)
+        || test_input(0, inst, divert, func, is_extended_reg);
```
iximeow: I think this also needs a `|| test_input(1, inst, divert, func, is_extended_reg)` in the case that the base register is `r8-r15`? I have admit I'm not sure which order registers are specified but the other part of the [recipe this is used in](https://github.com/bytecodealliance/cranelift/blob/ae77d49fa49c7dccceb16f0a71380a25cf45531a/cranelift-codegen/meta/src/isa/x86/recipes.rs#L1506) has me suspect. I'm also not sure how well our tests cover getting less-likely encodings like that - I'd not think about it if I didn't just try to do `[r11]` a few days ago.

abrown:
I think we're good-to-go. I'm not really looking forward to doing this same thing for 20+ other recipes so any suggestions are appreciated. The `"size_plus_maybe_sib_or_offset_for_inreg_0_plus_rex_prefix_for_inreg0_outreg0"` code is not code I would hope to make a career of... is there another way?

abrown:
...and by "good-to-go" I mean I ran the wasmtime tests and everything passed except some wasm_c_examples for unrelated reasons like `make: *** No rule to make target 'wasm-c-api/out/example/reflect-c'`

PR #1389, opened 2020-02-13
wasm: Add support for passive data and element segments

fitzgen:
- [X] This has been discussed in issue https://github.com/bytecodealliance/wasmtime/issues/928

- [X] A short description of what this does, why it is needed; if the
  first.

  This is part of the bulk memory and reference types proposals.

- [X] This PR contains test cases, if meaningful.

### DIFF
--- a/cranelift-wasm/src/environ/spec.rs
+++ b/cranelift-wasm/src/environ/spec.rs
@@ -8,7 +8,8 @@
 
 use crate::state::{FuncTranslationState, ModuleTranslationState};
 use crate::translation_utils::{
-    FuncIndex, Global, GlobalIndex, Memory, MemoryIndex, SignatureIndex, Table, TableIndex,
+    FuncIndex, Global, GlobalIndex, Memory, MemoryIndex, PassiveDataIndex, PassiveElemIndex,
+    SignatureIndex, Table, TableIndex,
 };
 use core::convert::From;
 use cranelift_codegen::cursor::FuncCursor;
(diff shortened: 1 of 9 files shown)


cranelift-wasm/src/environ/spec.rs
```diff
@@ -600,6 +601,29 @@ pub trait ModuleEnvironment<'data>: TargetEnvironment {
...
+        index: PassiveElemIndex,
+        elements: Box<[FuncIndex]>,
+    ) -> WasmResult<()>;
+
+    /// Provides the number of passive data segments up front.
+    ///
+    /// By default this does nothing, but implementations may use this to
+    /// pre-allocate memory if desired.
+    fn reserve_passive_data(&mut self, count: u32) -> WasmResult<()> {
+        let _ = count;
```
eqrion: What's idiomatic rust for unused function arguments? In `dummy.rs` I see unused function arguments prefixed with `_`, and here we're adding a `_` use inside of the function def. I'm fine with either, but would prefer it to be consistent.

bjorn3: Prefixing the argument with `_` is the most common. (I haven't every seen it this way :) )

fitzgen: It is true that one would usually use `_prefix`, but because it is in a trait method's default implementation, that underscore would become part of the variable names in the auto-generated rustdocs. The separate underscore is to avoid underscore-prefixed variable names in the rustdocs.

Issue #951, opened 2019-08-28
cranelift-reader should have an option to error on pathological cases


bjorn3:
The following function will attempt to allocate 477777777 ebbs.

```
function %a(){
ebb477777777:
}
```

_Originally posted by @bjorn3 in https://github.com/CraneStation/cranelift/pull/944#issuecomment-525686450_

bnjbvr:
Fixed by https://github.com/bytecodealliance/cranelift/commit/7cbf6e7d87b46e0a820369d1e69da4c2e986f701

PR #1391, opened 2020-02-13
Add maximum threshold for number of ebbs per function #951

nalmt:
To fix this case that may take forever to compile (issue #951):
function %a() {
ebb477777777:
}

We decide to define a maximum threshold for the number of ebbs in functions.

Based on a large WASM program (https://github.com/mozilla/perf-automation/blob/master/benchmarks/wasm-misc/AngryBots.wasm), its IR functions does not exceed 1414 ebbs.
A number 100 times greater seems (currently) enough to define our maximum threshold.

static MAX_EBBS_IN_A_FUNCTION: u32 = 100000;

if ebb_num.as_u32() > MAX_EBBS_IN_A_FUNCTION {
    return Err(self.error("too many ebbs"));
}

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "cranelift-tools"
 authors = ["The Cranelift Project Developers"]
-version = "0.56.0"
+version = "0.57.0"
 description = "Binaries for testing the Cranelift libraries"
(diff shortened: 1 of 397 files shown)


jgouly:
EBBs have been renamed to Blocks, so you'll want to rebase and also rename the variables!

PR #1397, opened 2020-02-19
Add maximum threshold for number of blocks per function #951

nalmt:
To fix this case (#951) that may take forever to compile:

```
function %a(){
ebb477777777:
}
```

We decide to define a maximum threshold for the number of blocks in functions.

Based on a large WASM program (https://github.com/mozilla/perf-automation/blob/master/benchmarks/wasm-misc/AngryBots.wasm),
its IR functions does not exceed 1414 blocks. A number 100 times greater (100,000 blocks) seems (currently) enough to define our maximum threshold.

To make this quick benchmark the cranelift-wasm/src/func_translator.rs file has been modified like this:

```
static mut MAX: usize = 0;

pub fn translate_from_reader<FE: FuncEnvironment + ?Sized>(...) {

    [...]

    builder.finalize();

    // the compiler is single threaded
    unsafe {
        if func.dfg.num_ebbs() > MAX {
            MAX = func.dfg.num_ebbs();
            println!("MAX {}", MAX);
        }
    }

    Ok(())
}
```

### DIFF
--- a/cranelift-reader/src/parser.rs
+++ b/cranelift-reader/src/parser.rs
@@ -27,6 +27,9 @@ use std::str::FromStr;
 use std::{u16, u32};
 use target_lexicon::Triple;
 
+/// After some quick benchmarks a program should never have more than 100,000 blocks.
+const MAX_BLOCKS_IN_A_FUNCTION: u32 = 100_000;
+
 /// Parse the entire `text` into a list of functions.
 ///
 /// Any test commands or target declarations are ignored.
@@ -1768,6 +1771,10 @@ impl<'a> Parser<'a> {
         let block_num = self.match_block("expected block header")?;
         let block = ctx.add_block(block_num, self.loc)?;
 
+        if block_num.as_u32() >= MAX_BLOCKS_IN_A_FUNCTION {
+            return Err(self.error("too many blocks"));
+        }
+
         if !self.optional(Token::Colon) {
             // block-header ::= Block(block) [ * block-params ] ":"
             self.parse_block_params(ctx, block)?;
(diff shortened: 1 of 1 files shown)


cranelift-reader/src/parser.rs
```diff
@@ -27,6 +27,9 @@ use std::str::FromStr;
 use std::{u16, u32};
 use target_lexicon::Triple;
 
+/// After some quick benchmarks a program should never have more than 100,000 blocks.
+static MAX_BLOCKS_IN_A_FUNCTION: u32 = 100000;
```
bjorn3:
```suggestion
const MAX_BLOCKS_IN_A_FUNCTION: u32 = 100_000;
```

cranelift-reader/src/parser.rs
```diff
@@ -1768,6 +1771,10 @@ impl<'a> Parser<'a> {
         let block_num = self.match_block("expected block header")?;
         let block = ctx.add_block(block_num, self.loc)?;
 
+        if block_num.as_u32() > MAX_BLOCKS_IN_A_FUNCTION {
```
bnjbvr: nit: since block numbers start at 0, there's an off-by-one here, and it will accept 100,000 blocks, but not 100,001, which is why the test had to contain the latter number. Can you use ">=" here instead, please?

cranelift-reader/src/parser.rs
```diff
@@ -2979,6 +2986,24 @@ mod tests {
...
 
+    #[test]
+    fn ebbs_number() {
+        let ParseError {
+            location,
+            message,
+            is_warning,
+        } = Parser::new(
+            "function %a() {
+                block100001:",
```
bnjbvr: (and then you can adjust this number to be 100000.)

Issue #1386, opened 2020-02-11
add "raw" function definition interface to cranelift-module


froydnj:
#### Feature

While looking at rewriting `lucetc` to use `object` instead of `faerie` (primarily to get Windows support "for free" in `lucetc`), I think the only reason that "raw" `faerie` interfaces need to be used instead of going through `cranelift-module` is that the definition of the stack probe cannot be done as clif:

https://github.com/bytecodealliance/lucet/blob/cf53a3ba5d89174c26ba4ccdd5043628664da72f/lucetc/src/output.rs#L56-L62

as the stack probe must be defined as a certain sequence of instructions that other code needs to know the layout of.  And various later things after the above point need to know about the existence of the stack probe, so they can't be written in terms of the `cranelift-module` interface and need to be written with "raw" `faerie`.

The straightforward way to rewrite this code (and other code using `faerie`) to use `object` is to make the appropriate translations between the two APIs.  But there's nothing about the non-stack probe code that `cranelift-module` can't already handle -- basically defining a bunch of tables with assorted relocations.  So I propose adding an interface to `cranelift-module::Backend` that enables you to define a function in terms of the raw bytes for that function, rather than going through cranelift's code generation.

#### Benefit

If you squint right, this interface is a primitive form of inline assembly, which has proven useful in other compilers for similar reasons (tightly controlled instruction sequences, as well as access to functionality not directly supported by the compiler or doing things that are Hard To Do in the original language).

For the above example, this interface enables the stack probe to be defined in terms of `cranelift-module` interfaces, and subsequently all the remaining pieces to be defined in terms of `cranelift-module` interfaces as well, thus eliminating the need to drop down to a lower-level interface.  Switching between `faerie` and `object` would get a lot easier as well -- a handful of lines changed, rather than rewriting a bunch of object definitions.

#### Implementation

This functionality should be straightforward to add: all of the `cranelift-module` implementations already have to have a way of taking bytes and associating a function name with those bytes, inside `Backend::define_function`.  This interface would just sidestep the code generation parts.

I don't know what the exact signature should look like.  Something like `define_function_bytes(&mut self, id: FuncId, name: &str, bytes: &[u8], namespace: &ModuleNamespace<Self>)` would work for the above use case.  Depending on how Right people think a first cut should be, maybe you'd want to include an argument specifying relocations so external functions could be called.  I'm inclined to do the simplest thing possible for now.

#### Alternatives

[...]

PR #1400, opened 2020-02-21
add a "raw" function definition interface to cranelift-module

froydnj:
- [X] This has been discussed in issue #1386.
- [X] A short description of what this does, why it is needed: for certain functions (`lucet`'s stack probe code is the motivating example), generating them via cranelift is not necessarily feasible.  Other compilers have recognized this and provided inline assembly for such cases.  This interface can be thought of as a very primitive form of inline assembly (and a limited form; e.g. relocations are not currently supported).  The overarching intent is to make `cranelift-module`'s API rich enough that `lucet` can operate entirely within `cranelift-module`'s API and not have to drop down to `faerie` or `object`.  See #1386 for more details.
- [X] This PR does not contain test cases.

### DIFF
--- a/cranelift-module/src/traps.rs
+++ b/cranelift-module/src/traps.rs
@@ -1,0 +1,14 @@
+//! Defines `TrapSite`.
+
+use cranelift_codegen::{binemit, ir};
+
+/// Record of the arguments cranelift passes to `TrapSink::trap`.
+#[derive(Clone, Debug)]
+pub struct TrapSite {
+    /// Offset into function.
+    pub offset: binemit::CodeOffset,
+    /// Source location given to cranelift.
+    pub srcloc: ir::SourceLoc,
+    /// Trap code, as determined by cranelift.
+    pub code: ir::TrapCode,
+}
--- a/cranelift-simplejit/src/backend.rs
+++ b/cranelift-simplejit/src/backend.rs
@@ -8,7 +8,7 @@ use cranelift_codegen::isa::TargetIsa;
 use cranelift_codegen::{self, ir, settings};
 use cranelift_module::{
     Backend, DataContext, DataDescription, DataId, FuncId, Init, Linkage, ModuleNamespace,
-    ModuleResult,
+    ModuleResult, TrapSite,
 };
 use cranelift_native;
 #[cfg(not(windows))]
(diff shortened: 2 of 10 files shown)


cranelift-module/src/traps.rs
```diff
@@ -0,0 +1,14 @@
+//! Defines `TrapSite`.
+
+use cranelift_codegen::{binemit, ir};
+
+/// Record of the arguments cranelift passes to `TrapSink::trap`
```
bjorn3: Nit: comments missing trailing dots.

cranelift-simplejit/src/backend.rs
```diff
@@ -299,6 +299,42 @@ impl<'simple_jit_backend> Backend for SimpleJITBackend {
...
+        _traps: Vec<TrapSite>,
+    ) -> ModuleResult<Self::CompiledFunction> {
+        let size = bytes.len();
+        let ptr = self
+            .memory
+            .code
+            .allocate(size, EXECUTABLE_DATA_ALIGNMENT)
+            .expect("TODO: handle OOM etc.");
+
+        if cfg!(target_os = "linux") && ::std::env::var_os("PERF_BUILDID_DIR").is_some() {
```
bjorn3: Maybe extract this as a function to be used in `define_function` too.

cranelift-simplejit/src/backend.rs
```diff
@@ -299,6 +299,42 @@ impl<'simple_jit_backend> Backend for SimpleJITBackend {
...
+        }
+
+        unsafe {
+            ptr::copy_nonoverlapping(bytes.as_ptr(), ptr, size);
+        }
+
+        Ok(Self::CompiledFunction {
+            code: ptr,
+            size,
+            relocs: vec!(),
```
bjorn3:
```suggestion
            relocs: vec![],
```

bjorn3:
When running in the context of `perf` the `PERF_BUILDID_DIR` env var is set. By creating a map file, perf is able to map addresses of jitted functions back to their name.

(original author of that code speaking here)

pchickey:
@bjorn3 thanks. In that case, I agree with your suggestion that it be factored out to a common func, and it would be good to see that explanation in a comment there

froydnj:
Thanks for the reviews!  I added a commit with a comment that I think satisfies @pchickey's request.

PR #1403, opened 2020-02-22
Update to wasmparser 0.51

joshtriplett:
wasmparser::BinaryReaderError now encapsulates its fields, so call the
accessors rather than destructuring to get the fields.

### DIFF
--- a/cranelift-wasm/Cargo.toml
+++ b/cranelift-wasm/Cargo.toml
@@ -11,7 +11,7 @@ keywords = ["webassembly", "wasm"]
 edition = "2018"
 
 [dependencies]
-wasmparser = { version = "0.49.0", default-features = false }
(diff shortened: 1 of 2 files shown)


cranelift-wasm/Cargo.toml
```diff
@@ -11,7 +11,7 @@ keywords = ["webassembly", "wasm"]
 edition = "2018"
 
 [dependencies]
-wasmparser = { version = "0.49.0", default-features = false }
+wasmparser = { version = "0.51.0", default-features = false }
```
abrown: wasmparser is at [0.51.2](https://crates.io/crates/wasmparser); do you want to bump to that?

joshtriplett: 0.51.2 is semver-compatible, and we don't require anything from 0.51.2, so it doesn't matter either way.

PR #1405, opened 2020-02-24
Bump version to 0.59.0

fitzgen:


fitzgen:
Note that I'm not an owner on crates.io so someone who is will need to actually run the `cargo publish` (or add me as an owner and I can do it -- either way).

fitzgen:
Did #1406 instead.

